PR #44, opened 2020-03-02
Stop class -- internal representation

fabrice-v:
**Summary:**

This PR relates to processing of entities from stops.txt. The goal is to add the required use cases to go from a ParsedEntity representing a row from stops.txt to a specific concrete internal representation, based on the `location_type` field. Any parsed entity not providing all required data (or refering invalid parents) will be skipped. Making sure there are no duplicate IDs will also be done at that level.

**Expected behavior:** 


domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/stops/LocationBase.java
```diff
@@ -0,0 +1,132 @@
...
+
+package org.mobilitydata.gtfsvalidator.domain.entity.stops;
+
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Base class for all entities defined in stops.txt
+ */
+public abstract class LocationBase {
```
lionel-nj: Should LocationBase derive from a generic GtfsEntity type? Since we want to add stop objects, agency objects etc to a GtfsDataRepository. That could ease the process of type resolution.

barbeau: I think it makes sense to have a `GtfsEntity` superclass or interface if there is a common implementation or interface for all objects. And the only fields I can think of that would make sense is an ID for the GTFS dataset (maybe a hash of the file?), a globally unique ID for the entity (something like [Transitland's Onestop ID](https://transit.land/documentation/onestop-id-scheme/)), or both.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/stops/Entrance.java
```diff
@@ -0,0 +1,122 @@
...
+            return this;
+        }
+
+        public EntityBuildResult<?> build() {
+            if (stopId == null || stopCode == null || stopLat == null || stopLon == null || parentStation == null) {
+                if (stopId == null) {
+                    noticeCollection.add(new MissingRequiredValueNotice("stops.txt", "stop_id", null));
+                }
+                if (stopCode == null) {
+                    noticeCollection.add(new MissingRequiredValueNotice("stops.txt", "stop_code", stopId));
```
LeoFrachet, line 90: `stop_code` is optional, not required.

Could you add a process to have somebody (maybe Lionel) double checking those things?

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/stops/LocationBase.java
```diff
@@ -0,0 +1,171 @@
...
+                           String stopTimezone) {
+        this.stopId = stopId;
+        this.stopCode = stopCode;
+        this.stopName = stopName;
+        this.stopDesc = stopDesc;
+        this.stopLat = stopLat;
+        this.stopLon = stopLon;
+        this.zoneId = zoneId;
+        this.stopUrl = stopUrl;
+        this.stopTimezone = stopTimezone;
```
LeoFrachet, line 98: Is there a validation done on the timezone and the URL?

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/stops/LocationBase.java
```diff
@@ -0,0 +1,156 @@
...
+        GENERIC_NODE(3),
+        BOARDING_AREA(4);
+
+        private int value;
+
+        LocationType(int value) {
+            this.value = value;
+        }
+
+        static public LocationType fromInt(Integer fromValue) {
```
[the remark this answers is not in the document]
LeoFrachet: (Sorry this was on old pending comment, you likely can ignore it)

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ProcessParsedStop.java
```diff
@@ -0,0 +1,143 @@
...
+
+    public ProcessParsedStop(final GtfsSpecRepository specRepo,
+                             final ValidationResultRepository resultRepo,
+                             final GtfsDataRepository gtfsRepo) {
+        this.specRepo = specRepo;
+        this.resultRepo = resultRepo;
+        this.gtfsRepo = gtfsRepo;
+    }
+
+    public void execute(final ParsedEntity validatedStopEntity) {
```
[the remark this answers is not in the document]
LeoFrachet: (Sorry this was on old pending comment, you likely can ignore it)

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ProcessParsedStop.java
```diff
@@ -0,0 +1,143 @@
...
+                        .stopCode(stopCode)
+                        .stopDesc(stopDesc)
+                        .zoneId(zoneId)
+                        .stopUrl(stopUrl)
+                        .stopTimezone(stopTimezone);
+                //TODO: ready to be built and added to gtfsDataRepo
+                //TODO: wheelchair value in a subsequent use case (FinalizeStopEntity)
+                break;
+            }
+            case ENTRANCE_OR_EXIT: {
```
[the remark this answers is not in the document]
LeoFrachet: (Sorry this was on old pending comment, you likely can ignore it)

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/stops/LocationBase.java
fabrice-v: Earlier in the process there are.

- URL have their format validated
- timezone are validated against a list of timezones packaged with the SDK (and that shall be updatable)

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/stops/Entrance.java
fabrice-v: Yes, the process is having unit testing. Tests to come in subsequent commits before merge.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/stops/Entrance.java
```diff
@@ -0,0 +1,117 @@
...
+        }
+
+        public EntranceBuilder wheelchairBoarding(@Nullable final Integer wheelchairBoarding) {
+            this.wheelchairBoarding = WheelchairBoarding.fromInt(wheelchairBoarding);
+            this.originalWheelchairBoarding = wheelchairBoarding;
+            return this;
+        }
+
+        public EntityBuildResult<?> build() {
+            if (stopId == null || stopName == null || stopLat == null || stopLon == null || parentStation == null) {
```
lionel-nj, line 82:
```suggestion
            if (stopId == null || stopName == null || stopLat == null || stopLon == null || parentStation == null || wheelchairBoarding == INVALID_VALUE) {
```

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/stops/LocationBase.java
```diff
@@ -0,0 +1,184 @@
...
+/**
+ * Base class for all entities defined in stops.txt
+ */
+public abstract class LocationBase extends GtfsEntity {
+
+    private @NotNull
+    final String stopId;
+    private final String stopCode;
+    private final String stopName;
+    private final String stopDesc;
```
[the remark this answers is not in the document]
barbeau: It's official, it should be added to this PR.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/stops/LocationBase.java
```diff
@@ -0,0 +1,184 @@
...
+            stopId = null;
+            stopCode = null;
+            stopName = null;
+            stopDesc = null;
+            stopLat = null;
+            stopLon = null;
+            zoneId = null;
+            stopUrl = null;
+            stopTimezone = null;
+            childrenIdList = null;
```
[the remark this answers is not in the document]
barbeau: For large lists it can be more efficient to nullify it and let the old one be garbage collected rather than calling `.clear()`, which doesn't resize the internal data structure and therefore doesn't reclaim the memory. But I don't think the children list is expected to be huge here? Either would probably work.

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ProcessParsedStopAll.java
```diff
@@ -0,0 +1,304 @@
...
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Stream;
+
+/*
+ * This use case turns a parsed entity to a concrete class depending on the 'type' field
+ * Further processing stop.txt related entities is required to validate parent stations <--> child stops relationships
```
barbeau, line 37: This sounds like a `TODO`, but this is implemented now. I'd either point to where this validation happens in the comment or remove the comment.

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ProcessParsedStopAll.java
```diff
@@ -0,0 +1,304 @@
...
+        Map<String, List<String>> childrenPerStationId = new HashMap<>();
+
+        parsedEntityByStopId.values().forEach(stop -> {
+            String childId = stop.getEntityId();
+            String parentId = (String) stop.get("parent_station");
+            if (parentId != null) {
+                if (childrenPerStationId.containsKey(parentId)) {
+                    childrenPerStationId.get(parentId).add(childId);
+                } else {
+                    List<String> childrenList = new ArrayList<>();
```
lionel-nj, line 104: Should we work with `HashSet` to ensure no duplicate?

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/stops/BoardingArea.java
```diff
@@ -0,0 +1,83 @@
...
+    public static class BoardingAreaBuilder extends LocationBaseBuilder {
+
+        private String parentStation;
+
+        public BoardingAreaBuilder parentStation(final String parentStation) {
+            this.parentStation = parentStation;
+            return this;
+        }
+
+        public EntityBuildResult<?> build() {
```
barbeau, line 62: If I'm reading the spec right, stop `location_types` of `2`, `3`, and `4` (entrances, nodes, and boarding areas) shouldn't have children stops (i.e., they must be the leaf nodes). Should we be checking for a non-null children list here (and in the other relevant builders) and returning this as a notice?

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ProcessParsedStopAll.java
barbeau: I think we check for duplicate stop_ids elsewhere when inserting into the repo? If so, then not for duplicate checks (because a duplicate record would be the only way it could end up with the same child twice).

But HashSet would be useful here for efficient lookup to see if a stop has a particular stop_id as a child - although I don't know if we would encounter that use case in practice.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/stops/LocationBase.java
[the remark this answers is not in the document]
fabrice-v: `childrenIdList` being nullable, that would require a null check. Seemed superfluous as well as inconsistent, hence goign that way.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/stops/LocationBase.java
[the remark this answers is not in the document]
fabrice-v: Does the extension include other changes? I would be inclined to fully implement in a specific PR and not do partial implementation if that's the case. Otherwise yes, that should be added.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/stops/LocationBase.java
```diff
@@ -0,0 +1,184 @@
...
+        protected String stopId;
+        protected String stopCode;
+        protected String stopName;
+        protected String stopDesc;
+        protected Float stopLat;
+        protected Float stopLon;
+        protected String zoneId;
+        protected String stopUrl;
+        protected String stopTimezone;
+        protected List<String> childrenIdList = new ArrayList<>();
```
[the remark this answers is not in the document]
fabrice-v: No good reason. Will do.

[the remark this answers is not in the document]
fabrice-v: Said list can be nullified by setter. Actually I'm not sure we should even call the ArrayList constructor here

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/stops/LocationBase.java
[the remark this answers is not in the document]
barbeau: No, it's just the single field. I think when you started this PR for some reason the `tts_stop_name` field wasn't showing up on gtfs.org at the time (auto-sync with google/transit broke?), which is probably how it was missed.

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ProcessParsedStopAll.java
fabrice-v: It seems that would happen only in the case of duplicated rows in the original dataset (same `stop_id`, same `parent_station`) and would be caught as such.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/stops/BoardingArea.java
fabrice-v: Oh right could point. Will change

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/stops/LocationBase.java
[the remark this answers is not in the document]
fabrice-v: Also, our current goal is to attain feature parity with the pyhon validator.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/stops/BoardingArea.java
fabrice-v: Thought more about that while implementing. Because we validate parent `location_type` for any stop with a `parent_station`, that should be equivalent. Also, because parent --> child is not modeled directly in the dataset, the resulting notice would run the risk of being hard to decipher by users of the validator. For those reason, I think checking children in the builder is not nescessary.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/stops/StopOrPlatform.java
```diff
@@ -0,0 +1,129 @@
...
+
+    public @Nullable String getPlatformCode() {
+        return platformCode;
+    }
+
+    public static class StopOrPlatformBuilder extends LocationBaseBuilder {
+
+        private String parentStation;
+        private WheelchairBoarding wheelchairBoarding = INVALID_VALUE;
+        private Integer originalWheelchairBoarding = Integer.MAX_VALUE; // required to distinguish optional field not set
```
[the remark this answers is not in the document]
fabrice-v: Opening an issue?

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/stops/StopOrPlatform.java
```diff
@@ -0,0 +1,129 @@
...
+            return this;
+        }
+
+        public StopOrPlatformBuilder platformCode(@Nullable String platformCode) {
+            this.platformCode = platformCode;
+            return this;
+        }
+
+        public EntityBuildResult<?> build() {
+            if (stopId == null || stopName == null || stopLat == null || stopLon == null) {
```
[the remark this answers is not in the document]
fabrice-v: We now have a test for it

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/stops/Station.java
```diff
@@ -0,0 +1,116 @@
...
+        }
+
+        public StationBuilder wheelchairBoarding(@Nullable final Integer wheelchairBoarding) {
+            this.wheelchairBoarding = WheelchairBoarding.fromInt(wheelchairBoarding);
+            this.originalWheelchairBoarding = wheelchairBoarding;
+            return this;
+        }
+
+        public EntityBuildResult<?> build() {
+            if (stopId == null || stopName == null || stopLat == null || stopLon == null) {
```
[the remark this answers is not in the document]
fabrice-v: We now have a test for it

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/stops/Entrance.java
fabrice-v: We now have a test for it

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ProcessParsedStopAll.java
```diff
@@ -0,0 +1,304 @@
...
+        this.entranceBuilder = entranceBuilder;
+        this.genericNodeBuilder = genericNodeBuilder;
+        this.boardingAreaBuilder = boardingAreaBuilder;
+    }
+
+    public void execute(final Map<String, ParsedEntity> parsedEntityByStopId) {
+        Map<String, List<String>> childrenPerStationId = new HashMap<>();
+
+        parsedEntityByStopId.values().forEach(stop -> {
+            String childId = stop.getEntityId();
```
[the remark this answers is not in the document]
fabrice-v: Also finalized `parentId`

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/notice/error/StationWithParentStationNotice.java
```diff
@@ -0,0 +1,40 @@
...
+import java.io.IOException;
+
+public class StationWithParentStationNotice extends ErrorNotice {
+
+    public StationWithParentStationNotice(final String childId,
+                                          final String parentId) {
+        super("stops.txt", E_037,
+                "Non null `parent_station`",
+                "Stop with id:`" + childId + " is a station (location_type=1) with parent_id:`" +
+                        parentId + "`. This is forbidden. Expected null",
```
barbeau, line 31: Suggest a slightly clearer message here:

```suggestion
                "Stop with id:`" + childId + " is a station (location_type=1) with parent_id:`" +
                        parentId + "`. Stations cannot have parent stops. Expected null",
```

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/stops/Entrance.java
fabrice-v: https://github.com/MobilityData/gtfs-validator/blob/45601d0b30da2e58ccdb5760b9b0c9124bf70e1e/domain/src/test/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/stops/EntranceTest.java#L59

RULES.md
```diff
@@ -36,6 +36,8 @@ Rules are declared in the [`Notice` module](https://github.com/MobilityData/gtfs
 | [E032](#E032) | `calendar.txt` `end_date` is before `start_date` |
 | [E033](#E033) | `route_id` not found |
 | [E035](#E035) | `agency_id` not found |
+| [E036](#E035) | Stop with invalid parent `location_type` |
```
barbeau, line 39: Suggest slight wording change for rule to be more consistent with others:

```suggestion
| [E036](#E035) | Invalid parent `location_type` for stop |
```

barbeau: Oh, the error code needs to be changed to `E036` too (I just fixed that in my suggested change above.

RULES.md
```diff
@@ -36,6 +36,8 @@ Rules are declared in the [`Notice` module](https://github.com/MobilityData/gtfs
...
+| [E037](#E035) | Stop with `location_type` of 2 (Station) defines a parent |
```
barbeau, line 40: Same here - slight change of wording, and update error code:

```suggestion
| [E037](#E037) | Station stop (`location_type=2`) has a parent stop |
```

adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/db/InMemoryGtfsDataRepository.java
```diff
@@ -788,4 +792,51 @@ public Translation getTranslationByTableNameFieldValueLanguage(final String tabl
                                                                    final String language) {
         return translationPerTableName.get(tableName).get(fieldName).get(language);
     }
+
+    /**
+     * Add a stop representing a row from stops.txt to this {@link GtfsDataRepository}. Return the entity added to the
+     * repository if the uniqueness constraint of stop based on stop_id is respected. Otherwise returns null.
+     *
+     * @param newStop the internal representation. A daughter of {@link LocationBase}
```
barbeau, line 800:
```suggestion
     * @param newStop the internal representation. A subclass of {@link LocationBase}
```

adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/db/InMemoryGtfsDataRepository.java
```diff
@@ -788,4 +792,51 @@ public Translation getTranslationByTableNameFieldValueLanguage(final String tabl
...
+        } else {
+            throw new IllegalArgumentException("Cannot add null stop to data repository");
+        }
+    }
+
+    /**
+     * Return an immutable map of LocationBase objects representing all the rows from stops.txt,
+     *
+     * @return an immutable map of LocationBase objects representing all the rows from stops.txt
+     */
```
barbeau, line 825:
```suggestion
    /**
     * Return an immutable map of LocationBase objects representing all the rows from stops.txt,
     * with stop_id as the key
     *
     * @return an immutable map of LocationBase objects representing all the rows from stops.txt,
     * with stop_id as the key
     */
```

adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/db/InMemoryGtfsDataRepository.java
```diff
@@ -788,4 +792,51 @@ public Translation getTranslationByTableNameFieldValueLanguage(final String tabl
...
+    @Override
+    public Map<String, LocationBase> getStopAll() {
+        return Collections.unmodifiableMap(stopPerId);
+    }
+
+    /**
+     * Return the {@link LocationBase} representing a row from stops.txt related to the id provided as parameter
+     *
+     * @param stopId the key from stops.txt related to the LocationBase to be returned
```
barbeau, line 834:
```suggestion
     * @param stopId the stop_id from stops.txt related to the LocationBase to be returned
```

RULES.md
```diff
@@ -150,6 +152,18 @@ Value of field `route_id` should exist in GTFS data.
...
+
+### E036 - Stop with invalid parent `location_type`
+
+Value of field `location_type` of parent found in field `parent_station` in invalid.
+
+<a name="E037"/>
+
+### E037 - Stop with `location_type` of 2 (Station) defines a parent
+
+Value of field `parent_station` must be null when `location_type` is 2
```
barbeau, line 165:
```suggestion
Value of field `parent_station` must be empty when `location_type` is 2
```

RULES.md
```diff
@@ -150,6 +152,18 @@ Value of field `route_id` should exist in GTFS data.
 
 Value of field `agency_id` should exist in GTFS data.
 
+<a name="E036"/>
+
+### E036 - Stop with invalid parent `location_type`
+
+Value of field `location_type` of parent found in field `parent_station` in invalid.
```
barbeau, line 159: I would be good to expand on why it's invalid here and explicitly list the cases we check.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/stops/LocationBase.java
```diff
@@ -0,0 +1,185 @@
...
+            this.stopTimezone = timezone;
+            return this;
+        }
+
+        public LocationBaseBuilder childrenList(@Nullable List<String> children) {
+            this.childrenIdList = children;
+            return this;
+        }
+
+        protected void clearBase() {
```
barbeau, line 171: Looks like the subclass methods are all named `clear()` - should this be named `clear()` too?

fabrice-v: I couldn't figure the syntax to make it work as a protected member. hence the workaround. I'll try to look into it some more.

fabrice-v: thks for the help fixing it, @barbeau

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/stops/BoardingArea.java
```diff
@@ -0,0 +1,84 @@
...
+            if (stopId == null || parentStation == null) {
+                if (stopId == null) {
+                    noticeCollection.add(new MissingRequiredValueNotice("stops.txt", "stop_id", null));
+                }
+                if (parentStation == null) {
+                    noticeCollection.add(new MissingRequiredValueNotice("stops.txt", "parent_station", stopId));
+                }
+                return new EntityBuildResult<>(noticeCollection);
```
barbeau, line 71: Should this include `if (wheelchairBoarding == INVALID_VALUE) {` block too?

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/stops/GenericNode.java
```diff
@@ -0,0 +1,84 @@
...
+        }
+
+        public EntityBuildResult<?> build() {
+            if (stopId == null || parentStation == null) {
+                if (stopId == null) {
+                    noticeCollection.add(new MissingRequiredValueNotice("stops.txt", "stop_id", null));
+                }
+                if (parentStation == null) {
+                    noticeCollection.add(new MissingRequiredValueNotice("stops.txt", "parent_station", stopId));
+                }
```
barbeau, line 70: Same - should this include `if (wheelchairBoarding == INVALID_VALUE) {` block too?

Looking at the spec it's not immediately obvious what the valid options are for nodes and boarding areas.

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ProcessParsedStopAll.java
```diff
@@ -0,0 +1,317 @@
...
+                        if (LocationType.fromInt(parentLocationType) != LocationType.STATION) {
+                            resultRepository.addNotice(
+                                    new ParentStationInvalidLocationTypeNotice(
+                                            stopId, locationType, parentStation, 1,
+                                            parentLocationType
+                                    )
+                            );
+                        }
+
+                        if (WheelchairBoarding.fromInt(wheelchairBoarding) ==
```
barbeau, line 157: Stops also inherit parent `stop_timezone` - from https://gtfs.org/reference/static#stopstxt:

>If the location has a parent station, it inherits the parent station’s timezone instead of applying its own. Stations and parentless stops with empty stop_timezone inherit the timezone specified by agency.agency_timezone. If stop_timezone values are provided, the times in stop_times.txt should be entered as the time since midnight in the timezone specified by agency.agency_timezone.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/stops/BoardingArea.java
fabrice-v: My understanding of the spec is that `wheelchair_boarding` don't apply to generic node or boarding area.
@timMillet ?

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/stops/GenericNode.java
fabrice-v: My understanding of the spec is that `wheelchair_boarding` don't apply to generic node or boarding area.
@timMillet ?

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ProcessParsedStopAll.java
fabrice-v: Good catch! Some additional work is required. I will not merge this PR then

fabrice-v: Do you know what's the expected behavior if there are multiple agencies in agency.txt?

fabrice-v: Ok I got my answer from the spec

> If multiple agencies are specified in the dataset, each must have the same agency_timezone.

fabrice-v: Implemented in last commit
- location_type=0 -> inherit from parent if there is one. Otherwise inherit from agency if stop_timezone is not provided
- location_type==1 -> inherit from agency if stop_timezone is not provided
- location_type==2 -> inherit from parent if there is one
- location_type==3 -> inherit from parent if there is one
- location_type==4 -> inherit from parent if there is one

fabrice-v:
Decision was made with @LeoFrachet to go ahead and merge as is so we can release a new version. Then we'll work on adjusting it.

Issue #40, opened 2020-02-27
[REQUEST] Add Javadoc comments to domain layer entities as well as use cases


fabrice-v:
**Describe the solution you'd like**
To document our code, Javadoc comments should be added to entities defined in the domain layer

PR #45, opened 2020-03-02
Implement #40 - Add javadoc comments

lionel-nj:
**Summary:**

This PR will add javadoc comments in order to clarify the use of the code developed at this stage of the project. 

Related issue #40 

### DIFF
--- a/adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/db/FromFileRawEntityProvider.java
+++ b/adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/db/FromFileRawEntityProvider.java
@@ -27,6 +27,11 @@ import java.io.File;
 import java.io.IOException;
 import java.util.Map;
 
+/**
+ * Provides methods to perform operations on rows of a GTFS CSV file. Includes feature to transform a row of a GTFS
+ * CSV file into a {@link RawEntity}.
+ * This is called in {@link InMemoryRawFileRepository} to retrieve the data provider for a specific GTFS CSV file.
+ */
 public class FromFileRawEntityProvider implements RawFileRepository.RawEntityProvider {
     private final MappingIterator<Map<String, String>> dataSource;
     private final int headerCount;
@@ -47,16 +52,31 @@ public class FromFileRawEntityProvider implements RawFileRepository.RawEntityPro
         headerCount = headerCountArray[0];
     }
 
+    /**
+     * Returns a boolean indicating if a GTFS CSV file has more rows that can be read. 1 if yes, else 0.
+     *
+     * @return a boolean indicating if a GTFS CSV file has more rows that can be read. 1 if yes, else 0.
+     */
     @Override
     public boolean hasNext() {
         return dataSource.hasNext();
     }
 
+    /**
+     * Returns the next row of a GTFS CSV file as a {@link RawEntity}
+     *
+     * @return the next row of a GTFS CSV file as a {@link RawEntity}
+     */
     @Override
     public RawEntity getNext() {
         return new RawEntity(dataSource.next(), dataSource.getCurrentLocation().getLineNr());
     }
 
+    /**
+     * Returns the number of header of a GTFS CSV file.
+     *
+     * @return the number of header of a GTFS CSV file.
+     */
     @Override
     public int getHeaderCount() {
         return headerCount;
--- a/adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/db/FromFileRawEntityProvider.java
+++ b/adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/db/FromFileRawEntityProvider.java
@@ -53,9 +53,9 @@ public class FromFileRawEntityProvider implements RawFileRepository.RawEntityPro
     }
 
     /**
-     * Returns a boolean indicating if a GTFS CSV file has more rows that can be read. 1 if yes, else 0.
+     * Returns a boolean indicating if a GTFS CSV file has more rows that can be read. true if yes, else false.
      *
-     * @return a boolean indicating if a GTFS CSV file has more rows that can be read. 1 if yes, else 0.
+     * @return a boolean indicating if a GTFS CSV file has more rows that can be read. true if yes, else false
      */
     @Override
     public boolean hasNext() {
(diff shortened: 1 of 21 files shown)


domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/RawEntity.java
```diff
@@ -18,24 +18,44 @@
 
 import java.util.Map;
 
+/**
+ * The {@code RawEntity} class represents a row of a GTFS file as raw string data.
```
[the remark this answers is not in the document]
lionel-nj: From the recommendations given at: https://blog.joda.org/2012/11/javadoc-coding-standards.html; I understood that it is better not using @code in the first sentences. 

I modify accordingly to your advice.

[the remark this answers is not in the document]
barbeau: Actually, here, you're right (I missed that this was referring the same class we're already in). So this can stay as `@code`.

And you're right regarding the article, although I'm not sure I agree with it. Generally, I think if the class is mentioned more than once then follow the article advice. If it's only mentioned once, then I would prefer `@link` instead of `@code` to be able to jump to that class easily.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/ParsedEntity.java
```diff
@@ -18,26 +18,54 @@
 
 import java.util.Map;
 
+/**
+ * The {@code ParsedEntity} class represents a row of a GTFS file that has been parsed: the type of each column of the
+ * raw has been determined.
```
barbeau: Missing a word here - `raw data`?

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/RawEntity.java
```diff
@@ -18,24 +18,50 @@
...
+    /**
+     * Key: header string from original CSV file
+     * Value: a String, the raw value read from the CSV file
+     */
     private final Map<String, String> contentByHeaderMap;
     private final int entityIndex;
 
+    /**
+     * @param contentByHeaderMap the object mapping rows of a GTFS .txt file on header name
+     * @param entityIndex        the 1 based index of the row
```
barbeau: I would recommend that you add documentation where this method is called from to also indicate there that the index is 1-based. It's a little strange to have the documentation here, being that the RawEntity class itself doesn't seem to care whether it is 0 or 1 based, but as long as it's also documented where the code is called from I suppose it's ok to leave this. On this topic, in this documentation, I would recommend adding a summary description that includes something like:

>This class is used by `{@link XXXX}` with a 1 based index to identify the row location within a GTFS CSV file.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/RawEntity.java
```diff
@@ -18,24 +18,50 @@
...
+     * @return the number of headers (columns)
+     */
     public int size() {
         return contentByHeaderMap.size();
     }
 
+    /**
+     * Returns the 1 based index of the row
+     *
+     * @return the 1 based index of the row
```
barbeau: Same as above - please add documentation where this method is called from that it's 1-based.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/RawFileInfo.java
```diff
@@ -16,16 +16,29 @@
 
 package org.mobilitydata.gtfsvalidator.domain.entity;
 
+/**
+ * The @code{RawFileInfo} class contains information regarding a file location and expected content (file name)
```
barbeau: If you're referring to the class this Javadoc is for, you can omit the name to make the description shorter. So this would be:

>Contains information regarding a file location and expected content (file name)

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/CleanOrCreatePath.java
```diff
@@ -25,17 +25,33 @@
 import java.nio.file.Path;
 import java.util.Comparator;
 
+/**
+ * Use case to create a path. This use case is triggered after parsing the rows of a specific file. The resultant path
```
barbeau: 👍 This is a really good Javadoc. It gives you a lot of context about what this class does. Because this can be a destructive operation, I would suggest adding the note you have later in `execute()` to this Javadoc too:

>If the target location is not empty, all files at target are deleted.

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateHeadersForFile.java
```diff
@@ -44,6 +54,12 @@ public ValidateHeadersForFile(final GtfsSpecRepository specRepo,
         this.resultRepo = resultRepo;
     }
 
+    /**
+     * Use case execution method: for a file, checks the presence of all headers marked as "required" in the
+     * GTFS specification. A {@code MissingHeaderNotice} is generated each time a required header is missing.
```
barbeau: I would `{@link` the classes in this Javadoc so you can click on them within the IDE.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/RawEntity.java
lionel-nj: The entity is used in various use cases. We do not have dependency on these in the entity layer, therefore does not seem possible to @link them. 
We could write something like:

> This class matches a 1 based index identifying the row location within a GTFS CSV file and its content as a map of strings.

The map being described in the next @param

barbeau: @lionel-nj Ok, I think that helps. I think the important part is that the 1-based index is documented where this class is used, because that code (not this class) is what's enforcing (and relying on) the 1-based indexing.

adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/db/FromFileRawEntityProvider.java
```diff
@@ -47,16 +52,31 @@ public FromFileRawEntityProvider(RawFileInfo file) throws IOException {
         headerCount = headerCountArray[0];
     }
 
+    /**
+     * Returns a boolean indicating if a GTFS CSV file has more rows that can be read. 1 if yes, else 0.
```
barbeau: Nit (a small detail) - `1` should be `true` in the comments and `0` should be `false`, as `true` and `false` are the boolean literal values.

lionel-nj: Indeed

PR #55, opened 2020-03-12
Fix #54  - Redefinition of filename list to validate to avoid crash

### LINKED ISSUE #54: [BUG] Validator crashes when trying to uncomment filenameList in Main method
**Describe the bug**
The validator crashes when trying to validate files that have been commented in Main method.  See
https://github.com/MobilityData/gtfs-validator/blob/aa2634e82478cf7ce88e79018d0eb787f923b782/application/cli-app/src/main/java/org/mobilitydata/gtfsvalidator/Main.java#L111

**To Reproduce**
Run the program with these arguments: 
-u https://openmobilitydata.org/p/mbta/64/latest/download -z input.zip -i input -o output

**Expected behavior**
The validation process should be performed on all required and optional files, IF these are included in the archive.

**Witnessed behavior**
The code actually tries to validate files flagged as optional in the GTFS specification that do not exist in the downloaded archive, hence the crashing behavior.

**Environment used**
- OS: macOs Catalina 10.15.3 
- Java version: javac 11.0.5
 
**Additional context**

lionel-nj:
**Summary:**

This PR modifies the definition of the filename list to validate. We implement a new use case ValidateAllOptionalFileName and modify ValidateAllRequiredFileName to define the list of files that should be validates.

NB: at present date types are not supported, therefore we exclude files "calendar.txt", "calendar_dates.txt", and "stop_times.txt" from the validation process in order to avoid the generation of too many notices (which also causes the validator to crash at the report generation step).

**Expected behavior:** 

The filename list to validate is generated from the list of required files that were found in the archive  and the list of optional files that were found in the same archive.

Capture d’écran, le 2020-03-12 à 15 18 22

Linked issue: https://github.com/MobilityData/gtfs-validator/issues/54

EDIT: I just realized that I should have make this PR a draft, it seems to be too late to do so. Consider that this PR is not ready for merging yet.


adapter/repository/in-memory-simple/src/main/resources/gtfs_spec.asciipb
```diff
@@ -429,7 +429,7 @@ csvspec: {
 csvspec: {
   name: "routes"
   filename: "routes.txt"
-  required: false  # required unless the feed is stops only.
+  required: true  # required unless the feed is stops only.
```
barbeau: Do we want to get rid of `unless the feed is stops only.`? This was a Google-thing - official GTFS spec doesn't support stop-only feeds.

adapter/repository/in-memory-simple/src/main/resources/gtfs_spec.asciipb
```diff
@@ -1311,7 +1311,7 @@ csvspec: {
     value_required: true
     type: { type: INT_DEC }
     intmin: 0
-    intmax: 300000
+    intmax: 5500000
```
barbeau: Why is this `5500000`?

application/cli-app/src/main/java/org/mobilitydata/gtfsvalidator/Main.java
```diff
@@ -22,7 +22,8 @@
 import org.mobilitydata.gtfsvalidator.config.DefaultConfig;
 import org.mobilitydata.gtfsvalidator.usecase.ParseSingleRowForFile;
 
-import java.io.*;
```
barbeau: Looks like there is some shuffling of imports here (i.e., this commit changes the imports committed by someone else), which usually means that we collectively aren't using the same IntelliJ formatting template. Or someone doesn't have "Optimize imports" checked when committing within IntelliJ. It would be good to standardized on a code format template.

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateAllRequiredFilePresence.java
```diff
@@ -21,6 +21,8 @@
 import org.mobilitydata.gtfsvalidator.usecase.port.RawFileRepository;
 import org.mobilitydata.gtfsvalidator.usecase.port.ValidationResultRepository;
 
+import java.util.List;
+
 public class ValidateAllRequiredFilePresence {
```
barbeau: The name of this should probably be `ValidateAllRequiredFilesPresence` (the `s` is missing).

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateAllOptionalFileName.java
```diff
@@ -0,0 +1,40 @@
+package org.mobilitydata.gtfsvalidator.usecase;
+
+import org.mobilitydata.gtfsvalidator.usecase.notice.OptionalFileFoundNotice;
+import org.mobilitydata.gtfsvalidator.usecase.port.GtfsSpecRepository;
+import org.mobilitydata.gtfsvalidator.usecase.port.RawFileRepository;
+import org.mobilitydata.gtfsvalidator.usecase.port.ValidationResultRepository;
+
+import java.util.List;
+
+public class ValidateAllOptionalFileName {
```
barbeau: I think this class should be renamed - it doesn't really validate all optional file names, but more discovers and adds optional files to the provided file list (if I'm understanding correctly looking at the code). Maybe `DiscoverAndAddOptionalFiles`?

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateAllOptionalFileName.java
```diff
@@ -0,0 +1,40 @@
...
+    public ValidateAllOptionalFileName(final GtfsSpecRepository specRepo, final RawFileRepository rawFileRepo,
+                                       final ValidationResultRepository resultRepo, List<String> filenameList) {
+        this.specRepo = specRepo;
+        this.rawFileRepo = rawFileRepo;
+        this.resultRepo = resultRepo;
+        this.filenameList = filenameList;
+    }
+
+    public List<String> execute() {
+
```
barbeau: Extra whitespace

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateAllOptionalFileName.java
```diff
@@ -0,0 +1,40 @@
...
+        specRepo.getOptionalFilenameList().stream()
+                .filter(optionalFilename -> rawFileRepo.getFilenameAll().contains(optionalFilename))
+                .forEach(optionalFilename -> {
+                            filenameList.add(optionalFilename);
+                            resultRepo.addNotice(new OptionalFileFoundNotice(optionalFilename,
+                                    "I002",
```
barbeau: This is a separate issue, but I'd refactor `I002` and all other INFO, WARNINGS, ERRORS to variables ASAP. Otherwise we're just creating more work for ourselves every time we add a new one as a string. If we don't have an issue for this yet (I didn't see one scanning the list), could you create one?

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateAllOptionalFileNameTest.java
```diff
@@ -0,0 +1,186 @@
...
+            return null;
+        }
+
+        @Override
+        public Notice addNotice(Notice newNotice) {
+            return null;
+        }
+    }
+
+    @Test
```
barbeau: I would put the `@Test` methods at the top of the file, and all other boilerplate code that supports them below it. You'll likely edit to the `@Test` methods far more than the others, and that way you don't have to scroll down to find it.

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateAllRequiredFilePresence.java
```diff
@@ -35,12 +37,15 @@ public ValidateAllRequiredFilePresence(final GtfsSpecRepository specRepo,
...
 
-    public void execute() {
+    public List<String> execute() {
         if (!rawFileRepo.getFilenameAll().containsAll(specRepo.getRequiredFilenameList())) {
 
             specRepo.getRequiredFilenameList().stream()
                     .filter(requiredFile -> !rawFileRepo.getFilenameAll().contains(requiredFile))
                     .forEach(missingFile -> resultRepo.addNotice(new MissingRequiredFileNotice(missingFile)));
+
+            return specRepo.getRequiredFilenameList();
```
barbeau: I don't think the same line of code should be within and outside of the IF statement here?

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateAllOptionalFileName.java
```diff
@@ -0,0 +1,40 @@
...
+import java.util.List;
+
+public class ValidateAllOptionalFileName {
+
+    private final GtfsSpecRepository specRepo;
+    private final RawFileRepository rawFileRepo;
+    private final ValidationResultRepository resultRepo;
+    private final List<String> filenameList;
+
+    public ValidateAllOptionalFileName(final GtfsSpecRepository specRepo, final RawFileRepository rawFileRepo,
```
barbeau: This and `execute()` need Javadocs. It's not clear from the calling class that the `filenameList` you pass in here will be appended to and returned from `execute()`. Also, you need to decide here whether you should really be modifying the list that is passed in, or if you should first create a copy of the provided list and then modify the copy. The way this is currently set up, calling code may not expect that the list that is passed into the constructor will be modified. Either way, the behavior should be documented.

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateAllOptionalFileName.java
```diff
@@ -0,0 +1,40 @@
...
+        this.resultRepo = resultRepo;
+        this.filenameList = filenameList;
+    }
+
+    public List<String> execute() {
+
+        specRepo.getOptionalFilenameList().stream()
+                .filter(optionalFilename -> rawFileRepo.getFilenameAll().contains(optionalFilename))
+                .forEach(optionalFilename -> {
+                            filenameList.add(optionalFilename);
```
barbeau: Why are we adding INFO notices for each optional file? Just curious - it's not something I would have thought to do. Are you using these notices later for something?

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateAllOptionalFileName.java
```diff
@@ -0,0 +1,40 @@
...
+import org.mobilitydata.gtfsvalidator.usecase.port.ValidationResultRepository;
+
+import java.util.List;
+
+public class ValidateAllOptionalFileName {
+
+    private final GtfsSpecRepository specRepo;
+    private final RawFileRepository rawFileRepo;
+    private final ValidationResultRepository resultRepo;
+    private final List<String> filenameList;
```
barbeau: You don't need to include the type in the variable name - this can be `filenames`

adapter/repository/in-memory-simple/src/main/resources/gtfs_spec.asciipb
lionel-nj: I think that we should, since that is a google specific thing. Will modify.

adapter/repository/in-memory-simple/src/main/resources/gtfs_spec.asciipb
lionel-nj: After a few iteration I realized that the previous value was too small. I set it to 5500000 in order to ensure that the program run correctly and does not generate too many out of range notice. 

This seems to be a calibration issue, that could require to explore a large quantity of GTFS datasets. Maybe we should to open an issue on this matter as I realize that we need to go further on this?

application/cli-app/src/main/java/org/mobilitydata/gtfsvalidator/Main.java
lionel-nj: I will check this with @fabrice-v

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateAllOptionalFileName.java
lionel-nj: Given the code's behavior, I agree

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateAllRequiredFilePresence.java
lionel-nj: @fabrice-v we should decide of a common naming convention

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateAllOptionalFileNameTest.java
lionel-nj: I understand. To ensure consistency the same manipulation should be done in every file.
Will modify.

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateAllRequiredFilePresence.java
lionel-nj: I agree, will modify.

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateAllOptionalFileName.java
lionel-nj: Good remark. I realize that the best option should be to add an Info notice for each missing optional file. Having this information, the person using the validator would be aware of any missing optional file he/she tried to validate. 

It might be a bit memory consuming, but at the same time we would generate at most 10 notices.

Maybe this is not something that we'd like .. we can discuss this with @fabrice-v  and @LeoFrachet

adapter/repository/in-memory-simple/src/main/resources/gtfs_spec.asciipb
barbeau: Is there a downside to setting this to the max int value for 32 bit, which would be `2147483647`?

That should handle any reasonable value.

fabrice-v: We can't see any downside. Will modify.

application/cli-app/src/main/java/org/mobilitydata/gtfsvalidator/Main.java
fabrice-v: We will check the `Optimize import` in the commit dialog box, in addition to `Reformat code`
image

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateAllRequiredFilePresence.java
fabrice-v: So, as a rule of thumb I try to avoid 's' postfixes as they tend to lead to sometimes hard to debug typos. I would suggest using explicit naming like `headerCollection`or `headerList` instead of `headers` for example. In this specific case maybe `ValidateRequiredFileAll` would fit? Very open to counter arguments.

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateAllOptionalFileName.java
fabrice-v: The intent here is to actually validate that all non required files names indeed are in the schema definition. Will make modifications as such.

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateAllOptionalFileName.java
fabrice-v: Ok assuming you meant refactoring Ids, that is already the case as can be seen here for errors
https://github.com/MobilityData/gtfs-validator/blob/aa2634e82478cf7ce88e79018d0eb787f923b782/usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/notice/base/ErrorNotice.java#L23
This specific Info Id was an overlook

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateAllOptionalFileNameTest.java
fabrice-v: #56, will be done in a separate PR as it will touch all test classes

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateAllOptionalFileName.java
fabrice-v: Excellent point. We will modify the behavior as follow
- `ValidateAllRequiredFilePresence` will return the list of all required filenames
- `ValidateAllOptionalFilename`will return the list of all validated (that is, present in the schema file) optional filename
Concatenation will happen in client code.

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateAllOptionalFileName.java
fabrice-v: The use for info notices is not very defined yet. The middle/long term idea is to have an optional verbosity option that would surface them.

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateAllOptionalFileName.java
fabrice-v: That was done at my express request. From my experience `s` postfixes can lead to code where singular and plural forms coexist, sometimes leading to hard to debug issue because of a bad autocomplete. Hence explicit postfix. I'm quite open to discussion on that one.

application/cli-app/src/main/java/org/mobilitydata/gtfsvalidator/Main.java
barbeau: :+1: FWIW, you can enable this as a default in IntelliJ via "File->Settings". Just search for "import" and it's under the version control section

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateAllRequiredFilePresence.java
barbeau: This is certainly a style issue, so if you have strong feelings against the `s` that's ok. Personally, I find it easier to read the code when it's more concise and more closely reflects conversational language.

So:

```java
for (File f : files)
```

...is more natural to me than:

```java
for (File f : fileList)
```

Also, if you change types of the variable (e.g., from a `List` to a `Map`), then you also have to refactor the variable name, which adds more unnecessary chatter to version control.

But, I don't have extremely strong feelings on this.

Not sure if I get the `s` issue with Class names, though...

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateAllOptionalFileName.java
barbeau: Ah, yes, good!

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateAllRequiredFilePresence.java
fabrice-v: On linking variable name to the implementation, we'll prefer using Collection. 
As for the class name I was thinking about being consistent.

Issue #63, opened 2020-03-16
[REQUEST] Extend tests of notice repository to find memory limits


fabrice-v:
**Is your feature request related to a problem? Please describe.**
Right now the validator quickly runs out of memory when generating notices

**Describe the solution you'd like**
We need automated tests to figure out. the memory limits of our notices repository implementation.
Two kinds of tests
- test that pass when memory is not exhausted for a given number of notices
- tests that pass when memory IS exhausted for a given number of notices
Those trests will allow to monitor memory footprint progress as well as prevent regressions.

barbeau:
Here are the three implementations high level implementations / adaptors I can think of for the notice repository:
1. In memory (`InMemoryValidationResultRepository`)
1. File-based
1. Database-based

I'm assuming the tests are intended to measure memory use of 1) at the moment, but could potentially be used for 2) and 3) too.

>tests that pass when memory IS exhausted for a given number of notices

How are you planning on testing this?

fabrice-v:
Yes, tests are intended for case 1

As for testing memory exhaustion, we're not sure yet how the testing framework behaves but we are expecting something along the line of an exception being raised. So a valid test would check such exception is raised.

fabrice-v:
PR #65 can be merged in it's current state

PR #65, opened 2020-03-16
Implement #63 -- Validate notices repository memory delimitation

lionel-nj:
**Summary:**

This PR comes provides implementation of tests to determine the limits of the memory while generating notices. 

**Expected behavior:** 

1. These tests ensure that the development of our validation tool does not deteriorate our current performance. They fail if we generate an exception for sizes of validation repository that are currently handled.
  - creationOf100NoticeShouldNotExceedMemoryLimit
  - creationOf1000NoticeShouldNotExceedMemoryLimit
  - creationOf10000NoticeShouldNotExceedMemoryLimit
  - creationOf100000NoticeShouldNotExceedMemoryLimit
  - creationOf1000000NoticeShouldNotExceedMemoryLimit
  - creationOf2000000NoticeShouldNotExceedMemoryLimit

These tests are machine independent given the fact that they assert the number of used bytes during the generation of notices.

Capture d’écran, le 2020-04-03 à 13 29 58


adapter/repository/in-memory-simple/src/test/java/org/mobilitydata/gtfsvalidator/db/NoticeMemoryConsumptionTest.java
```diff
@@ -0,0 +1,140 @@
...
+        mockValidationResultRepo.getAll();
+    }
+
+    @Test
+    public void creationOf3_000_000NoticeShouldRaiseException() {
+
+        MockValidationResultRepo mockValidationResultRepo = new MockValidationResultRepo();
+
+        try {
+            for (int i = 0; i < 3_000_000; i++) {
```
lionel-nj: Even tho I created a private method `generateNotices` to avoid code duplication. The tests did not seem to run when calling this method in this `creationOf3_000_000NoticeShouldRaiseException`. So I duplicated the code here...could be improved.

adapter/repository/in-memory-simple/src/test/java/org/mobilitydata/gtfsvalidator/db/NoticeMemoryConsumptionTest.java
```diff
@@ -0,0 +1,140 @@
...
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public class NoticeMemoryConsumptionTest {
+
+    private void generateNotices(MockValidationResultRepo mockValidationResultRepo, int numberOfNotices) {
```
barbeau: If you're trying to ensure that the validator itself doesn't throw an exception in production on a certain number of notices, then this method should use the actual repo implementation being used in production, not a mocked repo.

adapter/repository/in-memory-simple/src/test/java/org/mobilitydata/gtfsvalidator/db/NoticeMemoryConsumptionTest.java
```diff
@@ -0,0 +1,140 @@
...
+    public void creationOf1_000_000NoticeShouldNotRaiseException() {
+
+        MockValidationResultRepo mockValidationResultRepo = new MockValidationResultRepo();
+
+        generateNotices(mockValidationResultRepo, 1_000_000);
+        mockValidationResultRepo.getAll();
+    }
+
+    @Test
+    public void creationOf3_000_000NoticeShouldRaiseException() {
```
barbeau: If you're trying to make sure that this method throws an exception, then it should be annotated like:

`@Test(expected = OutOfMemoryError.class)`

...instead of catching the exception. So, there shouldn't be a try/catch block.

fabrice-v: Oh, thks for that!

fabrice-v: It seems though things changed a little for JUnit5
https://howtodoinjava.com/junit5/expected-exception-example/

adapter/repository/in-memory-simple/src/test/java/org/mobilitydata/gtfsvalidator/db/NoticeMemoryConsumptionTest.java
```diff
@@ -0,0 +1,140 @@
...
+    @Test
+    public void creationOf1_000_000NoticeShouldNotRaiseException() {
+
+        MockValidationResultRepo mockValidationResultRepo = new MockValidationResultRepo();
+
+        generateNotices(mockValidationResultRepo, 1_000_000);
+        mockValidationResultRepo.getAll();
+    }
+
+    @Test
```
barbeau: I'm not entirely sure I understand what this test is trying to do. Right now it just proves that we can't handle a certain number of records, but that's going to be very machine-dependent - it may pass on some machines and fail on others.

It seems like the following alternate versions would be more useful:
1. Benchmark and print memory consumption of the test - Maybe using something like `Runtime.getRuntime().totalMemory() - Runtime.getRuntime().freeMemory()` (see https://stackoverflow.com/questions/17374743/how-can-i-get-the-memory-that-my-java-program-uses-via-javas-runtime-api)? This would give us a machine-independent history of how much memory we're using
1. Fail gracefully - Have a test that ensures if we do run out of memory when validating a feed that we're able to fail gracefully and still provide results. This would need to be mocked to force the error to ensure that the test passes on all machines.

adapter/repository/in-memory-simple/src/test/java/org/mobilitydata/gtfsvalidator/db/NoticeMemoryConsumptionTest.java
barbeau: I'd suggest trying to use the private method again after getting rid of the try block and annotating the test to expect the exception. You may need to make the private method `throws Exception {`.

adapter/repository/in-memory-simple/src/test/java/org/mobilitydata/gtfsvalidator/db/NoticeMemoryConsumptionTest.java
barbeau: @fabrice-v Oh, good point! Looks like you could use either (i.e., annotation on the method should still work), but `assertThrows()` allows more granularity.

adapter/repository/in-memory-simple/src/test/java/org/mobilitydata/gtfsvalidator/db/NoticeMemoryConsumptionTest.java
lionel-nj: Nice catch! I'm adapting the code to your advice. Thanks

adapter/repository/in-memory-simple/src/test/java/org/mobilitydata/gtfsvalidator/db/NoticeMemoryConsumptionTest.java
lionel-nj: Okay, thanks. Will modify, however I suggest the private method to throw Error not an Exception since OutOfMemoryError is an Error.

barbeau: @lionel-nj Sounds good!

adapter/repository/in-memory-simple/src/test/java/org/mobilitydata/gtfsvalidator/db/NoticeMemoryConsumptionTest.java
fabrice-v: We will implement 1. right away, and shall open an issue for 2.

adapter/repository/in-memory-simple/src/test/java/org/mobilitydata/gtfsvalidator/db/NoticeMemoryConsumptionTest.java
```diff
@@ -2,23 +2,12 @@
...
-import org.mobilitydata.gtfsvalidator.usecase.notice.base.WarningNotice;
-import org.mobilitydata.gtfsvalidator.usecase.port.ValidationResultRepository;
-
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.List;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
 public class NoticeMemoryConsumptionTest {
```
lionel-nj: This should not be present in this commit, it seems like I am having a bit of trouble with the git commands. Before committing I checked the changes.

adapter/repository/in-memory-simple/src/test/java/org/mobilitydata/gtfsvalidator/db/NoticeMemoryConsumptionTest.java
```diff
@@ -2,23 +2,12 @@
...
 
     private void generateNotices(MockValidationResultRepo mockValidationResultRepo, int numberOfNotices) {
```
lionel-nj: "mockValidationResultRepo" should not be present in this commit, it seems like I am having a bit of trouble with the git commands. Before committing I checked the changes.

adapter/repository/in-memory-simple/src/test/java/org/mobilitydata/gtfsvalidator/db/NoticeMemoryConsumptionTest.java
```diff
@@ -5,7 +5,7 @@
 
 public class NoticeMemoryConsumptionTest {
```
lionel-nj: This is the commit to consider

adapter/repository/in-memory-simple/src/test/java/org/mobilitydata/gtfsvalidator/db/NoticeMemoryConsumptionTest.java
```diff
@@ -0,0 +1,151 @@
...
+
+        InMemoryValidationResultRepository resultRepository = new InMemoryValidationResultRepository();
+
+        generateNotices(resultRepository, 1);
+        resultRepository.getAll();
+
+        long freeMemory = Runtime.getRuntime().freeMemory();  // bytes
+
+        System.out.println(
+                String.format("Generating 100 notices: Total memory: %s megabytes, Free memory: %s megabytes, Used memory: %s megabytes",
```
barbeau: It doesn't seem like this is showing up in the Travis builds (https://github.com/MobilityData/gtfs-validator/runs/519805000). Are we using a logging framework that should be used here instead?

adapter/repository/in-memory-simple/src/test/java/org/mobilitydata/gtfsvalidator/db/NoticeMemoryConsumptionTest.java
```diff
@@ -18,12 +18,25 @@ private void generateNotices(MockValidationResultRepo mockValidationResultRepo,
...
-        generateNotices(mockValidationResultRepo, 100);
-        mockValidationResultRepo.getAll();
+        System.out.println(
+                String.format("Generating 100 notices: Total memory: %s megabytes, Free memory: %s megabytes, Used memory: %s megabytes",
+                        totalMemory / 1_000_000,
+                        freeMemory / 1_000_000,
+                        (totalMemory - freeMemory) / 1_000_000)
+        );
+
+        assert (totalMemory - freeMemory < 455_000_000);
```
lionel-nj: Regarding this 455_000_000  value, I am not sure how the memory process works. Logically generating 100 notices should consume less memory than generating 1000. 
But the order in which the tests are executed makes me wonder how the JVM memory works, does it extend at some point?  On the screenshot, you can see that the test are not executed in the same order they are written. 

After running the test for 2 000 000 notices, it looks like we are using "another memory" (total value increases). I might be wrong thinking that total memory is the total memory allowed by the JVM. The documentation for Runtime.totalMemory() indicates:

>      * Returns the total amount of memory in the Java virtual machine.


Capture d’écran, le 2020-03-19 à 12 22 58

barbeau: Java does dynamically resize memory as it executes (up to a limit imposed by the OS), as well as reclaims member and discards unused objects (garbage collection). This seems to be a good short explanation - https://stackify.com/java-heap-vs-stack/. Note that there is a `maxMemory()` method too which should give you the limit.

For our purposes in benchmarking here, we really care the most about heap memory, which is where objects are stored - we want to measure how much space entities are taking up.  From a quick look at the Java docs, it seems like the methods you're using are for the heap.

Before and after each test you should call `Runtime.gc()` to ask the garbage collector to run and get rid of any unused objects to avoid tests having an impact on each other. You should be able to do this via `@Before` and `@After` annotated methods. Try that and see what the numbers show

lionel-nj: Very interesting explanation, i'll adapt and see what happens. Thanks

adapter/repository/in-memory-simple/src/test/java/org/mobilitydata/gtfsvalidator/db/NoticeMemoryConsumptionTest.java
lionel-nj: We are using LogManager in the main method. Will use this framework here to enable the display of this information in the CI builds

adapter/repository/in-memory-simple/src/test/java/org/mobilitydata/gtfsvalidator/db/NoticeMemoryConsumptionTest.java
lionel-nj: After checking the resource, since we use JUnit5, @Before and @After become @BeforeEach and @AfterEach

lionel-nj:
It seems that tests that pass on my computer do not pass here. The logger used in the last commit does not display the information about the test in the CI / build, when it does on my computer.

I am investigating this, maybe I should use another logging framework since we are in a context of test using Gradle. At least seeing what happened will help debug the test: I might have set the memory limits too low.

barbeau:
Looks like the following tests are failing:

>org.mobilitydata.gtfsvalidator.db.NoticeMemoryConsumptionTest > creationOf10000NoticeShouldNotExceedMemoryLimit() FAILED
    java.lang.AssertionError at NoticeMemoryConsumptionTest.java:96
>...
>org.mobilitydata.gtfsvalidator.db.NoticeMemoryConsumptionTest > creationOf100NoticeShouldNotExceedMemoryLimit() FAILED
    java.lang.AssertionError at NoticeMemoryConsumptionTest.java:54

fabrice-v:
> Looks like the following tests are failing:
> 
> > org.mobilitydata.gtfsvalidator.db.NoticeMemoryConsumptionTest > creationOf10000NoticeShouldNotExceedMemoryLimit() FAILED
> > java.lang.AssertionError at NoticeMemoryConsumptionTest.java:96
> > ...
> > org.mobilitydata.gtfsvalidator.db.NoticeMemoryConsumptionTest > creationOf100NoticeShouldNotExceedMemoryLimit() FAILED
> > java.lang.AssertionError at NoticeMemoryConsumptionTest.java:54

Yeah it puzzles us a little because tests with more notices don't fail. We also can't seem to be able to see the logs.
I suggested to @lionel-nj to see if we can identify the test environment through some environment variable and set things up such as the test would be skipped on GitHub infrastructure.

barbeau:
Have either of you tried running `gradlew test` from the command line?

I can actually reproduce this doing the above on my machine:

image

barbeau:
Actually, I can repro inside IntelliJ too:

image

barbeau:
Also, you can see the log messages by using `gradlew test -i` (see https://docs.gradle.org/current/userguide/logging.html).

barbeau:
Here's the log with the failures:

image

barbeau:
So, looks like:
* `creationOf1000NoticeShouldNotExceedMemoryLimit()` uses 9 MB while test asserts that it's less than 8MB
* `creationOf100NoticeShouldNotExceedMemoryLimit()` uses 7 MB while test asserts that it's less than 7MB

barbeau:
I'm also seeing `creationOf2000000NoticeShouldNotExceedMemoryLimit()` occasionally fail, with 454 MB used but test asserts less than 454 MB used.

One issue here is that JVM memory management isn't guaranteed to be deterministic. For example, `Runtime.gc()` [docs](https://docs.oracle.com/javase/8/docs/api/java/lang/Runtime.html#gc--) say:

>Calling this method suggests that the Java virtual machine expend effort toward recycling unused objects in order to make the memory they currently occupy available for quick reuse. When control returns from the method call, the virtual machine has made its best effort to recycle all discarded objects.

So it's not guaranteed that the garbage collector will run, or that it will reclaim the same amount of memory.

So I'd suggest bumping the memory limits a bit more to allow for this behavior.

fabrice-v:
Ok, I'll let @lionel-nj reply with what he witnessed

lionel-nj:
@barbeau  I did try  to run with gradlew test from the command line, but the bash returns 

> bash: gradlew: command not found

However it works with
>  ./gradlew  test

So I am working with that command instead.

Thanks for your insights, this should be solved by introducing a buffer value (equals to 10%) in the next commit.

adapter/repository/in-memory-simple/src/test/java/org/mobilitydata/gtfsvalidator/db/NoticeMemoryConsumptionTest.java
fabrice-v: #77

adapter/repository/in-memory-simple/build.gradle
```diff
@@ -44,9 +44,14 @@ dependencies {
     implementation 'com.fasterxml.jackson.core:jackson-databind:2.10.1'
     implementation 'com.fasterxml.jackson.dataformat:jackson-dataformat-csv:2.10.1'
 
+
+    implementation 'commons-cli:commons-cli:1.4'
```
fabrice-v: Delete

adapter/repository/in-memory-simple/src/test/java/org/mobilitydata/gtfsvalidator/db/NoticeMemoryConsumptionTest.java
```diff
@@ -0,0 +1,163 @@
...
+    public void cleanMemoryAfterTest() {
+        System.gc();
+    }
+
+    @Test
+    public void creationOf100NoticeShouldNotExceedMemoryLimit() {
+
+        InMemoryValidationResultRepository resultRepository = new InMemoryValidationResultRepository();
+
+        generateNotices(resultRepository, 1);
```
fabrice-v: should be a hundred

adapter/repository/in-memory-simple/build.gradle
fabrice-v: Should not be here

adapter/repository/in-memory-simple/src/test/java/org/mobilitydata/gtfsvalidator/db/NoticeMemoryConsumptionTest.java
fabrice-v: Should be a 100

adapter/repository/in-memory-simple/src/test/java/org/mobilitydata/gtfsvalidator/db/NoticeMemoryConsumptionTest.java
```diff
@@ -0,0 +1,163 @@
...
+        resultRepository.getAll();
+
+        long totalMemory = Runtime.getRuntime().totalMemory(); // bytes
+        long freeMemory = Runtime.getRuntime().freeMemory();  // bytes
+
+        LOGGER.info(String.format("Generating 100 notices: Total memory: %s megabytes, Free memory: %s megabytes," +
```
fabrice-v: Should be a method with 3 parameters
`noticeCount`
`totalMemoryInBytes`
`freeMemoryInBytes`

adapter/repository/in-memory-simple/src/test/java/org/mobilitydata/gtfsvalidator/db/NoticeMemoryConsumptionTest.java
```diff
@@ -0,0 +1,163 @@
...
+
+        assert (totalMemory - freeMemory < 28_000_000 * BUFFER);
+    }
+
+    @Test
+    public void creationOf1000000NoticeShouldNotExceedMemoryLimit() {
+
+        InMemoryValidationResultRepository resultRepository = new InMemoryValidationResultRepository();
+
+        generateNotices(resultRepository, CONVERSION_FACTOR);
```
fabrice-v: Bad search and replace ;p

lionel-nj: God catch, modifying this. Thanks

lionel-nj:
merging this branch into master following @fabrice-v https://github.com/MobilityData/gtfs-validator/issues/63#issuecomment-607509324

lionel-nj:
Last but not least: tests for notice consumption seem to not work on Github, meanwhile they pass my computer when using commands ./gradlew test or ./gradlew check -i

PR #70, opened 2020-03-19
Shape entities -- internal representation

lionel-nj:
**Summary:**

This PR relates to processing of entities from shapes.txt. The goal is to use the required use cases implemented via PR #44 that allow to go from a ParsedEntity representing a row from shapes.txt to a specific concrete internal representation.

See #44 

**Expected behavior:** 

**TODO:**
- [x] add javadoc comments 

- [ ] ~Format the title like "Fix #<issue_number> - <short description of fix and changes>" (for example - "Fix #1111 - Check for null value before using field")~


domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/Shape.java
```diff
@@ -0,0 +1,102 @@
...
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class Shape {
+
+    @NotNull
+    private final String shapeId;
+
+    @NotNull
+    private final Float shapePtLat;
```
fabrice-v: Replace with native `float` type instead of using object type

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/Shape.java
```diff
@@ -0,0 +1,102 @@
...
+
+    @NotNull
+    private final Float shapePtLon;
```
fabrice-v: Replace with native `float` type instead of using object type

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/Shape.java
```diff
@@ -0,0 +1,102 @@
...
+
+    @NotNull
+    private final Integer shapePtSequence;
```
fabrice-v: Replace with native `int` type instead of using object type

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/Shape.java
```diff
@@ -0,0 +1,102 @@
...
+
+    private final int shapeDistTraveled;
```
fabrice-v: - should be a floating point
- should use `Float` as it is nullable

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/Shape.java
```diff
@@ -0,0 +1,102 @@
...
+
+
+    public Shape(@NotNull String shapeId,
```
fabrice-v: Make private to enforce Builder pattern

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/Shape.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    public int getShapeDistTraveled() {
+        return shapeDistTraveled;
+    }
+
+    public static class ShapeBuilder {
+
+        private String shapeId;
+        private Float shapePtLat;
```
fabrice-v: `float`

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/Shape.java
```diff
@@ -0,0 +1,102 @@
...
+        private Float shapePtLon;
```
fabrice-v: `float`

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/Shape.java
```diff
@@ -0,0 +1,102 @@
...
+        private Integer shapePtSequence;
+        private Integer shapeDistTraveled;
```
fabrice-v: `Float`

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/Shape.java
```diff
@@ -0,0 +1,102 @@
...
+    public int getShapeDistTraveled() {
+        return shapeDistTraveled;
+    }
+
+    public static class ShapeBuilder {
+
+        private String shapeId;
+        private Float shapePtLat;
+        private Float shapePtLon;
+        private Integer shapePtSequence;
```
fabrice-v: `int`

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/Shape.java
```diff
@@ -0,0 +1,102 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.mobilitydata.gtfsvalidator.domain.entity;
+
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class Shape {
```
barbeau: This class will eat up memory quickly due to the number of records in `shapes.txt` - I opened an issue discussing how to optimize memory usage at https://github.com/MobilityData/gtfs-validator/issues/79.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/Shape.java
```diff
@@ -0,0 +1,121 @@
...
+    public Float getShapePtLat() {
+        return shapePtLat;
+    }
+
+    @NotNull
+    public Float getShapePtLon() {
+        return shapePtLon;
+    }
+
+    @NotNull
```
barbeau: GTFS shapes.txt isn't guaranteed to be ordered, so at some point we'll need to sort `Shape` objects - grouped by `shapeId`, ordered by increasing `shapePtSequence`. To support this you may want to implement `Comparable` - see https://www.geeksforgeeks.org/comparable-vs-comparator-in-java/ for an example. We can revisit this later unless you have a good idea how this entity will be used within business validation rules now.

lionel-nj: I keep this in mind, and will think of this implementation when adding shape object to the Gtfs Data repository.

Opening an issue on this matter though.

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ProcessParsedShapePointTest.java
```diff
@@ -0,0 +1,195 @@
...
+        final ArgumentCaptor<DuplicatedEntityNotice> captor = ArgumentCaptor.forClass(DuplicatedEntityNotice.class);
+
+        verify(mockResultRepo, times(1)).addNotice(captor.capture());
+
+        final List<DuplicatedEntityNotice> noticeList = captor.getAllValues();
+
+        assertEquals("shapes.txt", noticeList.get(0).getFilename());
+        assertEquals(SHAPE_ID, noticeList.get(0).getNoticeSpecific(Notice.KEY_FIELD_NAME));
+        assertEquals("shape id", noticeList.get(0).getEntityId());
+
```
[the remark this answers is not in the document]
lionel-nj: Works for the two previous tests, but not in this one since there is an interaction with this mock.

See https://github.com/MobilityData/gtfs-validator/blob/59cebcf5b878ee41d21274e7896516039851c640/usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ProcessParsedShapePointTest.java#L168

adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/db/InMemoryGtfsDataRepository.java
```diff
@@ -482,4 +486,43 @@ public FareRule getFareRule(final String fareId, final String routeId, final Str
...
+    /**
+     * Add a shape point representing a row from shapes.txt to this. Return the entity added to the repository if not
+     * null, otherwise returns null.
+     *
+     * @param newShapePoint the internal representation of a row from shapes.txt to be added to the repository.
+     * @return the entity added to the repository if not null, otherwise returns null.
+     * @throws IllegalArgumentException if the shape point passed as argument is null
+     */
+    @Override
+    public ShapePoint addShapePoint(final ShapePoint newShapePoint) throws IllegalArgumentException {
```
fabrice-v, line 499: Here again, maybe the distinction between what's a Shape and what is a ShapePoint could be made clearer

adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/db/InMemoryGtfsDataRepository.java
```diff
@@ -482,4 +486,43 @@ public FareRule getFareRule(final String fareId, final String routeId, final Str
...
+        if (newShapePoint != null) {
+            final String shapeId = newShapePoint.getShapeId();
+            if (shapePerId.containsKey(shapeId)) {
+                if (shapePerId.get(shapeId).contains(newShapePoint)) {
```
fabrice-v, line 503: !! be wary of how `contains` behave. I suppose it does a filed by field comparison but that should be checked. Also it seems we are missing a test where trying to add twice two different ShapePoint *instances* that specify the same id and data.

adapter/repository/in-memory-simple/src/test/java/org/mobilitydata/gtfsvalidator/db/InMemoryGtfsDataRepositoryTest.java
```diff
@@ -447,4 +443,49 @@ void addFareRuleAndGetFareRuleShouldReturnSameEntity() {
...
+    void callToAddShapeShouldSameReturnEntity() {
+        final ShapePoint mockShapePoint = mock(ShapePoint.class);
+        when(mockShapePoint.getShapeId()).thenReturn("test id");
+        final InMemoryGtfsDataRepository underTest = new InMemoryGtfsDataRepository();
+
+        assertEquals(mockShapePoint, underTest.addShapePoint(mockShapePoint));
+    }
+
+    @Test
+    void addTwiceSameShapePointShouldReturnNull() {
```
fabrice-v, line 458: this only test adding a single instance twice. We could have input data that contains twice the same row, that would be represented by two ShapePoint instances containing the same data. I suggest we add another test for that case. In our other entity classes we don't need that because checking IDs is sufficient.

config/src/main/java/org/mobilitydata/gtfsvalidator/config/DefaultConfig.java
```diff
@@ -237,6 +233,10 @@ public ProcessParsedFareRule processParsedFareRule() {
         return new ProcessParsedFareRule(resultRepo, gtfsDataRepository, new FareRule.FareRuleBuilder());
     }
 
+    public ProcessParsedShapePoint processParserShapePoint() {
```
fabrice-v, line 236: Typo in the name, should be `Parsed` instead of `Parser`

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/ShapePoint.java
```diff
@@ -0,0 +1,240 @@
...
+import org.mobilitydata.gtfsvalidator.domain.entity.notice.error.MissingRequiredValueNotice;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Class for all entities defined in shapes.txt. Can not be directly instantiated: user must use the
+ * {@link ShapePoint.ShapeBuilder} to create this.
+ */
+public class ShapePoint extends GtfsEntity implements Comparable<ShapePoint> {
```
fabrice-v, line 33: I guess I have my answer to previous comments regarding how the contains work

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/ShapePoint.java
```diff
@@ -0,0 +1,240 @@
...
+     *
+     * @param shapePoint shape to compare to
+     * @return the value {@code 0} if this {@code Shape} field shape_pt_sequence is equal to field shape_pt_sequence of
+     * argument; a value less than {@code 0} if this field shape_pt_sequence is numerically less than field
+     * shape_pt_sequence of argument; and a value greater than {@code 0} if this field shape_pt_sequence is numerically
+     * greater than field shape_pt_sequence of argument
+     */
+    @Override
+    public int compareTo(@NotNull final ShapePoint shapePoint) {
+        return getShapePtSequence().compareTo(shapePoint.getShapePtSequence());
```
fabrice-v, line 104: It seems we are comparing Integers here, so it seems calling compareTo as opposed to a plain comparison makes the code harder to read for not much of a benefit. Thoughts?

barbeau: I'd be ok leaving it as-is and depending on the `Integer.compareTo()` implementation - otherwise we're just duplicating the same if/than/else clause that's already implemented.

adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/db/InMemoryGtfsDataRepository.java
barbeau: This `.contains()` is on the `ArrayList`, right? So the time complexity here is O(n) to find out if the new shape point exists for this shape_id. Given that shapes.txt can be huge, this is very expensive.

I'd strongly suggest a more efficient solution with `TreeMap` instead:
https://docs.oracle.com/javase/8/docs/api/java/util/TreeMap.html

That will give you O(log n) time complexity for `.contains()`, and you can still get sorted shape points back.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/ShapePoint.java
```diff
@@ -0,0 +1,240 @@
...
+         * are met. Otherwise, method returns an entity representing a list of notices.
+         */
+        public EntityBuildResult<?> build() {
+            noticeCollection.clear();
+
+            // suppressed warning regarding nullability of fields shapePtLat, shapePtLon, and shapePtSequence since
+            // these fields could be provided as null values
+            //noinspection ConstantConditions to avoid lint
+            if (shapeId == null ||
+                    shapePtLat == null || (shapePtLat != null && (shapePtLat < -90.0f || shapePtLat > 90.0f)) ||
```
barbeau, line 201: The `shapePtLat < -90.0f || shapePtLat > 90.0f` and `shapePtLon < -180.0f || shapePtLon > 180.0f` should be encapsulated in their own methods for `isValidLatitude()` and `isValidLongitude()` accordingly. You'll need these for stops.txt too.

adapter/repository/in-memory-simple/src/test/java/org/mobilitydata/gtfsvalidator/db/InMemoryGtfsDataRepositoryTest.java
```diff
@@ -447,4 +443,49 @@ void addFareRuleAndGetFareRuleShouldReturnSameEntity() {
...
+    @Test
+    void addNullShapePointShouldThrowException() {
+        final InMemoryGtfsDataRepository underTest = new InMemoryGtfsDataRepository();
+        final Exception exception = assertThrows(IllegalArgumentException.class,
+                () -> underTest.addShapePoint(null));
+        assertEquals("Cannot add null shape point to data repository", exception.getMessage());
+    }
+
+    @Test
+    void getShapeByIdShouldReturnRelatedListOfShapePoint() {
```
barbeau, line 476: There should be additional tests to test adding points with `shape_pt_sequence` out of order, and ensuring that you get them back in-order.

adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/db/InMemoryGtfsDataRepository.java
barbeau: And as implied above, we should store these points already sorted in the repo (see my comment later for adding a test for this).

adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/db/InMemoryGtfsDataRepository.java
```diff
@@ -80,6 +81,9 @@
     // Example of key after composition: fare_idroute_idorigin_iddestination_idcontains_id
     private final Map<String, FareRule> fareRuleCollection = new HashMap<>();
 
+    // Map containing Shape Entities. Entities are mapped on the value found in column shape_id of GTFS file shapes.txt
+    private final Map<String, ArrayList<ShapePoint>> shapePerId = new HashMap<>();
```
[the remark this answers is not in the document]
lionel-nj: True, modifying this.

adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/db/InMemoryGtfsDataRepository.java
```diff
@@ -482,4 +487,45 @@ public FareRule getFareRule(final String fareId, final String routeId, final Str
         return fareRuleCollection.get(FareRule.getFareRuleMappingKey(fareId, routeId, originId, destinationId,
                 containsId));
     }
+
+    /**
+     * Add a shape point to a shape. A shape is a list of shape points whereas a ShapePoint represents a row from
+     * shapes.txt to this. Return the entity added to the repository if not null, otherwise returns null.
```
barbeau, line 493: I would re-word this:

>Return the entity added to the repository if not null, otherwise returns null.

to better describe the exact behavior:

>Return the entity added to the repository if the entity was successfully added, and returns null if the provided newShapePoint already exists in the repository.

adapter/repository/in-memory-simple/src/test/java/org/mobilitydata/gtfsvalidator/db/InMemoryGtfsDataRepositoryTest.java
```diff
@@ -447,4 +445,108 @@ void addFareRuleAndGetFareRuleShouldReturnSameEntity() {
...
+    void addTwiceSameShapePointShouldReturnNull() {
+        final ShapePoint mockShapePoint = mock(ShapePoint.class);
+        when(mockShapePoint.getShapeId()).thenReturn("test id");
+        when(mockShapePoint.getShapePtLat()).thenReturn(50f);
+        when(mockShapePoint.getShapePtLon()).thenReturn(100f);
+        when(mockShapePoint.getShapePtSequence()).thenReturn(4);
+        when(mockShapePoint.getShapeDistTraveled()).thenReturn(56f);
+        final InMemoryGtfsDataRepository underTest = new InMemoryGtfsDataRepository();
+
+        underTest.addShapePoint(mockShapePoint);
```
barbeau, line 459: I would also pull the shape you just inserted into the repo back out again and then assert that the field values for the shape equals the one you put in - that way it's an end-to-end test.

adapter/repository/in-memory-simple/src/test/java/org/mobilitydata/gtfsvalidator/db/InMemoryGtfsDataRepositoryTest.java
```diff
@@ -447,4 +445,108 @@ void addFareRuleAndGetFareRuleShouldReturnSameEntity() {
...
+        final ShapePoint mockShapePoint01 = mock(ShapePoint.class);
+        when(mockShapePoint01.getShapeId()).thenReturn("test id01");
+        when(mockShapePoint01.getShapePtSequence()).thenReturn(8);
+
+        final InMemoryGtfsDataRepository underTest = new InMemoryGtfsDataRepository();
+
+        underTest.addShapePoint(mockShapePoint00);
+        underTest.addShapePoint(mockShapePoint01);
+
+        final TreeMap<Integer, ShapePoint> firstMapToCheck = new TreeMap<>();
```
barbeau, line 508: Shouldn't this use the repo instead of instatiating your own TreeMaps here?

adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/db/InMemoryGtfsDataRepository.java
```diff
@@ -482,4 +487,45 @@ public FareRule getFareRule(final String fareId, final String routeId, final Str
...
+    public ShapePoint addShapePoint(final ShapePoint newShapePoint) throws IllegalArgumentException {
+        if (newShapePoint != null) {
+            final String shapeId = newShapePoint.getShapeId();
+            if (shapePerIdShapePtSequence.containsKey(shapeId)) {
+                if (shapePerIdShapePtSequence.get(shapeId).containsKey(newShapePoint.getShapePtSequence())) {
+                    return null;
+                }
+                shapePerIdShapePtSequence.get(shapeId).put(newShapePoint.getShapePtSequence(), newShapePoint);
+            } else {
+                final TreeMap<Integer, ShapePoint> innerMap = new TreeMap<>();
```
barbeau, line 509: Nit:

```suggestion
                final Map<Integer, ShapePoint> innerMap = new TreeMap<>();
```

adapter/repository/in-memory-simple/src/test/java/org/mobilitydata/gtfsvalidator/db/InMemoryGtfsDataRepositoryTest.java
```diff
@@ -447,4 +445,108 @@ void addFareRuleAndGetFareRuleShouldReturnSameEntity() {
...
+
+        final TreeMap<Integer, ShapePoint> secondMapToCheck = new TreeMap<>();
+        secondMapToCheck.put(8, mockShapePoint01);
+
+        assertEquals(firstMapToCheck, underTest.getShapeById("test id00"));
+        assertEquals(secondMapToCheck, underTest.getShapeById("test id01"));
+    }
+
+    @Test
+    void addShapePointShouldMaintainOrder() {
```
barbeau, line 519: 👍

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/ShapePoint.java
```diff
@@ -0,0 +1,258 @@
...
+         * @param shapeDistTraveled actual distance traveled along the shape from the first shape point to the point
+         *                          specified in this record
+         * @return builder for future object creation
+         */
+        public ShapeBuilder shapeDistTraveled(@Nullable final Float shapeDistTraveled) {
+            this.shapeDistTraveled = shapeDistTraveled;
+            return this;
+        }
+
+        /**
```
barbeau, line 187: I left this comment on another entity in a different PR - if the `Builder` here is used as a singleton (single instance), it seems like there should be a `Builder.clear()` method that resets all the fields. Otherwise you're hoping that the calling class will know to reset all of the fields manually themselves, or always call all of the builder fields and pass in `null` values when the field is not provided.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/ShapePoint.java
```diff
@@ -0,0 +1,258 @@
...
+                        shapeDistTraveled));
+            }
+        }
+
+        /**
+         * Utility method to determine the range validity of latitude
+         * @return true if the shapePtLat is valid, otherwise returns false
+         */
+        private boolean isValidLatitude() {
+            return shapePtLat > -90.0f && shapePtLat < 90.0f;
```
barbeau, line 247: Should be inclusive:

```suggestion
            return shapePtLat >= -90.0f && shapePtLat <= 90.0f;
```

Tests need to be updated too, if there are some.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/ShapePoint.java
```diff
@@ -0,0 +1,258 @@
...
+        }
+
+        /**
+         * Utility method to determine the range validity of longitude
+         * @return true if shapePtLon is valid, otherwise returns false
+         */
+        private boolean isValidLongitude() {
+            return shapePtLon > -180.0f && shapePtLon < 180.0f;
```
barbeau, line 255: Same here - should be inclusive

adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/db/InMemoryGtfsDataRepository.java
```diff
@@ -482,4 +487,45 @@ public FareRule getFareRule(final String fareId, final String routeId, final Str
...
+                shapePerIdShapePtSequence.put(shapeId, innerMap);
+            }
+            return newShapePoint;
+        } else {
+            throw new IllegalArgumentException("Cannot add null shape point to data repository");
+        }
+    }
+
+    /**
+     * Return the list of shape points from shapes.txt related to the id provided as parameter; which represents a shape
```
barbeau, line 520: I would add that the returned list is ordered by shape point sequence.

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/port/GtfsDataRepository.java
```diff
@@ -71,4 +72,8 @@
 
     FareRule getFareRule(final String fareId, final String routeId, final String originId, final String destinationId,
                          final String containsId);
+
+    ShapePoint addShapePoint(final ShapePoint newShapePoint) throws IllegalArgumentException;
+
+    TreeMap<Integer, ShapePoint> getShapeById(final String shapeId);
```
barbeau, line 78: I would add Javadocs here saying that the returned map is ordered by shape point sequence.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/ShapePoint.java
lionel-nj: @fabrice-v, opening a issue on this matter since all gtfs entities will be affected by such change.

adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/db/InMemoryGtfsDataRepository.java
lionel-nj: This requires to cast innerMap as a TreeMap at line 513511 of this same file. Is it advisable?
See:
https://github.com/MobilityData/gtfs-validator/blob/7dcadaa7cc7d71cc04ecb6518d9ea72e26e10792/adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/db/InMemoryGtfsDataRepository.java#L511

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/port/GtfsDataRepository.java
lionel-nj: Not against it, but wouldn't that duplicate the javadoc written in `InMemoryGtfsDataRepository`?
Also, to ensure consistency if we add javadoc here, we should add javadoc for all methods of this interface (maybe to all interface?).

adapter/repository/in-memory-simple/src/test/java/org/mobilitydata/gtfsvalidator/db/InMemoryGtfsDataRepositoryTest.java
lionel-nj: Absolutely, I missed that point. Thank you!

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/ShapePoint.java
lionel-nj: Actually these methods are not directly tested. However, the result they produce In `build` method is tested

lionel-nj: Do you think that these methods should be `public` or`protected` so that we can individually verify there implementation?

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/ShapePoint.java
lionel-nj: See https://github.com/MobilityData/gtfs-validator/issues/290

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/ShapePoint.java
barbeau: I think these should be `public` methods centralized somewhere (a utility class?), as we know we'll need them for `stops.txt` too.

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/port/GtfsDataRepository.java
barbeau: Yes, but for an interface if there are implementation assumptions like an ordered map that should be documented in the interface too. Otherwise later someone may come along and implement the interface again, but not realized that there is an assumption that the method returns ordered results. `TreeMap` implies this, but someone could generalize this to `Map` at some point as well, also not realizing that order matters.

In other projects I've ended up duplicating Javadocs across interfaces and implementations in some cases. It's most important when there are assumptions that may not be obvious.

adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/db/InMemoryGtfsDataRepository.java
```diff
@@ -522,4 +525,51 @@ public Pathway addPathway(final Pathway newPathway) throws IllegalArgumentExcept
...
+    /**
+     * Return the collection of shape points from shapes.txt related to the id provided as parameter; which represents a
+     * shape object. The returned collection is ordered by shape_pt_sequence.
+     *
+     * @param shapeId the key from shapes.txt related to the Route to be returned
+     * @return  the collection of shape points from shapes.txt related to the id provided as parameter; which represents
+     * a shape object. The returned collection is ordered by shape_pt_sequence.
+     */
+    @Override
+    public Collection<ShapePoint> getShapeById(final String shapeId) {
```
lionel-nj, line 572: Here, I wonder if we should query ShapePoint on their shape_pt_sequence (this would required this method to return a TreeMap), or keep the collection of ShapePoint as a list

adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/db/InMemoryGtfsDataRepository.java
```diff
@@ -522,4 +525,51 @@ public Pathway addPathway(final Pathway newPathway) throws IllegalArgumentExcept
...
+        return new ArrayList<>(shapePerIdShapePtSequence.get(shapeId).values());
```
lionel-nj, line 573: Here, I wonder if we should query ShapePoint on their shape_pt_sequence (this would required this method to return a TreeMap), or keep the collection of ShapePoint as a list

barbeau: I think the main use case for this method will be to get a sequence of points (the "shape") given a shape ID, so I don't think we need to query on shape_pt_sequence here? Unless I'm misunderstanding the question.

I think there are two other items to discuss here (and generally for the repo for all entities), though:
1. Should we be returning a List vs Map? The former (as implemented) prevents calling code from doing lookups based on shape_pt_sequence - maybe that's what you're saying above @lionel-nj?
1. Should we be doing a deep copy here (as currently implemented) or can we use `Collections.unmodifiableMap()` to return an unmodifiable view of this (and avoid copying it)?

lionel-nj: Thanks, yes that answers my question.
We can discuss the points that you mentioned in our next meeting.

lionel-nj: [PR#88 ](https://github.com/MobilityData/gtfs-validator/pull/88), is linked to this same discussion also.

lionel-nj: As discussed, we will implement a solution using  `Collections.unmodifiableMap()` to avoid data duplication.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/ShapePoint.java
lionel-nj: As discussed, range verification is already performed at the parsing stage. Therefore, there is no need for these checks here. Removing this duplicate code.
Also, the utility class will be implemented in another PR.

adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/db/InMemoryGtfsDataRepository.java
```diff
@@ -102,6 +102,10 @@
     // Example of key after composition: attribution_idagency_idroute_idtrip_idorganization_nameis_produceris_operatoris_authorityattribution_urlattribution_emailattribution_phone
     private final Map<String, Attribution> attributionCollection = new HashMap<>();
 
+    // Map containing Shape Entities. A shape is a actually a collection ShapePoint.
```
fabrice-v, line 105:
```suggestion
    // Map containing Shape Entities. A shape is a actually a collection of ShapePoint.
```

adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/db/InMemoryGtfsDataRepository.java
```diff
@@ -610,4 +614,52 @@ public Attribution getAttribution(final String attributionId, final String agenc
                 organizationName,isProducer, isOperator, isAuthority, attributionUrl, attributionEmail,
                 attributionPhone));
     }
+
+    /**
+     * Add a {@link ShapePoint} to a shape. A shape is a list of{@link ShapePoint} whereas a {@link ShapePoint}
+     * represents a row from shapes.txt to this. Return the entity added to the repository if the entity was
```
fabrice-v, line 620:
```suggestion
     * represents a row from shapes.txt. Return the entity added to the repository if the entity was
```

PR #71, opened 2020-03-19
Transfer entities -- internal representation

lionel-nj:
Summary:

This PR relates to processing of entities from transfers.txt. The goal is to use the required use cases implemented via PR #44 that allow to go from a ParsedEntity representing a row from transfers.txt to a specific concrete internal representation.

See #44

**Expected behavior:** 

**TODO:**


domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/Transfer.java
```diff
@@ -0,0 +1,87 @@
...
+
+    @NotNull
+    private final String toStopId;
+
+    @NotNull
+    private final TransferType transferType;
+
+    private final int minTransferTime;
+
+    public Transfer(@NotNull String fromStopId,
```
fabrice-v: Make private to enforce usage of Builder

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/Transfer.java
```diff
@@ -0,0 +1,102 @@
...
+    @NotNull
+    private final String fromStopId;
+
+    @NotNull
+    private final String toStopId;
+
+    @NotNull
+    private final TransferType transferType;
+
+    private final int minTransferTime;
```
barbeau: Should this be `@Nullable`?

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/Transfer.java
```diff
@@ -0,0 +1,102 @@
...
+
+    private Transfer(@NotNull final String fromStopId,
+                     @NotNull final String toStopId,
+                     @NotNull final TransferType transferType,
+                     int minTransferTime) {
```
barbeau: See above

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/Transfer.java
```diff
@@ -0,0 +1,102 @@
...
+            this.fromStopId = fromStopId;
+            return this;
+        }
+
+        public TransferBuilder toStopId(@NotNull final String toStopId) {
+            this.toStopId = toStopId;
+            return this;
+        }
+
+        public TransferBuilder minTransferTime(@NotNull final TransferType transferType) {
```
barbeau: This method should be named `transferType()`

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/Transfer.java
lionel-nj: From what I witnessed, IntelliJ recommends to not annotate `@Nullable` primitives such as int. I had a similar discussion about object types with @fabrice-v .

lionel-nj: Indeed since it's optional, this should be annotated as `@Nullable`. Therefore, we'll use object type Integer instead of primitive int to allow use of such annotation.

barbeau: Oh, right, I missed that this was an `int`. But then how are you checking if you should validate this field? If other words, later in the pipeline, how do you know if you should check the optional field with some kind of rule?

barbeau: Ok, seeing your edited response now. Yes, if you switch to the object representation that works.

adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/db/InMemoryGtfsDataRepository.java
```diff
@@ -98,4 +100,47 @@ public Route addRoute(@NotNull final Route newRoute) throws IllegalArgumentExcep
     public Route getRouteById(final String routeId) {
         return routeCollection.get(routeId);
     }
+
+    /**
+     * Add a Transfer representing a row from transfers.txt to this. Return the entity added to the repository if the
```
barbeau, line 105: I think there is a word missing at the end of the first sentence here - `transfers.txt to this`

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/transfers/Transfer.java
```diff
@@ -0,0 +1,180 @@
...
+                if (toStopId == null) {
+                    noticeCollection.add(new MissingRequiredValueNotice("transfers.txt",
+                            "to_stop_id", fromStopId + ";" + toStopId));
+                }
+                if (!TransferType.isEnumValueValid(originalTransferTypeInteger)) {
+                    noticeCollection.add(new UnexpectedEnumValueNotice("transfers.txt",
+                            "transfer_type", fromStopId + ";" + toStopId,
+                            originalTransferTypeInteger));
+                }
+                if (minTransferTime != null && minTransferTime < 0) {
```
barbeau, line 169: I don't think `0` is a valid value for `minTransferTime` either. The [spec says](https://gtfs.org/reference/static#transferstxt):

>Amount of time, in seconds, that must be available to permit a transfer between routes at the specified stops. The min_transfer_time should be sufficient to permit a typical rider to move between the two stops, including buffer time to allow for schedule variance on each route.

So it seems like it should be 1 at a minimum. This field is also a candidate for another threshold that could be configurable for what the lower possible value is that makes sense. For example, I would like that most min_transfer_times are greater than 30 seconds, but I don't know that for sure.

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/port/GtfsDataRepository.java
```diff
@@ -27,4 +28,8 @@
     Route addRoute(final Route newRoute) throws IllegalArgumentException;
```
lionel-nj, line 28: Bad merge here I think, the order "add, get" will be re-established in the next commit

adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/db/InMemoryGtfsDataRepository.java
```diff
@@ -98,4 +100,47 @@ public Route addRoute(@NotNull final Route newRoute) throws IllegalArgumentExcep
...
+    }
+
+    /**
+     * Return the Transfer representing a row from transfers.txt related to the composite key provided as parameter
+     *
+     * @param fromStopId first part of the composite key: identifies a stop or station where a connection between
+     *                   routes begins
+     * @param toStopId   second part of the composite key: identifies a stop or station where a connection between
+     *                   routes ends
+     * @return the Transfer representing a row from transfers.txt related to the composite key provided as parameter
```
barbeau, line 140: I would suggest adding a more explicit note to this Javadoc saying that directionality of the transfer matters - in other words, querying fromStopId of `1` will only give you records where `1` is in the from_stop_id column, and will not return any records where `1` is in the to_stop_id column but not in the ftom_stop_id column.

adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/db/InMemoryGtfsDataRepository.java
```diff
@@ -98,4 +100,47 @@ public Route addRoute(@NotNull final Route newRoute) throws IllegalArgumentExcep
...
+     * from_stop_id and to_stop_id is respected, if this requirement is not met, returns null.
+     */
+    @Override
+    public Transfer addTransfer(final Transfer newTransfer) throws IllegalArgumentException {
+        if (newTransfer != null) {
+            if ((transferCollection.containsKey(newTransfer.getFromStopId())) &&
+                    (transferCollection.get(newTransfer.getFromStopId()).containsKey(newTransfer.getToStopId()))) {
+                return null;
+            } else {
+                String fromStopId = newTransfer.getFromStopId();
```
barbeau, line 120: There needs to be a check if `transferCollection.get(fromStopId)` exists (and if so, adding the `toStopId` to that existing `Map`) before doing `transferCollection.put(fromStopId, innerMap)`. Otherwise you're always overwriting the existing map for the given `fromStopId`, and the max size of the inner map will always be 1.

adapter/repository/in-memory-simple/src/test/java/org/mobilitydata/gtfsvalidator/db/InMemoryGtfsDataRepositoryTest.java
```diff
@@ -113,4 +114,53 @@ void getRouteByIdShouldReturnRelatedRoute() {
...
+        when(mockTransfer00.getToStopId()).thenReturn("stop id 1");
+
+        when(mockTransfer01.getFromStopId()).thenReturn("stop id 1");
+        when(mockTransfer01.getToStopId()).thenReturn("stop id 0");
+
+        underTest.addTransfer(mockTransfer00);
+        underTest.addTransfer(mockTransfer01);
+
+        assertEquals(mockTransfer00, underTest.getTransferByStopPair("stop id 0", "stop id 1"));
+        assertEquals(mockTransfer01, underTest.getTransferByStopPair("stop id 1", "stop id 0"));
```
barbeau, line 164: This test needs to be expanded to include multiple transfers per stop - it's missing a current bug when adding the record and more than one transfer for a stop exists.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/transfers/Transfer.java
lionel-nj: Good point, modifying this keeping in mind that the field could be configurable (via a .json file for example). 
Opening an issue on this matter for the future.

adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/db/InMemoryGtfsDataRepository.java
lionel-nj: Will modify

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/transfers/Transfer.java
lionel-nj: See https://github.com/MobilityData/gtfs-validator/issues/181

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/transfers/Transfer.java
```diff
@@ -0,0 +1,178 @@
...
+                    noticeCollection.add(new MissingRequiredValueNotice("transfers.txt",
+                            "to_stop_id", fromStopId + ";" + toStopId));
+                }
+                if (!TransferType.isEnumValueValid(originalTransferTypeInteger)) {
+                    noticeCollection.add(new UnexpectedEnumValueNotice("transfers.txt",
+                            "transfer_type", fromStopId + ";" + toStopId,
+                            originalTransferTypeInteger));
+                }
+                if (minTransferTime != null && minTransferTime < 1) {
+                    // here minTransferTime threshold should be user configurable
```
fabrice-v, line 167: Yes, let's do it!

domain/src/test/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/transfers/TransferTest.java
```diff
@@ -0,0 +1,145 @@
...
+
+        assertEquals("transfers.txt", notice.getFilename());
+        assertEquals("transfer_type", notice.getFieldName());
+        assertEquals("stop id 0;stop id 1", notice.getEntityId());
+        assertEquals("55", notice.getEnumValue());
+        assertEquals(1, noticeCollection.size());
+    }
+
+    @Test
+    void createTransferWithInvalidMinTransferTimeValueShouldGenerateNotice() {
```
fabrice-v, line 110: As discussed, range validation is already performed at type validation stage. This is more of a sucpicious values validation. Thresholds should come from configuration file and lead to emission of warning notices.

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/port/GtfsDataRepository.java
```diff
@@ -27,4 +28,8 @@
...
 
     Route getRouteById(final String routeId);
+
+    Transfer addTransfer(final Transfer newTransfer) throws IllegalArgumentException;
+
+    Transfer getTransferByStopPair(final String fromStopId, final String toStopId);
```
fabrice-v, line 34: Rename suggestion : `getTransferByStop**Id**Pair`

domain/src/test/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/transfers/TransferTest.java
lionel-nj: Related issue: https://github.com/MobilityData/gtfs-validator/issues/159#issuecomment-626788425

lionel-nj: It seems that we might have neglected the fact that Transfer is a domain entity: it has no access to the user configurable thresholds contained in the ExecParamRepo content (adapter layer). 

In this particular case, when building the Transfer entity, we already checked that  minTransferTime >0). Hence, it might be reasonable to not check the "suspiciousness" of this field inside the build method, and let to be checked in a future use case, that will be able to link the ExecParamRepo and the Transfer entity.

lionel-nj: After discussion, the use case ProcessParsedTransfer can be used to pass the values (retrieved from the ExecParamRepo) to the builder build method.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/transfers/Transfer.java
lionel-nj: This field has been made user configurable in 373cc25fbeb4ca9a8491885f7707a02e51591efe

config/src/main/resources/default-execution-parameters.json
```diff
@@ -5,5 +5,7 @@
   "proto": false,
   "url": null,
   "zipinput": null,
-  "exclude": null
+  "exclude": null,
+  "mintransfertime": 1,
```
fabrice-v, line 9: I think we should adopt a more excplicit format for our parameters name like filename__fieldname__qualifier
In that case `transfers__min_transfer_time__min` and `transfers__min_transfer_time__max`

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/transfers/Transfer.java
```diff
@@ -73,7 +73,7 @@ public TransferType getTransferType() {
         return transferType;
     }
 
-    @SuppressWarnings("unused")
+    @SuppressWarnings("unused") // to avoid lint
```
fabrice-v, line 76: The comment should explain why it's ok to suppress the lint

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/transfers/Transfer.java
```diff
@@ -163,11 +163,11 @@ public TransferBuilder minTransferTime(@Nullable final Integer minTransferTime)
                             "transfer_type", fromStopId + ";" + toStopId,
                             originalTransferTypeInteger));
                 }
-                if (minTransferTime != null && minTransferTime < 1) {
-                    // here minTransferTime threshold should be user configurable
-                    noticeCollection.add(new IntegerFieldValueOutOfRangeNotice("transfers.txt",
-                            "min_transfer_time", fromStopId + ";" + toStopId, 0,
-                            Integer.MAX_VALUE, minTransferTime));
+                if (minTransferTime != null && ((minTransferTime < minTransferTimeLowerBound) ||
+                        (minTransferTime > minTransferTimeUpperBound))) {
```
fabrice-v, line 167: Wondering if this should be done here or at a later stage in its own use case. What were the reasons that led to have this implicit rule checking at entity creation time?

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/notice/warning/SuspiciousIntegerValueNotice.java
```diff
@@ -0,0 +1,68 @@
...
+ */
+
+package org.mobilitydata.gtfsvalidator.domain.entity.notice.warning;
+
+import org.mobilitydata.gtfsvalidator.domain.entity.notice.NoticeExporter;
+import org.mobilitydata.gtfsvalidator.domain.entity.notice.base.WarningNotice;
+
+import java.io.IOException;
+
+public class SuspiciousIntegerValueNotice extends WarningNotice {
```
fabrice-v, line 24: Connected with the previous remark, we create this new and generic Notice for something that is quite specific.

domain/src/test/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/transfers/TransferTest.java
```diff
@@ -32,6 +32,8 @@
     private final static String TO_STOP_ID = "stop id 1";
     private final static int VALID_TRANSFER_TYPE_VALUE = 1;
     private final static int VALID_MIN_TRANSFER_TIME_VALUE = 20;
+    private final int UPPER_BOUND_MIN_TRANSFER_TIME = 40;
```
fabrice-v, line 35: Having a hard time
- why there are 4 values total
- why those two are not static like the two others

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ProcessParsedTransfer.java
```diff
@@ -54,26 +58,28 @@ public ProcessParsedTransfer(final ValidationResultRepository resultRepository,
...
+        );
 
         if (transfer.isSuccess()) {
             if (gtfsDataRepository.addTransfer((Transfer) transfer.getData()) == null) {
                 resultRepository.addNotice(new DuplicatedEntityNotice("transfers.txt",
                         "from_stop_id;to_stop_id", validatedParsedTransfer.getEntityId()));
             }
         } else {
-            //noinspection unchecked
+            //noinspection unchecked to avoid lint
```
fabrice-v, line 82: Extend comment to explain why we end up having to suppress the lint check

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/port/ExecParamRepository.java
```diff
@@ -35,6 +35,8 @@
     String URL_KEY = "url";
     String ZIP_KEY = "zipinput";
     String EXCLUSION_KEY = "exclude";
+    String LOWER_BOUND_MIN_TRANSFER_TIME = "mintransfertime";
```
fabrice-v, line 38: Try maintaining the same nomenclature throughout the code to ease legibility, ie have the same names on both sides of the equal sign, excluding camel case differences

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/transfers/Transfer.java
lionel-nj: As discussed, this will be implemented in a distinct use case.

config/src/main/resources/default-execution-parameters.json
lionel-nj: Keeping this in mind for future implementation of semantic validation of numeric fields in a distinct use case.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/transfers/Transfer.java
lionel-nj: See https://github.com/MobilityData/gtfs-validator/issues/246

domain/src/test/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/transfers/TransferTest.java
lionel-nj: See https://github.com/MobilityData/gtfs-validator/issues/246

adapter/exporter/src/main/java/org/mobilitydata/gtfsvalidator/exporter/JsonNoticeExporter.java
```diff
@@ -226,4 +226,4 @@ public void export(RouteLongNameEqualsShortNameNotice toExport) throws IOExcepti
     public void export(RouteLongNameContainsShortNameNotice toExport) throws IOException {
         jsonGenerator.writeObject(toExport);
     }
-}
+}
```
fabrice-v, line 229: It is best practice to have newline characters at the end of files

adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/db/InMemoryExecParamRepository.java
```diff
@@ -235,20 +235,20 @@ public String getExecParamValue(final String key) throws IllegalArgumentExceptio
      */
     @Override
     public Options getOptions() {
-        final Options options = new Options();
-        options.addOption(String.valueOf(URL_KEY.charAt(0)), URL_KEY, true,
-                "URL to GTFS zipped archive");
-        options.addOption(String.valueOf(ZIP_KEY.charAt(0)), ZIP_KEY, true,
+        final Options options = new Options()
```
fabrice-v, line 238: Those changes don't seem to be related to the PR subject

lionel-nj: Indeed, reverting these changes. Will open a new issue to use the chained method addOption

adapter/exporter/src/main/java/org/mobilitydata/gtfsvalidator/exporter/JsonNoticeExporter.java
lionel-nj: Opening an issue on this matter since various files are affected

PR #73, opened 2020-03-20
Pathways entities -- internal representation

lionel-nj:
**Summary:**

This PR relates to processing of entities from pathways.txt. The goal is to use the required use cases implemented via PR #44 that allow to go from a Pathway representing a row from pathways.txt to a specific concrete internal representation.

**Expected behavior:** 


domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/IsBidirectional.java
```diff
@@ -0,0 +1,26 @@
+package org.mobilitydata.gtfsvalidator.domain.entity;
+
+import org.jetbrains.annotations.NotNull;
+
+import java.util.stream.Stream;
+
+public enum IsBidirectional {
```
fabrice-v: Replace by a boolean

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/Pathway.java
```diff
@@ -0,0 +1,177 @@
...
+
+    private final float length;
+    private final int traversalTime;
+    private final int stairCount;
+    private final float maxSlope;
+    private final float minWidth;
+    private final String signpostedAs;
+    private final String reversedSignpostedAs;
+
+    public Pathway(@NotNull final String pathwayId,
```
fabrice-v: make private to enforce Builder usage

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/Pathway.java
```diff
@@ -0,0 +1,177 @@
...
+
+    public String getSignpostedAs() {
+        return signpostedAs;
+    }
+
+    public String getReversedSignpostedAs() {
+        return reversedSignpostedAs;
+    }
+
+    @SuppressWarnings("CanBeFinal")
```
fabrice-v: Remove and provide setters for all fields

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/IsBidirectional.java
barbeau: We may want to keep this as an `enum` to allow future flexibility. It's possible this GTFS field could be overloaded with another value in the future, even though it seems like it should only have two possible values.

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ProcessParsedPathway.java
```diff
@@ -0,0 +1,97 @@
...
+import org.mobilitydata.gtfsvalidator.domain.entity.gtfs.pathways.Pathway;
+import org.mobilitydata.gtfsvalidator.domain.entity.notice.base.Notice;
+import org.mobilitydata.gtfsvalidator.domain.entity.notice.error.DuplicatedEntityNotice;
+import org.mobilitydata.gtfsvalidator.usecase.port.GtfsDataRepository;
+import org.mobilitydata.gtfsvalidator.usecase.port.ValidationResultRepository;
+
+import java.util.List;
+
+/**
+ * This use case turns a parsed entity representing a row from agency.txt into a concrete class
```
fabrice-v, line 30: Not the right file name

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/pathways/Pathway.java
```diff
@@ -0,0 +1,400 @@
...
+         * @return an {@code EntityBuildResult} representing a row from pathways.txt if the requirements from the
+         * official GTFS specification are met. Otherwise, method returns a collection of notices specifying the issues.
+         */
+        public EntityBuildResult<?> build() {
+            noticeCollection.clear();
+
+            if (pathwayId == null ||
+                    fromStopId == null ||
+                    toStopId == null ||
+                    !IsBidirectional.isEnumValueValid(originalIsBiDirectionalInteger) ||
```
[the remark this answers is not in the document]
lionel-nj: I think this situation is handled:

if `originalIsBiDirectionalInteger` is null, method `fromInt` of enum `IsBidirectional` returns `null`. 

This being done, we then check if `originalIsBiDirectionalInteger=null` is valid with method `isEnumValueValid` which will return `false`.

 After that, since `isBidirectional` and `originalIsBiDirectionalInteger` are null we emit a new `MissingRequiredValueNotice`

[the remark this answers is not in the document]
lionel-nj: Also, I think that this test covers this situation: https://github.com/MobilityData/gtfs-validator/blob/b8e25a006d8c5ddd56e3733be34649aedaddd08f/domain/src/test/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/pathways/PathwayTest.java#L250

PR #75, opened 2020-03-20
Attribution entities -- internal representation

lionel-nj:
**Summary:**

This PR relates to processing of entities from attributions.txt. The goal is to use the required use cases implemented via PR #44 that allow to go from a Level representing a row from attributions.txt to a specific concrete internal representation.

**Expected behavior:** 


domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/Attribution.java
```diff
@@ -0,0 +1,168 @@
...
+
+    public static class AttributionBuilder {
+        private String attributionId;
+        private String agencyId;
+        private String routeId;
+        private String tripId;
+        private String organizationName;
+        private IsProducer isProducer;
+        private IsOperator isOperator;
+        private IsAuthority isAuthority;
```
fabrice-v: One of those three fields is mandatory set to 1. This fact should be translated as to have different *classes*, on the model of what has been done for stops entities.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/IsAuthority.java
```diff
@@ -0,0 +1,26 @@
+package org.mobilitydata.gtfsvalidator.domain.entity;
```
fabrice-v: Delete this file

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/IsOperator.java
```diff
@@ -0,0 +1,26 @@
+package org.mobilitydata.gtfsvalidator.domain.entity;
```
fabrice-v: Delete this file

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/IsProducer.java
```diff
@@ -0,0 +1,26 @@
+package org.mobilitydata.gtfsvalidator.domain.entity;
```
fabrice-v: Delete this file

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/Attribution.java
```diff
@@ -0,0 +1,168 @@
...
+    private final String organizationName;
+
+    private final IsProducer isProducer;
+    private final IsOperator isOperator;
+    private final IsAuthority isAuthority;
+    private final String attributionUrl;
+    private final String attributionEmail;
+    private final String attributionPhone;
+
+    public Attribution(String attributionId,
```
fabrice-v: Should be private to enforce using the Builder class

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/Attribution.java
lionel-nj: A nuance being that "at least one"of these fields must be set to 1. 

Let's imagine that a row from attributions.txt has both isOperator and isProducer booleans set to 1: following the model of what's been done for stop entities could lead to loss of information in a way that that either we would have a ProducerAttribution or an OperatorAttribution entity.

In the use case "ProcessParsedAttribution", shall we consider using a vector of three booleans (isProducer, isOperator, isAuthority) named as "attributionRole"? From that vector we could set fields isProducer, isOperator, and isAuthority (via a builder pattern) depending of the components of said vector. 

The case (0,0,0): all fields set to 0,  would generate a notice in the (future) use case ProcessParsedAttribution. 

This would allow us to only have one class "Attribution" instead of 7 sub-classes depending on a parent class "AttributionBase".

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/Attribution.java
```diff
@@ -0,0 +1,383 @@
...
+        }
+    }
+
+    /**
+     * Returns the key used to map {@code Attribution}
+     *
+     * @return  the key used to map {@link Attribution}
+     */
+    public String getAttributionKey() {
+        return getAttributionId() + "; " + getAgencyId() + "; " + getRouteId() + "; " + getTripId() + "; "
```
fabrice-v, line 378: Same as FareRule, see https://github.com/MobilityData/gtfs-validator/pull/72#discussion_r438366355 and https://github.com/MobilityData/gtfs-validator/pull/72#discussion_r438369254

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/Attribution.java
```diff
@@ -0,0 +1,383 @@
...
+                    (originalIsOperatorInteger != null &&
+                            (originalIsOperatorInteger < 0 || originalIsOperatorInteger > 1)) ||
+                    (originalIsAuthorityInteger != null &&
+                            (originalIsAuthorityInteger < 0 || originalIsAuthorityInteger > 1)) ||
+                    (originalIsProducerInteger != null &&
+                            (originalIsProducerInteger < 0 || originalIsProducerInteger > 1)) ||
+                    ((isAuthority == isProducer) && (isAuthority == isOperator) &&
+                            (originalIsProducerInteger == null || originalIsProducerInteger == 0))
+            ) {
+                final String entityId = attributionId + "; " + agencyId + "; " + routeId + "; " + tripId + "; " +
```
fabrice-v, line 332: Same as FareRule, see https://github.com/MobilityData/gtfs-validator/pull/72#discussion_r438366355 and https://github.com/MobilityData/gtfs-validator/pull/72#discussion_r438369254

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/Attribution.java
lionel-nj: As discussed, for now we use a static method to generate the key.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/Attribution.java
lionel-nj: see https://github.com/MobilityData/gtfs-validator/pull/75#discussion_r441179373

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/Attribution.java
lionel-nj: See 6604871d9210650ebc6ed618d8afb109d66cb44b

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/Attribution.java
```diff
@@ -0,0 +1,397 @@
...
+        return attributionId+agencyId+routeId+tripId+organizationName+isProducer+isOperator+isAuthority+attributionUrl+
+                attributionEmail+attributionPhone;
+    }
+
+    /**
+     * Returns the key corresponding to this {@link Attribution}
+     *
+     * @return the key corresponding to this {@link Attribution}
+     */
+    public String getAttributionMappingKey() {
```
fabrice-v, line 392: I seem to recognize the pattern that led to code duplication. This should be implemented through a static method

lionel-nj: The static method is implemented at line 377 of this same file, as done in FareRule.java

fabrice-v: Very true, my bad. thks for the quick follow up.

PR #76, opened 2020-03-20
Fix #14 Extends GTFS types support -- Add support for Time and Date types

### LINKED ISSUE #14: [REQUEST] Extend validation of GTFS types
Current validator only supports ID, Text, Float and Integer types from https://gtfs.org/reference/static/#field-types

ValidationUtils class extension to new types

**Additional context**
Extension will allow for more GTFS files validation

fabrice-v:
**Summary:**

This PR extends support of GTFS types to time and date
It also refactors some classes to make them injectable and the associated tests by introducing mockito framework use

**Expected behavior:** 

All GTFS files should be validated, here for MTBA
image

close #14 


adapter/parser/src/main/java/org/mobilitydata/gtfsvalidator/parser/GtfsEntityParser.java
```diff
@@ -36,23 +40,37 @@
...
+
+    public GtfsEntityParser(GtfsSpecificationProto.CsvSpecProto fileSchema, RawFileInfo rawFileInfo,
+                            FloatValidator floatValidator, IntegerValidator integerValidator,
+                            DateValidator dateValidator) {
         this.fileSchema = fileSchema;
         this.rawFileInfo = rawFileInfo;
+        this.floatValidator = floatValidator;
+        this.integerValidator = integerValidator;
+        if (!dateValidator.isStrict()) {
+            throw new IllegalArgumentException("Date validator must be strict");
```
fabrice-v: Add link to strictness definition
https://commons.apache.org/proper/commons-validator/apidocs/org/apache/commons/validator/routines/AbstractFormatValidator.html#isStrict()

adapter/parser/src/main/java/org/mobilitydata/gtfsvalidator/parser/GtfsEntityParser.java
```diff
@@ -36,23 +40,37 @@
...
+    private final FloatValidator floatValidator;
+    private final IntegerValidator integerValidator;
+    private final DateValidator dateValidator;
 
-    public GtfsEntityParser(GtfsSpecificationProto.CsvSpecProto fileSchema, RawFileInfo rawFileInfo) {
+    private static final String DATE_PATTERN = "yyyyMMdd";
+
+    public GtfsEntityParser(GtfsSpecificationProto.CsvSpecProto fileSchema, RawFileInfo rawFileInfo,
+                            FloatValidator floatValidator, IntegerValidator integerValidator,
+                            DateValidator dateValidator) {
```
fabrice-v: Make those constructor parameters @NonNull annotated

adapter/parser/src/main/java/org/mobilitydata/gtfsvalidator/parser/GtfsEntityParser.java
```diff
@@ -120,22 +150,38 @@ public ParsedEntity parse(RawEntity toParse) {
...
+
+                    //FIXME: retrieve locale from agency_lang in agency.txt and if that doesn't exist,
+                    //from feed_lang in feed_info.txt before defaulting to Locale.US
+                    //FIXME: retrieve timezone from agency.txt
+                    if (dateValidator.isValid(rawField, DATE_PATTERN, Locale.US)) {
+                        contentByHeaderMap.put(
+                                //https://programminghints.com/2017/05/still-using-java-util-date-dont/
+                                columnSpecProto.getName(), LocalDateTime.ofInstant(
+                                        dateValidator.validate(rawField, DATE_PATTERN, Locale.US).toInstant(),
+                                        ZoneId.of("America/Montreal")
```
fabrice-v: move timezone fixme at the end of this line

adapter/parser/src/main/java/org/mobilitydata/gtfsvalidator/parser/GtfsEntityParser.java
```diff
@@ -93,6 +113,20 @@ public GtfsEntityParser(GtfsSpecificationProto.CsvSpecProto fileSchema, RawFileI
                                 )
                         );
                     }
+                } else if (columnSpecProto.getType().getType() ==
+                        GtfsSpecificationProto.ColumnInputType.InputType.DATE) {
+
+                    //FIXME: retrieve locale from agency_lang in agency.txt and if that doesn't exist,
+                    //from feed_lang in feed_info.txt before defaulting to Locale.US
+                    if (!dateValidator.isValid(rawField, DATE_PATTERN, Locale.US)) {
```
barbeau: The good news is that you shouldn't have to handle locales here. The [GTFS spec](https://gtfs.org/reference/static/#field-types) says:

>Service day in the YYYYMMDD format.

So the `, Locale.US` should be removed.

adapter/validator/src/test/java/org/mobilitydata/gtfsvalidator/validator/GtfsTypeValidatorTest.java
```diff
@@ -167,870 +218,927 @@ void outOfRangeInt_decGenerateNotice() throws IOException {
...
+        assertEquals(0, result.size());
 
-        notice = new ArrayList<>(result).get(0);
-        assertThat(notice, instanceOf(InvalidColorNotice.class));
-        assertEquals("E014", notice.getId());
-        assertEquals("Invalid color", notice.getTitle());
-        assertEquals(TEST_FILE_TST, notice.getFilename());
-        assertEquals("Invalid color:AZ-FTJ in field:int_hex_with_regex for entity with id:test_id",
-                notice.getDescription());
+        verify(mockTimeValidator, times(1)).isValid(ArgumentMatchers.eq("01:02:03"));
```
barbeau: Time is tricky in GTFS - the [spec says](https://gtfs.org/reference/static/#field-types):

>Time - Time in the HH:MM:SS format (H:MM:SS is also accepted). The time is measured from "noon minus 12h" of the service day (effectively midnight except for days on which daylight savings time changes occur). For times occurring after midnight, enter the time as a value greater than 24:00:00 in HH:MM:SS local time for the day on which the trip schedule begins.
Example: 14:30:00 for 2:30PM or 25:35:00 for 1:35AM on the next day.

There need to be a few more assertions here to make sure we're covering all the valid times.

I suggest verifying that the following are correctly parsed as legit times:
* "22:15:09"
* "25:05:55"
* "5:15:35"

adapter/validator/src/test/java/org/mobilitydata/gtfsvalidator/validator/GtfsTypeValidatorTest.java
```diff
@@ -167,870 +218,927 @@ void outOfRangeInt_decGenerateNotice() throws IOException {
...
-                        "string_type_url", "https://mobilitydata.org"),
-                new RawFileInfo.RawFileInfoBuilder().filename(TEST_FILE_TST).build()
-        ));
+        assertEquals(1, result.size());
+        Notice notice = new ArrayList<>(result).get(0);
+        assertThat(notice, instanceOf(InvalidTimeNotice.class));
+        assertEquals("E016", notice.getId());
+        assertEquals("Invalid time", notice.getTitle());
+        assertEquals(TEST_FILE_TST, notice.getFilename());
+        assertEquals("Invalid time:001:2:00003 in field:time_with_regex for entity with id:test_id",
```
barbeau: See above - I'd suggest adding more times that should fail:
* "05:5:35"
* "05:05:5"
* "05:60:00"
* "05:00:60"

adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/db/InMemoryGtfsSpecRepository.java
```diff
@@ -42,6 +44,10 @@
     private final GtfsSpecificationProto.CsvSpecProtos inMemoryGTFSSpec;
     private final Map<String, ParsedEntityTypeValidator> validatorByFilenameCache = new HashMap<>();
 
+    private static final String[] VALID_URL_SCHEMES = {"http", "https"};
+    private static final String VALID_COLOR_REGEX_PATTERN = "[0-9a-fA-F]{6}";
+    private static final String VALID_TIME_REGEXP_PATTERN = "([0-9][0-9]|[0-9]):[0-5][0-9]:[0-5][0-9]";
```
barbeau: IMHO something like "999:00:00" should be flagged as a warning, although it's not clear where the cutoff should be. FWIW, we used 29 hours as a cutoff for the GTFS-RT validator. This can be pushed to a future issue.

adapter/validator/src/test/java/org/mobilitydata/gtfsvalidator/validator/GtfsTypeValidatorTest.java
fabrice-v: Ok, in those unit tests the isValid return value is mocked and the parameter check is just here to verify it is passed unchanged. That way we don't test the validator library code at the same time. Do you think we shall have integration tests as well?

adapter/validator/src/test/java/org/mobilitydata/gtfsvalidator/validator/GtfsTypeValidatorTest.java
fabrice-v: Ok, in those unit tests the isValid return value is mocked and the parameter check is just here to verify it is passed unchanged. That way we don't test the validator library code at the same time. Do you think we shall have integration tests as well?

adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/db/InMemoryGtfsSpecRepository.java
fabrice-v: Ok yes we shall have a notion for a user configurable suspicious threshold. I'll create an issue

fabrice-v: #91

PR #81, opened 2020-03-25
FeedInfo class -- internal representation

lionel-nj:
**Summary:**

This PR relates to processing of entities from feed_info.txt. The goal is to use the required use cases implemented via PR #44 that allow to go from a FeedInfo representing a row from feed_info.txt to a specific concrete internal representation.

**Expected behavior:** 


usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ProcessParsedFeedInfo.java
```diff
@@ -0,0 +1,89 @@
...
+        builder.feedPublisherName(feedPublisherName)
+                .feedPublisherUrl(feedPublisherUrl)
+                .feedLang(feedLang)
+                .feedStartDate(feedStartDate)
+                .feedEndDate(feedEndDate)
+                .feedVersion(feedVersion)
+                .feedContactEmail(feedContactEmail)
+                .feedContactUrl(feedContactUrl);
+
+        @SuppressWarnings("rawtypes") final EntityBuildResult agency = builder.build();
```
fabrice-v, line 77: Suppression should be explained with a comment

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ProcessParsedFeedInfoTest.java
```diff
@@ -0,0 +1,248 @@
...
+        verify(mockGenericObject, times(1)).isSuccess();
+        //noinspection ResultOfMethodCallIgnored
+        verify(mockGenericObject, times(1)).getData();
+
+        verify(mockResultRepo, times(1)).addNotice(isA(Notice.class));
+        verifyNoMoreInteractions(mockParsedFeedInfo, mockGtfsDataRepo, mockBuilder, mockResultRepo, mockGenericObject);
+    }
+
+    @Test
+    void duplicateFeedInfoShouldAddNoticeToResultRepoAndShouldNotBeAddedToGtfsDataRepo() {
```
fabrice-v, line 176: Same remark as the other one, could testing at this level could be simplified to repository add rejection, no matter the cause?

fabrice-v: No it can't. The duplication notice is generated inside the use case, hence needs testing at that level.

PR #82, opened 2020-03-25
Calendar class -- internal representation

lionel-nj:
**Summary:**

This PR relates to processing of entities from calendar.txt. The goal is to use the required use cases implemented via PR #44 that allow to go from a Calendar representing a row from calendar.txt to a specific concrete internal representation.

**Expected behavior:** 


domain/src/test/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/CalendarTest.java
```diff
@@ -0,0 +1,664 @@
...
+                .thursday(0)
+                .friday(0)
+                .saturday(0)
+                .sunday(0)
+                .startDate(LocalDateTime.now())
+                .endDate(LocalDateTime.now());
+
+        final EntityBuildResult<?> buildResult = underTest.build();
+
+        final ArgumentCaptor<MissingRequiredValueNotice> captor =
```
fabrice-v, line 269: I understand the use of the captor in conjunction with the `CalendarBuilder` parametric constructor taking a list, but reading this test I can't help but wonder if we shouldn't simply unwrap the list from `buildResult` (cast its `data` member to a List). Is there something I missed?

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ProcessParsedCalendarTest.java
```diff
@@ -0,0 +1,250 @@
...
+
+        inOrder.verify(mockBuilder, times(1)).build();
+
+        inOrder.verify(mockGtfsDataRepo, times(1)).addCalendar(ArgumentMatchers.eq(mockCalendar));
+
+        verifyNoMoreInteractions(mockBuilder, mockCalendar, mockParsedCalendar, mockGtfsDataRepo);
+    }
+
+    @Test
+    public void invalidCalendarShouldAddNoticeToResultRepoAndShouldNotBeAddedToGtfsDataRepo() {
```
fabrice-v, line 115: both those tests should be refactored as one testing for add rejection by the repository, no matter the reason

fabrice-v: Nope, the duplication notice is generated inside the use case, so it makes sens it would be tested here

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/Calendar.java
```diff
@@ -0,0 +1,429 @@
...
+         */
+        public CalendarBuilder tuesday(final Integer tuesday) {
+            if (tuesday == null) {
+                this.tuesday = null;
+            } else if (tuesday.equals(1)) {
+                this.tuesday = true;
+            } else if (tuesday.equals(0)) {
+                this.tuesday = false;
+            } else {
+                this.tuesday = null;
```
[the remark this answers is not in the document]
lionel-nj: Indeed, will modify.

domain/src/test/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/CalendarTest.java
lionel-nj: As discussed, we will use de default parametric constructor here

adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/db/InMemoryGtfsDataRepository.java
```diff
@@ -31,6 +31,9 @@
 public class InMemoryGtfsDataRepository implements GtfsDataRepository {
     private final Map<String, Agency> agencyCollection = new HashMap<>();
     private final Map<String, Route> routeCollection = new HashMap<>();
+
+    // Map containing Calendar entities. Entities are mapped on the value found in column service_id of GTFS file
+    // calendar.txt.
     private final Map<String, Calendar> calendarCollection = new HashMap<>();
```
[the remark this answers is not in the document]
lionel-nj: If i'm not wrong, this has been changed in 87530bca793372502978442f2470d459d8301a47

PR #83, opened 2020-03-25
Frequency class -- internal representation

lionel-nj:
**Summary:**

This PR relates to processing of entities from frequencies.txt. The goal is to use the required use cases implemented via PR #44 that allow to go from a Frequency representing a row from frequencies.txt to a specific concrete internal representation.

**Expected behavior:** 


domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/frequencies/ExactTimes.java
```diff
@@ -0,0 +1,65 @@
...
+     * @return the enum item matching the integer passed as parameter. Or null if the integer passed as parameter is
+     * null or does not match any {@link org.mobilitydata.gtfsvalidator.domain.entity.gtfs.frequencies.ExactTimes} enum item
+     */
+    static public ExactTimes fromInt(final Integer fromValue) {
+        if (fromValue == null || fromValue == 0) {
+            return FREQUENCY_BASED_TRIPS;
+        } else if (fromValue == 1) {
+            return SCHEDULE_BASED_TRIPS;
+        } else {
+            return null;
```
barbeau, line 48: If the frequencies.txt `exact_times` column is missing a value, or if the `exact_times` column is omitted from the file, the record should be interpreted as `FREQUENCY_BASED_TRIPS`. This was the purpose of the entire frequencies.txt table before the `exact_times` field was added.

barbeau: Ah - I think you already accounted for that above in line 43? Sorry, I missed that initially.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/frequencies/Frequency.java
```diff
@@ -0,0 +1,218 @@
...
+                    noticeCollection.add(new UnexpectedEnumValueNotice("frequencies.txt",
+                            "exact_times", tripId, originalExactTimes));
+                }
+                return new EntityBuildResult<>(noticeCollection);
+            }
+            return new EntityBuildResult<>(new Frequency(tripId, startTime, endTime, headwaySecs, exactTimes));
+        }
+    }
+
+    public static Integer convertTimeToInteger(String timeString) {
```
barbeau, line 200: This needs a lot of unit tests :)

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/frequencies/Frequency.java
```diff
@@ -0,0 +1,218 @@
...
+    public static class FrequencyBuilder {
+        private String tripId;
+        private Integer startTime;
+        private Integer endTime;
+        private Integer headwaySecs;
+        private ExactTimes exactTimes;
+        private String originalStartTime;
+        private String originalEndTime;
+        private Integer originalExactTimes;
+        private final List<Notice> noticeCollection = new ArrayList<>();
```
barbeau, line 97: Probably not a big deal for frequencies.txt due to the typical number of records, but you could potentially save a lot of memory by lazy-initializing this `List` for files like stop_times.txt and shapes.txt, and in some cases calendar_dates.txt. In most cases for valid data types this list will never be used, but it's being instantiated anyway.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/frequencies/Frequency.java
```diff
@@ -0,0 +1,218 @@
...
+         * @return builder for future object creation
+         */
+        public FrequencyBuilder exactTimes(@Nullable Integer exactTimes) {
+            this.exactTimes = ExactTimes.fromInt(exactTimes);
+            this.originalExactTimes = exactTimes;
+            return this;
+        }
+
+        public EntityBuildResult<?> build() {
+            noticeCollection.clear();
```
barbeau, line 159: Why is `noticeCollection` being cleared here? Doesn't seem like it's accessed anywhere besides this method.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/frequencies/Frequency.java
barbeau: This method also probably belongs in a utility class somewhere, as you'll need it for stop_times.txt as well.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/frequencies/Frequency.java
fabrice-v: I seem to remember it's because with use a single instance of the Builder class to build all entities.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/frequencies/Frequency.java
barbeau: Per below comment, if there is only one instance of the builder being used, then this doesn't really help.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/frequencies/Frequency.java
barbeau: Ah, yes - then it makes sense. But there isn't a general `.clear()` method on the builder? The current design is a bit dangerous when used as a single instance as it's assuming you always call all of the builder fields when building the object - if you don't then you keep one of the previous values when building.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/frequencies/Frequency.java
maximearmstrong: I agree. For now, we discussed that it should be move into the GtfsEntityParser, so the internal representation of an entity with a time type would implement the Integer form. I moved the code as is. I suggest a TimeType class or similar.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/frequencies/Frequency.java
```diff
@@ -0,0 +1,209 @@
...
+                if (tripId == null) {
+                    noticeCollection.add(new MissingRequiredValueNotice("frequencies.txt",
+                            "trip_id", tripId));
+                }
+
+                if (startTime == null) {
+                    noticeCollection.add(new MissingRequiredValueNotice("frequencies.txt",
+                            "start_time", tripId));
+                }
+                //else if (startTime == null && originalStartTime != null) InvalidTimeNotice?
```
lionel-nj, line 169: commented code to remove

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/frequencies/Frequency.java
```diff
@@ -0,0 +1,209 @@
...
+
+                if (endTime == null) {
+                    noticeCollection.add(new MissingRequiredValueNotice("frequencies.txt",
+                            "end_time", tripId));
+                }
+                //else if (endTime == null && originalEndTime != null) InvalidTimeNotice?
```
lionel-nj, line 175: commented code to remove

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/frequencies/Frequency.java
```diff
@@ -0,0 +1,209 @@
...
+                            "exact_times", tripId, originalExactTimes));
+                }
+                return new EntityBuildResult<>(noticeCollection);
+            }
+            return new EntityBuildResult<>(new Frequency(tripId, startTime, endTime, headwaySecs, exactTimes));
+        }
+    }
+
+    /**
+     * Returns the key corresponding to this {@link FareRule}
```
lionel-nj, line 198: `{@link Frequency}`

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/frequencies/Frequency.java
```diff
@@ -0,0 +1,209 @@
...
+     *
+     * @return the key corresponding to this {@link FareRule}
+     */
+    public static String getFrequencyMappingKey(final String tripId, final Integer startTime) {
```
lionel-nj, line 202: If we map `Frequency` entities on `trip_id`+`start_time` shouldn't we use this same key as `entity_id` when generating notices?

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/frequencies/Frequency.java
```diff
@@ -0,0 +1,209 @@
...
+            return this;
+        }
+
+        public EntityBuildResult<?> build() {
+            noticeCollection.clear();
+            if (tripId == null || startTime == null || endTime == null || headwaySecs == null ||
+                    headwaySecs < 0 || exactTimes == null) {
+
+                if (tripId == null) {
+                    noticeCollection.add(new MissingRequiredValueNotice("frequencies.txt",
```
lionel-nj, line 161: Should the `entity_id` be `trip_id+start_time`?

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/frequencies/ExactTimes.java
```diff
@@ -0,0 +1,65 @@
...
+        }
+    }
+
+    /**
+     * Returns true if the integer passed as parameter is expected for this enum, otherwise returns false
+     *
+     * @param value the integer to associate with this enum values
+     * @return true if the integer passed as parameter is expected for this enum, otherwise returns false
+     */
+    static public boolean isEnumValueValid(final Integer value) {
```
lionel-nj, line 58: Besides the test classes where this methods is tested, this method does not seem to be used. We could use it in `fromInt`, or simply remove it since the statement `(fromValue == null || fromValue == 0)` in `fromInt` seems to  do the job.

domain/src/test/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/frequencies/ExactTimesTest.java
```diff
@@ -0,0 +1,74 @@
...
+        assertNull(ExactTimes.fromInt(2));
+    }
+
+    @Test
+    void createExactTimesFromInvalidNegativeIntegerValueMinus1ShouldReturnNull() {
+        assertNull(ExactTimes.fromInt(-1));
+    }
+
+    @Test
+    void validNullValueShouldReturnTrue() {
```
lionel-nj, line 51: If method `isEnumValid` is removed these tests could also be removed

domain/src/test/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/frequencies/FrequencyTest.java
```diff
@@ -0,0 +1,227 @@
...
+
+        assertEquals("frequencies.txt", notice.getFilename());
+        assertEquals("exact_times", notice.getNoticeSpecific(KEY_FIELD_NAME));
+        assertEquals(STRING_TEST, notice.getEntityId());
+        assertEquals(2, notice.getNoticeSpecific(KEY_ENUM_VALUE));
+        assertEquals(1, noticeCollection.size());
+    }
+
+    @Test
+    void createFrequencyWithNullTransfersShouldNotGenerateNotice() {
```
lionel-nj, line 201: method should be renamed `createFrequencyWithNullExactTimeShouldNotGenerateNotice`

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ProcessParsedFrequency.java
```diff
@@ -0,0 +1,84 @@
...
+        this.builder = builder;
+    }
+
+    /**
+     * Use case execution method to go from a row from frequencies.txt to an internal representation.
+     * <p>
+     * This use case extracts values from a {@code ParsedEntity} and creates a {@code Frequency} object if the
+     * requirements from the official GTFS specification are met. When these requirements are not met, related notices
+     * generated in {@code Frequency.FrequencyBuilder} are added to the result repository provided in the
+     * constructor. This use case also adds a {@code EntityMustBeUniqueNotice} to said repository if the uniqueness
```
lionel-nj, line 52: Update javadoc the notice is actually named `DuplicatedEntityNotice`

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ProcessParsedFrequency.java
```diff
@@ -0,0 +1,84 @@
...
+     * constraint on fare attribute entities is not respected.
```
lionel-nj, line 53: frequency entities

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ProcessParsedFrequencyTest.java
```diff
@@ -0,0 +1,216 @@
...
+
+        verify(mockBuilder, times(1)).tripId(TRIP_ID);
+        verify(mockBuilder, times(1)).startTime(START_TIME_VALUE);
+        verify(mockBuilder, times(1)).endTime(END_TIME_VALUE);
+        verify(mockBuilder, times(1)).headwaySecs(HEADWAY_SECS_VALUE);
+        verify(mockBuilder, times(1)).exactTimes(EXACT_TIMES_VALUE);
+        verify(mockBuilder, times(1)).build();
+        verify(mockGtfsDataRepo, times(1)).addFrequency(ArgumentMatchers.eq(mockFrequency));
+
+        verifyNoMoreInteractions(mockBuilder, mockFrequency, mockParsedFrequency, mockGtfsDataRepo);
```
lionel-nj, line 91: No interaction should also be checked on `mockResultRepo`

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ProcessParsedFrequencyTest.java
```diff
@@ -0,0 +1,216 @@
...
+        verify(mockGenericObject, times(1)).getData();
+        verify(mockResultRepo, times(1)).addNotice(ArgumentMatchers.eq(mockNotice));
+
+        final ArgumentCaptor<MissingRequiredValueNotice> captor =
+                ArgumentCaptor.forClass(MissingRequiredValueNotice.class);
+
+        verify(mockResultRepo, times(1)).
+                addNotice(captor.capture());
+
+        verifyNoMoreInteractions(mockParsedFrequency, mockGtfsDataRepo, mockBuilder, mockResultRepo, mockGenericObject);
```
lionel-nj, line 147: No interaction should be checked on `mockGtfsDataRepo `

adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/db/InMemoryGtfsDataRepository.java
```diff
@@ -80,6 +81,13 @@
     // Example of key after composition: fare_idroute_idorigin_iddestination_idcontains_id
     private final Map<String, FareRule> fareRuleCollection = new HashMap<>();
 
+    // Map containing Frequency entities. Entities are mapped on a composite key made of the values found in the
+    // columns of GTFS file frequencies.txt:
+    // - trip_id
+    // - start_time
+    // Example of key after composition: trip_idstart_time
+    private final Map<String, Frequency> frequencyCollection = new HashMap<>();
```
lionel-nj, line 89: For legibility and consistency with most of the declarations here, `frequencyCollection` could be renamed as `frequencyPerTripIdStartTime`

adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/db/InMemoryGtfsDataRepository.java
```diff
@@ -482,4 +489,42 @@ public FareRule getFareRule(final String fareId, final String routeId, final Str
...
+     * @param newFrequency the internal representation of a row from frequencies.txt to be added to the
+     *                     repository.
+     * @return the entity added to the repository if the uniqueness constraint of frequency based on rows from
+     * frequencies.txt is respected, if this requirement is not met returns null.
+     * @throws IllegalArgumentException if the argument is null
+     */
+    @Override
+    public Frequency addFrequency(final Frequency newFrequency) throws IllegalArgumentException {
+        if (newFrequency != null) {
+            String key = newFrequency.getFrequencyMappingKey();
```
lionel-nj, line 507: variable should be `final`

adapter/parser/src/main/java/org/mobilitydata/gtfsvalidator/parser/GtfsEntityParser.java
```diff
@@ -185,6 +185,29 @@ public ParsedEntity parse(RawEntity toParse) {
                                 ));
                     }
 
+                /*
```
lionel-nj, line 188: I think this code works here, however in `GtfsTypeValidator` there is a fragment of code (related to `Time` validation) to remove to avoid duplication. Variables should be `final`, some of them could be renamed for legibility

```suggestion
                /*
                } else if (columnSpecProto.getType().getType() ==
                        GtfsSpecificationProto.ColumnInputType.InputType.TIME) {
                    //Shouldn't we implement this elsewhere? Custom validator or class?
                    Integer timeValue = null;
                    if (rawField != null && rawField.matches("([0-9][0-9]|[0-9]):[0-5][0-9]:[0-5][0-9]")) {
                        final String[] timeStringSplit = rawField.split(":");
                        final String hourString = timeStringSplit[0];
                        final String minuteString = timeStringSplit[1];
                        final String secondString = timeStringSplit[2];
                        final Integer hourValue = Integer.valueOf(hourString);
                        final Integer minuteValue = Integer.valueOf(minuteString);
                        final Integer secondValue = Integer.valueOf(secondString);
                        //Converting time to an Integer value representing number of seconds
                        final timeValueAsInt = hourValue * 3600 + minuteValue * 60 + secondValue;
                        //Setting noon as point zero, subtracting 12 * 3600 seconds
                        final timeValueAsIntFromNoonOfDayOfService = timeValue - 12 * 3600;
                    }
                    contentByHeaderMap.put(columnSpecProto.getName(), timeValue);
                    
```

lionel-nj: Also, this implementation requires additional tests

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/frequencies/Frequency.java
maximearmstrong: Good point, should we create a human readable string as we pass the entity id to the notice ex. `"" + tripId + ", " + startTime`? Or create a method?

maximearmstrong: Also, this is not implemented for other entities with similar case (ex. FareRule). 

@barbeau @fabrice-v What do you think? Should we go forward with this for every entity?

Maybe a `getHumanReadableUniqueKey()` method for every entity could be interesting if it is to be recurrent.

adapter/parser/src/main/java/org/mobilitydata/gtfsvalidator/parser/GtfsEntityParser.java
maximearmstrong: I modified it, but kept as commented code for now as we should discuss removing `Time` validation from `GtfsTypeValidator` (it implies removing lots of code related to the time validation)

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/frequencies/Frequency.java
barbeau: @lionel-nj Yes, if we're identifying entities uniquely internally with a combination of fields, then those same fields should be shared in any notices to let the user uniquely identify where the error is coming from in the GTFS file.

@maximearmstrong In the RT validator we ended with similar helper methods to generate the "field name + value" as human-readable text in errors. So the output in the notice would include the text (in some form) "...in frequencies.txt trip_id 100 startTime 12:12:12" to point the user to this record in the file.

adapter/parser/src/main/java/org/mobilitydata/gtfsvalidator/parser/GtfsEntityParser.java
```diff
@@ -185,6 +185,28 @@ public ParsedEntity parse(RawEntity toParse) {
...
+                } else if (columnSpecProto.getType().getType() ==
+                        GtfsSpecificationProto.ColumnInputType.InputType.TIME) {
+
+                    //Shouldn't we implement this elsewhere? Custom validator or class?
+                    Integer timeValueAsInt = null;
+                    if (rawField != null && rawField.matches("([0-9][0-9]|[0-9]):[0-5][0-9]:[0-5][0-9]")) {
```
barbeau, line 194: If this code stays, the regex in `.matches(...)` should be pulled out into a `Pattern` and `Matcher` which is re-used across rows. Regex can be expensive to compile and should be re-used when possible - see https://www.baeldung.com/java-regex-performance for details.

adapter/parser/src/main/java/org/mobilitydata/gtfsvalidator/parser/GtfsEntityParser.java
```diff
@@ -185,6 +185,28 @@ public ParsedEntity parse(RawEntity toParse) {
                                 ));
                     }
 
+                /*
+                } else if (columnSpecProto.getType().getType() ==
```
maximearmstrong, line 189: @fabrice-v @barbeau @lionel-nj Implementing time conversion here requires to remove time type validation from `GtfsTypeValidator`, and to implement validation here, which would breaks the architecture (validation done by the parser instead of the validator).

If we don't remove the validation from the `GtfsTypeValidator` and keep the conversion in the parser, we get an exception, which is why this code is commented at the moment.

Any suggestion on the implementation? I thought maybe keep time as a string and an Util class would be the best option after all.

adapter/parser/src/main/java/org/mobilitydata/gtfsvalidator/parser/GtfsEntityParser.java
maximearmstrong: It will be implemented in TimeConversionUtils class cc @lionel-nj

adapter/parser/src/main/java/org/mobilitydata/gtfsvalidator/parser/GtfsEntityParser.java
maximearmstrong: To be implemented in TimeConversionUtils class

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/frequencies/Frequency.java
maximearmstrong: @barbeau @fabrice-v I modified the builder regarding this in my latest commit. Let me know if it makes sense.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/frequencies/Frequency.java
```diff
@@ -0,0 +1,228 @@
...
+            EntityBuildResult<?> toReturn;
+            if (tripId == null || startTime == null || endTime == null || headwaySecs == null ||
+                    headwaySecs < 0 || exactTimes == null) {
+
+                if (tripId == null) {
+                    noticeCollection.add(new MissingRequiredValueNotice("frequencies.txt",
+                            "trip_id", tripId));
+                }
+
+                if (startTime == null) {
```
barbeau, line 163: My only concern would be if `MissingRequiredValueNotice` is always an accurate message here. I believe `startTime` could also be `null` if the conversion from HH:MM:SS to the int format fails? So the value may have been provided but not valid HH:MM:SS. But earlier in the type validation those types of errors should have been caught, right? So we know it's a valid HH:MM:SS value here?

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/frequencies/Frequency.java
```diff
@@ -0,0 +1,228 @@
...
+            try {
+                final int hourValue = Integer.parseInt(timeStringSplit[0]);
+                final int minuteValue = Integer.parseInt(timeStringSplit[1]);
+                final int secondValue = Integer.parseInt(timeStringSplit[2]);
+                // Converting time to an Integer value representing number of seconds
+                final int timeValueAsInt = hourValue * 3600 + minuteValue * 60 + secondValue;
+                // Setting noon as point zero, subtracting 12 * 3600 seconds
+                return timeValueAsInt - (12 * 3600);
+            } catch (ArrayIndexOutOfBoundsException | NumberFormatException e) {
+                return null;
```
barbeau, line 209: See my above comment - it would be good to have more comments about the interactions of this with type validation - for example, if we expect this to never happen because the string has already been validated as "HH:MM:SS" earlier in the validation pipeline.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/frequencies/Frequency.java
barbeau: @maximearmstrong That's what I had in mind 👍 , although two more comments about this in-line.

barbeau: Oh, wait, I thought this comment was referring to the "HH:MM:SS" conversion - let me look at clearing the notice collection.

barbeau: I don't have all the code in front of me (just looking at GitHub), but definitely make sure there is a unit test for the notices being returned, as this code may be an issue:

```java
                toReturn = new EntityBuildResult<>(noticeCollection);
                noticeCollection.clear();
```

If `noticeCollection` is being passed by reference, you're actually clearing the list inside the `toReturn` instance with the call to `noticeCollection.clear()` - so it will be empty when returned.

I think the original code with `noticeCollection.clear();` may be clearer to understand if you just add a comment above, like:

```java
// Notices are generated for each entity build, so clear the list in case this Builder instance is being re-used
noticeCollection.clear();
```

adapter/parser/src/main/java/org/mobilitydata/gtfsvalidator/parser/GtfsEntityParser.java
lionel-nj: As discussed, a Utility class will be implemented for time conversions. See https://github.com/MobilityData/gtfs-validator/pull/295.

adapter/parser/src/main/java/org/mobilitydata/gtfsvalidator/parser/GtfsEntityParser.java
lionel-nj: See 4aee60541096bf230bead2d0b3d9b3d3938bc84e

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/frequencies/Frequency.java
lionel-nj: a general `.clear()` method has been implemented in https://github.com/MobilityData/gtfs-validator/pull/293

adapter/parser/src/main/java/org/mobilitydata/gtfsvalidator/parser/GtfsEntityParser.java
lionel-nj: See 7577738eb1611d455514121fd20aebb3f5b31718

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/frequencies/Frequency.java
maximearmstrong: That code was implemented elsewhere, see 7577738eb1611d455514121fd20aebb3f5b31718

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/frequencies/Frequency.java
maximearmstrong: You're right, it should be an `InvalidTimeNotice`

fabrice-v: After discussion, we will implement the following strategy

- early fail on errors by default
- add an option to override that behavior, in which case it will be up to the person running the validator to understand what happened

As a consequence this code will stay unchanged

fabrice-v: #306

PR #84, opened 2020-03-25
CalendarDate class -- internal representation

lionel-nj:
**Summary:**

This PR relates to processing of entities from calendar_dates.txt. The goal is to use the required use cases implemented via PR #44 that allow to go from a CalendarDate representing a row from calendar_dates.txt to a specific concrete internal representation.

**Expected behavior:** 


adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/db/InMemoryGtfsDataRepository.java
```diff
@@ -30,6 +32,7 @@
 public class InMemoryGtfsDataRepository implements GtfsDataRepository {
     private final Map<String, Agency> agencyCollection = new HashMap<>();
     private final Map<String, Route> routeCollection = new HashMap<>();
+    private final Map<String, Map<LocalDateTime, CalendarDate>> calendarDateCollection = new HashMap<>();
```
fabrice-v, line 35: A comment describing the different parts of this declaration would be welcomed

adapter/repository/in-memory-simple/src/test/java/org/mobilitydata/gtfsvalidator/db/InMemoryGtfsDataRepositoryTest.java
```diff
@@ -114,4 +115,48 @@ void getRouteByIdShouldReturnRelatedRoute() {
         assertEquals(mockRoute00, underTest.getRouteById("route id0"));
         assertEquals(mockRoute01, underTest.getRouteById("route id1"));
     }
+
+    @Test
+    void callToAddCalendarDateShouldAddEntityToGtfsDataRepoAndReturnSameEntity() {
```
fabrice-v, line 120: Actual add is not tested, so maybe this test should be renamed

adapter/repository/in-memory-simple/src/test/java/org/mobilitydata/gtfsvalidator/db/InMemoryGtfsDataRepositoryTest.java
```diff
@@ -114,4 +115,48 @@ void getRouteByIdShouldReturnRelatedRoute() {
...
+    @Test
+    void addSameCalendarDateShouldReturnNull() {
+        final CalendarDate calendarDate = mock(CalendarDate.class);
+        final InMemoryGtfsDataRepository underTest = new InMemoryGtfsDataRepository();
+
+        underTest.addCalendarDate(calendarDate);
+        assertNull(underTest.addCalendarDate(calendarDate));
+    }
+
+
```
fabrice-v, line 136: extra line

adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/db/InMemoryGtfsDataRepository.java
barbeau: This map won't work as written because `service_id` isn't guaranteed to be unique in calendar_dates.txt.

[calendar_dates.txt](http://gtfs.org/reference/static#calendar_datestxt) is used in two ways in GTFS:
1. It defines exceptions to service defined in calendar.txt. So you may have weekdays that normally run service_id 1, but if a holiday Dec 25th falls on a Monday, you'd have 2 records here - one saying remove service_id 1 on Dec 25th, and a second one saying add service_id 2 on Dec 25th (for whatever the Dec 25th service is defined by service_id 2). In this case service_id isn't a primary key for records in calendar_dates.txt (it's a foreign key to calendar.txt). The combination of all 3 fields (service_id, date, and exception_type) would be a unique within calendar_dates.txt, if you're looking for a composite key.
2. If agencies don't use calendar.txt to define service, they can provide a massive dump of dates in calendar_dates.txt that says which service_id is running on which date. In this case, `service_id` is a primary key for calendar_dates.txt and all service will be exception_type `1` for added.

So in case number 1 above, `service_id` can't be used as the key in the Map, because it's not unique in calendar_dates.txt.

One option that would work across both cases is if you create a composite key of the 3 fields and use that as the key in the `Map`, and have just a single level map like `Map<String, CalendarDate> calendarDateCollection`.

lionel-nj: Thanks you for you insight, will implement the solution you suggest.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/calendardates/CalendarDate.java
```diff
@@ -0,0 +1,156 @@
...
+                }
+                return new EntityBuildResult<>(noticeCollection);
+            } else {
+                return new EntityBuildResult<>(new CalendarDate(serviceId, date, exceptionType));
+            }
+        }
+    }
+
+    public String getCalendarDateMappingKey() {
+        return getServiceId() + getDate().toString() + getExceptionType().toString();
```
barbeau, line 154: One note on the composite key here - thinking through this further, I believe we could get away with just having `getServiceId() + getDate().toString()` be the composite key. In both use cases of calendar_dates.txt (outlined [here](https://github.com/MobilityData/gtfs-validator/issues/189#issuecomment-626985584)) you shouldn't have the same service_id and date combination more than once - this would imply that you're adding and removing the same service on the same date.

The Python validator actually has a rule that checks to make sure there isn't more than one service_id + date entry with the same values - https://github.com/MobilityData/gtfs-validator/issues/189#issuecomment-626985584.

@timMillet Could you confirm this?

lionel-nj: Would considering just these two fields prevent us from detecting that a row from calendar.txt with same service_id, date and exception_type is already defined? Unless we check the 3rd element when trying to add a new service to the GtfsDataRepository.

barbeau: Correct, if we use two fields for the key, you'd need to fetch the existing record with the same key from the HashMap and do a manual comparison with the third field to see if the third field matched.

But, you can't do the reverse - key on three fields and then check if two fields match.

For example, this is wrong in GTFS (two field collision):

~~~
service_id,date,exception_type
ABCD,20200509,1
ABCD,20200509,2
~~~

And so is this:

~~~
service_id,date,exception_type
ABCD,20200509,1
ABCD,20200509,1
~~~

By keying on `getServiceId() + getDate().toString()` you can catch both situations (by manually checking the 3rd element), but when keying on `getServiceId() + getDate().toString() + getExceptionType().toString()` you can only catch the 2nd (because there would be no collision in the HashMap key for the first set of data).

One question is if we want to split these out as two different errors, or leave them under the same error. If they are the same error (e.g., "Invalid calendar_dates.txt data"), then you don't even need to check if the 3rd element matches.

lionel-nj: Thanks for the explanation, I understand the issue better now. Could we in a first approach consider the errors to be the same, them depending on the rules to implement make them different?

barbeau: IMHO yes. I'd suggest using `getServiceId() + getDate().toString()` as the key, and if there are collisions it's an error. This matches the current implementation in the Python validator and it's likely we won't need to go deeper than that.

timMillet: > you shouldn't have the same service_id and date combination more than once - this would imply that you're adding and removing the same service on the same date
I agree

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/calendardates/CalendarDate.java
```diff
@@ -0,0 +1,156 @@
...
+    public ExceptionType getExceptionType() {
+        return exceptionType;
+    }
+
+    /**
+     * Builder class to create {@link CalendarDate} objects. Allows an unordered definition of the different attributes
+     * of {@link CalendarDate}.
+     */
+    public static class CalendarDateBuilder {
+        private String serviceId;
```
fabrice-v, line 75: Shall there be nullability annotations?

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/calendardates/CalendarDate.java
```diff
@@ -0,0 +1,156 @@
...
+         */
+        public CalendarDateBuilder exceptionType(@NotNull final Integer exceptionType) {
+            this.exceptionType = ExceptionType.fromInt(exceptionType);
+            this.originalExceptionTypeInteger = exceptionType;
+            return this;
+        }
+
+        /**
+         * Returns an entity representing a row from calendar_dates.txt if the requirements from the official GTFS
+         * specification are met. Otherwise, method returns an entity representing a list of notices.
```
fabrice-v, line 117: There's ambiguity in the use of the term 'entity' at the end of the sentence.
Suggestion: Otherwise return a collection of notices describing the issue

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/calendardates/CalendarDate.java
```diff
@@ -0,0 +1,156 @@
...
+         * @return builder for future object creation
+         */
+        public CalendarDateBuilder exceptionType(@NotNull final Integer exceptionType) {
+            this.exceptionType = ExceptionType.fromInt(exceptionType);
+            this.originalExceptionTypeInteger = exceptionType;
+            return this;
+        }
+
+        /**
+         * Returns an entity representing a row from calendar_dates.txt if the requirements from the official GTFS
```
fabrice-v, line 116: It actually returns a an `EntityBuildResult`

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/calendardates/CalendarDate.java
```diff
@@ -0,0 +1,156 @@
...
+         * specification are met. Otherwise, method returns an entity representing a list of notices.
+         *
+         * @return entity representing a row from calendar_dates.txt if the requirements from the official GTFS
```
fabrice-v, line 119: See previous remark, it returns a result and not directly an entity

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/calendardates/CalendarDate.java
lionel-nj: As discussed, having annotations here would force us to initialize fields. For cosistency it's better to annotate parameters of builder methodsn

Issue #80, opened 2020-03-24
[REQUEST] Execute GTFS validation on a subset of files


lionel-nj:
**Is your feature request related to a problem? Please describe.**

We should think of a mechanism to be able to:

1.  execute the CSV validation on a limited subset of files
2. GTFS validation on this same subset of files

We must take into account that some files are linked fro example routes.txt, trips.txt, stop_times.txt. Therefore excluding trips.txt means that stop_times.txt would also be excluded from the GTFS validation.

NB: This process should take place in Main.java. See https://github.com/MobilityData/gtfs-validator/blob/30d4872d443aec3ca9024f14f1236362fe1fd0c7/application/cli-app/src/main/java/org/mobilitydata/gtfsvalidator/Main.java#L124 

**Describe the solution you'd like**

We think about excluding for example shapes.txt. The solution would allow to exclude specific files through command line.

**Describe alternatives you've considered**

**Additional context**

lionel-nj:
Working on this, PR to come.

fabrice-v:
After discussion today
- file list generation should happen in a specific use case, returning a list of String
- file interdependencies see what already exist in schema file and/or adapt/reuse it

fabrice-v:
First implementation should allow free text from the user as a command line option to exclude specific files. Main thing will be shapes.txt

PR #89, opened 2020-03-27
Implement #80 - Separate CSV from GTFS validation

lionel-nj:
**Summary:**

This PR comes provides implementation of tests to separate CSV from GTFS validation with command line options.

**Expected behavior:** 

CSV and GTFS validation should be separate, with the possibility to specify which files to exclude from the command line. Files excluded from the CSV validation should also be excluded from the GTFS validation process.


fabrice-v:
Iceboxed

Issue #95, opened 2020-03-30
[REQUEST] Add a ExecutionParameterRepository interface


fabrice-v:
**Is your feature request related to a problem? Please describe.**
There is a lot of code in the Main method relating to CLI parsing. It obfuscates the more important code as well as being error prone and untested.

**Describe the solution you'd like**
The object of this PR is to mimic what has been done with other repositories and have a generic CommandLineRepository  interface through which command lines values would be retrieved. 
- usage of Apache CLI parser should be encapsulated in the adapter layer/module
- repository should handle the presence/absence of options and have configurable defaults through a resource file in JSON
- all implementation files should come with tests done with mockito

fabrice-v:
Keep in mind having a POJO object for those parameters so that they could be deserialized directly with Jackson in the JSON implementation.

PR #110, opened 2020-04-08
Implement #95 -  Implement ExecutionParameterRepository 

lionel-nj:
**Summary:**

This PR provides implementation for an ExecutionParameterRepository storing execution parameters that have been parsed from a Json file or from Apache Command line.

This PR has been branched from tag version v1.1.0 according to advice from @fabrice-v

**Expected behavior:** 

- usage of Apache CLI parser is encapsulated in the adapter layer/module
- all implementation file come with tests done with mockito

Capture d’écran, le 2020-04-21 à 14 24 37


adapter/parser/src/test/resources/test-config.json
```diff
@@ -0,0 +1,22 @@
+[
+  {
+    "longName": "help",
```
barbeau, line 3: I would add tests with sample data that only provide the parameters that users are likely to provide, in particular just the `longName` or `shortname` and `value`:

```json
  {
    "longName": "input",
    "value": "input"
  },
```

adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/execution/InMemoryExecutionParameterRepository.java
```diff
@@ -0,0 +1,52 @@
+package org.mobilitydata.gtfsvalidator.execution;
```
barbeau, line 1: Seems like this should be in `package org.mobilitydata.gtfsvalidator.db` with the other repo implementation?

adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/execution/InMemoryExecutionParameterRepository.java
```diff
@@ -0,0 +1,52 @@
...
+import org.mobilitydata.gtfsvalidator.domain.entity.ExecutionParameter;
+import org.mobilitydata.gtfsvalidator.parser.ApacheExecutionParameterParser;
+import org.mobilitydata.gtfsvalidator.parser.JsonExecutionParameterParser;
+import org.mobilitydata.gtfsvalidator.usecase.port.ExecutionParameterRepository;
+
+import java.util.Collections;
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+public class InMemoryExecutionParameterRepository implements ExecutionParameterRepository {
```
barbeau, line 15: The expected behavior/usage of the `ExecutionParameterRepository` interface and `InMemoryExecutionParameterRepository` implementation aren't very clear to me.

For example, for a basic repo I'd expect [CRUD](https://en.wikipedia.org/wiki/Create,_read,_update_and_delete) operations that all operate on the same dataset.

I would expect that I could create/initialize the repo either empty or with data via the constructor (an array of string arguments), and then add more via `addExecutionParameter()`, and query the repo when executing code to determine what I should do based on the provided parameter and value.  It should be flexible to be initialize from either file or command-line arguments.

But from what I understand, it looks like you have to initialize with a string of arguments from the command-line, which are stored internally. That, via the parser, is then added back to the repo as parsed `ExecutionParameter` to the internal `Map` representation - with the end result being a duplication of the data in both the string array as well as the map. And, then you could add more parameters via `addExecutionParameter()` and the two copies of the data would be out of sync.

It seems like `arguments` String array is just a temporary representation of the data that should be discarded, not stored in the repo.

adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/execution/InMemoryExecutionParameterRepository.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    @Override
+    public ExecutionParameter addExecutionParameter(final ExecutionParameter newExecutionParameter) {
+        executionParameterCollection.put(newExecutionParameter.getShortName(), newExecutionParameter);
+        return newExecutionParameter;
+    }
+
+    @Override
+    public boolean hasExecutionParameter(String shortName) {
```
barbeau, line 40: I could see users using shortName for command-line parameters, but longName for config files (to make them more understandable). The best solution would probably be to have a key property for each `ExecutionParameter` (which could be a string), and then if the long or short name is provided by the user store the `ExecutionParameter` keyed on this (instead of being keyed on the shortName or longName. Then, you can always just query by the key instead of having to support querying for both long and short name.

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ParseAllExecutionParameterTest.java
```diff
@@ -0,0 +1,50 @@
...
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.mockito.Mockito.*;
+
+class ParseAllExecutionParameterTest {
+
+    @Test
+    public void allExecutionParameterFromCommandLineShouldBeParsedAndAddedToRepoAsExecutionParameterEntities() throws IOException {
```
maximearmstrong, line 18: Same as Sean mentioned for `ApacheExecutionParameterParser`, maybe we could uses abrevations for some words : Execution -> Exec, Parameter -> Param, Argument -> Arg. 
Also, maybe a name like `allExecParamToRepoTest` could do the trick and we could add some comments/javadoc to explain what should be the exact behaviour expected for the test.

adapter/parser/src/main/java/org/mobilitydata/gtfsvalidator/parser/ApacheExecutionParameterParser.java
```diff
@@ -0,0 +1,59 @@
...
+import org.apache.commons.cli.ParseException;
+import org.mobilitydata.gtfsvalidator.domain.entity.ExecutionParameter;
+import org.mobilitydata.gtfsvalidator.usecase.port.ExecutionParameterRepository;
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+
+public class ApacheExecutionParameterParser implements ExecutionParameterRepository.ExecutionParameterParser {
```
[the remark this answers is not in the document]
lionel-nj: Good point but I figured this would not be a problem since we are using IntelliJ that provided good autocompletion.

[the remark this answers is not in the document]
barbeau: It's more for code readability - longer names are harder to read quickly.

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ParseAllExecutionParameterTest.java
lionel-nj: Good point also!

domain/src/test/java/org/mobilitydata/gtfsvalidator/domain/entity/ExecutionParameterTest.java
```diff
@@ -0,0 +1,56 @@
+package org.mobilitydata.gtfsvalidator.domain.entity;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+class ExecutionParameterTest {
+
```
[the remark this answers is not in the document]
lionel-nj: If we use Mockito, we would need to stub the methods of ExecutionParameter which would prevent us from testing the real behavior of ExecutionParameter's class methods.

adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/execution/InMemoryExecutionParameterRepository.java
lionel-nj: Absolutely! I missed that while writing the code, thanks!

config/src/main/java/org/mobilitydata/gtfsvalidator/config/DefaultConfig.java
```diff
@@ -38,11 +40,15 @@
...
+    private final String[] arguments;
     private final GtfsSpecRepository specRepo = new InMemoryGtfsSpecRepository("gtfs_spec.asciipb");
     private final RawFileRepository rawFileRepo = new InMemoryRawFileRepository();
     private final ValidationResultRepository resultRepo = new InMemoryValidationResultRepository();
+    private final ExecutionParameterRepository executionParameterRepo;
 
-    public DefaultConfig() throws IOException {
+    public DefaultConfig(String[] arguments) throws IOException {
+        this.arguments = arguments;
+        this.executionParameterRepo = new InMemoryExecutionParameterRepository(arguments);
```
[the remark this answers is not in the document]
lionel-nj: We actually do not need to store these arguments since we do not use them after, removing this.

adapter/parser/src/test/resources/test-config.json
lionel-nj: Writing these test after handling default values for missing arguments

barbeau:
>json file containing execution parameters values for current run should be removed from resources so that validator users can tweak those values. I suggest simply hard coding working directory lookup. It will be distributed with our realeases.

If the JSON file is relatively small, I don't think it would hurt to leave a default config bundled in resources in the JAR (or have the default values otherwise hard-coded somewhere), but then add a command-line param to let the user override that with a user-provided path to a JSON config file. That way the default executable distribution is still one JAR file.

fabrice-v:
You are absolutely right. There are actually

- one defining default values for execution paramters
- one to provide an alternate way to configure the execution of the validator

Presently they both reside in the res folder.

My remark concerned the second file, the one allowing to replace default value for a specific execution without using command lines. It doesn't make much sense packaging it in our resources and I incorreclty imagined we would distribute it alongside the jar. But I stand corrected, we will document where and how one can provide a json configuration to override defaults and not distribute it.

lionel-nj:
Last but not least: tests for notice consumption seem to not work on Github, meanwhile they pass my computer when using commands ./gradlew test or ./gradlew check -i

PR #120, opened 2020-04-23
FareAttribute class -- internal representation 

lionel-nj:
**Summary:**

This PR relates to processing of entities from fare_attributes.txt. The goal is to use the required use cases implemented via PR #44 that allow to go from a row from fare_attributes.txt to a specific concrete internal representation.

**Expected behavior:** 


usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ProcessParsedFareAttribute.java
```diff
@@ -0,0 +1,96 @@
...
+                    .paymentMethod(paymentMethod)
+                    .transfers(transfers)
+                    .agencyId(agencyId)
+                    .transferDuration(transferDuration);
+
+            gtfsDataRepository.addFareAttribute(builder.build());
+
+        } catch (IllegalArgumentException e) {
+
+            if (fareId == null) {
```
barbeau, line 66: This code would be a lot more efficient if we avoided throwing any exceptions, and just checked for these conditions before adding the object to the `gtfsDataRepository`.

Also, the more I see code that's expected to run in our validation process in a `catch` block, the less I like it. IMHO code that handles data validation is expected to run as part of our application under normal conditions and shouldn't be in `catch` blocks.

adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/db/InMemoryGtfsDataRepository.java
```diff
@@ -60,4 +61,23 @@ public Route addRoute(final Route newRoute) throws SQLIntegrityConstraintViolati
...
+
+    private final Map<String, FareAttribute> fareAttributeCollection = new HashMap<>();
+
+    @Override
+    public FareAttribute getFareAttributeByFareId(final String fareId) {
+        return fareAttributeCollection.get(fareId);
+    }
+
+    @Override
+    public FareAttribute addFareAttribute(final FareAttribute newFareAttribute)
```
barbeau, line 73: While I appreciate you trying to find the right exception to throw here, I don't think `java.sql.SQLIntegrityConstraintViolationException ` is the right one. We're not using SQL here. 

If we stick to using exceptions for this type of data validation (which, IMHO, we shouldn't - see my other comment), then these types of errors should probably be our own subclass of `Exception`, something like `GtfsUniqueIdException`.

adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/db/InMemoryGtfsDataRepository.java
```diff
@@ -60,4 +61,23 @@ public Route addRoute(final Route newRoute) throws SQLIntegrityConstraintViolati
...
+            throws SQLIntegrityConstraintViolationException {
+        if (fareAttributeCollection.containsKey(newFareAttribute.getFareId())) {
+            throw new SQLIntegrityConstraintViolationException("fare attribute must be unique in dataset");
+        } else {
+            String fareId = newFareAttribute.getFareId();
```
barbeau, line 78: You don't need to allocate another `String` for `fareId` here - you can eliminate the first line and the next can just be:
`fareAttributeCollection.put(newFareAttribute.getFareId(), newFareAttribute);`

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/fareattributes/FareAttribute.java
```diff
@@ -0,0 +1,158 @@
...
+        }
+
+        public FareAttributeBuilder transferDuration(int transferDuration) {
+            this.transferDuration = transferDuration;
+            return this;
+        }
+
+        public FareAttribute build() {
+            if (fareId == null) {
+                throw new IllegalArgumentException("field `fare_id` in file `fare_attributes.txt` can not be null");
```
barbeau, line 139: Nit - `can not` - > `cannot`

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/fareattributes/FareAttribute.java
```diff
@@ -0,0 +1,158 @@
...
+            }
+            if (price == null) {
```
barbeau, line 141: `price` also can't be negative - you could check for that as well.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/fareattributes/Transfers.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+package org.mobilitydata.gtfsvalidator.domain.entity.gtfs.fareattributes;
+
+import java.util.stream.Stream;
+
+public enum Transfers {
+    NO_TRANSFERS_ALLOWED(0),
+    ONE_TRANSFER_ALLOWED(1),
+    UNLIMITED_TRANSFERS(2);
```
barbeau, line 24: Looking at http://gtfs.org/reference/static#fare_attributestxt:
>0 - No transfers permitted on this fare.
1 - Riders may transfer once.
2 - Riders may transfer twice.
empty - Unlimited transfers are permitted.

...`2` should be twice, and empty should be unlimited.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/fareattributes/FareAttribute.java
```diff
@@ -0,0 +1,158 @@
...
+            return this;
+        }
+
+        public FareAttributeBuilder transfers(Integer transfers) {
+            this.transfers = Transfers.fromInt(transfers);
+            return this;
+        }
+
+        public FareAttributeBuilder agencyId(@Nullable String agencyId) {
+            this.agencyId = agencyId;
```
barbeau, line 128: `agency_id` is conditionally required here like in the other files, if agency.txt contains multiple records. I'd suggest adding a `// TODO` statement here pointing to the issue that we have open for this so it's easier to track down later.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/fareattributes/FareAttribute.java
```diff
@@ -0,0 +1,158 @@
...
+                throw new IllegalArgumentException("field `currency_type` in file `fare_attributes.txt` can not be null");
+            }
+            if (paymentMethod == null) {
+                throw new IllegalArgumentException("unexpected value encountered for field `payment_method` in file" +
+                        " `fare_attributes.txt`");
+            }
+            if (transfers == null) {
+                throw new IllegalArgumentException("unexpected value encountered for field `transfers` in file" +
+                        " `fare_attributes.txt`");
+            }
```
barbeau, line 154: `transfer_duration` also can't be negative - you could check for that as well

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/fareattributes/FareAttribute.java
```diff
@@ -0,0 +1,158 @@
...
+        }
+
+        public FareAttribute build() {
+            if (fareId == null) {
+                throw new IllegalArgumentException("field `fare_id` in file `fare_attributes.txt` can not be null");
+            }
+            if (price == null) {
+                throw new IllegalArgumentException("field `price` in file `fare_attributes.txt` can not be null");
+            }
+            if (currencyType == null) {
```
barbeau, line 144: We should at some point be validating `currencyType` as a valid ISO 4217 alphabetical currency code.

One way to do this is try call `Currency.getInstance(currencyType)`, and if an `IllegalArgumentException` is thrown then it's not valid.

See https://docs.oracle.com/javase/8/docs/api/java/util/Currency.html#getInstance-java.lang.String-

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ProcessParsedFareAttribute.java
fabrice-v: I see your point regarding the exception, but it feels that checking before building would kinda defeat the point of having the builder, with the validation being done outside of it. If it were to be reused it would come with implicit expectations that you check the data as it would let you build anything. For the catch, I agree, the less we do in catch blocks, the better. In that case though it's a duplicate of what happens in the Builder class.
I proposed to @lionel-nj  to return a list of Notice instead of raising an exception. He pointed out that our entity classes are in the domain layer whereas our Notices are in the use case one, because of the visitor pattern used to export them (they all have an export method taking a NoticeExporter parameter, which also isn't in domain). Maybe we should focus on how to refactor so that notices can be in the domain as well, which you already mentionned would make more sense.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/fareattributes/FareAttribute.java
fabrice-v: Yes. I suggest to have a dedicated `ValidateCurrencyCode` use case that will be run after we are done filling the GtfsDataRepository

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ProcessParsedFareAttribute.java
lionel-nj: An issued on how to refactor notice to be part of the domain layer has been opened. See https://github.com/MobilityData/gtfs-validator/issues/122

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/fareattributes/FareAttribute.java
barbeau: That's one possible approach. I see this as a data type validation, though. Aren't we attempting to validate all data types prior to filling the repo with the record?

lionel-nj: Could that be done in class GtfsTypeValidator, at the same place where other gtfs types are checked?

fabrice-v: Good point. In my head data validation concerned what can be found in the schema (ie: range for numeric types) and this is more advanced, but I was mistaken as that is not even true for timezone validity which is a more advanced case not treated through a use case. So yes you are rigtht @lionel-nj, the correct place for it is in GtfsTypeValidator.

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ProcessParsedFareAttribute.java
lionel-nj: After discussion with @maximearmstrong we could in a first step implement @barbeau  solution: check values (using getters on the builder) in the usecase and add notices for the relevant cases, and in a second step refactor Notices.

Of course, this depends on our priorities

lionel-nj: Implementation comes in https://github.com/MobilityData/gtfs-validator/pull/126

adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/db/InMemoryGtfsDataRepository.java
lionel-nj: Finally, since we are not throwing Exceptions, this method returns null, then a null check if executed in the use case execution method, see: 
https://github.com/MobilityData/gtfs-validator/blob/9f6f3a0d99bbdc2f08ed9bce682efdbacdcd84c7/usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ProcessParsedFareAttribute.java#L75

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/fareattributes/FareAttribute.java
lionel-nj: Comment: closing this conversation when closing https://github.com/MobilityData/gtfs-validator/issues/123

lionel-nj: A proposition for currency code validation have been implemented here: https://github.com/MobilityData/gtfs-validator/pull/127

adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/db/InMemoryGtfsDataRepository.java
```diff
@@ -196,4 +201,40 @@ public Level addLevel(final Level newLevel) throws IllegalArgumentException {
     public Level getLevelById(final String levelId) {
         return levelPerId.get(levelId);
     }
+
+    /**
+     * Add an FareAttribute representing a row from fare_attributes.txt to this {@code GtfsDataRepository}.
+     * Return the entity added to the repository if the uniqueness constraint of agency based on fare_id is respected,
```
fabrice-v, line 207: agency

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/fareattributes/FareAttribute.java
```diff
@@ -0,0 +1,266 @@
...
+import java.util.List;
+
+/**
+ * Class for all entities defined in fare_attributes.txt. Can not be directly instantiated: user must use the
+ * {@code FareAttribute.FareAttributeBuilder} to create this.
+ */
+public class FareAttribute extends GtfsEntity {
+    @NotNull
+    final String fareId;
+    private final float price;
```
fabrice-v, line 39: This filed is the only one not to be null annotated

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/fareattributes/FareAttribute.java
```diff
@@ -0,0 +1,266 @@
...
+                }
+                if (fareId == null) {
+                    noticeCollection.add(new MissingRequiredValueNotice("fare_attributes.txt",
+                            "fare_id", fareId));
+                }
+                if (currencyType == null) {
+                    noticeCollection.add(new MissingRequiredValueNotice("fare_attributes.txt",
+                            "currency_type", fareId));
+                }
+                if (paymentMethod == null) {
```
fabrice-v, line 238: I figured .fromInt of PaymentMethod returns null when passed an invalid value. I leave my original comment as follow and suggest the following
```suggestion
                if (paymentMethod == null) { // True when PaymentMethod.isEnumValueValid(originalPaymentMethodInteger) returns false
```

ORIGINAL COMMENT
Would it be possible for `!PaymentMethod.isEnumValueValid(originalPaymentMethodInteger)` to be true while `if (paymentMethod == null)` is NOT?

If that is not the case, some comment explaining how it comes to be would be good. It's easy to understand the code flow checking for a `null` value before doing an if on it, less so to understand the flow when `isEnumValid` returns false

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/fareattributes/FareAttribute.java
```diff
@@ -0,0 +1,266 @@
...
+                    if (originalPaymentMethodInteger == null) {
+                        noticeCollection.add(new MissingRequiredValueNotice("fare_attributes.txt",
+                                "payment_method", fareId));
+                    } else {
+                        noticeCollection.add(new UnexpectedEnumValueNotice("fare_attributes.txt",
+                                "payment_method", fareId, originalPaymentMethodInteger));
+                    }
+                }
+                if (transfers == null) {
```
fabrice-v, line 247: same remark

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/fareattributes/FareAttribute.java
lionel-nj: Indeed, that part is not clear. Thanks for the suggestion, will add comments to clarify the code.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/fareattributes/FareAttribute.java
lionel-nj: https://github.com/MobilityData/gtfs-validator/pull/120#discussion_r437677654

Issue #122, opened 2020-04-24
[REQUEST] Refactor Notice into domain layer


lionel-nj:
**Is your feature request related to a problem? Please describe.**

Notices are part of the use case layer which prevents them to be used along within domain classes. 
This issue has been pointed out when trying to avoid to throw exceptions in client code (to process parsed entities into an internal representation) by generating a list of notices that will be processed in client code.

**Describe the solution you'd like**
Notice should be part of the domain layer. However, the visitor pattern used to export said notice needs to be reworked since these export method require a NoticeExporter (itself part of the adapt layer).

The end goal is to be able to attribute to each Gtfs entity's builder a list of notices that will be used to populate the GtfsResultRepository provided to the constructor of use cases that process gtfs entities.

**Additional context**

lionel-nj:
The summary of the decisions that have been made can be found in the link PR description.

PR #126, opened 2020-04-24
Implement #122 - Make notice classes part of the domain layer

lionel-nj:
**Summary:**

This PR provides context to make notice classes part of the domain layer instead of the usecase one.

The code provided here will be used to generate notices from the future gtfs entities.

**Expected behavior:** 

1. Notice classes have been moved to package domain
2. NoticeExporter interface has been moved to package domain to allow automatic type inference for export methods
  - here, the poor stability of notices makes it convenient to keep the visitor pattern
3. Visitor pattern for addNotice methods has been removed:
  - we determine the type of notice using `instanceof` statement because there is relatively few cases to handle: InfoNotice, WarningNotice, ErrorNotice - and, it is very unlikely that we add another type of notice
4. Tests have been adapted to the new architecture: import statements.

Capture d’écran, le 2020-04-24 à 17 35 11


domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/NoticeExporter.java
```diff
@@ -0,0 +1,58 @@
+package org.mobilitydata.gtfsvalidator.domain.entity;
```
fabrice-v, line 1: This file should reside in notice package

lionel-nj: File has been moved in f8aebbedbac0edf1e24c10bca243a15d4edd66b5

Issue #123, opened 2020-04-24
[REQUEST] Validate currency codes


lionel-nj:
**Is your feature request related to a problem? Please describe.**
File `fare_attributes.txt` contain a field `currency_type` with specific type `currency code` that should be checked.

**Describe the solution you'd like**
An implementation of this is needed in class `GtfsTypeValidator`
From [comment](https://github.com/MobilityData/gtfs-validator/pull/120#discussion_r414044769):

> One way to do this is try call Currency.getInstance(currencyType), and if an IllegalArgumentException is thrown then it's not valid.
> 
> See https://docs.oracle.com/javase/8/docs/api/java/util/Currency.html#getInstance-java.lang.String-

**Describe alternatives you've considered**

**Additional context**
See https://github.com/MobilityData/gtfs-validator/pull/120#discussion_r414044769

PR #127, opened 2020-04-28
Implement #123 - Validate currency code

lionel-nj:
**Summary:**

This PR suggests code to validate currency codes from parsed entities.

**Expected behavior:** 

The validity of field currency_type from fare_attributes.txt will be verified before operations on  parsed entities representing row from fare_attributes.txt.

Capture d’écran, le 2020-04-28 à 17 11 32


adapter/validator/src/test/java/org/mobilitydata/gtfsvalidator/validator/GtfsTypeValidatorTest.java
```diff
@@ -1141,4 +1141,43 @@ void emptyRequiredValueGenerateError() {
...
+                mock(IntegerValidator.class),
+                mock(UrlValidator.class),
+                mock(RegexValidator.class),
+                mock(RegexValidator.class),
+                Collections.emptySet()
+        );
+
+        Collection<Notice> result = underTest.validate(new ParsedEntity(
+                TEST_ID,
+                Map.of("currency_type", "JAN"), //
```
barbeau, line 1169: For all rules/notices, we should be checking for potential false positives too. So, we should test against a valid currency type and make sure that no notices are generated.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/notice/base/ErrorNotice.java
```diff
@@ -35,6 +35,7 @@
     protected static final String E_015 = "E015";
     protected static final String E_016 = "E016";
     protected static final String E_017 = "E017";
+    protected static final String E_018 = "E018";
```
barbeau, line 38: I would strongly suggest that we start documenting each of these rules in a human readable markdown file like https://github.com/CUTR-at-USF/gtfs-realtime-validator/blob/master/RULES.md along with each of these PRs.

From my experience with the RT validator, it's far easier to document these as we go, as opposed to trying to go back later and re-read the code to remember what we implemented.

adapter/validator/src/test/java/org/mobilitydata/gtfsvalidator/validator/GtfsTypeValidatorTest.java
fabrice-v: I agree. Our tests should always include the regular situation where everything goes according to plan.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/notice/base/ErrorNotice.java
lionel-nj: Opening an issue on this matter

fabrice-v:
@barbeau While I do agree with your points, I think here we are in a bit of a special case. The goal of the architecture is to have one rule = one use case = one test class.
Here is a special case as we are not following this general idea but we do have a `ValidateGtfsTypes` use case class. It does lacks a test class, and I think fixing that could satisfy the need for readability.

adapter/validator/src/test/java/org/mobilitydata/gtfsvalidator/validator/GtfsTypeValidatorTest.java
```diff
@@ -1141,4 +1141,51 @@ void emptyRequiredValueGenerateError() {
         assertEquals("Missing value for field:type_date marked as required in entity with id:test_id",
                 notice.getDescription());
     }
+
+    @Test
+    void checkCurrencyCodeValidation() {
```
fabrice-v, line 1146: After discussion, not sure we gain in clarity. Please refactor to seperate the two tests and give them specific methods names

PR #219, opened 2020-05-29
Fix NPE when not providing list of files to exclude

lionel-nj:
**Summary:**

This PR provides code to fix the Exception that is thrown when no user input (concerning the list of files to exclude) is specified.

To solve the issue, it was necessary to add support handling this situation. That is to say, the default value for --exclude option has been modified in `default-configuration-parameters.json`, and the client code in ExecParamRepo has been modified consequently to return an empty String instead of `null`. Therefore, operations on the value can be done, which avoids the NPE.

Following these changes, information logging has been made more precise:
- specific message when validating GTFS dataset without specification of list of files to exclude.

Class of test has been modified: reformat text.

**Expected behavior:** 

Capture d’écran, le 2020-05-29 à 12 34 59

### DIFF
--- a/adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/db/InMemoryExecParamRepository.java
+++ b/adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/db/InMemoryExecParamRepository.java
@@ -18,7 +18,6 @@ package org.mobilitydata.gtfsvalidator.db;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import org.apache.commons.cli.DefaultParser;
-import org.apache.commons.cli.Option;
 import org.apache.commons.cli.Options;
 import org.apache.logging.log4j.Logger;
 import org.mobilitydata.gtfsvalidator.domain.entity.ExecParam;
@@ -219,7 +218,9 @@ public class InMemoryExecParamRepository implements ExecParamRepository {
             }
 
             case EXCLUSION_KEY: {
-                return hasExecParamValue(EXCLUSION_KEY) ? getExecParamByKey(EXCLUSION_KEY).getValue().toString() : null;
+                return hasExecParamValue(EXCLUSION_KEY) ?
+                        getExecParamByKey(EXCLUSION_KEY).getValue().toString()
+                        : defaultValue.get(0);
             }
         }
         throw new IllegalArgumentException("Requested key is not handled");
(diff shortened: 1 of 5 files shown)


config/src/main/resources/default-execution-parameters.json
```diff
@@ -5,5 +5,5 @@
   "proto": false,
   "url": null,
   "zipinput": null,
-  "exclude": null
+  "exclude": ""
```
fabrice-v, line 8: It seems inconsistent that some default String value be defined as null while some others as an empty string. Could we settle on just a single way of doing things?

adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/db/InMemoryExecParamRepository.java
```diff
@@ -258,9 +259,6 @@ public Options getOptions() {
                     option.setArgs(0);
                     break;
                 }
-                case EXCLUSION_KEY: {
```
fabrice-v, line 261: I am not sure to understand why that had to be removed. It seems the EXCLUSION_KEY gets a very specific treatment now, or it it because UNLIMITED_VALUES is the default?

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/GenerateExclusionFilenameList.java
```diff
@@ -57,7 +54,11 @@ public GenerateExclusionFilenameList(final GtfsSpecRepository gtfsSpecRepo,
         gtfsFilenameList.addAll(gtfsSpecRepo.getRequiredFilenameList());
         gtfsFilenameList.addAll(gtfsSpecRepo.getOptionalFilenameList());
 
-        if (!gtfsFilenameList.containsAll(toExcludeFromValidation)) {
+        if(toExcludeFromValidation.size()==1 && Objects.equals(toExcludeFromValidation.get(0), "")){
```
fabrice-v, line 57: Would it be possible to handle a null at this point instead of having an empty default string?

adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/db/InMemoryExecParamRepository.java
lionel-nj: This had to be removed since EXCLUSION_KEY does not longer needs a specific treatment. As all options, it can only be defined once with an unique argument(which is a  String such as `agency.txt,route.txt`

Issue #241, opened 2020-06-04
[REQUEST] Implement semantic verification of fields route_sort_order in routes.txt


lionel-nj:
**Is your feature request related to a problem? Please describe.**
This numeric fields should be user configurable so that suspicious values can be reported.

**Describe the solution you'd like**

- [x] Add new Warning notice:
  - `SuspiciousRouteSortOrderNotice`

- [x] Add support to provide thresholds for this field via configuration file and/or command line

- [ ] Determine default values for said thresholds

barbeau:
Not sure how route_sort_order could be suspicious? It's just a integer for ordering records.

lionel-nj:
We could imagine that for an agency managing 300 routes, having a route with field `route_sort_order` equal to 6000 would be suspicious.

barbeau:
Yes, I guess so. I guess I'm just used to seeing people use somewhat arbitrary integers for ordering. I also haven't seen this field much myself in practice - I'd be interested in seeing actual values used in practice.

PR #248, opened 2020-06-04
Implement #241 - Implement semantic validation of field route_sort_order of file routes.txt

lionel-nj:
**Summary:**

This PR provides code to validate semantically field routes.route_sort_order.  To do so, thresholds have to be provided by user. If said thresholds are not provided, the implementation uses default values for these thresholds.

- new notice `SuspiciousRouteSortOrderNotice` has been added
- support to provide threshold for said field has been added: these are user configurable via configuration file and/or command line

**Expected behavior:** 

Field routes.route_sort_order should be semantically validated following the thresholds provided as input from the software user.

### DIFF
--- a/config/src/main/resources/default-execution-parameters.json
+++ b/config/src/main/resources/default-execution-parameters.json
@@ -5,5 +5,7 @@
   "proto": false,
   "url": null,
   "zipinput": null,
-  "exclude": null
+  "exclude": null,
+  "route__route_sort_order_lower_bound": 0,
+  "route__route_sort_order_upper_bound": 3000
 }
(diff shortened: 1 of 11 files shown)


config/src/main/resources/default-execution-parameters.json
```diff
@@ -5,5 +5,7 @@
   "proto": false,
   "url": null,
   "zipinput": null,
-  "exclude": null
+  "exclude": null,
+  "route__route_sort_order_lower_bound": 0,
+  "route__route_sort_order_upper_bound": 3000
```
barbeau, line 10: It's not clear to me what a reasonable upper threshold is for `route_sort_order`, as I haven't seen real data in use for this field. Maybe @timMillet could weigh in?

I don't know if agencies are using some type of scheme for order numbers - for example, first two digits being route number, second two digits being variant of route, like for a route 10:
* 1001
* 1002

...in which case `3000` really won't be a reasonable threshold. 

But if everyone is numbering them 1,2,3 then the current threshold is fine.

lionel-nj: After discussion with @fabrice-v, it became clear to me how this field is used. If i'm not wrong as you said, fields that represent a sequence order are arbitrary determined by the agencies: they're not numbered the same way from a dataset to another on, which makes it difficult to determine their default values..Hence, I wonder if this PR is relevant at this moment.

barbeau: Correct, it's a sequence number not a key to other data (same as `stop_sequence` and `shape_pt_sequence`. If the code is already implemented you could certainly keep it, although in practice I'm not sure how much it would really be used.

PR #253, opened 2020-06-05
Fix #221 - Run validator on CI

### LINKED ISSUE #221: [REQUEST] Add command-line execution to CI builds
There have been a few issues (most recently https://github.com/MobilityData/gtfs-validator/issues/218) where changes that have passed CI and been merged have broken runtime execution.

IMHO, it would be a good idea to add command-line execution on at least a small real-world GTFS dataset as part of CI to test end-to-end execution prior to merging PRs.

For example, we could add a few lines to the [GitHub CI config](https://github.com/MobilityData/gtfs-validator/blob/master/.github/workflows/gradle.yml) to download an archived GTFS dataset from a fixed time (e.g., MBTA at https://transitfeeds.com/p/mbta/64/20200521) and execute the basic validation against the set. I'm not sure what the current validation time against MBTA GTFS, so if that's too long we could choose a smaller dataset.

If this is implemented, the return code of the app would need to be tested to ensure that a crashing app actually fails the CI build - see https://stackoverflow.com/a/6775860/937715.

1. Do nothing
1. Add more integration tests

barbeau:
**Summary:**

*Please don't merge*

This is a test for https://github.com/MobilityData/gtfs-validator/issues/221 - running the validator on CI.

**Expected behavior:** 

Validator should now download GTFS data and execute on CI

Closes #221


barbeau:
The good news is that it's running the validator, the bad news is that it's just showing help text output and not actually validating the file:

~~~
Run java -jar application/cli-app/build/libs/gtfs-validator-v1.1.0-SNAPSHOT.jar -u https://transitfeeds.com/p/mbta/64/20200531/download -z input.zip -e input -o output

usage: Loads input GTFS feed from url or disk.
Checks files integrity, numeric type parsing and ranges as well as string
             format according to GTFS spec
Validation results are exported to JSON file by default
If no command line is passed as argument of main method, execution
             parameters will be retrieved from .json file located at
             /home/runner/work/gtfs-validator/gtfs-validator/execution-par
             ameters.json
In the case said file could not be found, gtfs-validator will use default
             values.
If command line is passed as argument of main method, execution parameters
             will be retrieved from this command line.
In both cases: if the value for an execution parameter could not be found,
             gtfs-validator will use defined default values
 -e,--extract <arg>    Relative path where to extract the zip content
 -h,--help             Print this message
 -o,--output <arg>     Relative path where to place output files
 -p,--proto            Export validation results as proto
 -u,--url <arg>        URL to GTFS zipped archive
 -x,--exclude <arg>    Exclude files from semantic GTFS validation
 -z,--zipinput <arg>   if --url is used, where to place the downloaded
                       archive. Otherwise, relative path pointing to a
                       valid GTFS zipped archive on disk

[INFO ] 2020-06-05 16:45:29.586 [main] Main - Took 1059ms
~~~

I'm using this as the command:

`java -jar application/cli-app/build/libs/gtfs-validator-v1.1.0-SNAPSHOT.jar -u https://transitfeeds.com/p/mbta/64/20200531/download -z input.zip -e input -o output`

Does that look right?

barbeau:
Per today's discussion looks like GitHub CI isn't seeing the command-line parameters, so it's just displaying the help text.

barbeau:
Actually, looks like the above issue was CI finding a bug, which today I bumped into on my desktop - https://github.com/MobilityData/gtfs-validator/issues/275. So this works! 👍 

Let's wait for https://github.com/MobilityData/gtfs-validator/issues/275 to be fixed before we review/merge this, just so we can test CI end-to-end.

fabrice-v:
It seems fixing #275 did the trick indeed. It's ready to be merged I think

Issue #229, opened 2020-06-03
[REQUEST] Implement Agency ID presence verification (GTFS rule)


maximearmstrong:
**Is your feature request related to a problem? Please describe.**
An agency ID must be present in agency.txt if there's multiple agencies.

**Describe the solution you'd like**
Verify if there's more than one agency, if yes agency IDs are required.

**Describe alternatives you've considered**

**Additional context**
GTFS Spec http://gtfs.org/reference/static/#agencytxt

Issue #228, opened 2020-06-03
[REQUEST] Implement unique Agency ID verification (GTFS Rule)


maximearmstrong:
**Is your feature request related to a problem? Please describe.**
Every Agency ID should be unique in agency.txt.

**Describe the solution you'd like**
Verify an Agency ID is not duplicated in agency.txt

**Describe alternatives you've considered**

**Additional context**
GTFS Spec http://gtfs.org/reference/static/#agencytxt

Issue #130, opened 2020-04-29
[REQUEST] Implement unique Agencies timezone (GTFS rule)


maximearmstrong:
**Is your feature request related to a problem? Please describe.**
Agencies in a same GTFS dataset must have the same Agency timezone. This is a GTFS rule implemented in Google Python validator and featured in Google Type Error as TYPE_AGENCIES_WITH_DIFFERENT_TIMEZONES.

**Describe the solution you'd like**
Actual Google GTFS validator behaviour : verifies if every agencies are in the same timezone by comparing (standard format)

**Describe alternatives you've considered**

**Additional context**
Line 35 in Error support priorities https://docs.google.com/spreadsheets/d/1vqe6wq7ctqk1EhYkgtZ0_TbcQ91vccfs2daSjn20BLE/edit#gid=0

PR #269, opened 2020-06-08
Use cases classes -- Agency related use cases

lionel-nj:
**Summary:**

This PR provides support to validate rules related to file `agency.txt`. 
- [x] [ Implement Agency ID presence verification](https://github.com/MobilityData/gtfs-validator/issues/229)
- [x] [Implement unique Agency ID verification](https://github.com/MobilityData/gtfs-validator/issues/228). 
This rule has been implemented in 5b05fd9c0aac8a4037ca4ff2b4adbae4d9cf06f6
- [x]  [Implement Agencies timezone ](https://github.com/MobilityData/gtfs-validator/issues/130)

**Expected behavior:** 

### DIFF
--- a/RULES.md
+++ b/RULES.md
@@ -30,7 +30,7 @@ Rules are declared in the [`Notice` module](https://github.com/MobilityData/gtfs
 | [E026](#E026) | Invalid route type | 
 | [E027](#E027) | Missing route short name and long name | 
 | [E028](#E028) | Route long name equals short name | 
-| [E029](#E029) | Missing field `agency_id` | 
+| [E029](#E029) | Missing field `agency_id` for file `agency.txt` with more than 1 record | 
 | [E030](#E030) | Inconsistent field `agency_timezone` | 
 | [E031](#E031) | Invalid `agency_id` | 
 | [E032](#E032) | `calendar.txt` `end_date` is before `start_date` |
@@ -111,7 +111,7 @@ A Route color and a Route text color should be contrasting. Minimum Contrast Rat
 
 <a name="E029"/>
 
-### E029 - Missing `agency_id` 
+### E029 - Missing `agency_id` for file `agency.txt` with more than 1 record
 
 All records of file `agency.txt` should have a non-null value for field [agency_id](https://gtfs.org/reference/static/#agencytxt) when this file counts more than one record.
 
--- a/RULES.md
+++ b/RULES.md
@@ -30,7 +30,7 @@ Rules are declared in the [`Notice` module](https://github.com/MobilityData/gtfs
 | [E026](#E026) | Invalid route type | 
 | [E027](#E027) | Missing route short name and long name | 
 | [E028](#E028) | Route long name equals short name | 
-| [E029](#E029) | Missing field `agency_id` for file agency.txt with more than 1 record | 
+| [E029](#E029) | Missing field `agency_id` | 
 | [E030](#E030) | Inconsistent field `agency_timezone` | 
 
 ### Table of Warnings
@@ -109,7 +109,7 @@ A Route color and a Route text color should be contrasting. Minimum Contrast Rat
 
 <a name="E029"/>
 
-### E029 - Missing field `agency_id` for file agency.txt with more than 1 record 
+### E029 - Missing `agency_id` 
 
 All records of file `agency.txt` should have a non-null value for field [agency_id](https://gtfs.org/reference/static/#agencytxt) when this file counts more than one record.
 
--- a/RULES.md
+++ b/RULES.md
@@ -30,6 +30,7 @@ Rules are declared in the [`Notice` module](https://github.com/MobilityData/gtfs
 | [E026](#E026) | Invalid route type | 
 | [E027](#E027) | Missing route short name and long name | 
 | [E028](#E028) | Route long name equals short name | 
+| [E029](#E029) | Missing field `agency_id` for file agency.txt with more than 1 record | 
 
 ### Table of Warnings
 
@@ -104,6 +105,12 @@ A Route color and a Route text color should be contrasting. Minimum Contrast Rat
 <a name="E028"/>
 
 ### E028 - Route long name equals short name
+
+<a name="E029"/>
+
+### E029 - Missing field `agency_id` for file agency.txt with more than 1 record 
+
+All records of file `agency.txt` should have a non-null value for field [agency_id](https://gtfs.org/reference/static/#agencytxt) when this file counts more than one record.
 
 # Warnings
 
--- a/RULES.md
+++ b/RULES.md
@@ -31,7 +31,7 @@ Rules are declared in the [`Notice` module](https://github.com/MobilityData/gtfs
 | [E027](#E027) | Missing route short name and long name | 
 | [E028](#E028) | Route long name equals short name | 
 | [E029](#E029) | Missing field `agency_id` for file agency.txt with more than 1 record | 
-| [E029](#E030) | Inconsistent field `agency_timezone` | 
+| [E030](#E030) | Inconsistent field `agency_timezone` | 
 
 ### Table of Warnings
 
@@ -112,6 +112,8 @@ A Route color and a Route text color should be contrasting. Minimum Contrast Rat
 ### E029 - Missing field `agency_id` for file agency.txt with more than 1 record 
 
 All records of file `agency.txt` should have a non-null value for field [agency_id](https://gtfs.org/reference/static/#agencytxt) when this file counts more than one record.
+
+<a name="E030"/>
 
 ### E030 - Inconsistent field `agency_timezone` 
 
--- a/RULES.md
+++ b/RULES.md
@@ -32,6 +32,7 @@ Rules are declared in the [`Notice` module](https://github.com/MobilityData/gtfs
 | [E028](#E028) | Route long name equals short name | 
 | [E029](#E029) | Missing field `agency_id` | 
 | [E030](#E030) | Inconsistent field `agency_timezone` | 
+| [E031](#E031) | Invalid `agency_id` | 
 
 ### Table of Warnings
 
@@ -118,6 +119,13 @@ All records of file `agency.txt` should have a non-null value for field [agency_
 ### E030 - Inconsistent field `agency_timezone` 
 
 All records of file `agency.txt` should have the same value for field `agency_timezone` [agency_id](https://gtfs.org/reference/static/#agencytxt) when this file counts more than one record.
+
+
+<a name="E031"/>
+
+### E031 - Invalid `agency_id` 
+
+When provided field `agency_id` should not be blank.
 
 # Warnings
 
--- a/RULES.md
+++ b/RULES.md
@@ -121,7 +121,6 @@ All records of file `agency.txt` should have a non-null value for field [agency_
 
 All records of file `agency.txt` should have the same value for field `agency_timezone` [agency_id](https://gtfs.org/reference/static/#agencytxt) when this file counts more than one record.
 
-
 <a name="E031"/>
 
 ### E031 - Invalid `agency_id` 
--- a/RULES.md
+++ b/RULES.md
@@ -31,6 +31,7 @@ Rules are declared in the [`Notice` module](https://github.com/MobilityData/gtfs
 | [E027](#E027) | Missing route short name and long name | 
 | [E028](#E028) | Route long name equals short name | 
 | [E029](#E029) | Missing field `agency_id` for file agency.txt with more than 1 record | 
+| [E029](#E030) | Inconsistent field `agency_timezone` | 
 
 ### Table of Warnings
 
@@ -111,6 +112,10 @@ A Route color and a Route text color should be contrasting. Minimum Contrast Rat
 ### E029 - Missing field `agency_id` for file agency.txt with more than 1 record 
 
 All records of file `agency.txt` should have a non-null value for field [agency_id](https://gtfs.org/reference/static/#agencytxt) when this file counts more than one record.
+
+### E030 - Inconsistent field `agency_timezone` 
+
+All records of file `agency.txt` should have the same value for field `agency_timezone` [agency_id](https://gtfs.org/reference/static/#agencytxt) when this file counts more than one record.
 
 # Warnings
 
--- a/adapter/exporter/src/main/java/org/mobilitydata/gtfsvalidator/exporter/ProtobufNoticeExporter.java
+++ b/adapter/exporter/src/main/java/org/mobilitydata/gtfsvalidator/exporter/ProtobufNoticeExporter.java
@@ -468,6 +468,7 @@ public class ProtobufNoticeExporter implements NoticeExporter {
         protoBuilder.clear()
                 .setCsvFileName(toExport.getFilename())
                 .setSeverity(GtfsValidationOutputProto.GtfsProblem.Severity.ERROR)
+                .setType(GtfsValidationOutputProto.GtfsProblem.Type.TYPE_AGENCIES_WITH_DIFFERENT_TIMEZONES)
                 .setValue(toExport.getFieldName())
                 .setEntityId(toExport.getConflictingTimezoneCollection())
                 .setEntityValue(String.valueOf(toExport.getDistinctTimezoneCount()))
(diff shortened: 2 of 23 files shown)


RULES.md
```diff
@@ -30,6 +30,8 @@ Rules are declared in the [`Notice` module](https://github.com/MobilityData/gtfs
 | [E026](#E026) | Invalid route type | 
 | [E027](#E027) | Missing route short name and long name | 
 | [E028](#E028) | Route long name equals short name | 
+| [E029](#E029) | Missing field `agency_id` for file agency.txt with more than 1 record | 
+| [E029](#E030) | Inconsistent field `agency_timezone` |
```
lionel-nj, line 34: E029 -> E030

RULES.md
```diff
@@ -105,6 +107,16 @@ A Route color and a Route text color should be contrasting. Minimum Contrast Rat
 
 ### E028 - Route long name equals short name
 
+<a name="E029"/>
+
+### E029 - Missing field `agency_id` for file agency.txt with more than 1 record 
+
+All records of file `agency.txt` should have a non-null value for field [agency_id](https://gtfs.org/reference/static/#agencytxt) when this file counts more than one record.
+
+### E030 - Inconsistent field `agency_timezone`
```
lionel-nj, line 116: Add anchor linking to E030

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateAgencyIdRequirement.java
```diff
@@ -0,0 +1,65 @@
...
+     * when said file counts more than one record. If a row from agency.txt has no value for field agency_id, a new
+     * error notice is generated and added to the {@code ValidationResultRepository} provided in the constructor
+     */
+    public void execute() {
+        logger.info("Validating rule 'E029 - Missing field `agency_id` for file agency.txt with more than 1 record'"
+                + System.lineSeparator());
+        final Collection<Agency> agencyPerId = dataRepo.getAgencyAll();
+        if (agencyPerId.size() > 1) {
+            agencyPerId.stream()
+                    .filter(agency -> agency.getAgencyId() == null)
```
maximearmstrong, line 60: Should we also check for empty strings?

adapter/exporter/src/main/java/org/mobilitydata/gtfsvalidator/exporter/ProtobufNoticeExporter.java
```diff
@@ -453,6 +453,29 @@ public void export(RouteLongNameContainsShortNameNotice toExport) throws IOExcep
                 .writeTo(streamGenerator.getStream());
     }
 
+    @Override
+    public void export(final MissingAgencyIdNotice toExport) throws IOException {
+        protoBuilder.clear()
+                .setCsvFileName(toExport.getFilename())
+                .setSeverity(GtfsValidationOutputProto.GtfsProblem.Severity.ERROR)
+                .setValue(toExport.getFieldName())
```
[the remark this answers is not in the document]
lionel-nj: For now, yes. @fabrice-v

RULES.md
```diff
@@ -30,6 +30,8 @@ Rules are declared in the [`Notice` module](https://github.com/MobilityData/gtfs
 | [E026](#E026) | Invalid route type | 
 | [E027](#E027) | Missing route short name and long name | 
 | [E028](#E028) | Route long name equals short name | 
+| [E029](#E029) | Missing field `agency_id` for file agency.txt with more than 1 record |
```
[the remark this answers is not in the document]
lionel-nj: Good point for the consistency. 

Also, as you said I'm not sure that adding more detail here is very useful. This could be in the rule description. Modifying this.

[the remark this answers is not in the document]
lionel-nj: However, to keep consistency with the GTFS specification, it would be good to format the field names. For example: `E026 - Invalid route type` ->  `E026 - Invalid route_type`.

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateAgenciesHaveSameAgencyTimezone.java
```diff
@@ -0,0 +1,65 @@
...
+    /**
+     * Use case execution method: checks if every record of file agency.txt have the same value for field agency_timezone
+     * If a row from agency.txt has a different value for said field an error notice is generated and added to the
+     * {@code ValidationResultRepository} provided in the constructor
+     */
+    public void execute() {
+        logger.info("Validating rule 'E030 - Different 'agency_timezone'" + System.lineSeparator());
+        final Set<String> timezoneCollection = new HashSet<>();
+        dataRepo.getAgencyAll()
+                .forEach(agency -> timezoneCollection.add(agency.getAgencyTimezone()));
```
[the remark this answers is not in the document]
lionel-nj: `agency_timezone` in a required field. At this step of the validation process we manipulate `Agency` entities: when building the entity, if the field is missing, a notice is generated, and the entity is not added to the data repo.

Long story short, all `Agency` entities have a non null value for this field. :)

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateAgencyIdRequirement.java
lionel-nj: As discussed, yes.

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateAgenciesHaveSameAgencyTimezoneTest.java
```diff
@@ -0,0 +1,144 @@
...
+
+        final List<InconsistentAgencyTimezoneNotice> noticeList = captor.getAllValues();
+
+        assertEquals("agency.txt", noticeList.get(0).getFilename());
+        assertEquals("agency_timezone", noticeList.get(0).getFieldName());
+        assertEquals("no id", noticeList.get(0).getEntityId());
+        assertEquals(2, noticeList.get(0).getDistinctTimezoneCount());
+        assertTrue(noticeList.get(0).getConflictingTimezoneCollection().contains("timezone 00"));
+        assertTrue(noticeList.get(0).getConflictingTimezoneCollection().contains("timezone 01"));
+
```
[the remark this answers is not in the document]
lionel-nj: Assertions are used here to check the content of the notice. Should this be done in another layer @fabrice-v?

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateAgencyIdRequirement.java
lionel-nj: See 6d7c8a3a780afbb6e76d4b18624649db5dc2be24


https://github.com/MobilityData/gtfs-validator/blob/6d7c8a3a780afbb6e76d4b18624649db5dc2be24/domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/Agency.java#L237

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/Agency.java
```diff
@@ -232,8 +233,11 @@ public AgencyBuilder agencyEmail(@Nullable final String agencyEmail) {
         public EntityBuildResult<?> build() {
             noticeCollection.clear();
 
-            if (agencyName == null || agencyUrl == null || agencyTimezone == null) {
-
+            if (agencyName == null || agencyUrl == null || agencyTimezone == null ||
+                    (agencyId!=null && agencyId.isBlank())) {
+                if (agencyId!=null && agencyId.isBlank()) {
+                    noticeCollection.add(new InvalidAgencyIdNotice(agencyId));
```
maximearmstrong, line 239: I'm not sure to understand the logic behind this. An agency ID null or blank is valid if there is only 1 agency in agency.txt. Maybe this should be checked in the use case instead.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/notice/error/InvalidAgencyIdNotice.java
```diff
@@ -0,0 +1,43 @@
...
+import java.io.IOException;
+
+public class InvalidAgencyIdNotice extends ErrorNotice {
+    private String fieldName;
+
+    public InvalidAgencyIdNotice(final String entityId) {
+        super("agency.txt", E_031,
+                "Invalid blank string",
+                "Field `agency_id` of file `agency.txt` is blank",
+                entityId);
```
maximearmstrong, line 31: What is the difference with Missing `agency_id`?

domain/src/test/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/AgencyTest.java
```diff
@@ -131,4 +133,32 @@ public void createAgencyWithValidValuesShouldNotGenerateNotice() {
...
+                .build();
+
+        assertTrue(entityBuildResult.getData() instanceof List);
+        // suppressed warning since cast check is not useful here: the test is designed so that method .getData()
+        // returns a list of notices.
+        //noinspection unchecked
+        final List<InvalidAgencyIdNotice> noticeCollection =
+                (List<InvalidAgencyIdNotice>) entityBuildResult.getData();
+        final InvalidAgencyIdNotice notice = noticeCollection.get(0);
+
```
maximearmstrong, line 158: Same, I'm not sure this should be done in the entity as a blank or null ID should be valid for a sole agency.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/Agency.java
lionel-nj: When building an `Agency` entity, field `agency_id` can be null. As `agency_id` is conditionally required, we check this condition in a use case.

> This field is required when the dataset contains data for multiple transit agencies, otherwise it is optional

However,  if provided, it can not be left blank. Therefore it is considered as non-valid in this implementation.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/notice/error/InvalidAgencyIdNotice.java
lionel-nj: `MissingAgencyId` refers to the fact that field `agency_id` is null, in a `agency.txt` file containing more than one record.

`InvalidAgencyId` refers to the fact that `agency_id` is provided but blank, for example: 
`"   "`

domain/src/test/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/AgencyTest.java
lionel-nj: See https://github.com/MobilityData/gtfs-validator/pull/269#discussion_r437533627

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateAgenciesHaveSameAgencyTimezoneTest.java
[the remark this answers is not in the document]
lionel-nj: After discussion, it seems relevant to verify the content of notices via assertions

RULES.md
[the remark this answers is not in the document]
barbeau: I believe agency_id isn't required if only one agency is defined in agency.txt? In that case, I'd leave the last part of the rule description, otherwise you could interpret the rule differently.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/Agency.java
lionel-nj: If provided, `agency_id` should not be blank or empty "`   `" or `""`. That's the reason why this check is done in the builder. We only check the presence of a valid `agency_id` (if there is more than one record in `agency.txt` in the related use case `ValidateAgencyIdRequirement`.

domain/src/test/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/AgencyTest.java
lionel-nj: See https://github.com/MobilityData/gtfs-validator/pull/269#discussion_r443566311

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/Agency.java
lionel-nj: Thinking more about this, the specification says that `agency_id` can be any sequence of UTF-8 characters. `SPACE` qualifies as a UTF-8 character, so my remark might not be relevant here (http://www.fileformat.info/info/charset/UTF-8/list.htm). 

Leaving this discussion open for further discussion.

barbeau: @lionel-nj I agree with your comment - the intent of the spec is that the agency_id shouldn't be empty or whitespace - it needs to be something visible to the end user.

adapter/exporter/src/main/java/org/mobilitydata/gtfsvalidator/exporter/ProtobufNoticeExporter.java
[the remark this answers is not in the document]
fabrice-v: For now we should as modeling protobuf issues beyond what the Python validator handles is not top priority

adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/db/InMemoryGtfsDataRepository.java
```diff
@@ -119,6 +119,16 @@ public Agency getAgencyById(final String agencyId) {
...
     }
 
+    /**
+     * Return a collection of Agency objects representing all the rows from agency.txt
+     *
+     * @return a collection of Agency objects representing all the rows from agency.txt
+     */
+    @Override
+    public Collection<Agency> getAgencyAll() {
+        return agencyPerId.values();
```
fabrice-v, line 129: Shall we apply the principle of exposing only an immutable version? I am not sure that's the case here

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateAgencyIdRequirement.java
```diff
@@ -0,0 +1,65 @@
...
+
+    /**
+     * Use case execution method: checks if every record of file agency.txt has a non null value for field agency_id
+     * when said file counts more than one record. If a row from agency.txt has no value for field agency_id, a new
+     * error notice is generated and added to the {@code ValidationResultRepository} provided in the constructor
+     */
+    public void execute() {
+        logger.info("Validating rule 'E029 - Missing field `agency_id` for file agency.txt with more than 1 record'"
+                + System.lineSeparator());
+        final Collection<Agency> agencyPerId = dataRepo.getAgencyAll();
```
fabrice-v, line 57: If it's just for a size check, prefer having a method `getAgencyCount` that would return the result. In that case I understand it may be for consistency or this method is required anyway, so as discussed we'd rather expose an immutable version of the underlying data structure.

adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/db/InMemoryGtfsDataRepository.java
lionel-nj: Absolutely. Indeed, here this is not the case. Modifying this.

adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/db/InMemoryGtfsDataRepository.java
```diff
@@ -135,6 +136,26 @@ public Agency getAgencyById(final String agencyId) {
         return agencyPerId.get(agencyId);
     }
 
+    /**
+     * Return a collection of Agency objects representing all the rows from agency.txt
```
fabrice-v, line 140: Mention immutability of the returned collection

lionel-nj: Good point, see 02f200bd43e833c2b79b2775665dc84f797c9427

PR #251, opened 2020-06-05
Add documentation on how to add a rule

fabrice-v:
**Summary:**

This PR adds documentation regarding the required steps to add a new validation rule to the validator. It is heavily inspired by https://github.com/CUTR-at-USF/gtfs-realtime-validator/blob/master/ADDING_NEW_RULES.md

PR #272, opened 2020-06-10
Fix #251 - New rule - `calendar.txt` `end_date` is before `start_date`

### LINKED ISSUE #133: [REQUEST] Implement Start/End dates verification for calendar (GTFS rule)
Start date must be before End date in calendar. This is a GTFS rule implemented in Google Python validator and featured in Google Type Error as TTYPE_CALENDAR_START_AND_END_DATE_OUT_OF_ORDER.

Actual Google GTFS validator behaviour : verifies if end date is smaller than start date.

**Additional context**
Line 40 in Error support priorities https://docs.google.com/spreadsheets/d/1vqe6wq7ctqk1EhYkgtZ0_TbcQ91vccfs2daSjn20BLE/edit#gid=0

barbeau:
**Summary:**

~~*Please don't merge - work-in-progress*~~

This PR implements a new validation rule for https://github.com/MobilityData/gtfs-validator/issues/133 - checking if `calendar.txt` `end_date` is before `start_date`.

I'm following the documentation at https://github.com/MobilityData/gtfs-validator/pull/251.

Done (Steps 1-4):
- [x] Rules.md
- [x] Notice impl
- [x] Exporters and tests impl

TODO (Steps 5-8):
- [x] Use case for rule impl
- [x] Use case test impl
- [x] Instantiate use case
- [x] Retrieve and execute use case
- [x] Resolve discussion for null test case on GTFS required fields (e.g., remove commented code)

Closes #133

**Expected behavior:** 

Execute a new rule for https://github.com/MobilityData/gtfs-validator/issues/133 checking if the calendar.txt record end date is before the start date.

**Other thoughts so far:** 
* Exporter test code, especially for protocol buffers, includes a lot of duplication
* In `ProtobufNoticeExporter`, maybe move inner class `ProtobufOutputStreamGenerator` to the top of the file so that new `export()` methods can just be added to the bottom of the file?
* Testing notice message output text would be useful - we hit a number of output message formatting bugs on the GTFS-RT validator, and it's always been on my TODO list to go back and add more unit tests for this. This would have also helped me catch an output issue related to https://github.com/MobilityData/gtfs-validator/issues/276 earlier.
* For use cases, each rule requires a loop through the entire list of entities being examined (e.g., loop through all routes). For `m` rules and `n` entities, that's a time complexity of O(m*n). You can probably get away with that for tables that typically have a small amount of records, but this design will definitely have a performance hit for tables with lots of records like shapes.txt, stop_times.txt, and (in some feed) calendar_dates.txt. We may want to consider an alternate design for rules on these tables where multiple rules are implemented in a single use case.
* For each use case, it would be helpful to document the rule ID (e.g., `E001`) in the Javadocs for the class.
* There should be an interface for rules (use cases) to enforce contract for methods
* In `DefaultConfig`, maybe move all validation rules to the end of the file so new rules just get added to the bottom? Right now it seems like they should be added in the middle of the class with the other `Validate*` methods
* Seems like it would greatly simplify the use of the configuration to support a `DefaultConfig.validateAll()` method for all semantic rules, rather than having to call all the individual `config.validate*()` methods

### DIFF
--- a/usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateCalendarEndDateBeforeStartDateNoticeTest.java
+++ b/usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateCalendarEndDateBeforeStartDateNoticeTest.java
@@ -1,0 +1,167 @@
+/*
+ * Copyright (c) 2020. MobilityData IO.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.mobilitydata.gtfsvalidator.usecase;
+
+import org.apache.logging.log4j.Logger;
+import org.junit.jupiter.api.Test;
+import org.mobilitydata.gtfsvalidator.domain.entity.gtfs.Calendar;
+import org.mobilitydata.gtfsvalidator.domain.entity.notice.error.CalendarEndDateBeforeStartDateNotice;
+import org.mobilitydata.gtfsvalidator.usecase.port.GtfsDataRepository;
+import org.mobilitydata.gtfsvalidator.usecase.port.ValidationResultRepository;
+
+import java.time.LocalDateTime;
+import java.util.List;
+
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.*;
+
+/**
+ * Unit tests to validate behavior of use case for "E032 - calendar.txt end_date is before start_date"
+ */
+class ValidateCalendarEndDateBeforeStartDateNoticeTest {
+
+    // TODO - below test throws an NPE, but calendar.txt start_date and end_date are required and marked @NonNull,
+    // so checking for them within the use case is flagged as "these fields are always non-null".
+    // Do we include a null test for required GTFS fields?
+
+//    @Test
+//    void nullCalendarStartAndEndDateShouldNotGenerateNotice() {
+//        Calendar mockCalendar = mock(Calendar.class);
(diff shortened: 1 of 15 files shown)


usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateCalendarEndDateBeforeStartDateNoticeTest.java
```diff
@@ -0,0 +1,167 @@
...
+
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.*;
+
+/**
+ * Unit tests to validate behavior of use case for "E032 - calendar.txt end_date is before start_date"
+ */
+class ValidateCalendarEndDateBeforeStartDateNoticeTest {
+
+    // TODO - below test throws an NPE, but calendar.txt start_date and end_date are required and marked @NonNull,
```
barbeau, line 37: Flagging this as a question for discussion

lionel-nj: IMO, we manipulate `Calendar` entities in the use case, given that these entities are only built if `non-null` fields are actually `non-null`; if I am correct, we could skip null tests for required GTFS fields.
However, we should include them for `optional` GTFS fields.

barbeau:
Ok, this should be ready for full review! 

@fabrice-v if you agree with @lionel-nj's comment on testing for null on GTFS required fields at https://github.com/MobilityData/gtfs-validator/pull/272#discussion_r438981672 I'll remove that commented-out code (I didn't want to remove prior to your comment to avoid removing discussion thread).

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateCalendarEndDateBeforeStartDateNoticeTest.java
fabrice-v: I agree with @lionel-nj , we have builders that shield us from having nulls downstream, hence the NPE is to be expected.

barbeau: Sounds good! I just removed this commented-out test in https://github.com/MobilityData/gtfs-validator/pull/272/commits/8925240802796cb7ed6b315e82196279af7dd3ac.

PR #273, opened 2020-06-11
Notice rework

fabrice-v:
**Summary:**

- encapsulate notice specific fields in a `noticeSpecific` object
- fix contrastRatio output as string instead of numeric value
- remove field `noticeId`
- add field `level`, static for each derived class
- add field `code`, defining specific Notice for a level

**Expected behavior:** 

Notice should have a noticeSpecific object in their JSON export

**additional context**
https://github.com/MobilityData/gtfs-validator/pull/251#discussion_r438356211

image


adapter/exporter/src/main/java/org/mobilitydata/gtfsvalidator/exporter/ProtobufNoticeExporter.java
```diff
@@ -63,7 +65,7 @@ public void export(final NonStandardHeaderNotice toExport) throws IOException {
                 .setCsvFileName(toExport.getFilename())
                 .setType(GtfsValidationOutputProto.GtfsProblem.Type.TYPE_UNKNOWN_COLUMN)
                 .setSeverity(GtfsValidationOutputProto.GtfsProblem.Severity.SUSPICIOUS_WARNING)
-                .setAltEntityValue(toExport.getExtraHeader())
+                .setAltEntityValue((String) toExport.getNoticeSpecific(NOTICE_SPECIFIC_KEY__EXTRA_HEADER_NAME))
```
barbeau, line 68: Looks like there is an extra `_` in `NOTICE_SPECIFIC_KEY__EXTRA_HEADER_NAME`? Or was that intentional to separate the key name?

barbeau: Looking at other entries below, I'm guessing it's intentional. Maybe it would be better to just remove most of the prefix, so you're left with something like `KEY_EXTRA_HEADER_NAME`?

adapter/parser/src/test/java/org/mobilitydata/gtfsvalidator/parser/GtfsEntityParserTest.java
```diff
@@ -381,7 +381,7 @@ void invalidIntegerValidationShouldGenerateError() {
 
         ErrorNotice notice = result.get(0);
         assertThat(notice, instanceOf(CannotParseIntegerNotice.class));
-        assertEquals("E005", notice.getId());
+        assertEquals(5, notice.getCode());
```
barbeau, line 384: If error level is removed from the ID, shouldn't the test also check for the error level?

fabrice-v: absolutely!

adapter/exporter/src/main/java/org/mobilitydata/gtfsvalidator/exporter/ProtobufNoticeExporter.java
fabrice-v: Streamlining indeed, will do.

PR #281, opened 2020-06-12
Fix #275 -- Handle absence of execution-parameters.json

### LINKED ISSUE #275: [BUG] Execution without execution-parameters.json always shows help text
**Describe the bug**
If I run the validator with the command-line parameters shown in the README:

`-u https://transitfeeds.com/p/mbta/64/latest/download -z input.zip -e input -o output`

..but *without* the `execution-parameters.json` file in my working directory, the validator executes but just shows me the help text - it doesn't parse any of the command-line parameters.

This is because internally it's throwing a `NoSuchFileException`, which is being caught in the try/catch block in `Main`:

image

**To Reproduce**
1. Make sure you don't have a `execution-parameters.json` file in your working directory
1. Execute the validator with any command-line parameters (e.g., `-u https://transitfeeds.com/p/mbta/64/latest/download -z input.zip -e input -o output`)

**Expected behavior**
Validator should parse command-line parameters and execute

**Witnessed behavior**
Command-line parameters aren't parsed, and internally a `java.nio.file.NoSuchFileException` is thrown.

**Screenshots**
See above

**Environment used**
- OS: Windows 10
- Java version: JDK 14

**Additional context**
Somewhat ironically, it looks like this is the problem we were seeing on the PR to execute the validator on CI - https://github.com/MobilityData/gtfs-validator/pull/253. So the good news is the CI works!

lionel-nj:
**Summary:**

This PR provides support to avoid the display of help text when not providing configuration file `execution-parameters.json`.

A new variable `executionParametersAsString` is defined. It is `null` when the file could not be find, or the `String `representation of the `.json` configuration file.

This variables is defined through a `try/catch`  block.

Additional logging information is provided regarding the absence/presence of said file.

**Expected behavior:** 

When file `execution-parameter.json` is not provided 2 possible behavior are expected:
- command line options have been provided -> we use them as execution parameters

- command line options have not been provided -> default values for execution parameters should be used.

In both cases message `"Configuration file execution-parameters.json not found in working directory"` should be logged.

Capture d’écran, le 2020-06-15 à 10 56 12

In the opposite scenario, if file `execution-parameters.json` is provided, message `"Configuration file execution-parameters.json found in working directory"` should be logged.

Capture d’écran, le 2020-06-15 à 10 54 11

### DIFF
--- a/adapter/parser/src/main/java/org/mobilitydata/gtfsvalidator/parser/JsonExecParamParser.java
+++ b/adapter/parser/src/main/java/org/mobilitydata/gtfsvalidator/parser/JsonExecParamParser.java
@@ -18,6 +18,7 @@ package org.mobilitydata.gtfsvalidator.parser;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectReader;
+import com.google.common.base.Strings;
 import org.apache.logging.log4j.Logger;
 import org.mobilitydata.gtfsvalidator.domain.entity.ExecParam;
 import org.mobilitydata.gtfsvalidator.usecase.port.ExecParamRepository;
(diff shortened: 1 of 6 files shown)


barbeau:
I'd also suggest changing the logging statements to make if very clear whether the config file or command-line parameters are being used - right now if both are provided it just says it's parsing both. In that same area of the log, I'd suggest being very explicit which is being used.

lionel-nj:
I updated the screenshots in the PR description, the old one were misleading. Could you check and tell me if this is what you had in mind please @barbeau ?

barbeau:
@lionel-nj Thanks, that's better 👍.

Issue #290, opened 2020-06-19
[REQUEST] implement `clear` method for GTFS entities builder


lionel-nj:
**Is your feature request related to a problem? Please describe.**

There should be a Builder.clear() method that resets all the fields of a builder. Otherwise it is implied that the calling classes will know to reset all of the fields manually themselves, or always call all of the builder fields and pass in null values when the field is not provided. 

This behavior can be confusing and induce future bugs. 

**Describe the solution you'd like**
Add a `clear` method to builders for all GTFS entities. Adapt existing tests.

**Additional context**
See PR discussion https://github.com/MobilityData/gtfs-validator/pull/70#discussion_r442940292

PR #293, opened 2020-06-22
Implement #290 - Improve GTFS entity builders

lionel-nj:
**Summary:**

This PR provides support to add a `clear` method to GTFS entities builder. Calling said method sets all builder's field to `null`. This PR also provides support to use static method to generate mapping key of `Attribution` GTFS entities inside `AttributionBuilder.build` method

**Expected behavior:** 

Fields of builders should be set to `null` after calling said method.
Attribution Mapping key should be generated via static method

Please note that these methods are not tested because builders' fields are private, and we chose  not to provide getters on builders.

### DIFF
--- a/domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/Agency.java
+++ b/domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/Agency.java
@@ -255,7 +255,7 @@ public class Agency extends GtfsEntity {
          * Method to reset all fields of builder. Returns builder with all fields set to null.
          * @return builder with all fields set to null;
          */
-        public AgencyBuilder clearFieldAll() {
+        public AgencyBuilder clear() {
             agencyId = null;
             agencyName = null;
             agencyUrl = null;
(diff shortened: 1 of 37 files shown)


domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/Attribution.java
```diff
@@ -370,6 +370,29 @@ public AttributionBuilder attributionPhone(@Nullable final String attributionPho
                         attributionPhone));
             }
         }
+
+        /**
+         * Method to reset all fields of builder. Returns builder with all fields set to null.
+         * @return builder with all fields set to null;
+         */
+        public AttributionBuilder clearFieldAll() {
```
fabrice-v, line 378: I'd suggest to name it `clear` to mimick the standard library

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/Agency.java
```diff
@@ -252,5 +250,22 @@ public AgencyBuilder agencyEmail(@Nullable final String agencyEmail) {
                         agencyPhone, agencyFareUrl, agencyEmail));
             }
         }
+
+        /**
+         * Method to reset all fields of builder. Returns builder with all fields set to null.
+         * @return builder with all fields set to null;
+         */
+        public AgencyBuilder clearFieldAll() {
```
fabrice-v, line 258: I'd suggest to name it `clear` to mimick the standard library

PR #294, opened 2020-06-23
Fix #288 - Color contrast validation bug

### LINKED ISSUE #288: [BUG] Validator crashes on color luminance check
**Describe the bug**
The validator crashes at the beginning of validation process.

**To Reproduce**
-

**Expected behavior**
The validator should not crash.

**Witnessed behavior**
The validator crashes.

Here is the command arguments, log output and stacktrace which might be useful

```
$ java -jar gtfs-validator-v1.2.0.jar -i Archive.zip -o gtfs_out -e gtfs_extract
[WARN ] 2020-06-19 23:15:37.112 [main] Main - Configuration file execution-parameters.json not found in working directory

[INFO ] 2020-06-19 23:15:37.254 [main] Main - Retrieving execution parameters from command-line

[INFO ] 2020-06-19 23:15:37.264 [main] Main - Unzipping archive

[INFO ] 2020-06-19 23:15:37.436 [main] Main - No file to exclude -- will execute validation process on all files

[INFO ] 2020-06-19 23:15:37.442 [main] Main - List of filenames to exclude is: []

[INFO ] 2020-06-19 23:15:37.443 [main] Main - Will execute validation on the following subset of files: [agency.txt, stops.txt, routes.txt, trips.txt, stop_times.txt, fare_attributes.txt, calendar.txt, fare_rules.txt, shapes.txt]

[INFO ] 2020-06-19 23:15:46.281 [main] Main - Validating rule 'W005 - Route short name too long'

[INFO ] 2020-06-19 23:15:46.284 [main] Main - Validating rule 'E025 - Insufficient route color contrast'

Exception in thread "main" java.lang.StringIndexOutOfBoundsException: begin 0, end 2, length 1
        at java.base/java.lang.String.checkBoundsBeginEnd(String.java:3319)
        at java.base/java.lang.String.substring(String.java:1874)
        at org.mobilitydata.gtfsvalidator.usecase.ValidateRouteColorAndTextContrast.luminance(ValidateRouteColorAndTextContrast.java:66)
        at org.mobilitydata.gtfsvalidator.usecase.ValidateRouteColorAndTextContrast.contrast(ValidateRouteColorAndTextContrast.java:82)
        at org.mobilitydata.gtfsvalidator.usecase.ValidateRouteColorAndTextContrast.areContrasting(ValidateRouteColorAndTextContrast.java:100)
        at org.mobilitydata.gtfsvalidator.usecase.ValidateRouteColorAndTextContrast.lambda$execute$0(ValidateRouteColorAndTextContrast.java:56)
        at java.base/java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:176)
        at java.base/java.util.HashMap$ValueSpliterator.forEachRemaining(HashMap.java:1675)
        at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
        at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
        at java.base/java.util.stream.ForEachOps$ForEachOp.evaluateSequential(ForEachOps.java:150)
        at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.evaluateSequential(ForEachOps.java:173)
        at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
        at java.base/java.util.stream.ReferencePipeline.forEach(ReferencePipeline.java:497)
        at org.mobilitydata.gtfsvalidator.usecase.ValidateRouteColorAndTextContrast.execute(ValidateRouteColorAndTextContrast.java:57)
        at org.mobilitydata.gtfsvalidator.Main.main(Main.java:140)
```

**Environment used**
- OS: Ubuntu 20.04 LTS
- Java version: openjdk 11.0.7 2020-04-14
- gtfs-validator version: v1.2.0
 
**Additional context**

maximearmstrong:
**Summary:**

Invalid values passed to ValidateRouteColorAndTextContrast class were raising a StringIndexOutOfBoundsException and Validator was crashing.

**Expected behavior:** 

This PR fixes the bug by implementing a regex validation with an Apache RegexValidator before testing the contrast. The regex validation is the same as the one implemented in GtfsTypeValidator and InMemoryGtfsSpecRepository.

- [x] Fix ValidateRouteColorAndTextContrast
- [x] Unit tests are modified and `gradle clean build` passes all tests.


usecase/build.gradle
```diff
@@ -34,6 +34,7 @@ dependencies {
     implementation 'commons-cli:commons-cli:1.4'
     implementation 'org.apache.logging.log4j:log4j-api:2.12.1'
     implementation 'org.apache.logging.log4j:log4j-core:2.12.1'
+    implementation 'commons-validator:commons-validator:1.6'
```
barbeau, line 37: We should standardize on regex implementations. Apache commons is fine, but if we use that, we should use it everywhere.

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateRouteColorAndTextContrast.java
```diff
@@ -96,9 +100,24 @@ private double contrast(String routeColor, String textColor) {
...
     private boolean areContrasting(String routeColor, String textColor) {
         boolean areContrasting = true;
-        if (routeColor != null && textColor != null) {
+        if (colorValidator.isValid(routeColor) && colorValidator.isValid(textColor)) {
             areContrasting = contrast(routeColor, textColor) >= 4.5;
         }
         return areContrasting;
     }
+
+    /**
```
barbeau, line 109: Javadoc needs to be updated

usecase/build.gradle
fabrice-v: Good point. I guess we shall open an issue of tracking down current regex usage?

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateRouteColorAndTextContrast.java
maximearmstrong: The method was removed

usecase/build.gradle
lionel-nj: See https://github.com/MobilityData/gtfs-validator/issues/305

adapter/parser/src/main/java/org/mobilitydata/gtfsvalidator/parser/GtfsEntityParser.java
```diff
@@ -185,6 +189,13 @@ public ParsedEntity parse(RawEntity toParse) {
                                 ));
                     }
 
+                } else if (columnSpecProto.getType().getType() ==
```
fabrice-v, line 192: My understanding of the logic is that line 200 `contentByHeaderMap.put(columnSpecProto.getName(), rawField);` was supposed to let through any color raw value and that invalid color would be caught by https://github.com/MobilityData/gtfs-validator/blob/e2e59464ea7955feffa5af8ed5a2b48fc95a5104/adapter/validator/src/main/java/org/mobilitydata/gtfsvalidator/validator/GtfsTypeValidator.java#L136
Here it seems we would discard invalid raw values without generating an output

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateRouteColorAndTextContrastTest.java
```diff
@@ -85,7 +85,7 @@ void nullRouteTextColorShouldNotGenerateNotice() {
         verify(mockRoute, times(1)).getRouteColor();
         verify(mockRoute, times(1)).getRouteTextColor();
         verify(mockLogger, times(1)).info("Validating rule 'E025 - Insufficient route color" +
-                " contrast'"+ System.lineSeparator());
+                " contrast'" + System.lineSeparator());
```
fabrice-v, line 88: I think formatting character shall not end up in our JSON output

barbeau: I agree with @fabrice-v.

I think generally our JSON output is compact right now without any extra whitespace?

I've seen other tools implement a `--pretty` command-line parameter that adds new-lines at the end of each JSON line to make it easier to read. I think that would be a good option for this tool as well.

fabrice-v: Yes, right now we try to avoid any formatting into the output JSON.
Maybe we would implement prettyfiing but I would see it as low priority because there are tools, some of them online, that can also do it

barbeau: Agreed in terms of priority. Although, looks like enabling pretty print via Jackson is one line of code:
https://mkyong.com/java/how-to-enable-pretty-print-json-output-jackson/

The main effort on our end would be setting up and passing the command-line parameter.

Issue #209, opened 2020-05-11
[REQUEST] Implement route associated to a trip exists in routes list


maximearmstrong:
**Is your feature request related to a problem? Please describe.**
A trip should have a route existing in schedule routes list. This is a GTFS rule implemented in Google Python validator as ValidateRouteIdExistsInRouteList.

**Describe the solution you'd like**
Actual Google GTFS validator behaviour : verifies if route ID is in schedule routes.
https://github.com/google/transitfeed/blob/d727e97cb66ac2ca2d699a382ea1d449ee26c2a1/transitfeed/trip.py#L531

**Describe alternatives you've considered**

**Additional context**
Line 199 in Error support priorities https://docs.google.com/spreadsheets/d/1vqe6wq7ctqk1EhYkgtZ0_TbcQ91vccfs2daSjn20BLE/edit#gid=0

PR #298, opened 2020-06-30
Implement #209 -- Validate route id reference in trips

lionel-nj:
**Summary:**

This PR provides support to check cross reference between files `trips.txt` and `routes.txt`.

**Expected behavior:** 

Records from `trips.txt` referring to non-existing records of `routes.txt` should generate a notice `NonExistingRouteIdNotice`.


RULES.md
```diff
@@ -34,6 +34,7 @@ Rules are declared in the [`Notice` module](https://github.com/MobilityData/gtfs
 | [E030](#E030) | Inconsistent field `agency_timezone` | 
 | [E031](#E031) | Invalid `agency_id` | 
 | [E032](#E032) | `calendar.txt` `end_date` is before `start_date` |
+| [E032](#E033) | Reference to non-existing `route_id` in `trips.txt` |
```
barbeau, line 37: I would suggest we adopt a broader definition of this rule, like "`route_id` missing from routes.txt" or "Invalid route_id" or "route_id does not exist in routes.txt" that would cover all instances of a route_id used in a different file that doesn't exist in routes.txt.

route_id is used in many several files, and given the number of rules that will reference different IDs across different files, we're doing to end up with a lot of rules just of bad ID references. I think this unnecessarily clutters the rule list and makes it harder to find other semantic rules. Plus, we can trace which file the error is coming from from the message itself, right?

Also, related to the terminology of the rule name, we ended up using the following convention in the RT validator:
"GTFS-rt trip_id does not exist in GTFS data"

After a lot of discussion this seemed easier to understand that using words like "reference" or "non-existing" (which we also used at some point).

lionel-nj: True, and yes in the notice description we can track back the problematic file.

fabrice-v: I agree with the stated points, I suppose they are not implemented yet and will pause my review of this PR

RULES.md
```diff
@@ -136,6 +137,12 @@ In `calendar.txt`, the `end_date` of a service record must not be earlier than t
 #### References:
 * [calendar.txt specification](https://gtfs.org/reference/static/#calendartxt)
 
+<a name="E033"/>
+
+### E033 - Invalid `route_id`
```
fabrice-v, line 142:
```suggestion
### E033 - `route_id` not found
```

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateTripRouteIdTest.java
```diff
@@ -0,0 +1,128 @@
...
+
+        underTest.execute();
+
+        verify(mockLogger, times(1)).info(ArgumentMatchers.eq(
+                "Validating rule E033 - Invalid `route_id`" + System.lineSeparator()));
+
+        verify(mockDataRepo, times(1)).getRouteAll();
+        verify(mockDataRepo, times(1)).getTripAll();
+
+        // suppress warning regarding ignored result of method since it is not necessary here.
```
fabrice-v, line 62: I think having just one explanatory comment would be enough. I'd suggest keeping that one and getting rid of the others

Issue #163, opened 2020-05-04
Implement trip distance vs. shape verification (GTFS rule)


maximearmstrong:
**Is your feature request related to a problem? Please describe.**
The distance traveled by a trip should be less or equal to the max length of its shape. This is a GTFS rule implemented in Google Python validator and featured in Google Type Error as TYPE_TRIP_WITH_OUT_OF_ORDER_SHAPE_DIST_TRAVELED.

**Describe the solution you'd like**
Actual Google GTFS validator behaviour : verifies if distance travelled is smaller than shape length. If not, generate an ERROR.

**Describe alternatives you've considered**

**Additional context**
Line 141 in Error support priorities https://docs.google.com/spreadsheets/d/1vqe6wq7ctqk1EhYkgtZ0_TbcQ91vccfs2daSjn20BLE/edit#gid=0

barbeau:
@maximearmstrong Could you point us to where this rule is implemented in the Python validator? I don't see it from a quick search.

maximearmstrong:
@barbeau Here https://github.com/google/transitfeed/blob/master/transitfeed/trip.py#L644

isabelle-dr:
Labeling as blocked because there is no mention of this in the spec or best practices

isabelle-dr:
We have a PR open to add this to the spec: https://github.com/google/transit/pull/380, which would unblock this issue for the validator 🥳.

PR #302, opened 2020-07-06
Implement #163 - Shape length verification

lionel-nj:
### **PR in progress**

**Summary:**

This PR provides support to execute cross verification of fields `stop_times.shape_dist_traveled`, `trips.shape_id` and `shapes.shape_id`.

**Expected behavior:** 

- [ ] For all records from `stop_times.txt` that have a non-null value for field `shape_dist_traveled`, `shape_dist_traveled` should be smaller or equal to the total shape length it is related to. See https://github.com/MobilityData/gtfs-validator/issues/163

  - [ ] write unit tests
  - [ ] add javadoc


adapter/utils/src/main/java/org/mobilitydata/gtfsvalidator/utils/GeodeticUtils.java
```diff
@@ -0,0 +1,106 @@
...
+     */
+    private static DistanceCalculator getDistanceCalculator() {
+        return GEO.getDistCalc();
+    }
+
+    /**
+     * Method returns the total length of a shape. As a reminder, a shape is a collection of {@code ShapePoint} ordered
+     * by shape_pt_sequence. The result is expressed in the specified unit.
+     *
+     * @param shape          the collection of {@link ShapePoint} whose total length is to be computed
```
barbeau, line 86:
```suggestion
     * @param shape          the collection of {@link ShapePoint} sorted by `shape_dist_traveled` whose total length is to be computed
```

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateShapeDistTraveled.java
```diff
@@ -59,10 +59,6 @@ public ValidateShapeDistTraveled(final GtfsDataRepository dataRepo,
      * Use case execution method: checks if `stop_times.shape_dist_traveled` is smaller or equal to the related shape
      * total length.
      */
-    // todo: javadoc implement tests:
-    // - stopTime with null `shape_dist_Travelled` should not generate notice
-    // - stopTime with too big shape dist traveled should generate notice
-
     public void execute() {
         logger.info("Validating rule 'E036 - `If provided, stop_times.shape_dist_traveled` must be smaller or equal to"+
```
barbeau, line 63: I think you can get rid of `If provided, ` - that should be assumed.

adapter/utils/src/main/java/org/mobilitydata/gtfsvalidator/utils/GeodeticUtils.java
```diff
@@ -0,0 +1,106 @@
...
+     *
+     * @param originLatitude        latitude of the origin point
+     * @param destinationLatitude   latitude of the destination point
+     * @param originLongitude       longitude of the origin point
+     * @param destinationLongitude  longitude of the destination point
+     * @param distanceUnit          unit of the desired result of computation
+     * @return the distance between two points given there lat/lon positions in the specified unit
+     */
+    @Override
+    public double distanceBetweenTwoPoints(final float originLatitude, final float destinationLatitude,
```
barbeau, line 49: Typically lat and lon for the same point are kept adjacent in method signatures - so I suggest changing this to:

```java
    public double distanceBetweenTwoPoints(final float originLatitude, final float originLongitude,
                                                  final float destinationLatitude,
                                                  final float destinationLongitude,
                                                  final DistanceUnit distanceUnit) {
```

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateShapeDistTraveled.java
```diff
@@ -71,23 +67,22 @@ public void execute() {
...
                                 final Trip trip = dataRepo.getTripById(tripId);
                                 if (trip != null) {
                                     final String shapeId = trip.getShapeId();
                                     if (shapeId != null) {
                                         final Map<Integer, ShapePoint> shape = dataRepo.getShapeById(shapeId);
                                         if (shape != null) {
-                                            if (stopTime.getShapeDistTraveled() >
-                                                    distanceCalculationUtils
-                                                            .getShapeTotalDistance(shape, DistanceUnit.KILOMETER)) {
+                                            final float shapeTotalDistance = (float) distanceCalculationUtils
```
barbeau, line 78: A few comments here. If I'm reading this right it seems really inefficient to me. Here's how I read it for each trip:

For each record in stop_times.txt for a given trip_id:
* Calculate the entire shape length using lat/lons from shapes.txt
* Compare that entire shape length against this stop_times.txt shapeDistanceTraveled value

You shouldn't need to recalculate the entire shape distance for each record - that can be computed once per trip before looping through the stop_times.txt records for that trip.

But - reading the spec for shape_dist_traveled more carefully, I don't think this rule is correct. If anything, the shape_dist_traveled may be *more* than the "connect the dots" length for the lat/lons in shapes.txt. My understanding is that typically shape_dist_traveled is calculated using odometers, which means the distance may have a "curve" to it in between points. I haven't really used this field much myself, so I'd like to get feedback from @timMillet if this is correct.

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateShapeDistTraveled.java
```diff
@@ -71,23 +67,22 @@ public void execute() {
...
+                                                    .getShapeTotalDistance(shape, DistanceUnit.KILOMETER);
+                                            if (shapeDistanceTraveled > shapeTotalDistance) {
                                                 resultRepo.addNotice(
                                                         new FloatFieldValueOutOfRangeNotice("stop_times.txt",
```
barbeau, line 82: I wouldn't think of this particular error being "out of range" as currently implemented, as there really isn't a "range". It would be more of just an invalid value (but see my comment above).

adapter/utils/src/main/java/org/mobilitydata/gtfsvalidator/utils/GeodeticUtils.java
lionel-nj: I get the idea, but aren't `ShapePoint` sorted by `shape_pt_sequence`?

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateShapeDistTraveled.java
lionel-nj: Good catch, will modify the implementation. Waiting for Tim's feedback though.

barbeau: The other complication is that the units for shape_dist_traveled aren't specified in the spec, which also makes this impossible to implement.

adapter/utils/src/main/java/org/mobilitydata/gtfsvalidator/utils/GeodeticUtils.java
barbeau: Yes, but it doesn't hurt to clarify on this method that the input needs to sorted to get a correct result. I typically try to document assumptions in input when it affects correctness of the output, even if they are documented elsewhere.

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateShapeDistTraveled.java
barbeau: Ok, after going back and reviewing https://github.com/MobilityData/gtfs-validator/issues/163, I don't think this implementation is the correct interpretation of that rule (I looked quickly for the Python implementation but didn't see it, but I asked @maximearmstrong to add a pointer to it in https://github.com/MobilityData/gtfs-validator/issues/163).

Here's my interpretation:

```
For each of the records in shapes.txt for a given trip_id, check that each shape_dist_traveled is less than the maximum shape_dist_traveled for that trip_id.
```

Since our shape data is already sorted, you could further improve on this as:

```
For each of the records in shapes.txt for a given trip_id sorted by shape_pt_sequence, check that each shape_dist_traveled is ascending as shape_pt_sequence increases.
```

Let's review the Python implementation before any changes to make sure we're understanding this correctly.

barbeau: Also, the above logic for the rule can also apply to stop_times.txt data, sorted by stop_sequence.

adapter/utils/src/main/java/org/mobilitydata/gtfsvalidator/utils/GeodeticUtils.java
```diff
@@ -0,0 +1,110 @@
...
+     * @param destinationLatitude   latitude of the destination point
+     * @param destinationLongitude  longitude of the destination point
+     * @param distanceUnit          unit of the desired result of computation
+     * @return the distance between two points given there lat/lon positions in the specified unit
+     */
+    @Override
+    public double distanceBetweenTwoPoints(final float originLatitude, final float originLongitude,
+                                           final float destinationLatitude, final float destinationLongitude,
+                                                  final DistanceUnit distanceUnit) {
+        final ShapeFactory shapeFactory= getShapeFactory();
```
barbeau, line 53:
```suggestion
        final ShapeFactory shapeFactory = getShapeFactory();
```

adapter/utils/src/main/java/org/mobilitydata/gtfsvalidator/utils/GeodeticUtils.java
```diff
@@ -0,0 +1,110 @@
...
+        final DistanceCalculator distanceCalculator = getDistanceCalculator();
+
+        final Point origin = shapeFactory.pointXY(originLongitude, originLatitude);
+        final Point destination = shapeFactory.pointXY(destinationLongitude, destinationLatitude);
+        // implementation uses harvesine formula which determines the great-circle distance between two points on a
```
barbeau, line 58: You should add this to the Javadoc for the method too - it's important to know how the distance is being calculated.

Also, just curious - how did you know this? I looked at the Javadocs for Spatial4j but didn't see the default distance implementation. It would be good to include a link so someone can look more info if needed.

adapter/utils/src/main/java/org/mobilitydata/gtfsvalidator/utils/GeodeticUtils.java
```diff
@@ -0,0 +1,110 @@
...
+        // sphere given their longitudes and latitudes. Note that the elevation of both points is not taken into
+        // consideration.
+        final double distance = DistanceUtils.DEG_TO_KM*distanceCalculator.distance(origin, destination);
```
barbeau, line 61:
```suggestion
        final double distance = DistanceUtils.DEG_TO_KM * distanceCalculator.distance(origin, destination);
```

adapter/utils/src/main/java/org/mobilitydata/gtfsvalidator/utils/GeodeticUtils.java
```diff
@@ -0,0 +1,110 @@
...
+        if (distanceUnit == DistanceUnit.KILOMETER) {
+                return distance;
+        } else {
+                return distance* KILOMETER_TO_METER_CONVERSION_FACTOR;
```
barbeau, line 65:
```suggestion
                return distance * KILOMETER_TO_METER_CONVERSION_FACTOR;
```

adapter/utils/src/main/java/org/mobilitydata/gtfsvalidator/utils/GeodeticUtils.java
```diff
@@ -0,0 +1,110 @@
...
+    /**
+     * private method returning a {@code ShapeFactory}
+     * @return a {@link ShapeFactory}
+     */
+    private static ShapeFactory getShapeFactory() {
+        return GEO.getShapeFactory();
+    }
+
+    /**
+     * private method returning a {@code DistanceCalculator}
```
barbeau, line 78: Generally you don't need to include things like "private method" in the Javadoc which are obvious just from looking at the method signature.

adapter/utils/src/test/java/GeodeticUtilsTest.java
```diff
@@ -0,0 +1,113 @@
...
+    private static final float NEW_YORK_LATITUDE = 40.7128f;
+    private static final float NEW_YORK_LONGITUDE = -74.0060f;
+    private static final float POINTE_A_PITRE_LATITUDE = 16.2650f;
+    private static final float POINTE_A_PITRE_LONGITUDE = -61.5510f;
+    private static final float FORT_DE_FRANCE_LATITUDE = 14.6161f;
+    private static final float FORT_DE_FRANCE_LONGITUDE = -61.0588f;
+
+    @Test
+    void distanceBetweenSamePointShouldBeZero(){
+        assertEquals(0, new GeodeticUtils().distanceBetweenTwoPoints(
```
barbeau, line 42: Generally, you shouldn't need to instantiate a utility class. Unless this is part of the clean architecture guidelines. Utility methods should be static.

adapter/utils/src/test/java/GeodeticUtilsTest.java
```diff
@@ -0,0 +1,113 @@
...
+                DistanceUnit.METER));
+    }
+
+    @Test
+    void distanceBetweenMontrealAndPointeAPitreShouldBeSufficientlyAccurate() {
+        final double distance =
+                new GeodeticUtils().distanceBetweenTwoPoints(POINTE_A_PITRE_LATITUDE,POINTE_A_PITRE_LONGITUDE,
+                        MONTREAL_LATITUDE, MONTREAL_LONGITUDE, DistanceUnit.KILOMETER);
+        // see https://www.movable-type.co.uk/scripts/latlong.html
+        assertTrue(3438-.5 <= distance && distance <= 3438+.5); // unit is kilometers
```
barbeau, line 56: Typically when testing float values you'd use the `assertEquals()` that includes the `delta`: https://junit.org/junit5/docs/5.0.1/api/org/junit/jupiter/api/Assertions.html#assertEquals-float-float-float-

adapter/utils/src/test/java/GeodeticUtilsTest.java
lionel-nj: True, however here we implement an interface which does not allow us to declare methods as `static` if we want to follow the clean architecture principles: if the method is declared static within the interface, we have to implement it there. Which means that external libraries will be present within the `usecase` layer.

adapter/utils/src/test/java/GeodeticUtilsTest.java
lionel-nj: Thank, modifying this :)

adapter/utils/src/test/java/GeodeticUtilsTest.java
fabrice-v: Yeah I had the same hunch yesterday but as @lionel-nj mentioned it seems we don't have a choice given we use Java and want to declare interfaces in our domain/use case layer and relegate the implementation to the adapter one. That being said I think that would be a good use case to implement a Singleton. https://www.geeksforgeeks.org/singleton-class-java/

adapter/utils/src/main/java/org/mobilitydata/gtfsvalidator/utils/GeodeticUtils.java
lionel-nj: It's  not straighforward from the documentation, however I did find this resources (https://locationtech.github.io/spatial4j/apidocs/org/locationtech/spatial4j/context/SpatialContext.html, section `Constructor Detail - SpatialContext`) that specifies the default implementation

barbeau: `SpatialContext` constructor isn't being used here though, right? Those methods are deprecated. You're just referencing the static `GEO.getDistCalc()`. And [those docs](https://locationtech.github.io/spatial4j/apidocs/org/locationtech/spatial4j/context/SpatialContext.html#GEO) just say:

>A popular default SpatialContext implementation for geospatial.

But poking around in [`SpatialContext`](https://github.com/locationtech/spatial4j/blob/63b318368ce7eca0e090af64cd49d134c8619c29/src/main/java/org/locationtech/spatial4j/context/SpatialContext.java#L159) and [`SpatialContextFactory`](https://github.com/locationtech/spatial4j/blob/63b318368ce7eca0e090af64cd49d134c8619c29/src/main/java/org/locationtech/spatial4j/context/SpatialContextFactory.java#L167) source code confirms you're correct - the default is haversine because `isGeo()` is `true`.

Also, note there is a typo in the comment above - `harvesine` -> `haversine`.

fabrice-v:
Iceboxed

Issue #303, opened 2020-07-07
[REQUEST] Execute use case ProcessParsedStopTime


lionel-nj:
**Is your feature request related to a problem? Please describe.**
Use case `ProcessParsedStopTime` has been written and tested, however it is not executed since there was a need for a time utility class to convert `HH:MM:SS` into Integer value.

**Describe the solution you'd like**
Use case should be retrieved and executed.

PR #304, opened 2020-07-07
Implement #303 - Execute usecase ProcessParsedStopTime

lionel-nj:
**Summary:**

This PR provides support to execute use case `ProcessParsedStopTime`.

**Expected behavior:** 

`GtfsDataRepository` should be filled with `StopTime` entities using time utility class. 
PR implements a singleton pattern for time utility class.

Capture d’écran, le 2020-07-07 à 13 02 12


config/src/main/java/org/mobilitydata/gtfsvalidator/config/DefaultConfig.java
```diff
@@ -51,6 +53,7 @@
     private final RawFileRepository rawFileRepo = new InMemoryRawFileRepository();
     private final ValidationResultRepository resultRepo = new InMemoryValidationResultRepository();
     private final GtfsDataRepository gtfsDataRepository = new InMemoryGtfsDataRepository();
+    private final TimeUtils timeUtils = new TimeConversionUtils();
```
fabrice-v, line 56: `TimeConversionUtils` shall implement a singleton pattern. https://www.geeksforgeeks.org/singleton-class-java/ 
We'd retrieve and pass the single instance here

Issue #165, opened 2020-05-04
[REQUEST] Implement service exists for a trip (GTFS rule)


maximearmstrong:
**Is your feature request related to a problem? Please describe.**
A trip should be related to an existing service. This is a GTFS rule implemented in Google Python validator and featured in Google Type Error as  TYPE_TRIP_WITH_UNKNOWN_SERVICE_ID.

**Describe the solution you'd like**
Actual Google GTFS validator behaviour : verifies if service id exists in services list.

**Describe alternatives you've considered**

**Additional context**
Line 145 in Error support priorities https://docs.google.com/spreadsheets/d/1vqe6wq7ctqk1EhYkgtZ0_TbcQ91vccfs2daSjn20BLE/edit#gid=0

barbeau:
@timMillet This rule is actually directly related to a discussion we're having right now with GTFS ServiceChanges v3.1:
https://bit.ly/gtfs-service-changes-v3_1

The question is what's the best practice for including inactive service in static GTFS that could be activated later via real-time.

Looks like Google's current rule (as well as community Python validator) is that service_ids defined for trips must exist in calendar.txt or calendar_dates.txt.

timMillet:
IMHO, `trips.service_id` that are not defined in `calendar.service_id` or `calendar_dates.service_id` should be flagged with a notice, like "Do you plan to trigger these service_ids with ServiceUdpates? If yes, ignore this message; if no, please resolve".

But if GTFS-ServiceUpdates within GTFS-ServiceChanges is not likely to be adopted anytime soon (in less than 1 year maybe?), maybe it would be worth to reject these values in the meantime. Idk how much it is on the dev side.

barbeau:
For now I'd suggest that we stick with the existing Python validation rule - I think that should shape our solution in ServiceChanges, unless we have a good reason to go against it.

PR #307, opened 2020-07-09
Implement #165 - Validate trips.service_id

lionel-nj:
**Summary:**

This PR provides support to check cross reference between files `trips.txt`, `calendar.txt` and `calendar_dates.txt`.

Note that this PR changes the implementaion of methods `getTripAll`, and `getCalendarAll`, in order to return an unmodifiable `Map`instead of a `Collection`

**Expected behavior:** 

`trips.service_id` should exist in GTFS data


adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/db/InMemoryGtfsDataRepository.java
```diff
@@ -299,6 +299,15 @@ public CalendarDate getCalendarDateByServiceIdDate(final String serviceId, final
         return calendarDatePerServiceIdAndDate.get(serviceId + date.toString());
     }
 
+    /**
+     * Return an immutable collection of {@code CalendarDate} objects representing all the rows from calendar_dates.txt
```
barbeau, line 303: The key of the returned `Map` should be specified in the Javadocs.

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateTripServiceId.java
```diff
@@ -0,0 +1,74 @@
...
+    public void execute() {
+        logger.info("Validating rule E036 - `service_id` not found" + System.lineSeparator());
+
+        final Collection<Calendar> calendarCollection = dataRepo.getCalendarAll();
+        final Set<String> calendarDatesKeyCollection = dataRepo.getCalendarDateAll().keySet();
+        dataRepo.getTripAll().forEach(trip -> {
+            final String serviceId = trip.getServiceId();
+            if (calendarCollection.stream().noneMatch(calendar -> calendar.getServiceId().equals(serviceId)) &&
+                    calendarDatesKeyCollection.stream()
+                            .noneMatch(calendarDateKey -> calendarDateKey.startsWith(serviceId))) {
```
barbeau, line 64: Shouldn't this be `.equals` instead of `.startsWith`?

lionel-nj: For `CalendarDate`,  entities are mapped on a key made of the concatenation of fields `service_id` and `date`, so it makes sense to check that the key starts with a given `service_id`. A solution could be to store `CalendarDate` entities as a nested map where the first key would be `service_id` and the second `date`, I think that this would help keeping consistency with the code base.

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateTripServiceId.java
```diff
@@ -0,0 +1,74 @@
...
+    /**
+     * Use case execution method: Checks if `service_id` of a trip refers to a record from files `calendar.txt` or
+     * `calendar_dates.txt`. A new notice is generated each time this condition is false.
+     * This notice is then added to the {@link ValidationResultRepository} provided in the constructor.
+     */
+    public void execute() {
+        logger.info("Validating rule E036 - `service_id` not found" + System.lineSeparator());
+
+        final Collection<Calendar> calendarCollection = dataRepo.getCalendarAll();
+        final Set<String> calendarDatesKeyCollection = dataRepo.getCalendarDateAll().keySet();
```
lionel-nj, line 59: Add comment regarding key usage for `CalendarDate`entities

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateTripServiceId.java
fabrice-v: As discussed, having maps would be less memory efficient for a marginal gain in that case.
We will
- add a comment/javadoc explicitely stating the expected form of keys
- add a test when key are generated (entity add to the repo) to check they are of expected form

barbeau: Ah, yes, thanks for the reminder. Yes, in retrospect, I agree and think that's a better design, primarily because you gain a lot of efficiency here (which I was also about to comment on above).

Right now you're lot levering the benefits of HashMaps at all - both calendar and calendar_dates are coverted to collection/set and you're iterating though them looking for matches.  When trips.txt and calendar_dates.txt are both huge, this is very inefficient.

If you get both `calendars` and `calendar_dates` as HashMaps here keyed on `service_id`, then it becomes a simple constant `.contains()` call to check if the service_id exists:

`if (!calendars.contains(service_id) && !calendar_dates.contains(service_id) {`

We were originally trying to optimize the case of checking for duplicates in calendar_dates.txt by creating the composite key, but given the above, I think it's worth changing. This is likely true for other composite keys when you need to reference the record from another table based on part of the composite key, although I'd wait to change any implementations until we run across the need to.

fabrice-v: As discussed, because in the majority of feeds have a relatively big trips file compared to calendar_date, the memory/performance tradeoff leans towards implementing @lionel-nj suggestion and have a constant lookup time.

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateTripServiceId.java
lionel-nj: Comment is no longer valid given the latest discussions

RULES.md
```diff
@@ -150,6 +151,12 @@ Value of field `route_id` should exist in GTFS data.
 
 Value of field `agency_id` should exist in GTFS data.
 
+<a name="E036"/>
+
+### E036 - `service_id` not found
+
+Value of field `service_id` should exist in GTFS data.
```
barbeau, line 158:
```suggestion
Value of field `service_id` should exist in GTFS `calendar.txt` or `calendar_dates.txt` data.
```

lionel-nj: Will report this precision for all the relevant rule descriptions

adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/db/InMemoryGtfsDataRepository.java
```diff
@@ -51,11 +51,9 @@
     // calendar.txt.
     private final Map<String, Calendar> calendarPerServiceId = new HashMap<>();
 
-    // CalendarDate entities container. Entities are mapped on key resulting from the concatenation of the values
-    // contained in the following columns (found in calendar_dates.txt gtfs file):
-    // - service_id
-    // - date
-    private final Map<String, CalendarDate> calendarDatePerServiceIdAndDate = new HashMap<>();
+    // CalendarDate entities container. Entities are mapped on the values found in column service_id and date of GTFS
+    // file calendar_dates.txt
```
barbeau, line 55:
```suggestion
    // CalendarDate entities container. The key for the outer map is the calendar_dates.txt service_id, with the key for
    // the inner map being the calendar_dates.txt date field
```

RULES.md
barbeau: Sounds good - you could probably get rid of the last word `data` too, re-reading it now.

adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/db/InMemoryGtfsDataRepository.java
```diff
@@ -296,7 +303,18 @@ public CalendarDate addCalendarDate(@NotNull final CalendarDate newCalendarDate)
      */
     @Override
     public CalendarDate getCalendarDateByServiceIdDate(final String serviceId, final LocalDate date) {
```
barbeau, line 305: Is this method used? If not you could remove it for now - it's not clear to me that it would be used in practice (unless I'm missing a use case).

fabrice-v: I think we tried to anticipate usage but I also lean towards extending APIs as needs arise and rely on your experience.

barbeau: I would say remove for now.

adapter/repository/in-memory-simple/src/test/java/org/mobilitydata/gtfsvalidator/db/InMemoryGtfsDataRepositoryTest.java
```diff
@@ -290,6 +314,24 @@ void getCalendarByServiceIdShouldReturnRelatedCalendar() {
...
+        underTest.addCalendar(mockCalendar01);
+
+        final Map<String, Calendar> toCheck = underTest.getCalendarAll();
+        assertEquals(2, toCheck.size());
+        assertTrue(toCheck.containsKey("service id00"));
+        assertTrue(toCheck.containsKey("service id01"));
+    }
+
     @Test
     public void addSameTripTwiceShouldReturnNull() {
```
fabrice-v, line 336: shall the public modifier also be removed here? Is it present on any other test method?

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateCalendarEndDateBeforeStartDate.java
```diff
@@ -51,10 +50,14 @@ public ValidateCalendarEndDateBeforeStartDate(final GtfsDataRepository dataRepo,
      */
     public void execute() {
         logger.info("Validating rule 'E032 - calendar.txt end_date is before start_date'" + System.lineSeparator());
-        Collection<Calendar> calendars = dataRepo.getCalendarAll();
-        calendars.stream()
-                .filter(calendar -> calendar.getEndDate().isBefore(calendar.getStartDate()))
-                .forEach(calendar -> resultRepo.addNotice(new CalendarEndDateBeforeStartDateNotice("calendar.txt",
-                        calendar.getServiceId(), calendar.getStartDate(), calendar.getEndDate())));
+
+        dataRepo.getCalendarAll().forEach((serviceId, calendar) -> {
```
fabrice-v, line 54: Is there a specific reason we dropped the streaming API? wouldn't that be more memory efficient? (I do not know the answer to that question, just thought relevant to raise it)

adapter/repository/in-memory-simple/src/test/java/org/mobilitydata/gtfsvalidator/db/InMemoryGtfsDataRepositoryTest.java
lionel-nj: We actually have this modifier in some tests, I tend to remove that when I encounter one.

adapter/exporter/src/main/java/org/mobilitydata/gtfsvalidator/exporter/ProtobufNoticeExporter.java
```diff
@@ -563,6 +563,19 @@ public void export(final RouteIdNotFoundNotice toExport) throws IOException {
...
     }
 
+    @Override
+    public void export(final ServiceIdNotFoundNotice toExport) throws IOException {
+        protoBuilder.clear()
+                .setCsvFileName(toExport.getFilename())
+                .setSeverity(GtfsValidationOutputProto.GtfsProblem.Severity.ERROR)
+                .setEntityId(toExport.getEntityId())
+                .setAltValue(String.valueOf(toExport.getNoticeSpecific(KEY_FIELD_NAME)))
+                .setEntityValue(KEY_UNKNOWN_SERVICE_ID)
```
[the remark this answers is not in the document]
lionel-nj: Good catch, thanks.

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateCalendarEndDateBeforeStartDate.java
lionel-nj: Discussing this on our slack channel, will report the final decision here.

lionel-nj: Will use the streaming API. It has been dropped just to manipulate the original content returned by method `getCalendarAll`, but it makes more sense to use the streaming model for code legibility and memory performance.

PR #312, opened 2020-07-13
Implement #313 - Rework gtfs entity retrieval methods

lionel-nj:
**Summary:**

This PR provides support to standardize on the output of methods used to retrieve all specific GTFS entities such as `Trip`, `Calendar`, `Agency`, and `Route`.

**Expected behavior:** 

The following methods should return an unmodifiable view of there related GTFS entity containers:
- getTripAll
- getRouteAll
- getCalendarAll
- getAgencyAll


lionel-nj:
The following question remains: should `getAgencyAll` be treated differently since `agency.txt`can contain a single record without `agency_id`, which makes it impossible to map on key `agency_id=null` ?

barbeau:
@lionel-nj IIRC you can still use `null` as a key value in a `Map`, but you just need to make sure there is only one `null` key. Or we could define a special constant value to use as the key in this case.

lionel-nj:
As discussed, we will allocate a default value to this field, and check in a second step the requirements on `agency_id`

adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/db/InMemoryGtfsDataRepository.java
```diff
@@ -125,17 +127,40 @@
...
-        if (newAgency != null) {
-            if (agencyPerId.containsKey(newAgency.getAgencyId())) {
-                return null;
+        if (newAgency == null) {
+            throw new IllegalArgumentException("Cannot add null agency to data repository");
+        } else {
+            final String agencyId = newAgency.getAgencyId();
+            if (agencyId == null) {
+                if (getAgencyCount() > 1) {
+                    final Agency replacementAgency = (Agency) builder
```
lionel-nj, line 140: Not sure about that one, the default value could be allocated when building the entity. Since we are planning on implementing an early fail mechanism when trying to add twice the `agency_ id` to the `agencyPerId` map we could skip this "rebuilding" process, and gracefully fail.

Having this made me remove use case E029 that becomes obsolete.

lionel-nj:
Converted to draft in order to re-adopt the streaming model in the use cases

adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/db/InMemoryGtfsDataRepository.java
barbeau: I don't think this will work as implemented, because you don't know how many total records are in the agencies.txt at this point, and which of those records you're currently processing. For example, if this is the first time `addAgency()` is invoked, `getAgencyCount()` will return `0`. If this is the only time `addAgency()` is going to be invoked then if `agency_id` is null you shouldn't generate a notice. But currently you don't know if `addAgency()` will be invoked again.

I believe the only way to implement this is to fetch the number of total agency records either before or after parsing the CSV file, and then pass that number into this method or earlier in `ProcessParsedAgency`, or after parsing all the agency.txt records. It may make sense to implement when first parsing the agency.txt CSV file and pass into `ProcessParsedAgency` because then you wouldn't have the redundant build as you mention.

barbeau: And I think the logic is reversed on the `DEFAULT_AGENCY_ID` - if there is only one record in agency.txt, then it's ok for agency_id to be null, so we should assign it the `DEFAULT_AGENCY_ID`. If there is more than one record in agency.txt and agency_id is null, we should fail early or generate an error notice.

PR #326, opened 2020-07-29
Implement Duplicate column (CSV rule)

### LINKED ISSUE #129: [REQUEST] Implement Duplicate column (CSV rule)
Duplicate column is a CSV rule implemented in Google Python validator and featured in Google Type Error as TYPE_CSV_DUPLICATE_COLUMN_NAME.

A column must not be duplicated in a GTFS file.

**Additional context**
Line 22 in Error support priority https://docs.google.com/spreadsheets/d/1vqe6wq7ctqk1EhYkgtZ0_TbcQ91vccfs2daSjn20BLE/edit#gid=0

fabrice-v:
**Summary:**

close #129 

This PR adds a rule to check if a column is duplicated in an input file.
A `DuplicatedHeaderNotice` is built for each duplicate found, even for out of spec fields.

**Expected behavior:** 

[MTBA_duplicated_agency_name.zip](https://github.com/MobilityData/gtfs-validator/files/4997168/MTBA_duplicated_agency_name.zip)

image

### DIFF
--- a/RULES.md
+++ b/RULES.md
@@ -240,7 +240,7 @@ Field `parent_station` must be empty when `location_type` is 2
 
 ### E043 - Duplicated field
 
-Some file contains the same column name twice
+A file cannot contain the same header value twice (i.e., duplicated column of data).
 
 # Warnings
 
--- a/RULES.md
+++ b/RULES.md
@@ -6,7 +6,7 @@ Rules are declared in the [`Notice` module](https://github.com/MobilityData/gtfs
 
 | Error ID      | Error Title         |
 |---------------|---------------------------|
-| [E001](#E001) | | 
+| [E001](#E001) | Missing required field | 
 | [E002](#E002) | | 
 | [E003](#E003) | | 
 | [E004](#E004) | | 
@@ -44,13 +44,14 @@ Rules are declared in the [`Notice` module](https://github.com/MobilityData/gtfs
 | [E040](#E040) | Dataset should be valid for at least the next 7 days | 
 | [E041](#E041) | Invalid parent `location_type` for stop |
 | [E042](#E042) | Station stop (`location_type`=2) has a parent stop |
+| [E043](#E043) | Duplicated field |
 
 ### Table of Warnings
 
 | Warning ID    | Warning Title             |
 |---------------|---------------------------|
 | [W001](#W001) | | 
-| [W002](#W002) | | 
+| [W002](#W002) | Non standard field name | 
 | [W003](#W003) | | 
 | [W004](#W004) | | 
 | [W005](#W005) | Route short name too long |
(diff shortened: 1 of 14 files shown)


RULES.md
```diff
@@ -222,15 +229,27 @@ Any other combination raise this error
...
 Field `parent_station` must be empty when `location_type` is 2
 
 #### References:
 * [stops.txt specification](http://gtfs.org/reference/static/#stopstxt)
 
+<a name="E043"/>
+
+### E043 - Duplicated field
+
+Some file contains the same column name twice
```
barbeau, line 243:
```suggestion
A file cannot contain the same header value twice (i.e., duplicated column of data).
```

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/notice/error/MissingHeaderNotice.java
```diff
@@ -25,7 +25,7 @@
 
     public MissingHeaderNotice(final String filename, final String missingHeaderName) {
         super(filename, E_001,
-                "Field name not defined in the GTFS specification",
```
barbeau, line 28: Just flagging unrelated changes - not sure if they are intentional. A few others exist in this PR as well.

adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/db/InMemoryRawFileRepository.java
```diff
@@ -69,15 +69,15 @@ public RawFileInfo create(RawFileInfo fileInfo) {
...
+        Collection<String> toReturn = new ArrayList<>();
 
         try {
             ((CsvSchema) (mapper.readerFor(Map.class)
                     .with(schema)
                     .readValues(csvFile).getParser().getSchema())).iterator().forEachRemaining(column -> toReturn.add(column.getName()));
         } catch (IOException e) {
             //TODO: this should go back up to use case level so it can be properly reported
-            return Collections.emptySet();
+            return Collections.emptyList();
```
barbeau, line 80: I'd suggest adding a note to the Javadoc that says that the returned list will contain all headers present in the CSV file, including any duplicates.

barbeau: Actually, thinking more, given that this actually affects program correctness, it's probably better to change the return value explicitly to a `List` (in addition to documenting in the Javadoc). For example, you wouldn't want someone to change the implementation of the method internally back to a `Set` in the future, thinking that it still satisfies the `Collection` interface, because E43 would never find any duplicate columns. Changing the return type, IMHO, is a more explicit statement that there is a reason why a `List` is used here.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/notice/error/MissingHeaderNotice.java
fabrice-v: Ok yes. This one is intentional  (the message is wrong)

PR #327, opened 2020-08-03
Implement start and end times verification for a trip

### LINKED ISSUE #205: [REQUEST] Implement start and end times verification for a trip
A trip start and end times should not be missing or set to none. This is a GTFS rule implemented in Google Python validator as ValidateTripStartAndEndTimes and as an inner verification in GetStartTime and GetEndTime.

Actual Google GTFS validator behaviour : verifies if start and end times are not missing or none
- when retrieving the start time of a trip
https://github.com/google/transitfeed/blob/d727e97cb66ac2ca2d699a382ea1d449ee26c2a1/transitfeed/trip.py#L303
- when retrieving the end time of a trip
https://github.com/google/transitfeed/blob/d727e97cb66ac2ca2d699a382ea1d449ee26c2a1/transitfeed/trip.py#L345

- inner verification
https://github.com/google/transitfeed/blob/d727e97cb66ac2ca2d699a382ea1d449ee26c2a1/transitfeed/trip.py#L586

**Additional context**
Line 194, 195 and 202 in Error support priorities https://docs.google.com/spreadsheets/d/1vqe6wq7ctqk1EhYkgtZ0_TbcQ91vccfs2daSjn20BLE/edit#gid=0

fabrice-v:
**Summary:**

closes #205 
This PR contains the required changes to validate each trip first and last stop time in `stop_times.txt` define either or both fields `departure_time` or `arrival_time`

**Expected behavior:** 

image

### DIFF
--- a/RULES.md
+++ b/RULES.md
@@ -44,7 +44,7 @@ Rules are declared in the [`Notice` module](https://github.com/MobilityData/gtfs
 | [E040](#E040) | Dataset should be valid for at least the next 7 days | 
 | [E041](#E041) | Invalid parent `location_type` for stop |
 | [E042](#E042) | Station stop (`location_type`=2) has a parent stop |
-| [E043](#E043) | Missing trip edge `arrival_time` and `departure_time` |
+| [E043](#E043) | Missing trip edge `arrival_time` or `departure_time` |
 
 ### Table of Warnings
 
@@ -234,7 +234,7 @@ Field `parent_station` must be empty when `location_type` is 2
 
 ### E043 - Missing trip edge `arrival_time` and `departure_time`
 
-First and last stop of a trip must define one or both fields
+First and last stop of a trip must define both fields
 
 # Warnings
 
--- a/RULES.md
+++ b/RULES.md
@@ -44,6 +44,7 @@ Rules are declared in the [`Notice` module](https://github.com/MobilityData/gtfs
 | [E040](#E040) | Dataset should be valid for at least the next 7 days | 
 | [E041](#E041) | Invalid parent `location_type` for stop |
 | [E042](#E042) | Station stop (`location_type`=2) has a parent stop |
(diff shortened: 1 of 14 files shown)


RULES.md
```diff
@@ -229,6 +230,12 @@ Field `parent_station` must be empty when `location_type` is 2
 #### References:
 * [stops.txt specification](http://gtfs.org/reference/static/#stopstxt)
 
+<a name="E043"/>
+
+### E043 - Missing trip edge `arrival_time` and `departure_time`
+
+First and last stop of a trip must define one or both fields
```
barbeau, line 237: My interpretation of the spec is that both `arrival_time` and `departure_time` for both the first and last stop_time of a trip *must* be populated. If any of these 4 field values is empty, it's an error.

https://gtfs.org/reference/static#stop_timestxt says:

and

To my knowledge you can't have an arrival time but not a departure time, and vice versa, for any stop_time in the trip - they either both need to be populated, or neither should be populated - the exception being the first and last stop_time, which always need both arrival and departure times.

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/port/GtfsDataRepository.java
```diff
@@ -158,7 +158,7 @@ Attribution getAttribution(final String attributionId, final String agencyId, fi
      * @param tripId identifies a trip
      * @return an immutable map of {@link StopTime} from stop_times.txt related to the trip_id provided as parameter
      */
-    Map<Integer, StopTime> getStopTimeByTripId(final String tripId);
+    SortedMap<Integer, StopTime> getStopTimeByTripId(final String tripId);
```
barbeau, line 161: 👍

RULES.md
fabrice-v: Alright, that is not how the python validator is implemented but it makes a lot of sense. Will modify

Issue #198, opened 2020-05-11
[REQUEST] Implement stop times arrival and departure seconds verification


maximearmstrong:
**Is your feature request related to a problem? Please describe.**
A stop time should have valid arrival and departure seconds (time from midnight). This is a GTFS rule implemented in Google Python validator as an inner verification in StopTimes __init__.

**Describe the solution you'd like**
Actual Google GTFS validator behaviour : verifies if arrival and departure seconds are not left blank and if are valid. Verifies if departure time is not before arrival time.
https://github.com/google/transitfeed/blob/d727e97cb66ac2ca2d699a382ea1d449ee26c2a1/transitfeed/stoptime.py#L81

**Describe alternatives you've considered**

**Additional context**
Line 186 in Error support priorities https://docs.google.com/spreadsheets/d/1vqe6wq7ctqk1EhYkgtZ0_TbcQ91vccfs2daSjn20BLE/edit#gid=0

barbeau:
Also note that values greater than 24:00:00 are allowed:
https://gtfs.org/reference/static#stop_timestxt

There are some other good rules related to the same fields in the same *init()* function.

fabrice-v:
Possibly most or all those checks done in `StopTimeBuilder`

PR #328, opened 2020-08-04
Implements #198 - Stop times Departure Time after Arrival Time verification

maximearmstrong:
**Summary:**

Implement validation use case for Departure Time after Arrival Time for a StopTime record (stop_times.txt)

**Expected behavior:** 

Validates and adds a StopTimeArrivalTimeAfterDepartureTimeNotice if needed.

### DIFF
--- a/adapter/exporter/src/test/java/org/mobilitydata/gtfsvalidator/exporter/ProtobufNoticeExporterTest.java
+++ b/adapter/exporter/src/test/java/org/mobilitydata/gtfsvalidator/exporter/ProtobufNoticeExporterTest.java
@@ -1778,7 +1778,7 @@ class ProtobufNoticeExporterTest {
                         "arrival_time",
                         "departure_time",
                         "stop time trip id",
-                        "stop time stop sequence"));
+                        514));
 
         verify(mockBuilder, times(1)).clear();
         verify(mockBuilder, times(1)).setCsvFileName(
@@ -1792,7 +1792,7 @@ class ProtobufNoticeExporterTest {
         verify(mockBuilder, times(1)).setOtherCsvFileName(
                 ArgumentMatchers.eq("stop time trip id"));
         verify(mockBuilder, times(1)).setOtherCsvKeyName(
-                ArgumentMatchers.eq("stop time stop sequence"));
+                ArgumentMatchers.eq(String.valueOf(514)));
         verify(mockBuilder, times(1)).setEntityValue(
                 ArgumentMatchers.eq("arrival_time"));
         verify(mockBuilder, times(1)).setAltEntityValue(
--- a/adapter/exporter/src/test/java/org/mobilitydata/gtfsvalidator/exporter/ProtobufNoticeExporterTest.java
+++ b/adapter/exporter/src/test/java/org/mobilitydata/gtfsvalidator/exporter/ProtobufNoticeExporterTest.java
@@ -1777,8 +1777,6 @@ class ProtobufNoticeExporterTest {
                 new StopTimeArrivalTimeAfterDepartureTimeNotice("stop_times.txt",
                         "arrival_time",
                         "departure_time",
-                        "stop_time_trip_id",
-                        "stop_time_stop_sequence",
                         "stop time trip id",
                         "stop time stop sequence"));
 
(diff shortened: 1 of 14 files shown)


domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/StopTimeArrivalTimeAfterDepartureTimeNotice.java
```diff
@@ -0,0 +1,54 @@
...
+                                                       final String compositeKeySecondPart,
+                                                       final String compositeKeyFirstValue,
+                                                       final String compositeKeySecondValue) {
+        super("stop_times.txt",
+                E_043,
+                "Fields `arrival_time` and `departure_time` out of order",
+                "`departure_time`: `" + departureTimeAsString + "` precedes `arrival_time`: `" +
+                        arrivalTimeAsString + "` in file `" + filename +
+                        "` for entity with composite id:" +
+                        "`" + compositeKeyFirstPart + "`: `" + compositeKeyFirstValue + "` -- " +
```
fabrice-v, line 38: Given that Notice class is very specific, shouldn't `*keyPart` parameters removed and string be directly built with `trip_id` and `stop_sequence` 

and `*keyValue` parameters be renamed 
`compositeKeyFirstValue` --> `tripId`
`compositeKeySecondValue` --> `stopSequence`
?

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateStopTimeDepartureTimeAfterArrivalTime.java
```diff
@@ -0,0 +1,86 @@
...
+     * {@code StopTimeArrivalTimeAfterDepartureTimeNotice} is added to the {@code ValidationResultRepo} provided in the
+     * constructor.
+     * This verification is executed if {@link StopTime} has non-null value for both fields `departure_time` and
+     * `arrival_time`.
+     */
+    public void execute() {
+        logger.info("Validating rule 'E043 - `departure_time` and `arrival_time` out of order");
+        dataRepo.getStopTimeAll().forEach((tripId, innerMap) -> {
+            final Map<Integer, StopTime> tripStopTimes = innerMap;
+            tripStopTimes.forEach((stopSequence, stopTime) -> {
```
barbeau, line 66: I'd suggest looking at implementing all the "time comparison" rules within stop_times.txt in the same use case. Looping through all stop_times.txt records for all trips multiple times will definitely have a negative performance impact, and some of the rules are actually part of the same IF/THEN structure.

fabrice-v: I think we can go ahead like this and will be able to refactor quite easily when we have a better picture of all the use cases, given the limited resources we have at the moment.

application/cli-app/src/main/java/org/mobilitydata/gtfsvalidator/Main.java
```diff
@@ -196,6 +196,8 @@ public static void main(String[] args) {
                 config.validateFeedCoversTheNext7ServiceDays().execute();
                 config.validateFeedCoversTheNext30ServiceDays().execute();
                 config.validateFeedInfoFeedEndDateIsPresent().execute();
+                config.validateFeedInfoFeedStartDateIsPresent().execute();
```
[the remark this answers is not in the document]
maximearmstrong: Yes, I noticed in was implemented in DefaultConfig, but not used in Main

adapter/exporter/src/test/java/org/mobilitydata/gtfsvalidator/exporter/ProtobufNoticeExporterTest.java
```diff
@@ -27,7 +28,6 @@
 import java.net.URL;
 import java.time.LocalDate;
 
-import static org.mobilitydata.gtfsvalidator.domain.entity.notice.base.Notice.KEY_UNKNOWN_ROUTE_ID;
```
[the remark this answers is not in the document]
maximearmstrong: I tried adding it back, but it doesn't seem used anymore and IntelliJ removes it

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/notice/error/StopTimeArrivalTimeAfterDepartureTimeNotice.java
```diff
@@ -0,0 +1,52 @@
...
+import org.mobilitydata.gtfsvalidator.domain.entity.notice.base.ErrorNotice;
+
+import java.io.IOException;
+
+public class StopTimeArrivalTimeAfterDepartureTimeNotice extends ErrorNotice {
+    public StopTimeArrivalTimeAfterDepartureTimeNotice(final String filename,
+                                                       final String arrivalTimeAsString,
+                                                       final String departureTimeAsString,
+                                                       final String tripId,
+                                                       final String stopSequence) {
```
fabrice-v, line 29: I will make the change but I think we should avoid Stringifying non String data at this level so that consumers of the report can leverage native JSON types.

Issue #206, opened 2020-05-11
[REQUEST] Implement trip frequency verification 


maximearmstrong:
**Is your feature request related to a problem? Please describe.**
A trip should have frequencies with valid start and end times, valid headway seconds and exact times. This is a GTFS rule implemented in Google Python validator as an inner verification in AddFrequency.

**Describe the solution you'd like**
Actual Google GTFS validator behaviour : verifies if start and ends times are not missing and valid (format), verifies if headway seconds is not missing and an integer greater than zero. Verifies start and end times order. Verifies that exact times should be set to 0 or 1.
https://github.com/google/transitfeed/blob/d727e97cb66ac2ca2d699a382ea1d449ee26c2a1/transitfeed/trip.py#L403

**Describe alternatives you've considered**

**Additional context**
Line 196 in Error support priorities https://docs.google.com/spreadsheets/d/1vqe6wq7ctqk1EhYkgtZ0_TbcQ91vccfs2daSjn20BLE/edit#gid=0

lionel-nj:
- `frequencies.start_time` checked in https://github.com/MobilityData/gtfs-validator/pull/83
- `frequencies.end_time` checked in https://github.com/MobilityData/gtfs-validator/pull/83
- `1>=frequencies.headway_secs=>0` checked in https://github.com/MobilityData/gtfs-validator/pull/83

lionel-nj:
Use case `ValidateFrequenciesStartTimeBeforeEndTime` upcoming.

PR #335, opened 2020-08-11
Implement #206 - Verify frequencies.start_time is before frequencies.end_time

lionel-nj:
**Summary:**

This PR provides support to verify frequencies.start_time is before frequencies.end_time.

**Expected behavior:** 

A new use case is created `ValidateFrequencyStartTimeBeforeEndTime`, `Frequency` entity with `start_time < end_time` should lead to the generation of a `FrequencyStartTimeAfterEndTimeNotice`.

Capture d’écran, le 2020-08-11 à 16 41 16


domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/notice/error/FrequencyStartTimeAfterEndTimeNotice.java
```diff
@@ -0,0 +1,55 @@
...
+import org.mobilitydata.gtfsvalidator.domain.entity.notice.NoticeExporter;
+import org.mobilitydata.gtfsvalidator.domain.entity.notice.base.ErrorNotice;
+
+import java.io.IOException;
+
+public class FrequencyStartTimeAfterEndTimeNotice extends ErrorNotice {
+    public FrequencyStartTimeAfterEndTimeNotice(final String filename,
+                                                final String startTimeAsString,
+                                                final String endTimeAsString,
+                                                final String compositeKeyFirstPart,
```
maximearmstrong, line 28: Here, Fabrice suggested on my PR that we use real identification instead of `compositeKeyFirstValue`, like `startTime`. Example here https://github.com/MobilityData/gtfs-validator/blob/master/domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/notice/error/StopTimeArrivalTimeAfterDepartureTimeNotice.java

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/notice/error/FrequencyStartTimeAfterEndTimeNotice.java
```diff
@@ -0,0 +1,55 @@
...
+                                                final String compositeKeySecondPart,
+                                                final String compositeKeyThirdPart,
+                                                final String compositeKeyFirstValue) {
+        super(filename,
+                E_046,
+                "Fields `start_time` and `end_time` out of order",
+                "`end_time`: `" + startTimeAsString + "` precedes `start_time`: `" +
+                        endTimeAsString + "` in file `" + filename +
+                        "` for entity with composite id:" +
+                        "`" + compositeKeyFirstPart + "`: `" + compositeKeyFirstValue + "` -- " +
```
maximearmstrong, line 38: Since `compositeKeyFirstPart` and others are fixed for this notice, we could use static text instead of using parameters. (you can also see an example in my link above)

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/notice/error/FrequencyStartTimeAfterEndTimeNotice.java
```diff
@@ -0,0 +1,55 @@
...
+                        "`" + compositeKeySecondPart + "`: `" + startTimeAsString + "` -- " +
+                        "`" + compositeKeyThirdPart + "`: `" + endTimeAsString + "`.",
+                null);
+
+        putNoticeSpecific(KEY_COMPOSITE_KEY_FIRST_PART, compositeKeyFirstPart);
+        putNoticeSpecific(KEY_COMPOSITE_KEY_FIRST_VALUE, compositeKeyFirstValue);
+        putNoticeSpecific(KEY_COMPOSITE_KEY_SECOND_PART, compositeKeySecondPart);
+        putNoticeSpecific(KEY_COMPOSITE_KEY_SECOND_VALUE, startTimeAsString);
+        putNoticeSpecific(KEY_COMPOSITE_KEY_THIRD_PART, compositeKeyThirdPart);
```
maximearmstrong, line 47: The Frequency mapping key includes only 2 parts, `trip_id` and `start_time`, should we do the same here?

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateFrequencyStartTimeBeforeEndTime.java
```diff
@@ -0,0 +1,76 @@
...
+     * `end_time` date does not precede the `start_time`. If this requirement is not met, a
+     * {@code StartTimeAfterEndTimeNotice} is added to the {@code ValidationResultRepo} provided in the
+     * constructor.
+     */
+    public void execute() {
+        logger.info("Validating rule 'E046 - `start_time` and `end_time` out of order");
+        dataRepo.getFrequencyAll().forEach((startTimeEndTime, frequency) -> {
+                    final Integer endTime = frequency.getEndTime();
+                    final Integer startTime = frequency.getStartTime();
+                    if (endTime < startTime) {
```
maximearmstrong, line 61: I think we should add null verification here.

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateFrequencyStartTimeBeforeEndTimeTest.java
```diff
@@ -0,0 +1,132 @@
...
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.mockito.Mockito.*;
+
+class ValidateFrequencyStartTimeBeforeEndTimeTest {
+
```
maximearmstrong, line 37: I would add null check in the tests as well.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/notice/error/FrequencyStartTimeAfterEndTimeNotice.java
lionel-nj: True, it adds clarity to the code. Will modify.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/notice/error/FrequencyStartTimeAfterEndTimeNotice.java
lionel-nj: Definitely yes, I did not catch that when writing this piece of code. Thanks.

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateFrequencyStartTimeBeforeEndTime.java
lionel-nj: Not sure about that one, when operating on `Frequency` entities, fields `start_time` and `end_time` are not null. 
I would rather add comments to make that point clear.

@barbeau @fabrice-v

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateFrequencyStartTimeBeforeEndTimeTest.java
lionel-nj: Related to https://github.com/MobilityData/gtfs-validator/pull/335#discussion_r469393010

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateFrequencyStartTimeBeforeEndTime.java
barbeau: Yes, start and end times are required fields in frequencies.txt, so they can't be `null` here (we would have caught that earlier when importing the file):
http://gtfs.org/reference/static#frequenciestxt

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateFrequencyStartTimeBeforeEndTime.java
```diff
@@ -0,0 +1,74 @@
...
+    /**
+     * Use case execution method: checks if for each {@code Frequency} contained in {@link GtfsDataRepository} that
+     * `end_time` date does not precede the `start_time`. If this requirement is not met, a
+     * {@code StartTimeAfterEndTimeNotice} is added to the {@code ValidationResultRepo} provided in the
+     * constructor.
+     */
+    public void execute() {
+        logger.info("Validating rule 'E046 - `start_time` and `end_time` out of order");
+        dataRepo.getFrequencyAll().forEach((startTimeEndTime, frequency) -> {
+            // endTime and startTime are assumed to be not null at this stage. See Frequency builder.
```
barbeau, line 59: I'd suggest even stronger language:

```suggestion:
            // endTime and startTime cannot be not null at this stage. See Frequency builder.
```

Also, you really don't need to declare another two `int`s here - you could just reference `frequency.get*Time()` throughout the method and I think it's still very understandable.

adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/db/InMemoryGtfsDataRepository.java
```diff
@@ -640,6 +640,17 @@ public Frequency getFrequency(final String tripId, final Integer startTime) {
         return frequencyPerTripIdStartTime.get(Frequency.getFrequencyMappingKey(tripId, startTime));
     }
 
+    /**
+     * Return an unmodifiable map of Frequency objects representing all the rows from frequencies.txt. Entities are
+     * mapped on values of fields trip_id and start_time of file `frequencies.txt`.
```
barbeau, line 645: So this is a composite key? I'd add to the Javadocs explicitly how to create the key (link to the utility method, if one exists) so future developers know.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/notice/base/ErrorNotice.java
```diff
@@ -64,6 +64,7 @@
     protected static final int E_043 = 43;
     protected static final int E_044 = 44;
     protected static final int E_045 = 45;
+    protected static final int E_046 = 46;
```
barbeau, line 67: IIRC #338 also has an E46.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/notice/error/FrequencyStartTimeAfterEndTimeNotice.java
```diff
@@ -0,0 +1,49 @@
...
+import java.io.IOException;
+
+public class FrequencyStartTimeAfterEndTimeNotice extends ErrorNotice {
+    public FrequencyStartTimeAfterEndTimeNotice(final String filename,
+                                                final String startTimeAsString,
+                                                final String endTimeAsString,
+                                                final String tripId) {
+        super(filename,
+                E_046,
+                "Fields `start_time` and `end_time` out of order",
```
barbeau, line 31: This doesn't match docs rule name.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/notice/error/FrequencyStartTimeAfterEndTimeNotice.java
```diff
@@ -0,0 +1,49 @@
...
+                "`end_time`: `" + startTimeAsString + "` precedes `start_time`: `" +
+                        endTimeAsString + "` in file `" + filename +
```
barbeau, line 33: I'd suggest always putting the file name last in the error log. This is because most developers and those familiar with GTFS will know the file just by looking at the field names, so the key info that they need (fields that are wrong and the ID fields) should go first as they are more important and needed for every error.

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateFrequencyStartTimeBeforeEndTime.java
```diff
@@ -0,0 +1,72 @@
...
+     * `end_time` date does not precede the `start_time`. If this requirement is not met, a
+     * {@code StartTimeAfterEndTimeNotice} is added to the {@code ValidationResultRepo} provided in the
+     * constructor.
+     */
+    public void execute() {
+        logger.info("Validating rule 'E046 - `start_time` and `end_time` out of order");
+        dataRepo.getFrequencyAll().forEach((startTimeEndTime, frequency) -> {
+            // endTime and startTime cannot be null at this stage. See Frequency builder.
+            // Reference: http://gtfs.org/reference/static#frequenciestxt
+            if (frequency.getEndTime() < frequency.getStartTime()) {
```
barbeau, line 61: Should this be `<=`? It doesn't seem like they should be allowed to be equal either.

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateFrequencyStartTimeBeforeEndTime.java
```diff
@@ -0,0 +1,72 @@
...
+
+    /**
+     * Use case execution method: checks if for each {@code Frequency} contained in {@link GtfsDataRepository} that
+     * `end_time` date does not precede the `start_time`. If this requirement is not met, a
+     * {@code StartTimeAfterEndTimeNotice} is added to the {@code ValidationResultRepo} provided in the
+     * constructor.
+     */
+    public void execute() {
+        logger.info("Validating rule 'E046 - `start_time` and `end_time` out of order");
+        dataRepo.getFrequencyAll().forEach((startTimeEndTime, frequency) -> {
```
barbeau, line 58: I think a better name for `startTimeEndTime` is `tripIdStartTime`, if that's the key here

lionel-nj: Indeed.

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateFrequencyStartTimeBeforeEndTime.java
lionel-nj: Technically yes, since the spec does not enforce these two fields to be different. However, it says that `end_time` is the
> Time at which service changes to a different headway (or ceases) at the first stop in the trip.

If these two fields are equal, does it mean that there is no service?

lionel-nj: As discussed, we will revisit this question after discussion with community

barbeau: Per my Slack comment, let's table the `==` discussion on this until we have further discussion with community.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/notice/error/FrequencyStartTimeAfterEndTimeNotice.java
lionel-nj: Will do so.

RULES.md
```diff
@@ -47,6 +47,7 @@ Rules are declared in the [`Notice` module](https://github.com/MobilityData/gtfs
 | [E043](#E043) | Duplicated field |
 | [E044](#E044) | Missing trip edge `arrival_time` or `departure_time` |
 | [E045](#E045) | `arrival_time` after `departure_time` in `stop_times.txt` |
+| [E046](#E046) | `end_time` after `start_time` in `frequencies.txt` |
```
fabrice-v, line 50: Should that be 48?

lionel-nj: Indeed, will provide modification

Issue #204, opened 2020-05-11
[REQUEST] Implement stop order verification for a trip


maximearmstrong:
**Is your feature request related to a problem? Please describe.**
A trip should have ordered stop/stop time and first stop time should not be none. This is a GTFS rule implemented in Google Python validator as an inner verification in AddStopTimeObject.

**Describe the solution you'd like**
Actual Google GTFS validator behaviour : verifies each stop added are in order by verifying stops sequence.
https://github.com/google/transitfeed/blob/d727e97cb66ac2ca2d699a382ea1d449ee26c2a1/transitfeed/trip.py#L110

**Describe alternatives you've considered**

**Additional context**
Line 193 in Error support priorities https://docs.google.com/spreadsheets/d/1vqe6wq7ctqk1EhYkgtZ0_TbcQ91vccfs2daSjn20BLE/edit#gid=0

barbeau:
Similar to https://github.com/MobilityData/gtfs-validator/issues/193, note that the actual CSV file contents for any GTFS file aren't guaranteed to be in order by any field. I believe the Python tool sorts by `stop_sequence` and then applies the validation rules.

LeoFrachet:
I think it is out of spec. @barbeau what do you think? If Google is enforcing it still day one it means we could add it to the spec.

barbeau:
If we're saying that the CSV files needs to be sorted by stop_sequence, correct, it's out of spec. CSV contents aren't guaranteed to be ordered (and if Google's current internal validator in PartnerDash still enforces this, then yes, it could be added to the spec).

However, looking at the Python code at https://github.com/google/transitfeed/blob/d727e97cb66ac2ca2d699a382ea1d449ee26c2a1/transitfeed/trip.py#L143:

```python
      stoptime.stop_sequence = row[0] + 1
      prev_secs = max(row[1], row[2])
      if new_secs != None and new_secs < prev_secs:
        problems.OtherProblem(
            'out of order stop time for stop_id=%s trip_id=%s %s < %s' %
            (util.EncodeUnicode(stoptime.stop_id),
             util.EncodeUnicode(self.trip_id),
             util.FormatSecondsSinceMidnight(new_secs),
             util.FormatSecondsSinceMidnight(prev_secs)))
    self._AddStopTimeObjectUnordered(stoptime, schedule)
```

[...]

PR #338, opened 2020-08-13
Implement #204 - Verify no backwards time travel in stop_times.txt

lionel-nj:
**Summary:**

This PR provides support to verify no backwards time travel in `stop_times.txt`.

**Expected behavior:** 

Bad combinations of fields `departure_time` and `arrival_time` should generate a notice. 

> IF arrival_time for stop[N] < departure_time for stop[N - 1] THEN it's an error. (see https://github.com/MobilityData/gtfs-validator/issues/204#issuecomment-673080081)

1. These situations **should not** generate a notice:

| tripId | stopSequence | arrivalTime | departureTime |
|--------|--------------|---------------|-------------|
| 0      | 2            | 07:10         | 09:20       |
| 0      | 5            | 9:45          | 10:11       |
| 0      | 8            | 11:44         | 12:01       |

| tripId | stopSequence | arrivalTime | departureTime |
|--------|--------------|---------------|-------------|
| 0      | 2            | 07:10         | 09:20       |
| 0      | 5            | null          | 10:00       |
| 0      | 8            | 10:20         | 12:01       |

| tripId | stopSequence | arrivalTime | departureTime |
|--------|--------------|---------------|-------------|
| 0      | 2            | 07:10         | 09:20       |
| 0      | 5            | 09:50         | null        |
| 0      | 8            | 10:20         | 12:01       |

| tripId | stopSequence | arrivalTime | departureTime |
|--------|--------------|---------------|-------------|
| 0      | 2            | 07:10         | 09:20       |
| 0      | 5            | null          | null        |
| 0      | 8            | 09:50         | 12:01       |

2. These situations **should** generate a notice:

| tripId | stopSequence | arrivalTime | departureTime |
|--------|--------------|---------------|-------------|
| 0      | 2            | 07:10         | 09:20       |
| 0      | 5            | 9:45          | 10:11       |
| 0      | 8            | 10:00         | 12:01       |

| tripId | stopSequence | arrivalTime | departureTime |
|--------|--------------|---------------|-------------|
| 0      | 2            | 07:10         | 09:20       |
| 0      | 5            | null          | 08:22       |
| 0      | 8            | 09:30         | 12:01       |

| tripId | stopSequence | departureTime | arrivalTime |
|--------|--------------|---------------|-------------|
| 0      | 2            | 07:10         | 09:20       |
| 0      | 5            | 09:30         | null        |
| 0      | 8            | 08:20         | 12:01       |

Capture d’écran, le 2020-08-13 à 12 58 51


barbeau:
FYI, if we're going strictly by the spec, https://gtfs.org/reference/static/#stop_timestxt says:

...and:

So you should always have both an arrival and a departure time at a stop - you can't have one of the two fields empty. You also must have both arrival and departures at the first and last stop of the trip (this is implemented in [E044](https://github.com/MobilityData/gtfs-validator/blob/master/RULES.md#E044)).

That means that the following representations aren't allowed in GTFS in terms of empty fields (xxxx means the field is populated):

| tripId | stopSequence | arrivalTime | departureTime |
|--------|--------------|---------------|-------------|
| 0      | 0            | xxxx         | xxxx       |
| 0      | 1            | | xxxx       |
| 0      | 2            | xxxx         | xxxx       |

| tripId | stopSequence | arrivalTime | departureTime |
|--------|--------------|---------------|-------------|
| 0      | 0            | xxxx         | xxxx       |
| 0      | 1            | xxxx|        |
| 0      | 2            | xxxx         | xxxx       |

| tripId | stopSequence | arrivalTime | departureTime |
|--------|--------------|---------------|-------------|
| 0      | 0            | | |
| 0      | 1            | xxxx         | xxxx       |
| 0      | 2            | xxxx         | xxxx       |

| tripId | stopSequence | arrivalTime | departureTime |
|--------|--------------|---------------|-------------|
| 0      | 0            | xxxx         | xxxx       |
| 0      | 1            | xxxx         | xxxx       |
| 0      | 2            | | |

So we shouldn't be considering the above valid GTFS. However, we should have null handling in the code, or have multiple rules embedded in IF/ELSE statements, so that we don't crash on an NPE.

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateStopTimeTimeCombination.java
```diff
@@ -0,0 +1,71 @@
...
+        };
+
+        stopTimeSequence.forEach((stopSequence, stopTime) -> {
+
+            // useful when value of departure_time is null, the comparison is therefore done using the arrival_time of
+            // the preceding stoptime in sequence. Here we consider that if departure_time is null, then it can be
+            // approached by the previous stoptime in sequence's arrival_time value.
+            // Note that both `arrival_time` and `departure_time` fields should not be null
+            // (see best practices: http://gtfs.org/best-practices/#stop_timestxt)
+            final Integer currentStopTimeDepartureTime = stopTime.getDepartureTime() != null ?
```
barbeau, line 49: See my note at https://github.com/MobilityData/gtfs-validator/pull/338#issuecomment-673714240 on what's considered valid GTFS. Valid GTFS shouldn't have only one of the two populated (arrival and departure times). I suggest removing the code that assumes this is a valid condition.

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/crossvalidation/StopTimeBasedCrossValidator.java
```diff
@@ -35,38 +38,50 @@
...
-                    // E034
-                    validateShapeIdReferenceInStopTime.execute(resultRepo, stopTime, shape, trip);
-                    // E037
-                    validateStopTimeTripId.execute(resultRepo, stopTime, dataRepo.getTripAll());
-                })
+        StopTime previousStopTime = null;
+
+        dataRepo.getStopTimeAll().values().forEach(stopTimeCollection -> {
+                    // E046
+                    validateStopTimeTimeCombination.execute(resultRepo, stopTimeCollection, timeUtils);
```
barbeau, line 75: I wouldn't really call this "cross validating", since it's fully enclosed within stop_times.txt. Maybe the `StopTimeBasedCrossValidator` class should be renamed? When I hear "cross validating" I picture comparing one file against another.

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateStopTimeTimeCombinationTest.java
```diff
@@ -0,0 +1,461 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.mobilitydata.gtfsvalidator.domain.entity.notice.base.Notice.*;
+import static org.mockito.Mockito.*;
+
+class ValidateStopTimeTimeCombinationTest {
+
+    @Test
+    void validTimeCombinationShouldNotGenerateNotice() {
+        // | tripId | stopSequence | departureTime | arrivalTime |
```
barbeau, line 37: 💯 This visual representation is very helpful!

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateStopTimeTimeCombinationTest.java
```diff
@@ -0,0 +1,461 @@
...
+        assertEquals("0", notice.getNoticeSpecific(KEY_COMPOSITE_KEY_FIRST_VALUE));
+        assertEquals(8, notice.getNoticeSpecific(KEY_COMPOSITE_KEY_SECOND_VALUE));
+        assertEquals("930", notice.getNoticeSpecific(KEY_STOP_TIME_ARRIVAL_TIME));
+        assertEquals("820", notice.getNoticeSpecific(KEY_STOP_TIME_DEPARTURE_TIME));
+        assertEquals(5, notice.getNoticeSpecific(KEY_STOP_TIME_STOP_SEQUENCE));
+        assertEquals("no id", notice.getEntityId());
+
+        verifyNoMoreInteractions(firstStopTimeInSequence, secondStopTimeInSequence, thirdStopTimeInSequence,
+                mockResultRepo);
+    }
```
barbeau, line 460: Looks like you're missing the test case here for when there is backwards time travel and the middle rows are null:

 | tripId | stopSequence | arrivalTime | departureTime |
 |--------|--------------|---------------|-------------|
 | 0      | 2            | 07:10         | 09:20       |
 | 0      | 5            | null                 | null        |
 | 0      | 8            | 08:20         | 12:01       |

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateStopTimeTimeCombinationTest.java
barbeau: Actually, looks like you have the arrival and departure times reversed in the header. Same for other tests:

| tripId | stopSequence | arrivalTime | departureTime |
|--------|--------------|---------------|-------------|
| 0      | 2            | 07:10         | 09:20       |
| 0      | 5            | 9:45          | 10:11       |
| 0      | 8            | 11:44         | 12:01       |

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateStopTimeTimeCombinationTest.java
```diff
@@ -0,0 +1,461 @@
...
+    @Test
+    void validTimeCombinationShouldNotGenerateNotice() {
+        // | tripId | stopSequence | departureTime | arrivalTime |
+        // |--------|--------------|---------------|-------------|
+        // | 0      | 2            | 07:10         | 09:20       |
+        // | 0      | 5            | 9:45          | 10:11       |
+        // | 0      | 8            | 11:44         | 12:01       |
+
+        final StopTime firstStopTimeInSequence = mock(StopTime.class);
+        when(firstStopTimeInSequence.getDepartureTime()).thenReturn(710);
```
barbeau, line 44: Looks like arrival and departure times are swapped here too (and in other test code as well) - `710` is arrival and `920` is departure.

barbeau: I would re-order the code too so arrival time always comes before departure time.

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateStopTimeTimeCombination.java
```diff
@@ -0,0 +1,71 @@
...
+                                    stopTime.getStopSequence(),
+                                    timeUtils.convertIntegerToHMMSS(previousStopTimeData.arrivalTime),
+                                    timeUtils.convertIntegerToHMMSS(stopTime.getDepartureTime()),
+                                    previousStopTimeData.stopSequence));
+                }
+            }
+            // useful when value of arrival_time is null, the comparison is therefore done using the departure_time of
+            // the preceding stoptime in sequence. Here we consider that if arrival_time is null, then it can be approa-
+            // ched by the current stoptime's departure_time value.
+            previousStopTimeData.arrivalTime = stopTime.getArrivalTime() != null ?
```
barbeau, line 66: See my note at https://github.com/MobilityData/gtfs-validator/pull/338#issuecomment-673714240 on what's considered valid GTFS. Valid GTFS shouldn't have only one of the two populated (arrival and departure times). I suggest removing the code that assumes this is a valid condition.

However, you need to check if both `stopTime.getArrivalTime() && stopTime.getDepartureTime()` are `null` here as well, and if so don't overwrite the values of `previousStopTimeData`.

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateStopTimeTimeCombinationTest.java
lionel-nj: Good catch, thanks! :)

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/crossvalidation/StopTimeBasedCrossValidator.java
lionel-nj: That's true, and it's actually kind of confusing by the end. I suggest renaming it as `StopTimeValidator`, we could add comments to specify when a use case executes cross validation for more legibility.

barbeau: I like `StopTimeValidator` if these are all related to stop_times.txt - that's very clear.

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateBackwardsTimeTravelForStopsTest.java
```diff
@@ -0,0 +1,653 @@
...
+
+        verifyNoMoreInteractions(firstStopTimeInSequence, secondStopTimeInSequence, thirdStopTimeInSequence,
+                mockResultRepo);
+    }
+
+    // suppressed warning regarding ignored result of method since it is not necessary here
+    @SuppressWarnings("ResultOfMethodCallIgnored")
+    @Test
+    void nullArrivalTimeWithinValidTimeCombinationShouldNotGenerateNotice() {
+        // | tripId | stopSequence | arrivalTime   | departureTime |
```
fabrice-v, line 170: That's a nice way to present the information

PR #339, opened 2020-08-13
Abort on error

### LINKED ISSUE #306: [REQUEST] Fail early on errors as default behavior
It can be ambiguous to pinpoint the source of a MissingRequiredValueNotice because early steps in the process can nullify values (if they are not parsable for example)

The validator should stop the validation process as early as possible when an error is detected in the dataset. It should output an additional notice (InfoNotice) clearly stating the fact it stopped  and that an option is available to override that behavior

The validator should also provide an option to override this behavior. The consequences of activating non default behavior should be clearly stated in the console, as well as in the generated report (InfoNotice)

Doing nothing

fabrice-v:
**Summary:**
closes #306 

This PR contains required changes to make the validator abort the validation process when encountering the first error. This default behavior can be skipped by passing an option
@lionel-nj I would require your help on making it configurable through the command line

**Expected behavior:** 

The validator should abort on encountering the first error and export partial validation results

image


fabrice-v:
@barbeau After pulling this PR, the validator default behavior will change by aborting the validation process on the first `ErrorNotice`. To allow the validation to continue as before, add `"abort_on_error": false` to `execution-parameters.json` or pass `-abort_on_error false` as a command line argument.

Issue #155, opened 2020-05-04
[REQUEST] Implement stop distance from trip shape verification (GTFS Rule)


maximearmstrong:
**Is your feature request related to a problem? Please describe.**
A stop should not be too far away from its trip shape. This is a GTFS rule implemented in Google Python validator and featured in Google Type Error as  TYPE_STOP_TOO_FAR_FROM_SHAPE and TYPE_STOP_TOO_FAR_FROM_SHAPE_USING_USER_DISTANCE.

**Describe the solution you'd like**
Actual Google GTFS validator behaviour : verifies if a stop is not 1000m far away from its trip shape.

**Describe alternatives you've considered**

**Additional context**
Line 107 and 108 in Error support priorities https://docs.google.com/spreadsheets/d/1vqe6wq7ctqk1EhYkgtZ0_TbcQ91vccfs2daSjn20BLE/edit#gid=0

barbeau:
Looks like GTFS best practices are actually more strict than 1000m:
https://gtfs.org/best-practices/#shapestxt

>Route alignments (in shapes.txt) should be within 100 meters of stop locations which a trip serves.

LeoFrachet:
hard

fabrice-v:
Using a geospatial library

- for each trip
  - retrieve shape point collection from trip shape
  - pass collection to library with and expand parameter of 100m
  - retrieve an expanded shape
  - for each stop in trip, check if polygon contains the stop coordinates

barbeau:
If you use spatial4j/JTS this is actually pretty straight-forward to implement. We implemented very a similar "is the vehicle position within X meters of the trip shape" in the GTFS-RT validator as part of rule ["E029 - Vehicle position far from trip shape"](https://github.com/CUTR-at-USF/gtfs-realtime-validator/blob/4070cbe50b0d69a27ae2506f1531ced2d37818c7/RULES.md#E029).

Basic logic:
1. Use spatial4j/JTS to buffer the trip shape line by X meters to create a polygon with an edge X meters from the line.
1. For each vehicle position, test if the position is contained within the polygon (also using spatial4j/JTS)

Here's the dependency:

```
        <!-- Spatial operations -->
        <dependency>
            <groupId>org.locationtech.spatial4j</groupId>
            <artifactId>spatial4j</artifactId>
            <version>0.6</version>
        </dependency>
        <dependency>
            <groupId>com.vividsolutions</groupId>
            <artifactId>jts</artifactId>
[...]

PR #340, opened 2020-08-14
Implement #155 - Check for stops too far from trip shape

barbeau:
**Summary:**

This PR implements the rule to check when a stop is too far from the trip shape.

[GTFS Best Practices](https://gtfs.org/best-practices/#shapestxt) say:

>Route alignments (in shapes.txt) should be within 100 meters of stop locations which a trip serves.

Closes https://github.com/MobilityData/gtfs-validator/issues/155

Data used in unit tests is available as GeoJSON on GitHub here (GitHub provides a toggle between text and map view):
https://gist.github.com/barbeau/d9c0b90a26a3e2ba105cae5f0e8aec4a

image

If desired, you can visualize the buffer by exporting a JTS-version of the buffer in WKT format using code at https://gist.github.com/barbeau/d9c0b90a26a3e2ba105cae5f0e8aec4a#gistcomment-3425554 and then visualizing the WKT using https://arthur-e.github.io/Wicket/sandbox-gmaps3.html.

image

The spatial4j version of the buffer can't easily be visualized using GeoJSON or WKT because it uses a LineString and a proprietary "buffer" extension to GeoJSON and WKT, which most tools don't support.

**Expected behavior:** 

image

TODO:

- [x] Add rule
- [x] Export tests
- [x] Manual test using Google Earth and geojson.io
- [x] Unit tests - utility method
- [x] Unit tests - use case
- [x] Update rule number based on latest in master - Change to E052
- [x] Merge with master to pick up `stops` package refactor
- [x] Clarify which `location_type`s this rule applies to (0 and 4?)
- [x] Fix failing test
- [x] Final diff review by Sean


adapter/utils/src/main/java/org/mobilitydata/gtfsvalidator/geoutils/GeospatialUtilsImpl.java
```diff
@@ -63,8 +79,72 @@ public int distanceBetweenMeter(double fromLat, double fromLng, double toLat, do
...
+
+        // Check if each stop is within the buffer polygon
+        stopTimes.forEach((integer, stopTime) -> {
+            LocationBase stop = stops.get(stopTime.getStopId());
+            if (stop == null || stop.getStopLat() == null || stop.getStopLon() == null) {
+                // Lat/lon are optional for location_type 4 - skip to the next stop if they aren't provided
+                return;
+            }
+            if (!(stop instanceof StopOrPlatform) && !(stop instanceof BoardingArea)) {
+                // This rule only applies to stops of location_type 0 and 4 - skip to next stop
```
barbeau, line 122: I believe this rule only applies to location_types 0 and 4, but I've asked the spec team to clarify.

barbeau: This is correct per spec team.

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateStopTooFarFromTripShapeNoticeTest.java
```diff
@@ -0,0 +1,171 @@
...
+        final ValidationResultRepository mockResultRepo = mock(ValidationResultRepository.class);
+        final Logger mockLogger = mock(Logger.class);
+        final GeospatialUtils mockGeoUtil = mock(GeospatialUtils.class);
+
+        final ValidateStopTooFarFromTripShape underTest =
+                new ValidateStopTooFarFromTripShape(mockDataRepo, mockResultRepo,
+                        mockGeoUtil, mockLogger);
+
+        underTest.execute();
+        verify(mockLogger, times(1)).info("Validating rule 'E047 - Stop too far from trip shape'");
```
barbeau, line 132: Per discussion with @lionel-nj - just test number of calls to utility method here. Don't need to mock/test spatial calculations.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/notice/base/ErrorNotice.java
```diff
@@ -70,6 +70,7 @@
     protected static final int E_049 = 49;
     protected static final int E_050 = 50;
     protected static final int E_051 = 51;
+    public static final int E_052 = 52;
```
barbeau, line 73: This is `public` so the detected error type in `GeospatialUtilsImplTest` can be asserted, which is in the `adapter` module.

adapter/exporter/src/main/java/org/mobilitydata/gtfsvalidator/exporter/ProtobufNoticeExporter.java
```diff
@@ -786,6 +786,21 @@ public void export(final StopTimeArrivalTimeAfterDepartureTimeNotice toExport) t
                 .writeTo(streamGenerator.getStream());
     }
 
+    @Override
+    public void export(final StopTooFarFromTripShape toExport) throws IOException {
+        // TODO - Needs review - Are these right?
```
barbeau, line 791: I'm not sure I got the protobuf export correct - please review!

adapter/repository/in-memory-simple/src/main/java/org/mobilitydata/gtfsvalidator/db/InMemoryGtfsDataRepository.java
```diff
@@ -110,7 +110,7 @@
 
     // Map containing Shape Entities. A shape is a actually a collection of ShapePoint.
     // Entities are mapped on the values found in column shape_id  and shape_pt_sequence of GTFS file shapes.txt
-    private final Map<String, Map<Integer, ShapePoint>> shapePerIdShapePtSequence = new HashMap<>();
+    private final Map<String, SortedMap<Integer, ShapePoint>> shapePerIdShapePtSequence = new HashMap<>();
```
barbeau, line 113: This interface should have originally been `SortedMap` instead of `Map, because we use a `TreeMap` to order the contents by `shape_pt_sequence`. I had to change it to be able to use `SortedMap` in the utility implementation, because you can't cast from a `Map` to a `SortedMap`.

adapter/utils/src/main/java/org/mobilitydata/gtfsvalidator/geoutils/GeospatialUtilsImpl.java
```diff
@@ -63,8 +77,71 @@ public int distanceBetweenMeter(double fromLat, double fromLng, double toLat, do
...
+     * @param trip        Trip for this GTFS trip
+     * @param stopTimes   a map of StopTimes for a trip, sorted by stop_sequence
+     * @param shapePoints a map of ShapePoints for a trip, sorted by shape_pt_sequence
+     * @param stops       a map of all stops (keyed on stop_id), needed to obtain the latitude and longitude for each stop
+     * @param testedCache a cache for previously tested shape_id and stop_id pairs (keyed on shape_id+stop_id). If the
+     *                    combination of shape_id and stop_id appears in this set, we shouldn't test it again. Shapes
+     *                    and stops tested in this method execution will be added to this testedCache.
+     * @return a list of E052 errors, one for each stop that is too far from the trip shapePoints
+     */
+    public List<StopTooFarFromTripShape> checkStopsWithinTripShape(Trip trip,
```
lionel-nj, line 92: variables should be final

adapter/utils/src/main/java/org/mobilitydata/gtfsvalidator/geoutils/GeospatialUtilsImpl.java
```diff
@@ -63,8 +77,71 @@ public int distanceBetweenMeter(double fromLat, double fromLng, double toLat, do
...
+                                                                   SortedMap<Integer, StopTime> stopTimes,
+                                                                   SortedMap<Integer, ShapePoint> shapePoints,
+                                                                   Map<String, LocationBase> stops,
```
lionel-nj, line 95: Should this be renamed as `stopsByStopId` or `stopCollectonById`?

adapter/utils/src/main/java/org/mobilitydata/gtfsvalidator/geoutils/GeospatialUtilsImpl.java
```diff
@@ -63,8 +77,71 @@ public int distanceBetweenMeter(double fromLat, double fromLng, double toLat, do
...
+        // Create a polyline from the GTFS shapes data
+        ShapeFactory.LineStringBuilder lineBuilder = getShapeFactory().lineString();
+        shapePoints.forEach((integer, shapePoint) -> lineBuilder.pointXY(shapePoint.getShapePtLon(), shapePoint.getShapePtLat()));
+        Shape shapeLine = lineBuilder.build();
+
+        // Create the buffered version of the trip as a polygon
+        Shape shapeBuffer = shapeLine.getBuffered(TRIP_BUFFER_DEGREES, shapeLine.getContext());
+
+        // Check if each stop is within the buffer polygon
+        stopTimes.forEach((integer, stopTime) -> {
```
lionel-nj, line 112: I suggest renaming `integer` as `stopSequence`

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/notice/error/StopTooFarFromTripShape.java
```diff
@@ -0,0 +1,32 @@
...
+import org.mobilitydata.gtfsvalidator.domain.entity.notice.NoticeExporter;
+import org.mobilitydata.gtfsvalidator.domain.entity.notice.base.ErrorNotice;
+
+import java.io.IOException;
+
+public class StopTooFarFromTripShape extends ErrorNotice {
+
+    public StopTooFarFromTripShape(final String filename, final String stopId, final int stopSequence,
+                                   final String tripId, final String shapeId, final double stopShapeThreshold) {
+        super(filename, E_052,
```
lionel-nj, line 12: in `GeospatialUtilsImpl` we use this method with filename=`shapes.txt`: we could remove this field, and use a static String directly in call to super method. 
Also, since the problematic file is `shapes.txt`, I think that entity id should be set to `shapeId`.

Should we create new static key for this notice? Since `shapes.txt` has a defined `id` field, there is no need to define composite key values, we could have keys such as `KEY_STOP_STOP_ID`, `KEY_TRIP_TRIP_ID`, `KEY_STOP_STOP_SEQUENCE`. The linked information would be the same, but I think that it improves the code consistency and legibility for the future.

adapter/utils/src/test/java/org/mobilitydata/gtfsvalidator/geoutils/GeospatialUtilsImplTest.java
```diff
@@ -39,4 +53,613 @@ void distanceReferenceCheck() {
...
+
+        List<StopTooFarFromTripShape> errorList =
+                GEO_UTILS.checkStopsWithinTripShape(trip, stopTimes, points, stopPerId, new HashSet<>());
+
+        assertEquals(1, errorList.size());
+
+        StopTooFarFromTripShape notice = errorList.get(0);
+        assertEquals(E_052, notice.getCode());
+        assertEquals("Stop too far from trip shape", notice.getTitle());
+        assertEquals(stopId3, notice.getNoticeSpecific(KEY_COMPOSITE_KEY_FIRST_VALUE));
```
lionel-nj, line 257: I would check for interactions happening on all mocked objects.

adapter/utils/src/test/java/org/mobilitydata/gtfsvalidator/geoutils/GeospatialUtilsImplTest.java
```diff
@@ -39,4 +53,613 @@ void distanceReferenceCheck() {
...
+
+        // Map containing Stop entities. Entities are keyed on GTFS stops.txt stop_id
+        final Map<String, LocationBase> stopPerId = new HashMap<>(Map.of(stopId1, stop1, stopId2, stop2));
+
+        // Entities are keyed on shape_pt_sequence of GTFS file shapes.txt
+        SortedMap<Integer, ShapePoint> points = new TreeMap<>(Map.of(1, pt1, 2, pt2, 3, pt3, 4, pt4, 5, pt5));
+
+        List<StopTooFarFromTripShape> errorList =
+                GEO_UTILS.checkStopsWithinTripShape(trip, stopTimes, points, stopPerId, new HashSet<>());
+        assertEquals(0, errorList.size());
```
lionel-nj, line 147: I would check for interactions happening on all mocked objects.

adapter/utils/src/test/java/org/mobilitydata/gtfsvalidator/geoutils/GeospatialUtilsImplTest.java
```diff
@@ -39,4 +53,613 @@ void distanceReferenceCheck() {
...
+
+        StopTooFarFromTripShape notice = trip1ErrorList.get(0);
+        assertEquals(E_052, notice.getCode());
+        assertEquals("Stop too far from trip shape", notice.getTitle());
+        assertEquals(stopId3, notice.getNoticeSpecific(KEY_COMPOSITE_KEY_FIRST_VALUE));
+
+        // Validate the 2nd trip - no new errors should be added, because the shapeId+stopId combination has already been flagged
+        List<StopTooFarFromTripShape> trip2ErrorList =
+                GEO_UTILS.checkStopsWithinTripShape(trip1, stopTimes, points, stopPerId, testedCache);
+        assertEquals(0, trip2ErrorList.size());
```
lionel-nj, line 381: I would check for interactions happening on all mocked objects.

adapter/utils/src/test/java/org/mobilitydata/gtfsvalidator/geoutils/GeospatialUtilsImplTest.java
```diff
@@ -39,4 +53,613 @@ void distanceReferenceCheck() {
...
+        final TreeMap<Integer, StopTime> stopTimes = new TreeMap<>(Map.of(1, stopTime1, 2, stopTime2, 3, stopTime3));
+        stopTimeCollection.put(tripId1, stopTimes);
+
+        // Map containing Stop entities. Entities are keyed on GTFS stops.txt stop_id
+        final Map<String, LocationBase> stopPerId = new HashMap<>(Map.of(stopId1, stop1, stopId2, stop2, stopId3, stop3));
+
+        List<StopTooFarFromTripShape> errorList =
+                GEO_UTILS.checkStopsWithinTripShape(trip1, stopTimes, null, stopPerId, new HashSet<>());
+
+        assertEquals(0, errorList.size());
```
lionel-nj, line 453: I would check for interactions happening on all mocked objects.

adapter/utils/src/test/java/org/mobilitydata/gtfsvalidator/geoutils/GeospatialUtilsImplTest.java
```diff
@@ -39,4 +53,613 @@ void distanceReferenceCheck() {
...
+        // Map containing Stop entities. Entities are keyed on GTFS stops.txt stop_id
+        final Map<String, LocationBase> stopPerId = new HashMap<>(Map.of(stopId1, stop1, stopId2, stop2, stopId3, stop3));
+
+        // Entities are keyed on shape_pt_sequence of GTFS file shapes.txt
+        SortedMap<Integer, ShapePoint> points = new TreeMap<>(Map.of(1, pt1, 2, pt2, 3, pt3, 4, pt4, 5, pt5));
+
+        List<StopTooFarFromTripShape> errorList =
+                GEO_UTILS.checkStopsWithinTripShape(trip, stopTimes, points, stopPerId, new HashSet<>());
+
+        assertEquals(0, errorList.size());
```
lionel-nj, line 558: I would check for interactions happening on all mocked objects.

adapter/utils/src/test/java/org/mobilitydata/gtfsvalidator/geoutils/GeospatialUtilsImplTest.java
```diff
@@ -39,4 +53,613 @@ void distanceReferenceCheck() {
...
```
lionel-nj, line 663: I would check for interactions happening on all mocked objects.

adapter/utils/src/test/java/org/mobilitydata/gtfsvalidator/geoutils/GeospatialUtilsImplTest.java
```diff
@@ -39,4 +53,613 @@ void distanceReferenceCheck() {
...
+    }
 }
```
lionel-nj, line 665: I think that we should have tests for:
- null stop should not generate notice
- null trip should not generate notice
- null stoptimes should not generate notice
- empty stoptimes should not generate notice
- null shapePoints should not generate notice
- empty shapePoints should not generate notice

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/notice/base/ErrorNotice.java
lionel-nj: I think that we should add in-line comments to explain why this error code is public instead of protected :)

adapter/exporter/src/main/java/org/mobilitydata/gtfsvalidator/exporter/ProtobufNoticeExporter.java
lionel-nj: If we keep the current structure of `StopTooFarFromTripShape`, this misses the export of `KEY_COMPOSITE_KEY_THIRD_VALUE`, `KEY_COMPOSITE_KEY_THIRD_PART`, `KEY_COMPOSITE_KEY_FOURTH_VALUE`, `KEY_COMPOSITE_KEY_THIRD_PART`.

adapter/exporter/src/test/java/org/mobilitydata/gtfsvalidator/exporter/ProtobufNoticeExporterTest.java
```diff
@@ -2132,4 +2133,49 @@ void exportUnusableTripNoticeShouldMapToCsvProblemAndWriteToStream() throws IOEx
...
+                ArgumentMatchers.eq("trip_id"));
+        verify(mockBuilder, times(1)).setOtherCsvFileName(
+                ArgumentMatchers.eq("1234"));
+        verify(mockBuilder, times(1)).setOtherCsvKeyName(
+                ArgumentMatchers.eq("A"));
+        verify(mockBuilder, times(1)).setEntityValue(
+                ArgumentMatchers.eq(String.valueOf(GeospatialUtils.TRIP_BUFFER_METERS)));
+
+        verify(mockBuilder, times(1)).build();
+        verify(mockProblem, times(1)).writeTo(ArgumentMatchers.eq(mockStream));
```
lionel-nj, line 2179: Here, we need to verify the export of the information linked to `KEY_COMPOSITE_KEY_THIRD_PART`, `KEY_COMPOSITE_KEY_THIRD_VALUE`, `KEY_COMPOSITE_KEY_FOURTH_PART`, and `KEY_COMPOSITE_KEY_FOURTH_VALUE`

barbeau:
@lionel-nj @fabrice-v Looks like this PR was passing all the unit tests, but then after @fabrice-v merged the master branch into this PR in https://github.com/MobilityData/gtfs-validator/pull/340/commits/c0f4bf45ff134cdf96915837191d857ab006f1d9 now it's failing the memory tests for 100 and 1000 notices:

```
InMemoryValidationResultRepositoryTest > addingNoticeShouldExtendNoticeList() PASSED

NoticeMemoryConsumptionTest > memoryLimitTest_100notices() FAILED
    org.opentest4j.AssertionFailedError at NoticeMemoryConsumptionTest.java:55

NoticeMemoryConsumptionTest > memoryLimitTest_1000notices() FAILED
    org.opentest4j.AssertionFailedError at NoticeMemoryConsumptionTest.java:55

NoticeMemoryConsumptionTest > memoryLimitTest_10_000notices() PASSED

NoticeMemoryConsumptionTest > memoryLimitTest_100_000notices() PASSED

NoticeMemoryConsumptionTest > memoryLimitTest_1_000_000notices() PASSED
```

So it looks like the cumulative memory requirements have gone up.

How do you want to handle this? Do you want to bump the test thresholds?

lionel-nj:
@barbeau We had a similar issue in https://github.com/MobilityData/gtfs-validator/pull/363. We ended up bumping the test thresholds by providing a safety buffer factor to each tests. 
We think that the addition of static constant to class `Notice` (E_045, E_046 etc) might have something to deal with that, but we are not sure yet, it is still under investigation.

adapter/exporter/src/main/java/org/mobilitydata/gtfsvalidator/exporter/ProtobufNoticeExporter.java
barbeau: Thanks! I added `KEY_COMPOSITE_KEY_THIRD_VALUE` and `KEY_COMPOSITE_KEY_THIRD_PART` based on other examples in the class, but I'm not sure what `protoBuilder` methods the `KEY_COMPOSITE_KEY_FOURTH_VALUE` and `KEY_COMPOSITE_KEY_FOURTH_PART` should map to. It looks like the other notices that use these same constants also don't export them to protobuf. Any suggestions?

barbeau:
@lionel-nj Ok, thanks for the info. I'll pull in those changes to this PR then to fix that failing test.

adapter/exporter/src/main/java/org/mobilitydata/gtfsvalidator/exporter/ProtobufNoticeExporter.java
lionel-nj: I'd say to pick `protoBuilder` methods that are not used

barbeau:
@lionel-nj Hmm, I pulled in https://github.com/MobilityData/gtfs-validator/pull/340/commits/1dc9c94238d7a7fea257609b10aee2666a349ff6 to this PR but https://github.com/MobilityData/gtfs-validator/runs/1129155610 is still failing:

```
InMemoryValidationResultRepositoryTest > abortOnErrorRepoShouldThrowOnError() PASSED

InMemoryValidationResultRepositoryTest > addingNoticeShouldExtendNoticeList() PASSED

NoticeMemoryConsumptionTest > memoryLimitTest_100notices() FAILED
    org.opentest4j.AssertionFailedError at NoticeMemoryConsumptionTest.java:57

NoticeMemoryConsumptionTest > memoryLimitTest_1000notices() FAILED
    org.opentest4j.AssertionFailedError at NoticeMemoryConsumptionTest.java:57

NoticeMemoryConsumptionTest > memoryLimitTest_10_000notices() PASSED

NoticeMemoryConsumptionTest > memoryLimitTest_100_000notices() PASSED

NoticeMemoryConsumptionTest > memoryLimitTest_1_000_000notices() PASSED
```

adapter/exporter/src/main/java/org/mobilitydata/gtfsvalidator/exporter/ProtobufNoticeExporter.java
barbeau: I thought about that, but seems like there are two problems with that approach:
1. It seems like that's abusing the protobuf model. How would another application using the same .proto file know that a particular value was written to a random field?
1. There don't seem to be any remaining String or int methods I could write to (unless I'm missing one?)

Seems like for composite keys the .proto file should really be changed so that there is `OtherCsvKey1Name`, `OtherCsvKey2Name`, `OtherCsvKey3Name`, etc. fields. Or maybe better yet change to using fields named `CompositeKeyPart1`, `CompositeKeyPart2`, etc.

lionel-nj:
@barbeau  my bad I referenced the wrong commit .. we actually considered a safety buffer of 15%, instead of a 11% one (as implemented in the commit I mentioned). I think that changing `1.11` to `1.15` should float our boat for now

barbeau:
@lionel-nj Thanks! Looks like that fixed the memory test.

adapter/exporter/src/main/java/org/mobilitydata/gtfsvalidator/exporter/ProtobufNoticeExporter.java
lionel-nj: 1. that is question that has been raised during one of the monthly meetings, and if I recall correctly there was no clear instruction regarding this. However, we had the approval to add new error types to `gtfs_validation.proto`. I guess that we should do the same for the `protoBuilder` methods to make clear which `protoBuilder` method is used to write a particular field.

lionel-nj: 2. We could use these methods: 
```                .setAltEntityValue()
                .setEntityName()
                .setValue()
                .setParentEntityName()``` to export the rest of the information for now. And then see for adding more fields to the `.proto` file and refactoring the existing code to use new generated methods.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/notice/error/StopTooFarFromTripShape.java
barbeau: >in GeospatialUtilsImpl we use this method with filename=shapes.txt: we could remove this field, and use a static String directly in call to super method.

👍  Will change this

>Also, since the problematic file is shapes.txt, I think that entity id should be set to shapeId.
>
>Should we create new static key for this notice? Since shapes.txt has a defined id field, there is no need to define composite key values, we could have keys such as KEY_STOP_STOP_ID, KEY_TRIP_TRIP_ID, KEY_STOP_STOP_SEQUENCE. The linked information would be the same, but I think that it improves the code consistency and legibility for the future.

It's possible the problem that this rule detects is actually a stop in a wrong location, not the shape being wrong. So I think it's best to keep the composite key, as the individual error is a combination of the shape, stop sequence, and stop ID.

adapter/exporter/src/main/java/org/mobilitydata/gtfsvalidator/exporter/ProtobufNoticeExporter.java
barbeau: >I guess that we should adopt the same approach for the protoBuilder fields, to make clear which protoBuilder method is used to write a particular field.

Yes, I believe the `protoBuilder` should be autogenerated again after the .proto file is updated, which will give us the getters/setters for those fields.

Given that this is an issue beyond just the notice in this PR, I'd prefer to put off implementation of this for another PR as the changes will touch a lot of files. I've opened an issue here to track - https://github.com/MobilityData/gtfs-validator/issues/364.

adapter/exporter/src/test/java/org/mobilitydata/gtfsvalidator/exporter/ProtobufNoticeExporterTest.java
barbeau: Done for `...THIRD...` variables, but see https://github.com/MobilityData/gtfs-validator/pull/340#discussion_r477547434 for `...FOURTH`.

barbeau:
Thanks @lionel-nj for the great feedback! I believe I've addressed everything with either code changes or commenting on this PR, so it should be ready for review again. Please let me know if I missed anything!

adapter/exporter/src/main/java/org/mobilitydata/gtfsvalidator/exporter/ProtobufNoticeExporter.java
lionel-nj: Thanks

lionel-nj:
LGTM! Thanks for your work in this :) @barbeau

lionel-nj (approved):

Issue #162, opened 2020-05-04
[REQUEST] Implement trip has time and is usable verification (GTFS rule)


maximearmstrong:
**Is your feature request related to a problem? Please describe.**
A trip should have a beginning time and ending time, and be usable. This is a GTFS rule implemented in Google Python validator and featured in Google Type Error as TYPE_TRIP_WITH_NO_TIME_FOR_FIRST_STOP_TIME,  TYPE_TRIP_WITH_NO_TIME_FOR_LAST_STOP_TIME and TYPE_TRIP_WITH_NO_USABLE_STOPS.

**Describe the solution you'd like**
Actual Google GTFS validator behaviour : verifies if a trip has a time set for first stop time and last stop time, and that the trip is usable (more than 1 stop).
https://github.com/google/transitfeed/blob/master/transitfeed/schedule.py line 1298

**Describe alternatives you've considered**

**Additional context**
Line 136, 137 and 138 in Error support priorities https://docs.google.com/spreadsheets/d/1vqe6wq7ctqk1EhYkgtZ0_TbcQ91vccfs2daSjn20BLE/edit#gid=0

barbeau:
@maximearmstrong I'm not sure I understand TYPE_TRIP_WITH_NO_USABLE_STOPS here. Could you please explain that in more detail, and perhaps link to the Python code?

maximearmstrong:
Sure, I added the link above. The trip must have a time for first stop, another one for last stop, and must have more than one stop to be usable.

lionel-nj:
From my understanding, all `trips` should be refered to in table `StopTime`

https://github.com/google/transitfeed/blob/d727e97cb66ac2ca2d699a382ea1d449ee26c2a1/transitfeed/schedule.py#L1299

Am I correct to assume this @barbeau ?

barbeau:
>From my understanding, all Trips should be referred to (at least once) in table StopTime
>
>https://github.com/google/transitfeed/blob/d727e97cb66ac2ca2d699a382ea1d449ee26c2a1/transitfeed/schedule.py#L1299 (line 1299)

Correct - looks like this is a warning in the Python validator:

```python
      if not count_stop_times:
        problems.OtherProblem('The trip with the trip_id "%s" doesn\'t have '
                              'any stop times defined.' % trip.trip_id,
                              type=problems_module.TYPE_WARNING)
```

[...]

PR #317, opened 2020-07-22
Use case to check if a csv file is empty

fabrice-v:
**Summary:**

This PR adds a new use case to check for empty csv files.
Closes #213 

**Expected behavior:** 
 - the file is completely empty --> that's an error
 - the file defines headers but no row of data
   - required file --> that's an error
   - optional file --> that's a warning

In consequence, an `EmptyFileNotice` is added to the generated report

Datasets used for test taken from https://github.com/MobilityData/gtfs-validator/pull/317#issuecomment-663042812
one_empty_gtfs_file.zip
image

mixed_empty_full_gtfs_files.zip
image

Please make sure these boxes are checked before submitting your pull request - thanks!

- [ ] Run the unit tests with `gradle test` to make sure you didn't break anything
- [ ] Format the title like "Fix #<issue_number> - <short description of fix and changes>" (for example - "Fix #1111 - Check for null value before using field")
- [x] Linked all relevant issues
- [x] Include screenshot(s) showing how this pull request works and fixes the issue(s)

PR #344, opened 2020-08-18
Implement #162 - Verify trip usability and usage

lionel-nj:
**Summary:**

This PR provides support to verify that:
- all trips are referred to at least once in `stop_times.txt` (rule E050)
- all trip serve more than 1 stop (rule E051)

**Expected behavior:** 
Use cases E050 and E051 should be executed:

**E050:** if `trips.txt` defines a `trip_id` that is not used in `stop_times.txt` a `TripNotUsedNotice` is generated and added to the validation result repository

**E051:** if a trip does not serve more than one stop in `stop_times.txt` a `UnusableTripNotice` is generated and added to the validation result repository

Capture d’écran, le 2020-08-18 à 09 32 53

- [X] Run the unit tests with `gradle test` to make sure you didn't break anything
- [X] Format the title like "Fix #<issue_number> - <short description of fix and changes>" (for example - "Fix #1111 - Check for null value before using field")
- [X] Linked all relevant issues
- [X] Include screenshot(s) showing how this pull request works and fixes the issue(s)

### DIFF
--- a/domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/notice/error/BackwardsTimeTravelInStopNotice.java
+++ b/domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/notice/error/BackwardsTimeTravelInStopNotice.java
@@ -27,7 +27,7 @@ public class BackwardsTimeTravelInStopNotice extends ErrorNotice {
                                            final String arrivalTimeAsString,
                                            final String previousStopDepartureTimeAsString,
                                            final Integer previousStopTimeStopSequence) {
-        super("stop_times.txt", E_049,
+        super("stop_times.txt", E_047,
                 "Bad stoptime time combination",
                 String.format("The `arrival_time`: `%s` (`stop_sequence`: `%s`) occurs before `departure_time`:" +
                                 "`%s` (`stop_sequence`: `%s`) in `trip_id`: `%s`", arrivalTimeAsString, stopSequence,
--- a/domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/notice/error/BackwardsTimeTravelInStopNotice.java
+++ b/domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/notice/error/BackwardsTimeTravelInStopNotice.java
@@ -27,7 +27,7 @@ public class BackwardsTimeTravelInStopNotice extends ErrorNotice {
                                            final String arrivalTimeAsString,
                                            final String previousStopDepartureTimeAsString,
                                            final Integer previousStopTimeStopSequence) {
-        super("stop_times.txt", E_047,
+        super("stop_times.txt", E_049,
                 "Bad stoptime time combination",
                 String.format("The `arrival_time`: `%s` (`stop_sequence`: `%s`) occurs before `departure_time`:" +
                                 "`%s` (`stop_sequence`: `%s`) in `trip_id`: `%s`", arrivalTimeAsString, stopSequence,
--- a/domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/notice/error/EmptyFileErrorNotice.java
+++ b/domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/notice/error/EmptyFileErrorNotice.java
@@ -23,7 +23,7 @@ import java.io.IOException;
 
 public class EmptyFileErrorNotice extends ErrorNotice {
     public EmptyFileErrorNotice(String filename) {
-        super(filename, E_039,
+        super(filename, E_047,
                 "Empty csv file",
                 String.format("`%s` does not have any headers or data.", filename),
                 null);
--- a/domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/notice/error/EmptyFileErrorNotice.java
+++ b/domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/notice/error/EmptyFileErrorNotice.java
@@ -23,7 +23,7 @@ import java.io.IOException;
 
 public class EmptyFileErrorNotice extends ErrorNotice {
     public EmptyFileErrorNotice(String filename) {
-        super(filename, E_047,
+        super(filename, E_039,
                 "Empty csv file",
                 String.format("`%s` does not have any headers or data.", filename),
                 null);
--- a/domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/notice/error/UnusableTripNotice.java
+++ b/domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/notice/error/UnusableTripNotice.java
@@ -1,0 +1,36 @@
+/*
+ *  Copyright (c) 2020. MobilityData IO.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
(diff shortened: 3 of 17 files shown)


usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateTripUsageTest.java
```diff
@@ -0,0 +1,102 @@
...
+import static org.mockito.Mockito.*;
+
+class ValidateTripUsageTest {
+
+    @Test
+    void usedTripShouldNotGenerateNotice() {
+        final Logger mockLogger = mock(Logger.class);
+        final GtfsDataRepository mockDataRepo = mock(GtfsDataRepository.class);
+        final Map<String, Trip> mockTripCollection = new HashMap<>();
+        mockTripCollection.put("trip id", mock(Trip.class));
```
barbeau, line 43: I would suggest using something that looks like a real trip ID here, like `12345`, to make the code easier to read and distinguish from the field name `trip_id`.

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateTripUsageTest.java
```diff
@@ -0,0 +1,102 @@
...
+        when(mockDataRepo.getTripAll()).thenReturn(mockTripCollection);
+
+        final Map<String, TreeMap<Integer, StopTime>> mockStopTimeCollection = new HashMap<>();
+        final TreeMap<Integer, StopTime> innerStopTimeMap = new TreeMap<>();
+        mockStopTimeCollection.put("trip id", innerStopTimeMap);
```
barbeau, line 48: Same as above

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateTripUsageTest.java
```diff
@@ -0,0 +1,102 @@
...
+        verifyNoMoreInteractions(mockDataRepo, mockLogger);
+        verifyNoInteractions(mockResultRepo);
+    }
+
+    @Test
+    void unusedTripShouldGenerateNotice() {
+        final Logger mockLogger = mock(Logger.class);
+        final GtfsDataRepository mockDataRepo = mock(GtfsDataRepository.class);
+        final Map<String, Trip> mockTripCollection = new HashMap<>();
+        mockTripCollection.put("unused trip id", mock(Trip.class));
```
barbeau, line 70: Same as above

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateTripUsageTest.java
```diff
@@ -0,0 +1,102 @@
...
+        when(mockDataRepo.getTripAll()).thenReturn(mockTripCollection);
+
+        final Map<String, TreeMap<Integer, StopTime>> mockStopTimeCollection = new HashMap<>();
+        final TreeMap<Integer, StopTime> innerStopTimeMap = new TreeMap<>();
+        mockStopTimeCollection.put("trip id", innerStopTimeMap);
```
barbeau, line 75: Same as above

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/notice/error/UnusableTripNotice.java
```diff
@@ -0,0 +1,36 @@
...
+import org.mobilitydata.gtfsvalidator.domain.entity.notice.NoticeExporter;
+import org.mobilitydata.gtfsvalidator.domain.entity.notice.base.ErrorNotice;
+
+import java.io.IOException;
+
+public class UnusableTripNotice extends ErrorNotice {
+    public UnusableTripNotice(final String tripId) {
+        super("stop_times.txt", E_051,
+                "Trips must have more than one stop to be usable",
+                String.format("Trip with `trip_id`: `%s` should at least have two stops in stop_times.txt", tripId),
```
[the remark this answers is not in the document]
lionel-nj: @LeoFrachet suggested to add ticks around the value we refer to improve the error report legibility and ease the task of data providers when it comes to fixing issues related to their datasets. 
For example, if we do not have these ticks and the error report claims an error for trip with trip_id `trip id 00  `, it can be difficult to track down this entity given the "invisible" (space) postfixing the "visible" characters. Although this particular situation is borderline, it can happen, and that is why I put the ticks around both the field name and value.

However, I'm totally open to change this if required, for consistency we should update all notice using the same pattern.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/notice/error/EmptyFileErrorNotice.java
```diff
@@ -23,7 +23,7 @@
 
 public class EmptyFileErrorNotice extends ErrorNotice {
     public EmptyFileErrorNotice(String filename) {
-        super(filename, E_039,
+        super(filename, E_047,
```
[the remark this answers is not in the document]
lionel-nj: It appeared that this error code E039 was not the error code to be used for `EmptyFileErrorNotice` to ensure consistency with `RULES.md` doc. This probably comes an error when merging https://github.com/MobilityData/gtfs-validator/pull/317. 

Should I rather open a new issue referencing this, and apply this change in a new PR, so that this PR does not contain changes unrelated to  https://github.com/MobilityData/gtfs-validator/issues/162? @barbeau

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/notice/error/UnusableTripNotice.java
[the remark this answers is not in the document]
barbeau: That's fine, it can remain as-is. I would suggest getting feedback from providers at some point to make sure the assumption is correct.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/notice/error/EmptyFileErrorNotice.java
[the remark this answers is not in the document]
barbeau: Yes, I'd reference this in a new issue and separate PR, to avoid mixing unrelated things there.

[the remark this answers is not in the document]
lionel-nj: See https://github.com/MobilityData/gtfs-validator/issues/345

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/notice/error/BackwardsTimeTravelInStopNotice.java
```diff
@@ -27,7 +27,7 @@ public BackwardsTimeTravelInStopNotice(final String tripId,
                                            final String arrivalTimeAsString,
                                            final String previousStopDepartureTimeAsString,
                                            final Integer previousStopTimeStopSequence) {
-        super("stop_times.txt", E_047,
```
barbeau, line 30: Same here in terms of an unrelated change - was this intended, a bug fix, or an bad merge?

lionel-nj: Thanks, good catch I missed that one. It was intentional, so I'm removing this. Changes will be provided in a separate PR as previously discussed.

Issue #132, opened 2020-04-29
[REQUEST] Implement overlapping Stop times in Block trips (GTFS rule)


maximearmstrong:
**Is your feature request related to a problem? Please describe.**
Stop times should not be overlapping in Block trips. This is a GTFS rule implemented in Google Python validator and featured in Google Type Error as TYPE_BLOCK_TRIPS_WITH_OVERLAPPING_STOP_TIMES.

**Describe the solution you'd like**
Actual Google GTFS validator behaviour : verifies trips in a block. Verifies that a trip does not overlap with another one (same stop time for a stop).

**Describe alternatives you've considered**

**Additional context**
Line 39 in Error support priorities https://docs.google.com/spreadsheets/d/1vqe6wq7ctqk1EhYkgtZ0_TbcQ91vccfs2daSjn20BLE/edit#gid=0

barbeau:
@lionel-nj Looks like these are the Python unit tests for this rule, which match the interpretation we discussed:
https://github.com/google/transitfeed/blob/1c7f8c03612ac5c9c2b3f9a4cac6c229a04376b9/tests/transitfeed/testschedule.py#L62

One additional item to check, though, before calling it an error, is to make sure that the trips are assigned to the same service_id. You can see these checks in the Python tests.

lionel-nj:
Thank you. Looking at these tests it seems that an error is raised whether trips belong to the same service_id or not. @barbeau 

https://github.com/google/transitfeed/blob/1c7f8c03612ac5c9c2b3f9a4cac6c229a04376b9/tests/transitfeed/testschedule.py#L129

https://github.com/google/transitfeed/blob/1c7f8c03612ac5c9c2b3f9a4cac6c229a04376b9/tests/transitfeed/testschedule.py#L152

From my understanding the following situations would generate a notice: 

| routeId | tripId | serviceId | blockId | first stop time | last stop time |
|---------|--------|-----------|---------|-----------------|----------------|
| 0       | 2      | **a**         | 5       | 07:00           | 10:00          |
| 0       | 5      | **b**         | 5       | _08:45_           | _11:20_          |
| 0       | 8      | **c**         | 5       | 11:40           | 13:50          |

| routeId | tripId | serviceId | blockId | first stop time | last stop time |
|---------|--------|-----------|---------|-----------------|----------------|
| 0       | 2      | **a**         | 5       | 07:00           | 10:00          |
[...]

PR #346, opened 2020-08-20
Implement #132 - Verify no overlapping trips in block

lionel-nj:
**Summary:**

This PR provides support to verify that trips from same block do not overlap
**Expected behavior:** 

- [x] trips from same block with same service id that overlap should generate a notice
    - [x] rule logic
    - [x] unit tests
- [x] trips from same block with same service_id that do not overlap should not generate notice
    - [x] rule logic
    - [x] unit tests
- [x] trips from same block with with different service_id that overlap should generate notice if:
  - [x] only calendar.txt is provided
    - [x] rule logic
    - [x] unit tests
  - [x] only calendar_dates.txt is provided
    - [x] rule logic
    - [x] unit tests
- [x] trips from same block with with different service_id that do not overlap should not generate notice if:
  - [x] only calendar.txt is provided
    - [x] rule logic
    - [x] unit tests
  - [x] only calendar_dates.txt is provided
    - [x] rule logic
    - [x] unit tests

- [x] rework newly added notice constructor
- [x] update rules doc
- [x] retrieve and execute use case
- [x] additional unit tests for newly implemented methods in `Calendar.java`
- [x] additional test for notice export as json and pb
- [x] remove duplicate code
  - [x] update tests


lionel-nj:
Note: It seems that there is a bug in the time conversion utilities.

lionel-nj:
Note: It seems that there is a bug in the time conversion  utilities thi should be solved so that we can go forward on this PR, see https://github.com/MobilityData/gtfs-validator/issues/347. 

This results on bad calculation when it comes to determine if two stop times overlap. I will need to think further on this matter and will edit this comment with more information!

@barbeau @fabrice-v @maximearmstrong

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlock.java
```diff
@@ -0,0 +1,264 @@
...
+        this.logger = logger;
+        this.timeUtils = timeUtils;
+    }
+
+    /**
+     * Use case execution method: checks for each trip of the same block that no stop times overlap:
+     * - trips from same block with same service id that overlap should generate a notice
+     * - trips from same block with same service_id that do not overlap should not generate notice
+     * - trips from same block with with different service_id that overlap should generate notice whether calendar or
+     * calendar_Dates is provided
```
barbeau, line 56:
```suggestion
     * calendar_dates is provided
```

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlock.java
```diff
@@ -0,0 +1,264 @@
...
+                final StopTime currentTripFirstStopTimeInSequence =
+                        stopTimeCollectionForCurrentTrip.get(stopTimeCollectionForCurrentTrip.firstKey());
+                final StopTime currentTripLastStopTimeInSequence =
+                        stopTimeCollectionForCurrentTrip.get(stopTimeCollectionForCurrentTrip.lastKey());
+                final Integer currentTripFirstTime = currentTripFirstStopTimeInSequence.getArrivalTime();
+                final Integer currentTripLastTime = currentTripLastStopTimeInSequence.getDepartureTime();
+
+                visitedTripIdCollection.add(currentTripId);
+
+                if (currentTripFirstTime != null && currentTripLastTime != null) {
```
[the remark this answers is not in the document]
barbeau: For starters, you could do an early return `if (currentTripFirstTime != null && currentTripLastTime != null)` to skip to the next trip, so we know that none of the rest of the code block is executed in that case. That would remove one level of indent from the rest of the code and make it easier to read.

RULES.md
```diff
@@ -308,6 +309,13 @@ Trips must be referred to at least once in `stop_times.txt`.
 
 A trip must visit more than one stop in `stop_times.txt` to be usable by passengers for boarding and alighting.
 
+
+<a name="E054"/>
+
+### E054 - Block trips must not have overlapping stop times
+
+Stop times should not be overlapping in Block trips.
```
barbeau, line 327:
```suggestion
Trip stop times should not overlap when they are part of the same block operating on the same day.
```

RULES.md
```diff
@@ -308,6 +309,13 @@ Trips must be referred to at least once in `stop_times.txt`.
 
 A trip must visit more than one stop in `stop_times.txt` to be usable by passengers for boarding and alighting.
 
+
+<a name="E054"/>
+
+### E054 - Block trips must not have overlapping stop times
+
+Stop times should not be overlapping in Block trips. 
+
```
barbeau, line 328: I would add a reference here too to trips.txt

adapter/utils/src/main/java/org/mobilitydata/gtfsvalidator/timeutils/TimeUtilsImpl.java
```diff
@@ -118,4 +119,10 @@ public String convertIntegerToHMMSS(final Integer elapsedDurationSinceNoonInSeco
                     secondsAfterMidnight);
         }
     }
+
+    @Override
+    public boolean arePeriodsOverlapping(final int firstPeriodFirstTime, final int firstPeriodLastTime,
+                                         final int secondPeriodFirstTime, final int secondPeriodLastTime) {
+        return !(firstPeriodFirstTime <= secondPeriodLastTime && firstPeriodLastTime <= secondPeriodFirstTime);
```
barbeau: I believe this requires the same change as the frequency overlap logic to account for when the second period times occur before the first.

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlock.java
```diff
@@ -0,0 +1,264 @@
...
+        final Map<String, List<Trip>> tripPerBlockId = dataRepo.getAllTripByBlockId();
+        final boolean areCalendarProvided = dataRepo.getCalendarAll().size() != 0;
+
+        tripPerBlockId.forEach((blockId, tripCollectionFromSameBlockId) -> {
+            final List<String> visitedTripIdCollection = new ArrayList<>();
+            tripCollectionFromSameBlockId.forEach(currentTrip -> {
+
+                final String currentTripId = currentTrip.getTripId();
+                final SortedMap<Integer, StopTime> stopTimeCollectionForCurrentTrip =
+                        dataRepo.getStopTimeByTripId(currentTrip.getTripId());
```
barbeau, line 72:
```suggestion
                        dataRepo.getStopTimeByTripId(currentTripId);
```

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlock.java
```diff
@@ -0,0 +1,264 @@
...
+
+                final StopTime currentTripFirstStopTimeInSequence =
```
barbeau, line 74: I would try to shorten some of these variable names, as it becomes difficult to read. Maybe `tripFirstStopTime`?

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlock.java
```diff
@@ -0,0 +1,264 @@
...
+     */
+    public void execute() {
+        logger.info("validating rule 'E054 - Trips from same block overlap'");
+
+        final Map<String, List<Trip>> tripPerBlockId = dataRepo.getAllTripByBlockId();
+        final boolean areCalendarProvided = dataRepo.getCalendarAll().size() != 0;
+
+        tripPerBlockId.forEach((blockId, tripCollectionFromSameBlockId) -> {
+            final List<String> visitedTripIdCollection = new ArrayList<>();
+            tripCollectionFromSameBlockId.forEach(currentTrip -> {
```
barbeau, line 68: I would try to shorten some of these variable names, as it becomes difficult to read. Maybe just `trips` instead of `tripCollectionFromSameBlockId`?

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlock.java
```diff
@@ -0,0 +1,264 @@
...
+     * - trips from same block with with different service_id that overlap should generate notice whether calendar or
+     * calendar_Dates is provided
+     * - trips from same block with with different service_id that do not overlap should not generate notice whether
+     * calendar or calendar_dates is provided
+     */
+    public void execute() {
+        logger.info("validating rule 'E054 - Trips from same block overlap'");
+
+        final Map<String, List<Trip>> tripPerBlockId = dataRepo.getAllTripByBlockId();
+        final boolean areCalendarProvided = dataRepo.getCalendarAll().size() != 0;
```
barbeau, line 64: `areCalendarProvided` -> `isCalendarProvided`

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlock.java
```diff
@@ -0,0 +1,264 @@
...
+
+        tripPerBlockId.forEach((blockId, tripCollectionFromSameBlockId) -> {
+            final List<String> visitedTripIdCollection = new ArrayList<>();
```
barbeau, line 67: Change to `HashMap`

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlock.java
```diff
@@ -0,0 +1,264 @@
...
+                                final StopTime unvisitedTripLastStopTimeInSequence =
+                                        stopTimesForUnvisitedTrip.get(stopTimesForUnvisitedTrip.lastKey());
+
+                                final Integer unvisitedTripFirstTime =
+                                        unvisitedTripFirstStopTimeInSequence.getArrivalTime();
+                                final Integer unvisitedTripLastTime =
+                                        unvisitedTripLastStopTimeInSequence.getDepartureTime();
+
+                                if (unvisitedTrip.getServiceId().equals(currentTrip.getServiceId())) {
+
```
barbeau, line 110: Extra blank line

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlock.java
```diff
@@ -0,0 +1,264 @@
...
+                                    // stop_times arrival_time and departure_time should not be null to perform the
+                                    // subsequent operations
+                                    if (unvisitedTripFirstTime != null && unvisitedTripLastTime != null) {
```
barbeau, line 113: Same here - I believe this could be an early return?

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlock.java
```diff
@@ -0,0 +1,264 @@
...
+                                        // stop_times arrival_time and departure_time are supposed to be ordered
+                                        if (unvisitedTripFirstTime <= unvisitedTripLastTime) {
```
barbeau, line 115: The trip first and last time can't be equal, right? Otherwise the trip is `0` seconds long?

Also, if this is another sanity check, I'd suggest refactoring to `areTripTimesValid()` method to better explain what's happening.

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlock.java
```diff
@@ -0,0 +1,264 @@
...
+                        stopTimeCollectionForCurrentTrip.get(stopTimeCollectionForCurrentTrip.firstKey());
+                final StopTime currentTripLastStopTimeInSequence =
+                        stopTimeCollectionForCurrentTrip.get(stopTimeCollectionForCurrentTrip.lastKey());
+                final Integer currentTripFirstTime = currentTripFirstStopTimeInSequence.getArrivalTime();
+                final Integer currentTripLastTime = currentTripLastStopTimeInSequence.getDepartureTime();
+
+                visitedTripIdCollection.add(currentTripId);
+
+                if (currentTripFirstTime != null && currentTripLastTime != null) {
+                    if (currentTripFirstTime <= currentTripLastTime) {
```
barbeau, line 84: The trip first and last time can't be equal, right? Otherwise the trip is `0` seconds long?

Also, if this is another sanity check, I'd suggest refactoring to `areTripTimesValid()` method to better explain what's happening.

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlock.java
```diff
@@ -0,0 +1,264 @@
...
+
+                final String currentTripId = currentTrip.getTripId();
+                final SortedMap<Integer, StopTime> stopTimeCollectionForCurrentTrip =
+                        dataRepo.getStopTimeByTripId(currentTrip.getTripId());
+
+                final StopTime currentTripFirstStopTimeInSequence =
+                        stopTimeCollectionForCurrentTrip.get(stopTimeCollectionForCurrentTrip.firstKey());
+                final StopTime currentTripLastStopTimeInSequence =
+                        stopTimeCollectionForCurrentTrip.get(stopTimeCollectionForCurrentTrip.lastKey());
+                final Integer currentTripFirstTime = currentTripFirstStopTimeInSequence.getArrivalTime();
```
barbeau, line 78: Same here on name length - I think you could just shorten this to `tripFirstTime`?

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlock.java
```diff
@@ -0,0 +1,264 @@
...
+                                        stopTimesForUnvisitedTrip.get(stopTimesForUnvisitedTrip.firstKey());
+                                final StopTime unvisitedTripLastStopTimeInSequence =
+                                        stopTimesForUnvisitedTrip.get(stopTimesForUnvisitedTrip.lastKey());
+
+                                final Integer unvisitedTripFirstTime =
+                                        unvisitedTripFirstStopTimeInSequence.getArrivalTime();
+                                final Integer unvisitedTripLastTime =
+                                        unvisitedTripLastStopTimeInSequence.getDepartureTime();
+
+                                if (unvisitedTrip.getServiceId().equals(currentTrip.getServiceId())) {
```
barbeau, line 109: Is service ID guaranteed not to be `null` here?

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlock.java
```diff
@@ -0,0 +1,264 @@
...
+                                                        new BlockTripsWithOverlappingStopTimesNotice(
+                                                                currentTripId,
+                                                                timeUtils.convertIntegerToHMMSS(currentTripFirstTime),
+                                                                timeUtils.convertIntegerToHMMSS(currentTripLastTime),
+                                                                unvisitedTripId,
+                                                                timeUtils.convertIntegerToHMMSS(unvisitedTripFirstTime),
+                                                                timeUtils.convertIntegerToHMMSS(unvisitedTripLastTime),
+                                                                blockId)
+                                                );
+
```
barbeau, line 131: Extra blank lines

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlock.java
```diff
@@ -0,0 +1,264 @@
...
+
+                                            }
+                                        }
+                                    }
+                                } else {
+                                    // if trips have different service_id additional check needs to be done to determine
+                                    // whether they operate on the same day or noy
```
barbeau, line 138:
```suggestion
                                    // whether they operate on the same day or not
```

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlock.java
```diff
@@ -0,0 +1,264 @@
...
+                                    }
+                                    // interpret data from `calendar.txt` to determine if currentTrip and unvisitedTrip
+                                    // operate on same days
+                                    if (areCalendarProvided) {
+                                        // first check if `start_date` and `end_date` overlap for both trips
+                                        if (currentTripCalendar.areCalendarOverlapping(unvisitedTripCalendar)) {
+                                            // determine the if the trips operate on same days
+                                            if (currentTripCalendar.getCalendarsCommonOperationDayCollection(
+                                                    unvisitedTripCalendar).size() >= 1) {
+
```
barbeau, line 169: extra blank line

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlock.java
```diff
@@ -0,0 +1,264 @@
...
+                                                if (unvisitedTripFirstTime != null && unvisitedTripLastTime != null) {
+
```
barbeau, line 171: extra blank line

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlock.java
```diff
@@ -0,0 +1,264 @@
...
+                                    // interpret data from `calendar.txt` to determine if currentTrip and unvisitedTrip
+                                    // operate on same days
+                                    if (areCalendarProvided) {
+                                        // first check if `start_date` and `end_date` overlap for both trips
+                                        if (currentTripCalendar.areCalendarOverlapping(unvisitedTripCalendar)) {
+                                            // determine the if the trips operate on same days
+                                            if (currentTripCalendar.getCalendarsCommonOperationDayCollection(
+                                                    unvisitedTripCalendar).size() >= 1) {
+
+                                                if (unvisitedTripFirstTime != null && unvisitedTripLastTime != null) {
```
barbeau, line 170: Same - early return?

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlock.java
```diff
@@ -0,0 +1,264 @@
...
+
+                                                    if (unvisitedTripFirstTime <= unvisitedTripLastTime) {
```
barbeau, line 172: Same as previous - The trip first and last time can't be equal, right? Otherwise the trip is 0 seconds long?

Also, if this is another sanity check, I'd suggest refactoring to areTripTimesValid() method to better explain what's happening.

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlock.java
```diff
@@ -0,0 +1,264 @@
...
+
+                                        // get the list of common dates during which trips with different service_id
+                                        // operate: this represent the list of dates where there could be potential time
+                                        // overlap.
+                                        final Set<String> potentialConflictingDates =
+                                                new HashSet<>(Set.copyOf(currentTripDatePossibilityCollection));
+                                        potentialConflictingDates.retainAll(unvisitedTripDatePossibilityCollection);
+
+                                        if (potentialConflictingDates.size() >= 1) {
+
```
barbeau, line 227: extra blank line

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlock.java
```diff
@@ -0,0 +1,264 @@
...
+                                            if (unvisitedTripFirstTime != null && unvisitedTripLastTime != null) {
+
```
barbeau, line 229: extra blank line

application/cli-app/src/main/java/org/mobilitydata/gtfsvalidator/Main.java
```diff
@@ -187,30 +187,31 @@ public static void main(String[] args) {
...
-                config.validateFeedCoversTheNext30ServiceDays().execute();
-                config.validateFeedInfoFeedEndDateIsPresent().execute();
-                config.validateFeedInfoFeedStartDateIsPresent().execute();
-                config.validateStopTimeDepartureTimeAfterArrivalTime().execute();
-                config.validateTripEdgeArrivalDepartureTime().execute();
-                config.validateTripTravelSpeed().execute();
-                config.validateTripUsage().execute();
-                config.validateTripNumberOfStops().execute();
-                config.validateFrequencyStartTimeBeforeEndTime().execute();
+//                config.validateRouteShortNameLength().execute();
```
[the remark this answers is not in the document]
lionel-nj: Indeed, that was for testing.

adapter/utils/src/test/java/org/mobilitydata/gtfsvalidator/timeutils/TimeUtilsImplTest.java
```diff
@@ -240,4 +241,19 @@ void convertHHMMSStoIntegerAndIntegerToHHMMSSShouldBeConsistent() {
         assertEquals(toCheck, TIME_CONVERSION_UTILS.convertIntegerToHMMSS(
                 TIME_CONVERSION_UTILS.convertHHMMSSToIntFromNoonOfDayOfService(toCheck)));
     }
+
```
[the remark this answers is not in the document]
lionel-nj: Merge of master provided all the time period permutation checks

adapter/utils/src/main/java/org/mobilitydata/gtfsvalidator/timeutils/TimeUtilsImpl.java
lionel-nj: Indeed

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlock.java
lionel-nj: Indeed, these cannot be equal. Will modify.

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlock.java
lionel-nj: https://github.com/MobilityData/gtfs-validator/pull/346#discussion_r481122879

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlock.java
lionel-nj: Indeed, the GTFS specification required this field to be provided, and we have a null check in `TripBuilder` `.build` method.

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlock.java
```diff
@@ -0,0 +1,364 @@
...
+                                                                            final StopTime otherTripFirstStopTime,
+                                                                            final StopTime otherTripLastStopTime) {
+        // interpret data from `calendar_dates.txt` to determine if currentTrip and
+        // otherTrip operate on same days
+        final Map<String, CalendarDate> tripCalendarDateCollection =
+                dataRepo.getCalendarDateAll().get(trip.getServiceId());
+        final Map<String, CalendarDate> unvisitedTripCalendarDateCollection =
+                dataRepo.getCalendarDateAll().get(otherTrip.getServiceId());
+        final Set<String> currentTripServiceDates = new HashSet<>();
+        tripCalendarDateCollection.forEach((date, calendarDate) -> {
```
barbeau, line 319: I believe this should be:

```java
        tripCalendarDateCollection.forEach((serviceId, calendarDate) -> {
```

...which means that the below line is adding the service ID and not the LocalDate to the HashSet, and needs to be:

```java
                currentTripServiceDates.add(calendarDate);
```

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlock.java
```diff
@@ -0,0 +1,364 @@
...
+            if (calendarDate.getExceptionType() == ExceptionType.ADDED_SERVICE) {
+                currentTripServiceDates.add(date);
+            }
+        });
+        final Set<String> unvisitedTripServiceDates = new HashSet<>();
+        unvisitedTripCalendarDateCollection.forEach((date, calendarDate) -> {
```
barbeau, line 325: Same as above

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlock.java
```diff
@@ -0,0 +1,364 @@
...
+            if (calendarDate.getExceptionType() == ExceptionType.ADDED_SERVICE) {
+                unvisitedTripServiceDates.add(date);
+            }
+        });
+        // get the list of common dates during which trips with different service_id
+        // operate: this represent the list of dates where there could be potential time
+        // overlap.
+        final Set<String> potentialConflictingDates =
```
barbeau, line 333: Copying the set and then iterating over it to remove duplicates via `retainAll()` is a potentially very expensive operation on datasets like the Netherlands that put all service in calendar_dates.txt.

I think you can combine this code with the loop through the `unvisitedTripCalendarDateCollection` to have a more efficient implementation in time and memory. Basically, as you loop through `unvisitedTripCalendarDateCollection`, do `currentTripServiceDates.contains(calendarDate)`. If it's `true`, then add the date to the potentially conflicting date list (`potentialConflictingDates`).

Like:

```java
        final Set<String> potentialConflictingDates = new HashSet<>();
        unvisitedTripCalendarDateCollection.forEach((serviceId, calendarDate) -> {
            if (calendarDate.getExceptionType() == ExceptionType.ADDED_SERVICE) {
                if (currentTripServiceDates.contains(calendarDate)) {
                    potentialConflictingDates.add(calendarDate);
                }
            }
        });
```

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlock.java
barbeau: Oh, looking at `InMemoryGtfsDataRepository`, this string is actually a string representation of the date. Which is kind of confusing here, although I understand why we did it (to check for duplicates when importing calendar_dates.txt).

I think it's safer to add the `LocalDate` to the set here (i.e., the `calendarDate` field) as it's not a (somewhat) arbitrary string key based on the repository implementation and is more specific to our domain of date checking in this method.

barbeau: Actually, in the in memory repo, could we instead do:

```java
Map<String, Set<CalendarDate>> calendarDatePerServiceIdAndDate = new HashMap<>();
```

For data like the Netherlands with all service in calendar_dates.txt it would save a ton of duplicate strings in memory. Although we'd need to implement hashcode() and equals() for CalendarDate. Maybe open a new issue for this change?

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlock.java
barbeau: Actually, reading https://docs.oracle.com/javase/8/docs/api/java/time/LocalDate.html, it says:

>use of identity-sensitive operations (including reference equality (==), identity hash code, or synchronization) on instances of LocalDate may have unpredictable results and should be avoided.

So you'd need to change the above to use the `String` representation of the date instead of `LocalDate`.

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlock.java
```diff
@@ -0,0 +1,364 @@
...
+        // interpret data from `calendar.txt` to determine if currentTrip and otherTrip
+        // operate on same days
+        Calendar tripCalendar = dataRepo.getCalendarByServiceId(trip.getServiceId());
+        Calendar unvisitedTripCalendar = dataRepo.getCalendarByServiceId(otherTrip.getServiceId());
+        // first check if `start_date` and `end_date` overlap for both trips
+        if (!tripCalendar.isOverlapping(unvisitedTripCalendar)) {
+            return;
+        }
+        // determine the if the trips operate on same days
+        if (tripCalendar.getOverlappingDays(unvisitedTripCalendar).size() == 0) {
```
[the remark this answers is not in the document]
lionel-nj: Will add the comment, also I am opening an issue on this for later.

domain/src/test/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/CalendarTest.java
```diff
@@ -626,4 +625,277 @@ public void createCalendarWithValidValuesShouldNotGenerateNotice() {
...
+                .sunday(0)
+                .startDate(LocalDate.of(2020, 3, 1))
+                .endDate(LocalDate.of(2020, 4, 30))
+                .build()
+                .getData();
+
+        assertFalse(firstCalendar.isOverlapping(nonOverlappingCalendar));
+        // permute periods `a` and `b`
+        //            b  |----|    2020/03/01 -> 2020/04/30
+        // a |---      -----|      2020/01/01 -> 2020/01/31
```
barbeau, line 665: `a` needs to be moved after `b` in the ASCII art here

domain/src/test/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/CalendarTest.java
```diff
@@ -626,4 +625,277 @@ public void createCalendarWithValidValuesShouldNotGenerateNotice() {
...
+    }
+
+    @Test
+    void periodContainedOtherPeriodShouldReturnTrueWhenCallingIsOverlapping() {
+        // periods `b` is contained within period `a`
+        //  a |--------| 2020/01/01 -> 2020/12/31
+        //   b  |----|   2020/03/01 -> 2020/04/30
+        final Calendar.CalendarBuilder underTest = new Calendar.CalendarBuilder();
+
+        final Calendar firstCalendar = (Calendar) underTest.serviceId(SERVICE_ID)
```
barbeau, line 676: I'd suggest naming these variables like in the ASCII art, so it's easy to match the two. So this would be `...Calendar a...`.

domain/src/test/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/CalendarTest.java
```diff
@@ -626,4 +625,277 @@ public void createCalendarWithValidValuesShouldNotGenerateNotice() {
...
+                .sunday(0)
+                .startDate(LocalDate.of(2020, 5, 6))
+                .endDate(LocalDate.of(2020, 12, 31))
+                .build()
+                .getData();
+
+        assertTrue(firstCalendar.isOverlapping(overlappingCalendar));
+
+        // permute periods `a` and `b` in method call
+        //  a |--------|     2020/01/01 -> 2020/12/31
```
barbeau, line 788: ASCII art needs to be updated? It's the same as the above test.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/notice/error/BlockTripsWithOverlappingStopTimesNotice.java
```diff
@@ -0,0 +1,87 @@
...
+        super("trips.txt",
+                E_054,
+                "Overlapping stop times in trip block",
+                String.format("Trips with `trip_id`: `%s` (first time: `%s`, last time: `%s`)" +
+                                " and `%s` (first time: `%s`, last time: `%s`) from block: `%s` overlap.",
+                        tripId, tripFirstTime, tripLastTime, previousTripId, previousTripFirstTime,
+                        previousTripLastTime, blockId),
+                tripId);
+
+        putNoticeSpecific(KEY_TRIP_TRIP_ID, tripId);
```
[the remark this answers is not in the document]
lionel-nj: I think that this could be done by uing `Object` typ in the notice constructor for field `conflictingDateCollection`. Then, the null check can be done just right before editing the error message:
This could work:
```
    public BlockTripsWithOverlappingStopTimesNotice(final String tripId,
                                                    final String tripFirstTime,
                                                    final String tripLastTime,
                                                    final String previousTripId,
                                                    final String previousTripFirstTime,
                                                    final String previousTripLastTime,
                                                    final String blockId,
                                                    final Object conflictingDateCollection) {
        super("trips.txt",
                E_054,
                "Block trips must not have overlapping stop times",
                conflictingDateCollection != null ?
                String.format("Trips with `trip_id`: `%s` (first time: `%s`, last time: `%s`)" +
                                " and `%s` (first time: `%s`, last time: `%s`) from block: `%s` overlap." +
                                " Conflicts happen on: %s.",
                        tripId, tripFirstTime, tripLastTime, previousTripId, previousTripFirstTime,
                        previousTripLastTime, blockId, ((Set) conflictingDateCollection).toString()) :
                String.format("Trips with `trip_id`: `%s` (first time: `%s`, last time: `%s`)" +
                                " and `%s` (first time: `%s`, last time: `%s`) from block: `%s` overlap.",
                        tripId, tripFirstTime, tripLastTime, previousTripId, previousTripFirstTime,
                        previousTripLastTime, blockId),
                tripId);

        putNoticeSpecific(KEY_TRIP_TRIP_ID, tripId);
        putNoticeSpecific(KEY_TRIP_BLOCK_ID, blockId);
        putNoticeSpecific(KEY_TRIP_FIRST_TIME, tripFirstTime);
        putNoticeSpecific(KEY_TRIP_LAST_TIME, tripLastTime);
        putNoticeSpecific(KEY_TRIP_PREVIOUS_TRIP_ID, previousTripId);
        putNoticeSpecific(KEY_PREVIOUS_TRIP_FIRST_TIME, previousTripFirstTime);
        putNoticeSpecific(KEY_PREVIOUS_TRIP_LAST_TIME, previousTripLastTime);
        putNoticeSpecific(KEY_CONFLICTING_DATE_LIST, conflictingDateCollection);
    }
```

barbeau:
Hmmm...I think because in this solution you lose the strong typing for the
collection and that you still have to duplicate the majority of the text
I'd stick with the original two constructors.

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlockTest.java
```diff
@@ -0,0 +1,1523 @@
...
+        verify(mockTimeUtils, times(3))
+                .arePeriodsOverlapping(ArgumentMatchers.anyInt(), ArgumentMatchers.anyInt(),
+                        ArgumentMatchers.anyInt(), ArgumentMatchers.anyInt());
+
+        final ArgumentCaptor<BlockTripsWithOverlappingStopTimesNotice> captor =
+                ArgumentCaptor.forClass(BlockTripsWithOverlappingStopTimesNotice.class);
+
+        verify(mockResultRepo, times(1)).addNotice(captor.capture());
+
+        final List<BlockTripsWithOverlappingStopTimesNotice> noticeList = captor.getAllValues();
```
barbeau, line 752: I would assert the size of the `noticeList` with the expected number of errors.

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlockTest.java
```diff
@@ -0,0 +1,1523 @@
...
+        // | 0       | 8      | c         | 7       | 09:40           | 13:50          |
+
+        // calendar.txt
+        // | serviceId | monday | tuesday | wednesday | thursday | friday | saturday | sunday | start_date | end_date |
+        // |-----------|--------|---------|-----------|----------|--------|----------|--------|------------|----------|
+        // | a         | 1      | 1       | 0         | 0        | 0      | 0        | 0      | 20200801   | 20200831 |
+        // | b         | 1      | 0       | 1         | 0        | 0      | 0        | 0      | 20200901   | 20209030 |
+        // | c         | 0      | 1       | 0         | 1        | 1      | 1        | 0      | 20200801   | 20200831 |
+        //
+        // Here, services a and c overlap: they share the same date range and are both active on tuesday. Specifically
```
barbeau, line 803: 💯 for the additional written explanation

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlockTest.java
```diff
@@ -0,0 +1,1523 @@
...
+        when(mockTimeUtils.convertIntegerToHMMSS(1350)).thenReturn("1350");
+        when(mockTimeUtils.arePeriodsOverlapping(700, 1000,
+                940, 1350)).thenReturn(true);
+
+        final Trip firstMockTrip = mock(Trip.class);
+        when(firstMockTrip.getTripId()).thenReturn("2");
+        when(firstMockTrip.getServiceId()).thenReturn("a");
+        when(firstMockTrip.getBlockId()).thenReturn("7");
+        when(firstMockTrip.hasSameServiceId(any())).thenReturn(false);
+
```
barbeau, line 825: Extra blank line

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlockTest.java
```diff
@@ -0,0 +1,1523 @@
...
+        when(mockTimeUtils.convertIntegerToHMMSS(940)).thenReturn("940");
+        when(mockTimeUtils.convertIntegerToHMMSS(1350)).thenReturn("1350");
+        when(mockTimeUtils.arePeriodsOverlapping(700, 1000,
+                940, 1350)).thenReturn(true);
+
+        final Trip firstMockTrip = mock(Trip.class);
+        when(firstMockTrip.getTripId()).thenReturn("2");
+        when(firstMockTrip.getServiceId()).thenReturn("a");
+        when(firstMockTrip.getBlockId()).thenReturn("7");
+        when(firstMockTrip.hasSameServiceId(any())).thenReturn(false);
```
barbeau, line 824: Same on not mocking domain class methods

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlockTest.java
```diff
@@ -0,0 +1,1523 @@
...
+
+
+        final Trip secondMockTrip = mock(Trip.class);
+        when(secondMockTrip.getTripId()).thenReturn("5");
+        when(secondMockTrip.getServiceId()).thenReturn("b");
+        when(secondMockTrip.getBlockId()).thenReturn("7");
+        when(secondMockTrip.hasSameServiceId(any())).thenReturn(false);
```
barbeau, line 831: Same on not mocking domain class methods

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlockTest.java
```diff
@@ -0,0 +1,1523 @@
...
+
+        final Trip thirdMockTrip = mock(Trip.class);
+        when(thirdMockTrip.getTripId()).thenReturn("8");
+        when(thirdMockTrip.getServiceId()).thenReturn("c");
+        when(thirdMockTrip.getBlockId()).thenReturn("7");
+        when(thirdMockTrip.hasSameServiceId(any())).thenReturn(false);
```
barbeau, line 837: Same on not mocking domain class methods

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlockTest.java
```diff
@@ -0,0 +1,1523 @@
...
+        when(mockDataRepo.getStopTimeByTripId("8")).thenReturn(mockThirdTripStopTimeCollection);
+
+        final Map<String, Calendar> mockCalendarCollection = new HashMap<>();
+        final Calendar calendarForServiceA = mock(Calendar.class);
+        final Calendar calendarForServiceB = mock(Calendar.class);
+        final Calendar calendarForServiceC = mock(Calendar.class);
+        mockCalendarCollection.put("a", calendarForServiceA);
+        mockCalendarCollection.put("b", calendarForServiceB);
+        mockCalendarCollection.put("c", calendarForServiceC);
+        when(calendarForServiceA.isOverlapping(calendarForServiceB)).thenReturn(false);
```
barbeau, line 901: Same on not mocking domain class methods

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlockTest.java
```diff
@@ -0,0 +1,1523 @@
...
+        when(calendarForServiceA.isOverlapping(calendarForServiceC)).thenReturn(true);
+        when(calendarForServiceB.isOverlapping(calendarForServiceC)).thenReturn(false);
+        final LocalDate serviceAStartDate = LocalDate.of(2020, 8, 1);
```
barbeau, line 904: Same on not declaring extra variables here

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlockTest.java
```diff
@@ -0,0 +1,1523 @@
...
+        final LocalDate serviceCStartDate = LocalDate.of(2020, 8, 1);
+        final LocalDate serviceCEndDate = LocalDate.of(2020, 8, 31);
+
+        when(calendarForServiceA.getStartDate()).thenReturn(serviceAStartDate);
+        when(calendarForServiceA.getEndDate()).thenReturn(serviceAEndDate);
+        when(calendarForServiceB.getStartDate()).thenReturn(serviceBStartDate);
+        when(calendarForServiceB.getEndDate()).thenReturn(serviceBEndDate);
+        when(calendarForServiceC.getStartDate()).thenReturn(serviceCStartDate);
+        when(calendarForServiceC.getEndDate()).thenReturn(serviceCEndDate);
+        when(calendarForServiceA.getOverlappingDays(calendarForServiceC))
```
barbeau, line 917: Same on not mocking domain class methods

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlockTest.java
```diff
@@ -0,0 +1,1523 @@
...
+        verify(mockTimeUtils, times(1))
+                .arePeriodsOverlapping(ArgumentMatchers.anyInt(), ArgumentMatchers.anyInt(),
+                        ArgumentMatchers.anyInt(), ArgumentMatchers.anyInt());
+
+        final ArgumentCaptor<BlockTripsWithOverlappingStopTimesNotice> captor =
+                ArgumentCaptor.forClass(BlockTripsWithOverlappingStopTimesNotice.class);
+
+        verify(mockResultRepo, times(1)).addNotice(captor.capture());
+
+        final List<BlockTripsWithOverlappingStopTimesNotice> noticeList = captor.getAllValues();
```
barbeau, line 985: Same on asserting number of expected errors in `noticeList`

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlockTest.java
```diff
@@ -0,0 +1,1523 @@
...
+
+        final GtfsDataRepository mockDataRepo = mock(GtfsDataRepository.class);
+        final Map<String, List<Trip>> mockTripPerBlockIdCollection = new HashMap<>();
+        final List<Trip> mockTripCollection = new ArrayList<>();
+
+        final Trip firstMockTrip = mock(Trip.class);
+        when(firstMockTrip.getTripId()).thenReturn("2");
+        when(firstMockTrip.getServiceId()).thenReturn("a");
+        when(firstMockTrip.getBlockId()).thenReturn("7");
+        when(firstMockTrip.hasSameServiceId(any())).thenReturn(false);
```
barbeau, line 1053: Same on not mocking domain class methods

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlockTest.java
```diff
@@ -0,0 +1,1523 @@
...
+
+        final Trip secondMockTrip = mock(Trip.class);
+        when(secondMockTrip.getTripId()).thenReturn("5");
+        when(secondMockTrip.getServiceId()).thenReturn("b");
+        when(secondMockTrip.getBlockId()).thenReturn("7");
+        when(secondMockTrip.hasSameServiceId(any())).thenReturn(false);
```
barbeau, line 1059: Same on not mocking domain class methods

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlockTest.java
```diff
@@ -0,0 +1,1523 @@
...
+
+        final Trip thirdMockTrip = mock(Trip.class);
+        when(thirdMockTrip.getTripId()).thenReturn("8");
+        when(thirdMockTrip.getServiceId()).thenReturn("c");
+        when(thirdMockTrip.getBlockId()).thenReturn("7");
+        when(thirdMockTrip.hasSameServiceId(any())).thenReturn(false);
```
barbeau, line 1065: Same on not mocking domain class methods

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlockTest.java
```diff
@@ -0,0 +1,1523 @@
...
+
+        mockTripCollection.add(firstMockTrip);
+        mockTripCollection.add(secondMockTrip);
+        mockTripCollection.add(thirdMockTrip);
+        mockTripPerBlockIdCollection.put("7", mockTripCollection);
+
+        final SortedMap<Integer, StopTime> mockFirstTripStopTimeCollection = new TreeMap<>();
+
+        final StopTime firstTripFirstStopTime = mock(StopTime.class);
+        when(firstTripFirstStopTime.getArrivalTime()).thenReturn(700);
```
barbeau, line 1075: Same on negative representation of seconds before noon

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlockTest.java
```diff
@@ -0,0 +1,1523 @@
...
+        when(calendarDateForServiceC.getExceptionType()).thenReturn(ExceptionType.ADDED_SERVICE);
+        mockCalendarDateCollection.put("a", calendarDateCollectionServiceA);
+        mockCalendarDateCollection.put("b", calendarDateCollectionServiceB);
+        mockCalendarDateCollection.put("c", calendarDateCollectionServiceC);
+        when(mockDataRepo.getCalendarDateAll()).thenReturn(mockCalendarDateCollection);
+
+        final ValidationResultRepository mockResultRepo = mock(ValidationResultRepository.class);
+        final Logger mockLogger = mock(Logger.class);
+        final TimeUtils mockTimeUtils = mock(TimeUtils.class);
+        when(mockTimeUtils.convertIntegerToHMMSS(700)).thenReturn("700");
```
barbeau, line 1146: Same on "HH:MM:SS" values

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlockTest.java
```diff
@@ -0,0 +1,1523 @@
...
+        verify(secondTripLastStopTime, times(3)).getDepartureTime();
+
+        verify(thirdTripLastStopTime, times(5)).getDepartureTime();
+
+        final ArgumentCaptor<BlockTripsWithOverlappingStopTimesNotice> captor =
+                ArgumentCaptor.forClass(BlockTripsWithOverlappingStopTimesNotice.class);
+
+        verify(mockResultRepo, times(1)).addNotice(captor.capture());
+
+        final List<BlockTripsWithOverlappingStopTimesNotice> noticeList = captor.getAllValues();
```
barbeau, line 1204: Same on asserting number of errors in `noticeList`

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlockTest.java
```diff
@@ -0,0 +1,1523 @@
...
+
+        assertEquals("trips.txt", noticeList.get(0).getFilename());
```
barbeau, line 1206: I would suggest asserting the exact error code for all tests too.

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlockTest.java
```diff
@@ -0,0 +1,1523 @@
...
+        when(mockDataRepo.getStopTimeByTripId("8")).thenReturn(mockThirdTripStopTimeCollection);
+
+        final Map<String, Calendar> mockCalendarCollection = new HashMap<>();
+        final Calendar calendarForServiceA = mock(Calendar.class);
+        final Calendar calendarForServiceB = mock(Calendar.class);
+        final Calendar calendarForServiceC = mock(Calendar.class);
+        mockCalendarCollection.put("a", calendarForServiceA);
+        mockCalendarCollection.put("b", calendarForServiceB);
+        mockCalendarCollection.put("c", calendarForServiceC);
+        when(calendarForServiceA.isOverlapping(calendarForServiceB)).thenReturn(false);
```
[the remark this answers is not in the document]
lionel-nj: As discussed, implementation of domain layer classes is tested separately, so that tests of use case layer only verify interaction with mocked objects.

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlockTest.java
```diff
@@ -0,0 +1,1523 @@
...
+        when(calendarDateForServiceB.getExceptionType()).thenReturn(ExceptionType.REMOVED_SERVICE);
+        when(calendarDateForServiceC.getExceptionType()).thenReturn(ExceptionType.ADDED_SERVICE);
+        mockCalendarDateCollection.put("a", calendarDateCollectionServiceA);
+        mockCalendarDateCollection.put("b", calendarDateCollectionServiceB);
+        mockCalendarDateCollection.put("c", calendarDateCollectionServiceC);
+        when(mockDataRepo.getCalendarDateAll()).thenReturn(mockCalendarDateCollection);
+
+        final ValidationResultRepository mockResultRepo = mock(ValidationResultRepository.class);
+        final Logger mockLogger = mock(Logger.class);
+        final TimeUtils mockTimeUtils = mock(TimeUtils.class);
```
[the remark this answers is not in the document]
lionel-nj: Actually there is no interaction with `mockTimeUtils`, so I chose not to mock `timeUtils.arePeriodsOverlapping()`

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlockTest.java
lionel-nj: As discussed, implementation of domain layer classes is tested separately, so that tests of use case layer only verify interactions with mocked objects.

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlockTest.java
lionel-nj: As discussed, implementation of domain layer classes is tested separately, so that tests of use case layer only verify interactions with mocked objects.

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlockTest.java
lionel-nj: As discussed, implementation of domain layer classes is tested separately, so that tests of use case layer only verify interactions with mocked objects.

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlockTest.java
lionel-nj: As discussed, implementation of domain layer classes is tested separately, so that tests of use case layer only verify interactions with mocked objects.

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlockTest.java
lionel-nj: As discussed, implementation of domain layer classes is tested separately, so that tests of use case layer only verify interactions with mocked objects.

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlockTest.java
lionel-nj: As discussed, implementation of domain layer classes is tested separately, so that tests of use case layer only verify interactions with mocked objects.

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlockTest.java
```diff
@@ -0,0 +1,1523 @@
...
+        final LocalDate serviceCStartDate = LocalDate.of(2020, 8, 1);
+        final LocalDate serviceCEndDate = LocalDate.of(2020, 8, 31);
+
+        when(calendarForServiceA.getStartDate()).thenReturn(serviceAStartDate);
+        when(calendarForServiceA.getEndDate()).thenReturn(serviceAEndDate);
+        when(calendarForServiceB.getStartDate()).thenReturn(serviceBStartDate);
+        when(calendarForServiceB.getEndDate()).thenReturn(serviceBEndDate);
+        when(calendarForServiceC.getStartDate()).thenReturn(serviceCStartDate);
+        when(calendarForServiceC.getEndDate()).thenReturn(serviceCEndDate);
+        when(calendarForServiceA.getOverlappingDays(calendarForServiceC))
```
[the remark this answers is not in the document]
lionel-nj: As discussed, implementation of domain layer classes is tested separately, so that tests of use case layer only verify interactions with mocked objects.

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlockTest.java
lionel-nj: As discussed, implementation of domain layer classes is tested separately, so that tests of use case layer only verify interactions with mocked objects.

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlockTest.java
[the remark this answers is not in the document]
barbeau: Ah, you're right, because no conflicting dates are detected so it doesn't need to compare times. Sounds good 👍

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlockTest.java
```diff
@@ -0,0 +1,1523 @@
...
+
+        mockTripCollection.add(firstMockTrip);
+        mockTripCollection.add(secondMockTrip);
+        mockTripCollection.add(thirdMockTrip);
+        mockTripPerBlockIdCollection.put("7", mockTripCollection);
+
+        final SortedMap<Integer, StopTime> mockFirstTripStopTimeCollection = new TreeMap<>();
+
+        final StopTime firstTripFirstStopTime = mock(StopTime.class);
+        when(firstTripFirstStopTime.getArrivalTime()).thenReturn(700);
```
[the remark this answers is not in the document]
lionel-nj: I actually chose to use class constant, having constant names super clear. The constants are defined as the true negative representation here of seconds before noon.
For example 07:00 am is stored under a constant named `SEVEN_AM_AS_SECS_BEFORE_NOON` which values as `(7-12) * HOURS_TO_SECS`; having `HOURS_TO_SECS = 3600`.

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlock.java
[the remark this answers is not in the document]
lionel-nj: See https://github.com/MobilityData/gtfs-validator/issues/362

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlockTest.java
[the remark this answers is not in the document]
barbeau: Great idea!

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/calendardates/CalendarDate.java
```diff
@@ -158,4 +159,29 @@ public CalendarDateBuilder clear() {
...
     }
+
+    /**
+     * Generates an hash code for this {@link CalendarDate} based on service_id and date fields
+     *
+     * @return an hash code for this {@link CalendarDate} based on service_id and date fields
+     */
+    @Override
+    public int hashCode() {
+        return Objects.hash(serviceId.hashCode(), date.hashCode());
```
barbeau, line 170: While this might technically work for our current purposes of enforcing uniqueness of serviceId and date combinations, after looking at this I think it's a dangerous change that could result in strange bugs later. It would be easy for a future developer to call `CalendarDate.equals()` and assume that two instances of `CalendarDate` that have the same serviceId and date but different exception type would return `false`. But with this implementation it would return `true`.

Let's stick with the existing `HashMap` impl in the GTFS repo instead of the new `Set` that relies on this for now, and we can revisit later in #360.

adapter/utils/src/test/java/org/mobilitydata/gtfsvalidator/timeutils/TimeUtilsImplTest.java
```diff
@@ -180,65 +180,80 @@ void convertIntegerToHMMSSOnNullValueShouldThrowException() {
...
+        toCheck = TIME_CONVERSION_UTILS.convertIntegerToHHMMSS(twentyEightFortySevenPm20sec);
+        assertEquals(toCheck, TIME_CONVERSION_UTILS.convertIntegerToHHMMSS(
                 TIME_CONVERSION_UTILS.convertHHMMSSToIntFromNoonOfDayOfService(toCheck)));
     }
 
+    @Test
+    void commonFirstPeriodLastTimeAndSecondPeriodFirstTimeShouldReturnFalse() {
+        //   a |--------|         07:00 am -> 10:00 am
+        //   b          |----|    10:00 am -> 11:00 am
+        // period `b` partially overlaps with period `a`
```
barbeau, line 246:
```suggestion
        // period `b` shares edge with period `a`
```

adapter/utils/src/main/java/org/mobilitydata/gtfsvalidator/timeutils/TimeUtilsImpl.java
```diff
@@ -121,7 +121,25 @@ public String convertIntegerToHMMSS(final Integer elapsedDurationSinceNoonInSeco
     }
 
     /**
-     * This method checks if two periods of times overlap
+     * This method checks if two periods of times overlap:
+     * a |--------|            07:00 am to 10:00 am     here, method returns false according to the Python Validator im-
```
barbeau, line 125: 👍

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlockTest.java
```diff
@@ -0,0 +1,1645 @@
...
+import java.time.LocalDate;
+import java.util.*;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.mockito.Mockito.*;
+
+class ValidateNoOverlappingStopTimeInTripBlockTest {
+    private static final int HOURS_TO_SECS = 3600;
+    private static final int MINS_TO_SECS = 60;
+    private static final int NOON = 12;
```
barbeau, line 43: Using constants is a very good idea, but I think defining `NOON` here as `12` is a little confusing given that we define it as `0` throughout the rest of the project. 

I'd prefer to use `TimeUnit` and `LocalTime.NOON` and the built-in addition/subtraction functions there (`.until()`, etc. like we do in `TimeUtilsImpl`) rather than our own conversation factors.

barbeau (commented):
Thanks @lionel-nj, this is coming along nicely. A few comments below.

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlockTest.java
```diff
@@ -32,31 +32,46 @@
...
-    private static final int TEN_AM_20_MIN_AS_SECS_BEFORE_NOON = -(HOURS_TO_SECS + 40 * MINS_TO_SECS);
-    private static final int TEN_AM_30_MIN_AS_SECS_BEFORE_NOON = -(HOURS_TO_SECS + 30 * MINS_TO_SECS);
-    private static final int ELEVEN_AM_03_MIN_AS_SECS_BEFORE_NOON = -(57 * MINS_TO_SECS);
-    private static final int ELEVEN_AM_15_MIN_AS_SECS_BEFORE_NOON = -(45 * MINS_TO_SECS);
-    private static final int ELEVEN_AM_16_MIN_AS_SECS_BEFORE_NOON = -(44 * MINS_TO_SECS);
-    private static final int ELEVEN_AM_20_MIN_AS_SECS_BEFORE_NOON = -(40 * MINS_TO_SECS);
-    private static final int ELEVEN_AM_40_MIN_AS_SECS_BEFORE_NOON = -(20 * MINS_TO_SECS);
+    private static final LocalTime NOON = LocalTime.NOON;
+
+    private static final int FIVE_AM_AS_SECS_BEFORE_NOON =
```
barbeau, line 44: 💯

barbeau (approved):
💯! Great work @lionel-nj, LGTM! 🚀

PR #348, opened 2020-08-21
Fix #347 - Time conversion of AM times 

### LINKED ISSUE #347: [BUG] TimeUtils conversion to/from integer is wrong
**Describe the bug**
Wrong time conversion

**To Reproduce**
`timeUtils.convertIntegerToHMMSS(timeUtils.convertHHMMSSToIntFromNoonOfDayOfService("06:01:00"))`

**Expected behavior**
this should return `06:01:00` 

**Witnessed behavior**
Returns `06:60:00` instead

**Screenshots**

**Environment used**
- OS: macOS Catalina
- Java version: 11.0.5
- gtfs-validator version: 1.2.1

**Additional context**
https://github.com/MobilityData/gtfs-validator/pull/346

lionel-nj:
**Summary:**

This PR provides support to solve to avoid wrong conversion of AM time
**Expected behavior:** 

`timeUtils.convertIntegerToHMMSS(timeUtils.convertHHMMSSToIntFromNoonOfDayOfService("06:01:00"))`
returns `06:01:00` instead of `06:60:00`

### DIFF
--- a/adapter/utils/src/main/java/org/mobilitydata/gtfsvalidator/timeutils/TimeUtilsImpl.java
+++ b/adapter/utils/src/main/java/org/mobilitydata/gtfsvalidator/timeutils/TimeUtilsImpl.java
@@ -89,29 +89,32 @@ public class TimeUtilsImpl implements TimeUtils {
     public String convertIntegerToHMMSS(final Integer elapsedDurationSinceNoon) {
         if (elapsedDurationSinceNoon != null) {
             // Determine number of hours elapsed
-            int hourValue = (int) (TimeUnit.SECONDS.toHours(elapsedDurationSinceNoon) + 12);
+            int hourValue;
             int minuteValue;
             int secondValue;
             // elapsedDurationSinceNoon > 0 for times after noon, e.g 14:00PM is represented a +2*3600s=+7200s
             // elapsedDurationSinceNoon < 0 for times before noon, e.g 11:00AM is represented as -1*3600s=-3600s
-            if (elapsedDurationSinceNoon > 0) {
+            if (elapsedDurationSinceNoon >= 0) {
+                hourValue = (int) (TimeUnit.SECONDS.toHours(elapsedDurationSinceNoon) + 12);
                 // Determine number of minutes and seconds elapsed
                 minuteValue = (int) (TimeUnit.SECONDS.toMinutes(elapsedDurationSinceNoon) -
                         TimeUnit.HOURS.toMinutes(TimeUnit.SECONDS.toHours(elapsedDurationSinceNoon)));
                 secondValue = (int) (TimeUnit.SECONDS.toSeconds(elapsedDurationSinceNoon) -
                         TimeUnit.MINUTES.toSeconds(TimeUnit.SECONDS.toMinutes(elapsedDurationSinceNoon)));
             } else {
+                hourValue = (int) (TimeUnit.SECONDS.toHours(elapsedDurationSinceNoon + 12 * 3600));
                 // Determine number of minutes elapsed minus the number of hours elapsed
                 minuteValue = ((int) (TimeUnit.SECONDS.toMinutes(elapsedDurationSinceNoon) -
                         TimeUnit.HOURS.toMinutes(TimeUnit.SECONDS.toHours(elapsedDurationSinceNoon)))) % 60;
                 if (minuteValue < 0) {
-                    hourValue = hourValue - 1;
-                    minuteValue = 1 + ((int) -(TimeUnit.SECONDS.toMinutes(elapsedDurationSinceNoon) -
-                            TimeUnit.HOURS.toMinutes(TimeUnit.SECONDS.toHours(elapsedDurationSinceNoon))));
+                    minuteValue = 60 + minuteValue;
                 }
                 // Determine number of seconds elapsed minus the number of hours and minutes elapsed
                 secondValue = (int) (TimeUnit.SECONDS.toSeconds(elapsedDurationSinceNoon) -
                         TimeUnit.MINUTES.toSeconds(TimeUnit.SECONDS.toMinutes(elapsedDurationSinceNoon)));
+                if (secondValue < 0) {
+                    minuteValue = minuteValue - 1;
+                }
                 if ((int) (TimeUnit.SECONDS.toSeconds(elapsedDurationSinceNoon) -
                         TimeUnit.MINUTES.toSeconds(TimeUnit.SECONDS.toMinutes(elapsedDurationSinceNoon))) < 0) {
                     secondValue = 60 + secondValue;
--- a/adapter/utils/src/main/java/org/mobilitydata/gtfsvalidator/timeutils/TimeUtilsImpl.java
+++ b/adapter/utils/src/main/java/org/mobilitydata/gtfsvalidator/timeutils/TimeUtilsImpl.java
@@ -19,6 +19,7 @@ package org.mobilitydata.gtfsvalidator.timeutils;
 import org.mobilitydata.gtfsvalidator.usecase.utils.TimeUtils;
 
 import java.time.LocalTime;
+import java.time.temporal.ChronoUnit;
 import java.util.concurrent.TimeUnit;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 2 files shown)


lionel-nj:
Thanks for the review, I'll make the changes.

adapter/utils/src/main/java/org/mobilitydata/gtfsvalidator/timeutils/TimeUtilsImpl.java
```diff
@@ -88,36 +90,33 @@ public Integer convertHHMMSSToIntFromNoonOfDayOfService(final String timeAsStrin
...
             // elapsedDurationSinceNoon > 0 for times after noon, e.g 14:00PM is represented a +2*3600s=+7200s
             // elapsedDurationSinceNoon < 0 for times before noon, e.g 11:00AM is represented as -1*3600s=-3600s
-            if (elapsedDurationSinceNoon > 0) {
-                // Determine number of minutes and seconds elapsed
-                minuteValue = (int) (TimeUnit.SECONDS.toMinutes(elapsedDurationSinceNoon) -
-                        TimeUnit.HOURS.toMinutes(TimeUnit.SECONDS.toHours(elapsedDurationSinceNoon)));
-                secondValue = (int) (TimeUnit.SECONDS.toSeconds(elapsedDurationSinceNoon) -
-                        TimeUnit.MINUTES.toSeconds(TimeUnit.SECONDS.toMinutes(elapsedDurationSinceNoon)));
+//            LocalTime toReturn = elapsedDurationSinceNoon < 24*3600 ?
+            LocalTime toReturn = noon.plusHours(TimeUnit.SECONDS.toHours(elapsedDurationSinceNoon));
```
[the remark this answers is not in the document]
lionel-nj: That did the tric, reduced the size of the code, and improved its clarity at the same time. Thanks :)

adapter/utils/src/main/java/org/mobilitydata/gtfsvalidator/timeutils/TimeUtilsImpl.java
```diff
@@ -81,43 +84,36 @@ public Integer convertHHMMSSToIntFromNoonOfDayOfService(final String timeAsStrin
...
-            // elapsedDurationSinceNoon > 0 for times after noon, e.g 14:00PM is represented a +2*3600s=+7200s
-            // elapsedDurationSinceNoon < 0 for times before noon, e.g 11:00AM is represented as -1*3600s=-3600s
-            if (elapsedDurationSinceNoon > 0) {
-                // Determine number of minutes and seconds elapsed
-                minuteValue = (int) (TimeUnit.SECONDS.toMinutes(elapsedDurationSinceNoon) -
-                        TimeUnit.HOURS.toMinutes(TimeUnit.SECONDS.toHours(elapsedDurationSinceNoon)));
-                secondValue = (int) (TimeUnit.SECONDS.toSeconds(elapsedDurationSinceNoon) -
-                        TimeUnit.MINUTES.toSeconds(TimeUnit.SECONDS.toMinutes(elapsedDurationSinceNoon)));
+    public String convertIntegerToHMMSS(final Integer elapsedDurationSinceNoonInSeconds) {
+        if (elapsedDurationSinceNoonInSeconds != null) {
```
barbeau, line 93: Instead of wrapping this entire method in the `if`, I'd use an early return and throw an IllegalArgumentException instead (assuming that we shouldn't be getting `null` values at this point):

```java
if (elapsedDurationSinceNoon == null) {
 throw new IllegalArgumentException("elapsedDurationSinceNoonInSeconds cannot be null");
}
```

If we have to guard against `null` values here, then still use the early return, but return `null` instead and update the Javadocs to say it returns `null` if the input is `null`.

adapter/utils/src/main/java/org/mobilitydata/gtfsvalidator/timeutils/TimeUtilsImpl.java
```diff
@@ -81,43 +84,36 @@ public Integer convertHHMMSSToIntFromNoonOfDayOfService(final String timeAsStrin
...
+            final LocalTime noon = LocalTime.NOON;
+            // elapsedDurationSinceNoonInSeconds > 0 for times after noon, e.g 14:00PM is represented a +2*3600s=+7200s
+            // elapsedDurationSinceNoonInSeconds < 0 for times before noon, e.g 11:00AM is represented as -1*3600s=-3600s
+            final boolean isTimeGreaterThan24hours =
+                    elapsedDurationSinceNoonInSeconds >= TimeUnit.HOURS.toSeconds(12);
+
+            if (!isTimeGreaterThan24hours) {
+                return String.format("%02d:%02d:%02d", noon.plusSeconds(elapsedDurationSinceNoonInSeconds).getHour(),
```
barbeau, line 101: Just curious - did `return noon.plusSeconds(elapsedDurationSinceNoon).toString();` not work here?

adapter/utils/src/main/java/org/mobilitydata/gtfsvalidator/timeutils/TimeUtilsImpl.java
```diff
@@ -81,43 +84,36 @@ public Integer convertHHMMSSToIntFromNoonOfDayOfService(final String timeAsStrin
...
-                }
+                // This wraps around midnight
+                LocalTime timeAfterMidnight = noon.plusSeconds(elapsedDurationSinceNoonInSeconds);
+                // Get the number of hours, minutes, and seconds between midnight and timeAfterMidnight, which is the
+                // time to add to 24:00:00
+                long hoursAfterMidnight = LocalTime.MIDNIGHT.until(timeAfterMidnight, ChronoUnit.HOURS);
+                long minutesAfterMidnight = LocalTime.MIDNIGHT.until(timeAfterMidnight, ChronoUnit.MINUTES) % 60;
+                long secondsAfterMidnight = LocalTime.MIDNIGHT.until(timeAfterMidnight, ChronoUnit.SECONDS) % 60;
+                return String.format("%02d:%02d:%02d",
+                        hoursAfterMidnight + 24,
```
barbeau, line 113: In retrospect, this is probably clearer as:

```suggestion
                        24 + hoursAfterMidnight,
```

adapter/utils/src/main/java/org/mobilitydata/gtfsvalidator/timeutils/TimeUtilsImpl.java
lionel-nj: The String returned by `toString` depends on whether the time has seconds. If the time has seconds we end up with a String formatted as HHMMSS, but if it does not we end up with a String formatted as HHMM. As we want to maintain the HHMMSS format in any case, I had to make this change.

barbeau: Ah, thanks, makes sense 👍

adapter/utils/src/main/java/org/mobilitydata/gtfsvalidator/timeutils/TimeUtilsImpl.java
barbeau: I just updated my above comment - if you do throw an `IllegalArgumentException`, make sure you add it the method signature so calling code knows it might throw.

adapter/utils/src/main/java/org/mobilitydata/gtfsvalidator/timeutils/TimeUtilsImpl.java
```diff
@@ -81,45 +84,39 @@ public Integer convertHHMMSSToIntFromNoonOfDayOfService(final String timeAsStrin
...
-                // Determine number of minutes and seconds elapsed
-                minuteValue = (int) (TimeUnit.SECONDS.toMinutes(elapsedDurationSinceNoon) -
-                        TimeUnit.HOURS.toMinutes(TimeUnit.SECONDS.toHours(elapsedDurationSinceNoon)));
-                secondValue = (int) (TimeUnit.SECONDS.toSeconds(elapsedDurationSinceNoon) -
-                        TimeUnit.MINUTES.toSeconds(TimeUnit.SECONDS.toMinutes(elapsedDurationSinceNoon)));
+    public String convertIntegerToHMMSS(final Integer elapsedDurationSinceNoonInSeconds) throws
+            IllegalArgumentException {
+        if (elapsedDurationSinceNoonInSeconds == null) {
+            throw new IllegalArgumentException("elapsedDurationSinceNoonInSeconds cannot be null");
+        } else {
```
barbeau, line 96: For early returns like this, you typically don't wrap the rest of the method in an `else`.  It would just be:

```java
if (elapsedDurationSinceNoonInSeconds == null) {
    throw new IllegalArgumentException("elapsedDurationSinceNoonInSeconds cannot be null");
}
```

...followed by the rest of the code. This way the rest of the code is simpler to trace, doesn't indent as much, and indicates that you can't execute the remainder of the body if `null`.

lionel-nj: Oh okay! I understand, thanks.

PR #349, opened 2020-08-21
Implement#146 - Verify trip frequencies do not overlap

lionel-nj:
**Summary:**

This PR provides support to verify that time periods provided in `frequencies.txt` do not overlap.

This PR also provides support for a new use of the time util class: check that two time periods overlap.

**Expected behavior:** 

Overlapping time periods should generate a notice
Capture d’écran, le 2020-08-21 à 18 18 02

-  [x] Run the unit tests with `gradle test` to make sure you didn't break anything

### DIFF
--- a/adapter/utils/src/test/java/org/mobilitydata/gtfsvalidator/timeutils/TimeUtilsImplTest.java
+++ b/adapter/utils/src/test/java/org/mobilitydata/gtfsvalidator/timeutils/TimeUtilsImplTest.java
@@ -240,17 +240,68 @@ public class TimeUtilsImplTest {
     }
 
     @Test
-    void overlappingPeriodsShouldReturnTrue() {
+    void partialOverlapWithOneCommonBoundaryShouldReturnTrue() {
+        //   a |--------|         07:00 am -> 10:00 am
+        //   b     |----|         09:00 am -> 10:00 am
+        // period `b` partially overlaps with period `a`
         assertTrue(TIME_CONVERSION_UTILS.arePeriodsOverlapping(700, 1000,
-                800, 1200));
+                900, 1000));
+
+        // permute periods `a` and `b` in method call
+        assertTrue(TIME_CONVERSION_UTILS.arePeriodsOverlapping(900, 1000,
+                700, 1000));
+
+        //   a |--------|         07:00 am -> 10:00 am
+        //   b |----|             07:00 am -> 09:00 am
+        // period `b` partially overlaps with period `a`
+        assertTrue(TIME_CONVERSION_UTILS.arePeriodsOverlapping(700, 1000,
+                700, 900));
+
+        // permute periods `a` and `b` in method call
+        assertTrue(TIME_CONVERSION_UTILS.arePeriodsOverlapping(700, 900,
+                700, 1000));
     }
 
     @Test
-    void nonOverlappingPeriodsShouldReturnFalse() {
+    void partialOverlapPeriodsShouldReturnTrue() {
+        //   a |--------|       07:00 am -> 10:00 am
+        //     b      |----|    08:00 am -> 12:00 am
+        // period `b` partially overlaps with period `a`
+        assertTrue(TIME_CONVERSION_UTILS.arePeriodsOverlapping(700, 1000,
+                800, 1200));
+
+        // permute periods `a` and `b` in method call
+        assertTrue(TIME_CONVERSION_UTILS.arePeriodsOverlapping(800, 1200,
+                700, 1000));
+    }
+
+    @Test
+    void firstPeriodContainedWithTheSecondOneShouldReturnTrueWhenArePeriodsOverlappingIsCalled() {
+        //   a |--------|       07:00 am -> 12:00 am
+        //   b   |----|         08:00 am -> 11:00 am
+        // period `b` is contained within period `a`
+        assertTrue(TIME_CONVERSION_UTILS.arePeriodsOverlapping(700, 1200,
+                800, 1100));
+    }
+
+    @Test
+    void secondPeriodContainedWithinTheFirstOneShouldReturnTrueWhenArePeriodsOverlappingIsCalled() {
+        //   a   |----|         08:00 am -> 11:00 am
+        //   b |--------|       07:00 am -> 12:00 am
+        // period `a` is contained within period `b`
+        assertTrue(TIME_CONVERSION_UTILS.arePeriodsOverlapping(800, 1100,
+                700, 1200));
+    }
+
+    @Test
+    void disjointPeriodsShouldReturnFalse() {
+        // Disjoint periods of time, with `a` being the first period and `b` the second one
+        // a |--------|                 07:00 am -> 10:00 am
+        //             b  |----|        11:00 am -> 12:00 am
         assertFalse(TIME_CONVERSION_UTILS.arePeriodsOverlapping(700, 1000,
                 1100, 1200));
-        // common end time and start time
-        assertFalse(TIME_CONVERSION_UTILS.arePeriodsOverlapping(700, 1000,
-                1000, 1200));
+        // permute periods a and b in method call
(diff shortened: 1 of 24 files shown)


usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateTripFrequenciesOverlap.java
```diff
@@ -0,0 +1,77 @@
...
+        this.resultRepo = resultRepo;
+        this.logger = logger;
+        this.timeUtils = timeUtils;
+    }
+
+    public void execute() {
+        logger.info("Validating rule 'E053 - Trip frequencies overlap'");
+
+        dataRepo.getFrequencyAllByTripId().forEach((tripId, frequencyCollection) -> {
+            final List<String> visitedFrequencyTripIdStartTimeCollection = new ArrayList<>();
```
barbeau, line 48: This should be a `HashMap` instead of a `ArrayList` to change search from O(n) to O(1) time.

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateTripFrequenciesOverlap.java
```diff
@@ -0,0 +1,77 @@
...
+
+            frequencyCollection.forEach(currentFrequency -> {
+                visitedFrequencyTripIdStartTimeCollection.add(currentFrequency.getFrequencyMappingKey());
+                final int currentFrequencyStartTime = currentFrequency.getStartTime();
+                final int currentFrequencyEndTime = currentFrequency.getEndTime();
+                if (currentFrequencyStartTime < currentFrequencyEndTime) {
```
barbeau, line 54: I believe both of these checks for if the start time is before the end time for the same frequency are just sanity checks to make sure the frequency time itself is valid, right? If so, I'd move this implementation into a private method like `isFrequencyValid()` so it's clearer what you're testing.

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateTripFrequenciesOverlapTest.java
```diff
@@ -0,0 +1,285 @@
...
+
+        final GtfsDataRepository mockDataRepo = mock(GtfsDataRepository.class);
+        final ValidationResultRepository mockResultRepo = mock(ValidationResultRepository.class);
+        final Logger mockLogger = mock(Logger.class);
+        final TimeUtils mockTimeUtil = mock(TimeUtils.class);
+        when(mockTimeUtil.convertIntegerToHMMSS(1200)).thenReturn("1200");
+        when(mockTimeUtil.convertIntegerToHMMSS(1800)).thenReturn("1800");
+        when(mockTimeUtil.convertIntegerToHMMSS(1544)).thenReturn("1544");
+        when(mockTimeUtil.convertIntegerToHMMSS(2143)).thenReturn("2143");
+        when(mockTimeUtil.arePeriodsOverlapping(1200, 1800,
```
barbeau, line 58: I think mocking the implementation of the time utility class here is potentially dangerous, in that it is potentially giving the developer false confidence the permutations of the periods is passing when you don't in fact know if that is the case (because the actual implementation that determines this isn't a dependency in this module, but in the adapter module).

I'd prefer to see all of these tests be in the time utility implementation module, and limit the test for the use case to determine that each combination of frequencies are tested (e.g., the number of calls is correct).

barbeau:
Oh, I also forgot to mention that technically to be correct we also need to determine if the service for both frequencies is running on the same calendar day (from calendar.txt and calendar_dates.txt). This makes the rule much harder to implement, though. Did the Python validator check calendar for this rule?

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateTripFrequenciesOverlap.java
lionel-nj: Indeed, these checks are sanity checks.

lionel-nj:
@barbeau it does not seem that the Python validator checks for day of service via `calendar.txt` and `calendar_dates.txt`

Code for frequency validation: https://github.com/MobilityData/transitfeed/blob/54a4081b59bfa015d5f0405b68203e61762d4a52/transitfeed/trip.py#L686

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateTripFrequenciesOverlapTest.java
barbeau: Following our discussion, I do see value in testing to make sure the method returns the proper notices based on the values the utility method returns. So I would say let's leave the code, but add explicit comments/Javadocs saying that the testing of the time check logic happens in the utility method test.

lionel-nj:
Verification of day of service will be done in another PR. See open issue: https://github.com/MobilityData/gtfs-validator/issues/359

adapter/utils/src/test/java/org/mobilitydata/gtfsvalidator/timeutils/TimeUtilsImplTest.java
```diff
@@ -240,17 +240,68 @@ void convertHHMMSStoIntegerAndIntegerToHHMMSSShouldBeConsistent() {
...
+    void secondPeriodContainedWithinTheFirstOneShouldReturnTrueWhenArePeriodsOverlappingIsCalled() {
+        //   a   |----|         08:00 am -> 11:00 am
+        //   b |--------|       07:00 am -> 12:00 am
+        // period `a` is contained within period `b`
+        assertTrue(TIME_CONVERSION_UTILS.arePeriodsOverlapping(800, 1100,
+                700, 1200));
+    }
+
+    @Test
+    void disjointPeriodsShouldReturnFalse() {
```
barbeau, line 297: I think you're still missing the opposite of this test case, where they are disjoint but `b` is before `a`?

adapter/utils/src/test/java/org/mobilitydata/gtfsvalidator/timeutils/TimeUtilsImplTest.java
```diff
@@ -240,17 +240,68 @@ void convertHHMMSStoIntegerAndIntegerToHHMMSSShouldBeConsistent() {
...
+        assertTrue(TIME_CONVERSION_UTILS.arePeriodsOverlapping(700, 1000,
+                700, 900));
+
+        // permute periods `a` and `b` in method call
+        assertTrue(TIME_CONVERSION_UTILS.arePeriodsOverlapping(700, 900,
+                700, 1000));
+    }
+
+    @Test
+    void partialOverlapPeriodsShouldReturnTrue() {
```
barbeau, line 266: I think you're missing the opposite of this test?

```
b |--------| 
      a  |----|
```

adapter/utils/src/test/java/org/mobilitydata/gtfsvalidator/timeutils/TimeUtilsImplTest.java
barbeau: Oh, nevermind, I see it below. Could you add another comment visual above that test to explain?

adapter/utils/src/test/java/org/mobilitydata/gtfsvalidator/timeutils/TimeUtilsImplTest.java
barbeau: Oh, nevermind, I see it below. Could you add another comment visual above that test to explain?

barbeau: Same for the other methods where you swap numbers for the permutation at the end.

Issue #212, opened 2020-05-11
[REQUEST] Implement stop times time and distance verification for a trip


maximearmstrong:
**Is your feature request related to a problem? Please describe.**
Stop times in a trip should be have increasing time and distance. This is a GTFS rule implemented in Google Python validator as ValidateStopTimesSequenceHasIncreasingTimeAndDistance.

**Describe the solution you'd like**
Actual Google GTFS validator behaviour : verifies the time and distance travelled at a stop time in a trip is greater than the previous ones. (verifies the stop time and the distance travelled according to the shape)
https://github.com/google/transitfeed/blob/d727e97cb66ac2ca2d699a382ea1d449ee26c2a1/transitfeed/trip.py#L595

**Describe alternatives you've considered**

**Additional context**
Line 203 in Error support priorities https://docs.google.com/spreadsheets/d/1vqe6wq7ctqk1EhYkgtZ0_TbcQ91vccfs2daSjn20BLE/edit#gid=0

barbeau:
cc @timMillet 

Note that the Python validator allows sequential arrival and departure times in a trip to be the same - so you can have:

~~~
1, 12:00:00, 12:00:00
2, 12:10:00, 12:10:00
3, 12:10:00, 12:10:00
4, 12:15:00, 12:15:00
~~~

The Python validator only flags times that go backwards in time. This caught us in the GTFS-RT validator, as we initially implemented a rule that said that predictions needed to move forward in time for each stop (see https://github.com/CUTR-at-USF/gtfs-realtime-validator/issues/365).

Allowing sequential times being the same dates back to a major scheduling software package called HASTUS rounding times to the nearest minute:
https://github.com/google/transitfeed/blob/d727e97cb66ac2ca2d699a382ea1d449ee26c2a1/transitfeed/trip.py#L733

I don't know for sure if HASTUS still has this issue, although someone flagged that issue with our GTFS-RT validator, so I assume it (or another GTFS exporter) does.

It would be nice to update the validator to be more strict rules for this if the state of the industry allows. If not, we may have to implement some of the same heuristics in the Python validator:

>     # HASTUS makes it hard to output GTFS with times to the nearest second
>     # it rounds times to the nearest minute. Therefore stop_times at the
>     # same time ending in :00 are fairly common. These times off by no more
>     # than 30 have not caused a problem. See
>     # https://github.com/google/transitfeed/issues/193
>     # Show a warning if times are not rounded to the nearest minute or
[...]

PR #350, opened 2020-08-25
Implement #212 - Increasing distance stop times verification

lionel-nj:
**Summary:**

This PR provides support to verify that for each stop time of a trip have increasing `shape_dist_travelled` valued

**Expected behavior:** 

The Python validator code treats this rule as follows:
- [x] two consecutive stops times have same value for `shape_dist_travelled`-> this is an error
- [x] two consecutive stop times have values of `shape_dist_travelled` not ordered by ascending order -> this is an error

Code fro the Python valdiator https://github.com/MobilityData/transitfeed/blob/54a4081b59bfa015d5f0405b68203e61762d4a52/transitfeed/trip.py#L594

Tests from the python validator

https://github.com/MobilityData/transitfeed/blob/54a4081b59bfa015d5f0405b68203e61762d4a52/tests/transitfeed/testtrip.py#L357

Capture d’écran, le 2020-08-25 à 11 29 47


adapter/exporter/src/main/java/org/mobilitydata/gtfsvalidator/exporter/ProtobufNoticeExporter.java
```diff
@@ -854,6 +855,36 @@ public void export(final UnusableTripNotice toExport) throws IOException {
...
+    }
+
+    @Override
+    public void export(final DecreasingStopTimeDistanceWarningNotice toExport) throws IOException {
+        decreasingStopTimeDistanceToProto(GtfsValidationOutputProto.GtfsProblem.Severity.WARNING, toExport);
+    }
+
+    private void decreasingStopTimeDistanceToProto(final GtfsValidationOutputProto.GtfsProblem.Severity severity,
+                                                   final Notice toExport)
+            throws IOException {
```
barbeau, line 871: Could be moved to above line

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/notice/error/DecreasingStopTimeDistanceErrorNotice.java
```diff
@@ -0,0 +1,50 @@
...
+
+public class DecreasingStopTimeDistanceErrorNotice extends ErrorNotice {
+    public DecreasingStopTimeDistanceErrorNotice(final String tripId,
+                                                 final int stopSequence,
+                                                 final Float shapeDistTraveled,
+                                                 final int conflictingStopSequence,
+                                                 final Float conflictingShapeDistTraveled) {
+        super("stop_times.txt", E_054,
+                "Decreasing `shape_dist_traveled` values",
+                String.format("Trip with `trip_id` `%s` the stop `%s` has shape_dist_traveled=`%s`, which should be" +
```
barbeau, line 32: I'd suggest a more succinct message that includes the values close the front of the message, so you don't have to scroll to see them, as well as the exact field names:

>`trip_id` 1234 `stop_sequence` 1 has a larger `shape_dist_traveled` (15.21) than `stop_sequence` 2 (12.35). `shape_dist_traveled` must increase with `stop_sequence`.

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateStopTimeIncreasingDistance.java
```diff
@@ -0,0 +1,81 @@
...
+    public void execute(final TreeMap<Integer, StopTime> stopTimeSequence) {
+        // get previous stoptime relevant information
+        final var previousStopTimeData = new Object() {
+            Float shapeDistTraveled = null;
+            Integer stopSequence = null;
+        };
+        stopTimeSequence.forEach((stopSequence, stopTime) -> {
+            final Float shapeDistTraveled = stopTime.getShapeDistTraveled();
+            if (shapeDistTraveled != null && previousStopTimeData.shapeDistTraveled != null) {
+                if (previousStopTimeData.shapeDistTraveled > shapeDistTraveled) {
```
barbeau, line 55: See my above comment - I believe this should be `>=`

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/notice/error/DecreasingStopTimeDistanceErrorNotice.java
```diff
@@ -0,0 +1,50 @@
...
+                                " larger than the previous ones. In this case, the previous distance was `%s` (for the " +
+                                "stop `%s`).",
+                        tripId, conflictingStopSequence, conflictingShapeDistTraveled, shapeDistTraveled, stopSequence),
```
barbeau, line 35: I would try to keep the order of variables here and in the constructor the same - it could be easy to accidentally swap them.

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/notice/error/DecreasingStopTimeDistanceErrorNotice.java
```diff
@@ -0,0 +1,50 @@
...
+package org.mobilitydata.gtfsvalidator.domain.entity.notice.error;
+
+import org.mobilitydata.gtfsvalidator.domain.entity.notice.NoticeExporter;
+import org.mobilitydata.gtfsvalidator.domain.entity.notice.base.ErrorNotice;
+
+import java.io.IOException;
+
+public class DecreasingStopTimeDistanceErrorNotice extends ErrorNotice {
+    public DecreasingStopTimeDistanceErrorNotice(final String tripId,
+                                                 final int stopSequence,
```
barbeau, line 26: I would rename these variables to define the order of sequence, like:
* `stopSequence`
* `shapeDistTraveled`
* `previousStopSequence`
* `previousShapeDistTraveled`

Per my comment below, without explicit order in the names it would be easy to accidentally swap the variables when calling the method, unless you looked at the implementation.

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateStopTimeIncreasingDistanceTest.java
```diff
@@ -0,0 +1,321 @@
...
+import java.util.TreeMap;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.mobilitydata.gtfsvalidator.domain.entity.notice.base.Notice.*;
+import static org.mockito.Mockito.*;
+
+class ValidateStopTimeIncreasingDistanceTest {
+
+    // suppressed warning regarding ignored result of method used in assertions
+    @SuppressWarnings("ResultOfMethodCallIgnored")
```
barbeau, line 35: I would add one more test where the first stop sequence shape_dist_traveled is 10 and the second 5, just to cover all corner cases (i.e., make sure a large value in the first sequence index is detected).

adapter/exporter/src/main/java/org/mobilitydata/gtfsvalidator/exporter/ProtobufNoticeExporter.java
lionel-nj: Since the warning notice has been removed, this method is no more used.

RULES.md
```diff
@@ -307,6 +309,12 @@ Trips must be referred to at least once in `stop_times.txt`.
 
 A trip must visit more than one stop in `stop_times.txt` to be usable by passengers for boarding and alighting.
 
+<a name="E054"/>
+
+### E054 - Decreasing travelled distance
```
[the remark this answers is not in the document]
lionel-nj: Will implement this rule in another PR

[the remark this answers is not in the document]
lionel-nj: See https://github.com/MobilityData/gtfs-validator/issues/358

Issue #149, opened 2020-05-01
[REQUEST] Implement unique route names verification (GTFS rule)


maximearmstrong:
**Is your feature request related to a problem? Please describe.**
Route short and long names should be unique, and their combination should be unique. This is a GTFS rule implemented in Google Python validator and featured in Google Type Error as TYPE_ROUTE_NAME_REUSED.

**Describe the solution you'd like**
Actual Google GTFS validator behaviour : verifies if route_short_name and route_long_name combination is not used multiple times.

**Describe alternatives you've considered**

**Additional context**
Line 76 in Error support priorities https://docs.google.com/spreadsheets/d/1vqe6wq7ctqk1EhYkgtZ0_TbcQ91vccfs2daSjn20BLE/edit#gid=0

barbeau:
To help differentiate this vs https://github.com/MobilityData/gtfs-validator/issues/150 - my understanding is that this issue is comparing across multiple records from the routes.txt file, while https://github.com/MobilityData/gtfs-validator/issues/150 is looking at route short and long name within the same record.

maximearmstrong:
Exactly, here it is the combination that should be unique, i.e. a route short name can't be associated to 2 different route long names

timMillet:
It happens that different routes have the same `routes.route_short_name` values:
- if they are not part of the same agency (e.g: feed IDFM, Paris FR)
- they are not part of the same "service brand" like regular bus routes, school bus routes (feed: Auckland Transport, Auckland NZ)

It happens that different routes have the same `routes.route_long_name` values:
- When routes do not have actual long names, but the service brand like Express or Commuter is included within the `routes.route_long_name` (feed: some US agencies), as the Best Practices say.

So the validator should issue a warning but not an error. The warning message must explicit that if routes have the same route_short_name or route_long_name, they **must** be different routes serving different areas; and **must not** be different trips of the same route or different directions of the same route.

PR #351, opened 2020-08-25
Implement #149 - Verify route long name and route short name unicity

lionel-nj:
**Summary:**

This PR provides support to verify that all `route_long_name` and `route_short_name` used are unique, under certain conditions described in the linked issue.

**Expected behavior:** 

- [x] `route_long_name` should be unique, if not this is a warning (see https://github.com/MobilityData/gtfs-validator/issues/149#issuecomment-634829017) 
  - [x] rule logic
  - [x] unit tests
  - [x] documentation 

Capture d’écran, le 2020-08-25 à 17 13 13

- [x] `route_short_name` should be unique, if not this is a warning (see https://github.com/MobilityData/gtfs-validator/issues/149#issuecomment-634829017)
  - [x] rule logic
  - [x] unit tests
  - [x] documentation 

- [x] `route_long_name` + `route_short_name` should be unique regarding the Python validator rules (https://github.com/MobilityData/transitfeed/blob/54a4081b59bfa015d5f0405b68203e61762d4a52/transitfeed/schedule.py#L1070)
  - [x] rule logic (to be committed later)
  - [x] unit tests
  - [x] documentation 

 
Capture d’écran, le 2020-08-26 à 11 22 08


RULES.md
```diff
@@ -361,4 +364,22 @@ If possible, the GTFS dataset should cover at least the next 30 days of service
...
 Empty csv optional file found in the archive: file contains header but does not have data.  
-This is related to [E047](https://github.com/MobilityData/gtfs-validator/blob/master/RULES.md#E047).
\ No newline at end of file
+This is related to [E047](https://github.com/MobilityData/gtfs-validator/blob/master/RULES.md#E047).
+
+<a name="W014"/>
+
+### W014 - Duplicate `routes.route_long_name`
+
+All routes should have different `route_long_name`. If routes have same `route_long_name`, they must be different routes serving different areas; and must not be different trips of the same route or different directions of the same route.
```
barbeau, line 373:
```suggestion
All routes should have different `route_long_name`. If routes have the same `route_long_name`, they must be different routes serving different areas; and must not be different trips of the same route or different directions of the same route.
```

RULES.md
```diff
@@ -361,4 +364,22 @@ If possible, the GTFS dataset should cover at least the next 30 days of service
...
+
+<a name="W015"/>
+
+### W015 - Duplicate `routes.route_short_name`
+
+All routes should have different `route_short_name`. If routes have same `route_long_name`, they must be different routes serving different areas; and must not be different trips of the same route or different directions of the same route.
```
barbeau, line 379:
```suggestion
All routes should have different `route_short_name`. If routes have the same `route_short_name`, they must be different routes serving different areas; and must not be different trips of the same route or different directions of the same route.
```

RULES.md
```diff
@@ -361,4 +364,22 @@ If possible, the GTFS dataset should cover at least the next 30 days of service
...
+
+<a name="W016"/>
+
+### W016 - Duplicate combination of fields `route_long_name` and `routes.route_short_name`
+
+`route_short_name` and `route_long_name` 'should not be used for more than once.
```
barbeau, line 385:
```suggestion
The same combination of `route_short_name` and `route_long_name` should not be used for more than one route.
```

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/notice/warning/DuplicateRouteShortNameNotice.java
```diff
@@ -0,0 +1,44 @@
...
+
+import java.io.IOException;
+
+public class DuplicateRouteShortNameNotice extends WarningNotice {
+    public DuplicateRouteShortNameNotice(final String routeId,
+                                         final String conflictingRouteId,
+                                         final String duplicateRouteShortName) {
+        super("routes.txt",
+                W_015,
+                "Duplicate `route_long_name`",
```
barbeau, line 30:
```suggestion
                "Duplicate `route_short_name`",
```

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateRouteShortNameAreUnique.java
```diff
@@ -0,0 +1,84 @@
...
+        this.dataRepo = dataRepo;
+        this.resultRepo = resultRepo;
+        this.logger = logger;
+    }
+
+    /**
+     * Use case execution method: checks unicity of `route_short_name` in file `routes.txt`. A notice is
+     * generated and added to the {@code ValidationResultRepository} provided in the constructor each time this
+     * requirement is not satisfied.
+     * Note that two {@link Route} can have the same `route_short_name` if they do not belong to the same agency
```
barbeau, line 55: This should be added to the rule documentation markdown too as well as the notice output.

barbeau: Actually, this should apply to `route_long_name` too, right? This logic regarding same agency isn't in that rule.

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateRouteShortNameAreUnique.java
```diff
@@ -0,0 +1,84 @@
...
+
+        final Map<String, Route> routeByRouteShortName = new HashMap<>();
+        dataRepo.getRouteAll().forEach((routeId, route) -> {
+            final String routeShortName = route.getRouteShortName();
+            if (routeShortName != null) {
+                if (routeByRouteShortName.containsKey(routeShortName)) {
+                    final String routeAgencyId = route.getAgencyId();
+                    final String otherRouteAgencyId = routeByRouteShortName.get(routeShortName).getAgencyId();
+                    final boolean routesAreFromSameAgency =
+                            dataRepo.getAgencyCount() == 1 ||
```
barbeau, line 68: I'd consider moving this logic to a private method to make it easier to read, and then just calling that method in the below `if ()`

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateUniqueRouteLongNameRouteShortNameCombination.java
```diff
@@ -0,0 +1,78 @@
...
+        this.logger = logger;
+    }
+
+    /**
+     * Use case execution method: checks unicity of combination of fields `route_long_name` and `route_short_name`
+     * in file `routes.txt`. A notice is generated and added to the {@code ValidationResultRepository} provided in the
+     * constructor each time this requirement is not satisfied.
+     */
+    public void execute() {
+        logger.info("Validating rule 'W016 - Duplicate combination od fields`route_short_name` and `route_long_name`'");
```
barbeau, line 57:
```suggestion
        logger.info("Validating rule 'W016 - Duplicate combination of fields `route_short_name` and `route_long_name`'");
```

RULES.md
```diff
@@ -361,4 +364,22 @@ If possible, the GTFS dataset should cover at least the next 30 days of service
...
+
+<a name="W015"/>
+
+### W015 - Duplicate `routes.route_short_name`
+
+All routes should have different `route_short_name`. If routes have same `route_long_name`, they must be different routes serving different areas; and must not be different trips of the same route or different directions of the same route.
+
+<a name="W016"/>
+
+### W016 - Duplicate combination of fields `route_long_name` and `routes.route_short_name`
```
barbeau, line 383:
```suggestion
### W016 - Duplicate combination of fields `routes.route_long_name` and `routes.route_short_name`
```

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateUniqueRouteLongNameRouteShortNameCombination.java
```diff
@@ -0,0 +1,78 @@
...
+
+        final Map<String, Route> routeByRouteLongNameRouteShortName = new HashMap<>();
+        dataRepo.getRouteAll().forEach((routeId, route) -> {
+            final String routeShortName = route.getRouteShortName();
+            final String routeLongName = route.getRouteLongName();
+            if (routeShortName != null && routeLongName != null) {
```
barbeau, line 63: I think you should actually consider `null` value combinations as well - for example - for routeShortName, routeLongName:
* `A`, `null`
* `A`, `null`

...would be a warning, as would:
* `null`, `A`
* `null`, `A`

I see this differently than the other two warning rules, which are only considering one of two fields.

So I believe you could just eliminate this line with the null check, if we are already guaranteed to have at least one of the fields being not null from previous validation. You would need to add more unit tests to verify behavior though.

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateRouteShortNameAreUniqueTest.java
```diff
@@ -0,0 +1,329 @@
...
+        verify(mockRoute3, times(1)).getRouteShortName();
+
+        verifyNoInteractions(mockResultRepo);
+        verifyNoMoreInteractions(mockRoute0, mockRoute1, mockRoute2, mockRoute3, mockLogger, mockDataRepo);
+    }
+
+    // suppressed warning regarding ignored result of method because method is called in assertions
+    @SuppressWarnings("ResultOfMethodCallIgnored")
+    @Test
+    void duplicateRouteShortNameWhenRoutesBelongToUniqueAgencyShouldGenerateNotice() {
```
barbeau, line 143:
```suggestion
    void duplicateRouteShortNameWhenRoutesBelongToNullAgencyShouldGenerateNotice() {
```

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateRouteShortNameAreUnique.java
lionel-nj: Indeed!

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateUniqueRouteLongNameRouteShortNameCombination.java
lionel-nj: Indeed, will do. With additional comments to explain our assumptions regarding the nullability of the two fields, as you mentioned.

lionel-nj: However, I think that the logic here might be more complex: we would still need to check if both fields are not null since the previous validation did not exclude routes with null names from the data repository, these would still be processed and we would need a way to exclude them from the execution of this use case.

barbeau: Doesn't the spec say that you must provide a route short name or route long name - you can't leave them both empty? I would have thought we flagged this as an error early on when importing the CSV.

lionel-nj: Indeed, I checked our code in the builder for `Route` entities, and this requirement was not checked. So I will implement more verification in the `route` builder `.build` method to avoid this situation.

lionel-nj: Thanks for the heads-up

barbeau (approved):
LGTM! 👍

Issue #131, opened 2020-04-29
[REQUEST] Implement Agency language vs Feed language (GTFS rule)


maximearmstrong:
**Is your feature request related to a problem? Please describe.**
Agency language and Feed language must be the same. This is a GTFS rule implemented in Google Python validator and featured in Google Type Error as TYPE_AGENCY_LANG_AND_FEED_LANG_MISMATCH.

**Describe the solution you'd like**
Actual Google GTFS validator behaviour : verifies if language for agencies is consistent with feedinfo.txt

**Describe alternatives you've considered**

**Additional context**
Line 36 in Error support priorities https://docs.google.com/spreadsheets/d/1vqe6wq7ctqk1EhYkgtZ0_TbcQ91vccfs2daSjn20BLE/edit#gid=0

barbeau:
This rule should be reviewed by @timMillet before being implemented. I'm not sure it is an official requirement in the current GTFS spec.

maximearmstrong:
Maybe we could flag it as a warning if it occurs? Since it's in both Google GTFS validator and Google Type Error?

barbeau:
@maximearmstrong We'll need to go through the process of determining whether it's something we want to adopt into the official GTFS spec. If so, and there is consensus among the community, then it would be an error. If not, then this could be one of the configurable values that gets changed depending on which .json configuration you load into the validator. So if you load a "Google" configuration it could indicate it as an error, but others may be just a warning.

timMillet:
`agency.feed_lang` (many) may not match `feed_info.feed_lang` (unique).  It's not a requirement in the spec and I see normal situations where these fields could be different.

I agree I could help on defining what should be an error, and what should/could be a warning.

fabrice-v:
Ok so I guess the question is how the spec handles these files being different, is there a clear precedence rule as to which one to use over the others? It can have an impact on deserialization of numerical values (decimal . vs , for example), and probably other things I dont have in mind.

timMillet:
`feed_info.feed_lang` is the default language of the dataset: "This setting helps GTFS consumers choose capitalization rules and other language-specific settings for the dataset." It has precedence on the others. Heads-up: `mul` meaning multilingual is allowed.

`feed_info.feed_lang` is required but may not be provided as `feed_info.txt` is optional. In this case, `agency.feed_lang` should be used as the default language of the dataset. Heads-up: many `agency.feed_lang` may be provided, with different values. There are no rules to order them.

[...]

PR #352, opened 2020-08-26
Implement #131 - Check agency.agency_lang and feed_info.feed_lang match

lionel-nj:
**Summary:**

This PR provides support to check that `agency.agency_lang` and `feed_info.feed_lang` match.

**Expected behavior:** 

- [x] Write rule logic
- [x] Unit tests
- [x] Update documentation  

### DIFF
--- a/RULES.md
+++ b/RULES.md
@@ -314,6 +314,16 @@ A trip must visit more than one stop in `stop_times.txt` to be usable by passeng
 ### E055 - Mismatching feed and agency language fields
 
 Files `agency.txt` and `feed_info.txt` must define matching `agency.agency_lang` and `feed_info.feed_lang`.
+The default language may be multilingual for datasets with the original text in multiple languages. In such cases, the feed_lang field should contain the language code mul defined by 
+the norm ISO 639-2.
+* If `feed_lang` is not `mul` and does not match with `agency_lang`, that's an error
+* If there is more than one `agency_lang` and `feed_lang` isn't `mul`, that's an error
+* If `feed_lang` is `mul` and there isn't more than one `agency_lang`, that's an error
+
+
+#### References:
+* [feed_info.txt specification](http://gtfs.org/reference/static/#feed_infotxt)
+* [agency.txt specification](http://gtfs.org/reference/static/#agencytxt)
 
 <a name="E056"/>
 
--- a/RULES.md
+++ b/RULES.md
@@ -332,8 +332,8 @@ the norm ISO 639-2.
 * If `feed_lang` is `mul` and there isn't more than one `agency_lang`, that's an error
 
 #### References:
-* [feed_info.txt specification](http://gtfs.org/reference/static/#feed_infotxt)
-* [agency.txt specification](http://gtfs.org/reference/static/#agencytxt)
+* [GTFS feed_info.txt specification](http://gtfs.org/reference/static/#feed_infotxt)
+* [GTFS agency.txt specification](http://gtfs.org/reference/static/#agencytxt)
 
 <a name="E056"/>
 
--- a/RULES.md
+++ b/RULES.md
@@ -309,7 +309,6 @@ Trips must be referred to at least once in `stop_times.txt`.
 
 A trip must visit more than one stop in `stop_times.txt` to be usable by passengers for boarding and alighting.
 
-
 <a name="E055"/>
 
 ### E055 - Mismatching feed and agency language fields
--- a/RULES.md
+++ b/RULES.md
@@ -330,8 +330,7 @@ Trip frequencies must not overlap in time
 ### E055 - Mismatching feed and agency language fields
 
 Files `agency.txt` and `feed_info.txt` must define matching `agency.agency_lang` and `feed_info.feed_lang`.
-The default language may be multilingual for datasets with the original text in multiple languages. In such cases, the feed_lang field should contain the language code mul defined by 
-the norm ISO 639-2.
+The default language may be multilingual for datasets with the original text in multiple languages. In such cases, the feed_lang field should contain the language code mul defined by the norm ISO 639-2.
 * If `feed_lang` is not `mul` and does not match with `agency_lang`, that's an error
 * If there is more than one `agency_lang` and `feed_lang` isn't `mul`, that's an error
 * If `feed_lang` is `mul` and there isn't more than one `agency_lang`, that's an error
--- a/RULES.md
+++ b/RULES.md
@@ -52,6 +52,7 @@ Rules are declared in the [`Notice` module](https://github.com/MobilityData/gtfs
 | [E049](#E049) | Backwards time travel between stops in `stop_times.txt` |
 | [E050](#E050) | Trips must be used in `stop_times.txt` |
 | [E051](#E051) | Trips must have more than one stop to be usable |
+| [E055](#E055) | Mismatching feed and agency language fields |
 
 ### Table of Warnings
 
@@ -306,6 +307,13 @@ Trips must be referred to at least once in `stop_times.txt`.
 ### E051 - Trips must have more than one stop to be usable
 
 A trip must visit more than one stop in `stop_times.txt` to be usable by passengers for boarding and alighting.
+
+
+<a name="E055"/>
+
+### E055 - Mismatching feed and agency language fields
+
+Files `agency.txt` and `feed_info.txt` must define matching `agency.agency_lang` and `feed_info.feed_lang`.
 
 # Warnings
 
--- a/domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/notice/error/FeedInfoLangAgencyLangMismatchNotice.java
+++ b/domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/notice/error/FeedInfoLangAgencyLangMismatchNotice.java
@@ -20,6 +20,7 @@ import org.mobilitydata.gtfsvalidator.domain.entity.notice.NoticeExporter;
 import org.mobilitydata.gtfsvalidator.domain.entity.notice.base.ErrorNotice;
 
 import java.io.IOException;
+import java.util.Set;
 
 public class FeedInfoLangAgencyLangMismatchNotice extends ErrorNotice {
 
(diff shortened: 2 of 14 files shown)


lionel-nj:
sorry @barbeau  please disregard my request for review here until further notice

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateAgencyLangAndFeedInfoFeedLangMatch.java
```diff
@@ -0,0 +1,75 @@
...
+    public void execute() {
+        logger.info("Validating rule 'E055 - Mismatching feed and agency language fields'");
+
+        if (dataRepo.getFeedInfoAll().size() > 0) {
+            // .get can be used without isPresent check here since line 55 ensures the presence of at least 1 element
+            // in the map
+            final String feedInfoFeedLang =
+                    dataRepo.getFeedInfoAll().values().stream().findFirst().get().getFeedLang();
+            dataRepo.getAgencyAll().forEach((agencyId, agency) -> {
+                if (!feedInfoFeedLang.equals("mul")) {
```
barbeau, line 61: `"mul"` should be a static final constant

RULES.md
```diff
@@ -308,6 +309,12 @@ Trips must be referred to at least once in `stop_times.txt`.
 
 A trip must visit more than one stop in `stop_times.txt` to be usable by passengers for boarding and alighting.
 
+<a name="E055"/>
+
+### E055 - Mismatching feed and agency language fields
+
+Files `agency.txt` and `feed_info.txt` must define matching `agency.agency_lang` and `feed_info.feed_lang`.
+
```
barbeau, line 317: The rule description here should have a full description of the logic in the use case, including how `mul` is handled. You could just quote the spec here.

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateAgencyLangAndFeedInfoFeedLangMatch.java
barbeau, line 61: If the general rule logic stays as-is, you could move the `if (!feedInfoFeedLang.equals("mul")) {` before the loop for agencies and just early return if it is `mul`.

Alternately, looking at the spec you could add more checks here:
1. If there is more than one agency_lang and feed_lang isn't `mul`, that's an error
1. If feed_lang is `mul` and there isn't more than one agency_lang, that's an error

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/notice/error/FeedInfoLangAgencyLangMismatchNotice.java
```diff
@@ -0,0 +1,49 @@
...
+public class FeedInfoLangAgencyLangMismatchNotice extends ErrorNotice {
+
+    public FeedInfoLangAgencyLangMismatchNotice(final String agencyId,
+                                                final String agencyName,
+                                                final String agencyLang,
+                                                final String feedInfoFeedLang) {
+        super("feed_info.txt",
+                E_055,
+                "Mismatching feed and agency language fields",
+                String.format("Agency with `agency_name`: `%s` have mismatching `agency.agency_lang`: `%s` and" +
```
barbeau, line 33:
```suggestion
                String.format("Agency `agency_name`: `%s` has mismatching `agency.agency_lang`: `%s` and" +
```

RULES.md
```diff
@@ -308,6 +309,12 @@ Trips must be referred to at least once in `stop_times.txt`.
...
 <a name="E056"/>
```
barbeau, line 318: Reference links to agency.txt and feed_info.txt would be useful.

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateAgencyLangAndFeedInfoFeedLangMatchTest.java
```diff
@@ -0,0 +1,183 @@
...
+        verify(mockAgency1, times(1)).getAgencyLang();
+
+        final ArgumentCaptor<FeedInfoLangAgencyLangMismatchNotice> captor =
+                ArgumentCaptor.forClass(FeedInfoLangAgencyLangMismatchNotice.class);
+
+        verify(mockResultRepo, times(1)).addNotice(captor.capture());
+
+        final List<FeedInfoLangAgencyLangMismatchNotice> noticeList = captor.getAllValues();
+
+        assertEquals("feed_info.txt", noticeList.get(0).getFilename());
```
barbeau, line 175: I'd check the number of notices in the `noticeList` too

barbeau (commented):
Thanks @lionel-nj. Generally looks good, a few comments below.

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateAgencyLangAndFeedInfoFeedLangMatch.java
```diff
@@ -0,0 +1,105 @@
...
+                // If there is more than one agency_lang and feed_lang isn't mul, that's an error
+                if (agencyLangCollection.size() > 1 && !feedInfoFeedLang.equals(agency.getAgencyLang())) {
+                    resultRepo.addNotice(
+                            new FeedInfoLangAgencyLangMismatchNotice(
+                                    agencyId,
+                                    agency.getAgencyName(),
+                                    agency.getAgencyLang(),
+                                    feedInfoFeedLang));
+                    return;
+                }
```
barbeau, line 90: This check should be above this loop through the agencies.txt data - it only needs to be checked once after adding all the agency_lang to `agencyLangCollection`, and should have logic like:

```java
// If there is more than one agency_lang and feed_lang isn't mul, that's an error
 if (agencyLangCollection.size() > 1 && !feedInfoFeedLang.equals(MUL))) {
                    resultRepo.addNotice(
                            new FeedInfoLangAgencyLangMismatchNotice(
                                    agencyId,
                                    agency.getAgencyName(),
                                    agency.getAgencyLang(),
                                    feedInfoFeedLang));
                    return;
}
```

There shouldn't be

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateAgencyLangAndFeedInfoFeedLangMatch.java
```diff
@@ -0,0 +1,105 @@
...
+            } else {
+                // If feed_lang is mul and there isn't more than one agency_lang, that's an error
```
barbeau, line 92: This check can also be moved above this loop

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateAgencyLangAndFeedInfoFeedLangMatch.java
barbeau: After looking at this further I think I see that you're trying to combine some of the conditionals within a single `if (!feedInfoFeedLang.equals(MUL)) {` statement, and some of those could early return on the first loop iteration, so it might be logically equivalent to moving the IFs outside the loop. However, for clarity I'd still suggest you move them above this loop, as it's a clearer indication that you only expect the logic to execute once, and not on every item in agencies.txt. I think it's also easier to match the code to the comments if the logic is the same, even if you might be repeating some of the conditionals.

RULES.md
```diff
@@ -319,6 +320,21 @@ Trip frequencies must not overlap in time
 
 * [GTFS frequencies.txt specification](http://gtfs.org/reference/static/#frequenciestxt)
 
+<a name="E055"/>
+
+### E055 - Mismatching feed and agency language fields
+
+Files `agency.txt` and `feed_info.txt` must define matching `agency.agency_lang` and `feed_info.feed_lang`.
+The default language may be multilingual for datasets with the original text in multiple languages. In such cases, the feed_lang field should contain the language code mul defined by 
+the norm ISO 639-2.
```
barbeau, line 329: Extra line break

barbeau (commented):
Thanks @lionel-nj! This mostly looks good, see comments below.

barbeau (commented):
@lionel-nj Looks good! One last comment.

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateAgencyLangAndFeedInfoFeedLangMatch.java
```diff
@@ -0,0 +1,95 @@
...
+        }
+        // If there is more than one agency_lang and feed_lang isn't mul, that's an error
+        // check if feed_lang is `mul` here is redundant but is provided for future software maintenance
+        if (!feedInfoFeedLang.equals(MUL) && agencyLangCollection.size() > 1 && !agencyLangCollection.contains(feedInfoFeedLang)) {
+            resultRepo.addNotice(new FeedInfoLangAgencyLangMismatchNotice(agencyLangCollection, feedInfoFeedLang));
+            return;
+        }
+
+        dataRepo.getAgencyAll().forEach((agencyId, agency) -> {
+            // If feed_lang is not mul and differs from agency_lang, that's an error
```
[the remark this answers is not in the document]
lionel-nj: I'll remove the redundant checks, since the unit tests fails if the initial `if (feedInfoFeedLang.equals(MUL)) {` is removed (it has been manually tested).

usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateAgencyLangAndFeedInfoFeedLangMatch.java
```diff
@@ -0,0 +1,94 @@
...
+        if (feedInfoFeedLang.equals(MUL)) {
+            // If feed_lang is mul and there isn't more than one agency_lang, that's an error
+            if (agencyLangCollection.size() <= 1) {
+                resultRepo.addNotice(new FeedInfoLangAgencyLangMismatchNotice(agencyLangCollection));
+                return;
+            }
+            return;
+        }
+        // If there is more than one agency_lang and feed_lang isn't mul, that's an error
+        if (agencyLangCollection.size() > 1 && !agencyLangCollection.contains(feedInfoFeedLang)) {
```
barbeau, line 77: Sorry, just catching this - I believe `!agencyLangCollection.contains(feedInfoFeedLang)` should be removed from this IF statement. It should be an error regardless if at least one of the agency langs matches the feed lang, because in all cases when there is more than one agency lang the feed lang should be `mul`.

lionel-nj: What happens is that with the current implementation, the following situation would not be caught:
- `feed_info..feed_lang = `US` and `agency.txt` defines `FR` 

which leads tests `feedLangNotMulAndOnlyOneMismatchingAgencyLangShouldGenerateNotice` ans `nonMatchingFeedInfoFeedLandShouldGenerateNotice` to fail

So I cam to the conclusion that we needed an additional check to pass handle these situations and pass these tests.

barbeau: Nice! That's what unit tests are for :).

lionel-nj: Sorry I misunderstood something. I discussed, yes, the code will be changed.

barbeau (approved):
LGTM, thanks!

Issue #237, opened 2020-06-03
Implement Zone ID presence verification (GTFS Rule)


maximearmstrong:
**Is your feature request related to a problem? Please describe.**
Zone IDs should be present in stops.txt if fare rules are defined in fare_rules.txt.

**Describe the solution you'd like**
Verify if rules are defined in fare_rules.txt. If yes, Zone IDs are required.

**Describe alternatives you've considered**

**Additional context**
GTFS Spec http://gtfs.org/reference/static/#stopstxt

PR #369, opened 2020-09-18
Implement #237 - Implement Zone ID presence verification 

lionel-nj:
**Summary:**

`stops.zone_id` is conditionally required. If `fare_rules.txt` is provided, then is it required for all types of stops except `STATION` and `ENTRANCE`

**Expected behavior:** 

A `MissingRequiredValueNotice` should be generated and added to the result repo each time the aforementioned condition is not respected.


fabrice-v:
@lionel-nj What is the status of this PR?

e-lo:
Why was this closed – Is this no longer a valid rule that needs enforcing?

fabrice-v:
The status of this PR was unclear. The associated issue #237 is still opened.

lionel-nj:
@fabrice-v it was in-progress but we had to switch to other tasks. Looking at it, it seems like a few improvements could be done before code review. We should add the PR to the board so that we can plan work for the next sprint.

fabrice-v:
PR don't go to the board, only issues. Make sure the associated issue is in the right column. We will reopen when resuming work.

PR #414, opened 2020-10-08
End to end testing workflow

### LINKED ISSUE #405: [REQUEST] Add functional tests to testing suite
As a local user, I would appreciate end-to-end functional tests be included in *repository* testing suite (as opposed to in a GitHub-triggered-action) in order to make sure the code *in my environment* is working appropriately from end-to-end. 

A set of end-to-end functional tests for a variety of feed sources and settings run within local tests. 
https://guides.gradle.org/testing-gradle-plugins/ has some suggestions. 

While the GitHub triggered functional test is useful, it is not easily run locally by non-contributors *and* isn't necessarily run as a matter of course on local machines prior to pushing code.

### LINKED ISSUE #406: [REQUEST] Add additional feeds for end-to-end, functional tests
I've been frustrated when the validator fails and is brittle to different feeds other than MBTA. 

~5sh additional feeds of varying complexity/quality added to the validator's testing suite  (see issue #405) and continuous integration process. 

Running them myself and submitting bugs for each, which isn't ideal b/c developers would ideally catch these bugs before users.

f8full:
closes #406  

**Summary:**

Add a worflow to run the validator on multiple datasets when code is merged on master

**Expected behavior:** 

On future merge to master, the validator should run on both MBTA and STM datasets

### DIFF
--- a/.github/workflows/endtoend.yml
+++ b/.github/workflows/endtoend.yml
@@ -15,6 +15,8 @@ jobs:
         uses: actions/setup-java@v1
         with:
           java-version: 1.11
+      - name: Package jar with Gradle
+        run: ./gradlew shadowJar
       - name: Run validator on STM data
         run: java -jar application/cli-app/build/libs/gtfs-validator-v1.3.0-SNAPSHOT.jar -u http://www.stm.info/sites/default/files/gtfs/gtfs_stm.zip -i input.zip -e input -o output -abort_on_error false
       - name: Run validator on MBTA data
--- a/.github/workflows/endtoend.yml
+++ b/.github/workflows/endtoend.yml
@@ -1,0 +1,21 @@
+name: End to end
+
+on:
+  push:
+    branches: [ master ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v1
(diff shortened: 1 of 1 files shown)


.github/workflows/endtoend.yml
```diff
@@ -0,0 +1,23 @@
+name: End to end
```
lionel-nj, line 1: Should we add a precision: `End to end testing workflow`?

lionel-nj (approved):
One in-line comment but generally LGTM .

e-lo:
Thank for working on this. A few questions:

1. Is there a reason why you wouldn't add a list of feeds and then iterate through them?
2. Are you limited in compute time for tests? Especially if it is only running when merging to master it would seem appropriate to add many more feeds using a variety of techniques to check for brittleness across multiple dimensions:  running from local data, from a variety of common feed aggregators, using a handful of settings, etc.
3. See #405 for additional suggestions re using tests rather than just relying on workflows which aren't replicable locally.

.github/workflows/endtoend.yml
fabrice-v: I'd arbitrate against for 2 reasons
- the file is located in a folder named workflow
- this name is what ends up in the badges we display in our readme

Issue #397, opened 2020-09-30
"otherCalendar" is null


mcplanner:
**Describe the bug**
The validator was unable to process a feed and gave an error.

**To Reproduce**
Run the validator on the following dataset: http://transitfeeds.com/p/sonoma-marin-area-rail-transit/1050/latest

**Expected behavior**
No notice should be generated in this case.

**Witnessed behavior**
```
Exception in thread "main" java.lang.NullPointerException: Cannot invoke "org.mobilitydata.gtfsvalidator.domain.entity.gtfs.Calendar.getStartDate()" because "otherCalendar" is null
	at org.mobilitydata.gtfsvalidator.domain.entity.gtfs.Calendar.isOverlapping(Calendar.java:500)
	at org.mobilitydata.gtfsvalidator.usecase.ValidateNoOverlappingStopTimeInTripBlock.checkOverlappingTripsDifferentServiceIdCalendarProvided(ValidateNoOverlappingStopTimeInTripBlock.java:260)
	at org.mobilitydata.gtfsvalidator.usecase.ValidateNoOverlappingStopTimeInTripBlock.lambda$execute$0(ValidateNoOverlappingStopTimeInTripBlock.java:128)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
	at org.mobilitydata.gtfsvalidator.usecase.ValidateNoOverlappingStopTimeInTripBlock.lambda$execute$1(ValidateNoOverlappingStopTimeInTripBlock.java:84)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
	at org.mobilitydata.gtfsvalidator.usecase.ValidateNoOverlappingStopTimeInTripBlock.lambda$execute$2(ValidateNoOverlappingStopTimeInTripBlock.java:69)
	at java.base/java.util.HashMap.forEach(HashMap.java:1425)
	at java.base/java.util.Collections$UnmodifiableMap.forEach(Collections.java:1521)
	at org.mobilitydata.gtfsvalidator.usecase.ValidateNoOverlappingStopTimeInTripBlock.execute(ValidateNoOverlappingStopTimeInTripBlock.java:67)
	at org.mobilitydata.gtfsvalidator.Main.main(Main.java:216)
```
**Environment used**
- OS: macOS Catalina 10.15.6
- Java version: 11.0.5
- gtfs-validator version: v1.2.2

**Additional context**
[...]

PR #423, opened 2020-10-13
Fix #399 - Handle NPE

### LINKED ISSUE #399: [BUG] "this.originalRouteTypeInteger" is null in v1.3.0 SNAPSHOT
**Describe the bug**
The validator was unable to process a feed and gave an error.

**To Reproduce**
Run the validator on the following dataset: http://www.mst.org/google/google_transit.zip

**Expected behavior**

**Witnessed behavior**
```
Exception in thread "main" java.lang.NullPointerException: Cannot invoke "java.lang.Integer.intValue()" because "this.originalRouteTypeInteger" is null
	at org.mobilitydata.gtfsvalidator.domain.entity.gtfs.routes.Route$RouteBuilder.build(Route.java:291)
	at org.mobilitydata.gtfsvalidator.usecase.ProcessParsedRoute.execute(ProcessParsedRoute.java:80)
	at org.mobilitydata.gtfsvalidator.Main.lambda$main$0(Main.java:114)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
	at org.mobilitydata.gtfsvalidator.Main.main(Main.java:94)
```

**Environment used**
- OS: macOS Catalina 10.15.6
- Java version: 11.0.5
- gtfs-validator version: gtfs-validator-v1.3.0-SNAPSHOT.jar

lionel-nj:
**Summary:**

This support provides support to handle NPE when processing file `routes.txt` with null `routes.route_type`: 

- implementation of null checks in `Route.RouteBuilder.build` method is modified
- additional unit tests are provided to check said implementation

**Expected behavior:** 

No NPE


lionel-nj:
Note that fixing this bug led the way to discovering another one (which seems similar/related to https://github.com/MobilityData/gtfs-validator/issues/397).
It will be further investigated when fixing #397 

@fabrice-v

fabrice-v:
Ok I'll hold my review until we also have the fix for #397

fabrice-v (changes requested):
As discussed, this is not what we want because said value can't be null per specification.

The real fix would be to diagnose why we let a `null` slip through and generate an error in the ouput report

fabrice-v (dismissed):
After further discussion, this is actually the correct fix. Now on this dataset we output an error notice for a missing required value, and then instead of crashing we output a second notice for an unexpected enum value.

Note that the behavior of the validator is undefined after the first error.

lionel-nj:
To do: check if this is a pattern.

fabrice-v:
As discussed, some code is missing around having is `isEnumValid` calls from the rest of the code

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/routes/RouteType.java
```diff
@@ -66,10 +65,13 @@ static public RouteType fromInt(final Integer fromValue) {
...
      */
-    static public boolean isEnumValueValid(final int value) {
+    static public boolean isEnumValueValid(final Integer value) {
+        if (value == null) {
+            return false;
+        }
         return Stream.of(RouteType.values())
                 .filter(enumItem -> enumItem.value == value)
                 .findAny()
                 .orElse(null) != null;
```
fabrice-v, line 75: Isn't `.orElse(null) != null` supposed to cover the case where  value is null?

fabrice-v (commented):
It seems the explicit null check is extra?

domain/src/main/java/org/mobilitydata/gtfsvalidator/domain/entity/gtfs/routes/RouteType.java
lionel-nj: Indeed, the null check is redundant. I suggest in https://github.com/MobilityData/gtfs-validator/pull/423/commits/06018615c93a36e998a364d551fd01def54f3b76 to use `anyMatch` coupled with a try/catch block that will prevent NPE to be thrown when `isEnumValueMethod` arguent is `null`. This NPE is thrown by the filter `enumItem -> enumItem.value == value`.

fabrice-v (approved):
lgtm

PR #425, opened 2020-10-14
Fix #397 - Handle NPE when checking overlapping trips

### LINKED ISSUE #397: "otherCalendar" is null

lionel-nj:
**Summary:**

This PR provides support to avoid NPE when checking overlapping trips by:
- modifying rule E054 handle cases where comparing trips that refer to bith `calendar.txt` and `calendar_dates.txt` files.
- providing additional unit tests to verify said implementations

**Expected behavior:** 

No NPE


lionel-nj:
To do: check if this is a pattern.

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlockTest.java
```diff
@@ -1615,37 +1609,427 @@ void tripWithNullLastTimeShouldNotGenerateNotice() {
...
+        // | a         | 20200723 | 1              |
+        // | a         | 20200801 | 1              |
+        // | b         | 20200904 | 0              |
+        // | c         | 20201008 | 1              |
+
+        // Here no service overlap:
+        // - service a is active on two days: 2020-07-23 and 2020-08-01
+        // - service b is active only on 2020-09-04
+        // - service c is active only on 2020-01-08
+        // - service d is active from 2020-08-15 to 2020-08-31 on thursdays
```
barbeau, line 1701:
```suggestion
        // - service d is active from 2020-08-15 to 2020-08-31 on thursdays, fridays, and saturdays
```

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlockTest.java
```diff
@@ -1615,37 +1609,427 @@ void tripWithNullLastTimeShouldNotGenerateNotice() {
...
+        // | serviceId | monday | tuesday | wednesday | thursday | friday | saturday | sunday | start_date | end_date |
+        // |-----------|--------|---------|-----------|----------|--------|----------|--------|------------|----------|
+        // | d         | 0      | 0       | 0         | 1        | 1      | 1        | 0      | 20200815   | 20200831 |
+
+        // calendar_dates.txt:
+        // | serviceId | date     | exception_type |
+        // |-----------|----------|----------------|
+        // | a         | 20200723 | 1              |
+        // | a         | 20200801 | 1              |
+        // | b         | 20200904 | 0              |
```
barbeau, line 1694: This record for serviceId `b` doesn't make sense - it's removing service on Sept 4 2020 but there isn't any service scheduled on this date in `calendar.txt`.

I'd suggest changing this to:

```suggestion
        // | d         | 20200820 | 0              |
```

...so it's removing service `d` on 20200820 (a Thursday) that would otherwise normally run on a Thurs (according to it's definition in calendar.txt).

Then, I would change service `c` in calendar_dates.txt to be added on 20200820. This would then mirror the most likely real-world scenario here - 20200820 is a "holiday" that is running different service on one particular day of the service period by removing service `d` and running service `c` instead.

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlockTest.java
```diff
@@ -1615,37 +1609,427 @@ void tripWithNullLastTimeShouldNotGenerateNotice() {
...
+        // | c         | 20201008 | 1              |
```
barbeau, line 1695: Per my above comment:

```suggestion
        // | c         | 20200820 | 1              |
```

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlockTest.java
```diff
@@ -1615,37 +1609,427 @@ void tripWithNullLastTimeShouldNotGenerateNotice() {
...
+
+        // Here services c and d overlap
+        // - service c is active only on 2020-01-08
```
barbeau, line 1890: The above table says that service `c` is active on Oct. 8, 2020. But that still wouldn't overlap service `d`, which only runs from Aug 15 to Aug 31.

usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateNoOverlappingStopTimeInTripBlockTest.java
```diff
@@ -1615,37 +1609,427 @@ void tripWithNullLastTimeShouldNotGenerateNotice() {
...
+        // | 9      | 07:00       | 07:40         | 105    | 14            |
+        // | 9      | 07:40       | 07:59         | 106    | 15            |
+
+        // trips.txt + stop_times.txt
+        // | routeId | tripId | serviceId | blockId | first stop time | last stop time |
+        // |---------|--------|-----------|---------|-----------------|----------------|
+        // | 0       | 2      | a         | 7       | 05:00           | 06:30          |
+        // | 0       | 5      | b         | 7       | 07:00           | 10:00          |
+        // | 0       | 8      | c         | 7       | 11:03           | 11:20          |
+        // | 1       | 9      | c         | 7       | 07:00           | 10:00          |
```
barbeau, line 1682: I think this should be:

```suggestion
        // | 0       | 9      | d         | 7       | 06:00           | 10:00          |
```

That means that "normally" you'd run one long trip from 6am to 10am Thurs, Fri, and Sat.

Based on my other comments, the only exception would be 20200820 (service `c`), which instead runs only from 11:03 to 11:20. And then for service `a` we also have a trip running on 20200723 and 20200801 from 5:00 to 6:30 which shouldn't overlap due to the dates.

If you set it up this way, then for the other test `overlappingTripsWithDifferentServiceIdShouldGenerateNoticeWhenComparingTripsReferencingBothFiles()` where you want to detect overlap you can change the service `a` dates to 20200827 and 20200828 in calendar_dates.txt. That should cause an overlap on Aug 27 and Aug 28 between 6 and 6:30.

barbeau:
@lionel-nj Is this PR still relevant after the annotation refactor? Should it be closed and referenced from the issue to port this rule over to the annotation-based validator?

lionel-nj:
Just seing your question now @barbeau: we can close it for now, but we surely want to implement this in the annotation-based-validator.

Issue #435, opened 2020-10-19
[REQUEST] Rename validation report


lionel-nj:
**Is your feature request related to a problem? Please describe.**
At present, the validation report is named `results.json`.

**Describe the solution you'd like**
Name the validation report as `agency.agency_id` + `timestamp.json` or  `agency.agency_id` + `timestamp.pb`

PR #438, opened 2020-10-20
Implement #435 - Rename validation report

lionel-nj:
**Summary:**

This PR provides support to rename the validation report as the `feed_info.feed_publisher_name` + `__` + timestamp + file extension (`.json` or `.pb`). All whitespace characters are replaced by `_` before using the resulting path.

**Expected behavior:** 

**1. Validation report as `.json`:**
Validation reports should no longer be named `results.json`. Instead the name of said file should be `feed_info.feed_publisher_name` + `__` `timestamp`(at the time of validation in `America/Montreal` timezone) +`.json`.

If `feed_info.txt` is not provided then the name of the validation report is limited to the timestamp at the moment of validation in `America/Montreal` timezone.

**Sample filename:** after validating [MBTA's GTFS archive](https://cdn.mbta.com/MBTA_GTFS.zip) on 2020-10-20 at 09:07:48 (America/Montreal timezone), the validation report will be named as follows `MBTA__2020-10-20_09/07/48.442365.json` 

**2. Validation report as `.pb`:** 
Validation reports should no longer be named  `1.pb` /  `2.pb`/ ... / `n.pb` . Instead the same of said file should be `feed_info.feed_publisher_name` + `__` + `timestamp` (at the time of validation in `America/Montreal` timezone) + `-` + fileCounter + `.pb`.

If `feed_info.txt` is not provided then the name of the validation reports are limited to: `__` + `timestamp` (at the moment of validation in `America/Montreal` timezone) + filecounter + `.pb`

**Sample filenames:** after validating [MBTA's GTFS archive](https://cdn.mbta.com/MBTA_GTFS.zip) on 2020-10-20 at 09:07:48 (America/Montreal timezone), the validation reports will be named as follows 
- `MBTA__2020-10-20 09/07/48.442365-1.pb` 
- `MBTA__2020-10-20 09/07/48.442365-2.pb` 
- ...
- `MBTA__2020-10-20 09/07/48.442365-n.pb` 

image

### DIFF
--- a/adapter/exporter/src/main/java/org/mobilitydata/gtfsvalidator/exporter/ProtobufNoticeExporter.java
+++ b/adapter/exporter/src/main/java/org/mobilitydata/gtfsvalidator/exporter/ProtobufNoticeExporter.java
@@ -1013,7 +1013,7 @@ public class ProtobufNoticeExporter implements NoticeExporter {
 
         public OutputStream getStream() throws IOException {
             OutputStream newStream = Files.newOutputStream(Paths.get(
-                    targetPath + File.separator + streamCounter +
+                    targetPath + "-" + streamCounter +
                             ProtobufNoticeExporter.FILE_EXTENSION
             ));
             ++streamCounter;
--- a/README.md
+++ b/README.md
@@ -28,6 +28,16 @@ java -jar gtfs-validator-v1.1.0.jar -i relative/path/to/zipped_dataset -o relati
  1. Extract the zip content to a directory located at `relative/extraction/path`
  1. Validate the GTFS data and output the results to the directory located at `relative/output/path`. Validation results are exported to JSON by default. The validation process will not be executed on the enumeration of files provided via option `-x` and the files that rely on them.
 
+Note: 
+ - **export validation report as `.json` file**: After validating [MBTA's GTFS archive](https://cdn.mbta.com/MBTA_GTFS.zip) on 2020-10-20 at 09:07:48 (America/Montreal timezone), the validation report will be named as follows `MBTA_2020-10-20 09/07/48.442365.json`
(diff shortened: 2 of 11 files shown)


adapter/exporter/src/main/java/org/mobilitydata/gtfsvalidator/exporter/ProtobufNoticeExporter.java
```diff
@@ -1013,7 +1013,7 @@ public ProtobufOutputStreamGenerator(final String outputPath) {
 
         public OutputStream getStream() throws IOException {
             OutputStream newStream = Files.newOutputStream(Paths.get(
-                    targetPath + File.separator + streamCounter +
+                    targetPath + "-" + streamCounter +
```
fabrice-v, line 1016: Just out of curiosity: what is this a fix for and is there any reason we'd run into troubles using the `-` (like it's an invalid character on some file systems or something)

fabrice-v (changes requested):
Code looks good to me. 

2 changes required
- provide a sample filename in the PR description
- update the README to provide a sample filename

adapter/exporter/src/main/java/org/mobilitydata/gtfsvalidator/exporter/ProtobufNoticeExporter.java
lionel-nj: Using `File.separator` caused an issue: the system could not find the path to `targetPath/streamcounter.pb/`.

fabrice-v (approved):
Good thks. As discussed, I'll see if I can reformat the string to exclude ` ` characters, they may come from `feed_info.feed_publisher_name` in addition to the timestamp.

I will also do the merge.

fabrice-v:
This could be hardened by modifying the replace all regex to also replace those forbidden characters
`{ '\\', '/', ':', '*', '?', '"', '<', '>', '|', '[', ']', '\'', ';',
            '=', ',' }`

PR #433, opened 2020-10-19
Fix 398 - Implement solution to handle URL redirections

lionel-nj:
**Summary:**

This PR provides support to handle URL redirections.

**Expected behavior:** 
gtfs-validator should be able to download `.zip` files when content to download has been moved.

Please make sure these boxes are checked before submitting your pull request - thanks!

- [x] Run the unit tests with `gradle test` to make sure you didn't break anything
- [x] Format the title like "Fix #<issue_number> - <short description of fix and changes>" (for example - "Fix #1111 - Check for null value before using field")
- [x] Linked all relevant issues
- ~[] Include screenshot(s) showing how this pull request works and fixes the issue(s)~

PR #462, opened 2020-10-27
URL from OpenMobilityData BART GTFS --> empty zip

fabrice-v:
Please create the PR and a member of the Mobility Data team will take it from here - thanks!

- [x] This PR only touches `end_to_end.ym` in `.github/workflows`
- [x] `-abort_on_error` to **false**
- [ ] Null Pointer Exception --> add bug and NPE tag
- [x] Attach the dataset and report from the workflow run to this PR

***
***

**Summary:**

- [x] optimize GitHub Action worflow run for this specific PR

image
DONE: See https://github.com/MobilityData/gtfs-validator/pull/462/commits/095b5199635eb94e5f8983a954cd008cd4d6ab1a

- [ ] **compute** Path in a separate _use case_ (mmmm that actually was for json report path generation (name of the json report file), so not sure it applies here. I'll let you decide and let me know what you implemented/issues created/... you ended up putting or not in this PR in a convenient way for you, @lionel-nj 
- [x] **put** BART dataset _back in place_ (**ordering by issue number**). 
- [x] **remove** `-x agency.txt`

**Expected behavior:** 

This PR provides support to prevent `empty zip`  error when executing the validator on BART dataset.


fabrice-v:
[bart.zip](https://github.com/MobilityData/gtfs-validator/files/5448713/bart.zip)

[validation_report_all.zip](https://github.com/MobilityData/gtfs-validator/files/5448717/validation_report_all.zip)

.github/workflows/end_to_end.yml
```diff
@@ -51,6 +51,8 @@ jobs:
         run: java -jar application/cli-app/build/libs/*.jar -u https://octa.net/current/google_transit.zip -i octa.zip -e input -o output
       - name: Validate dataset from issue 400 -- Siskiyou Transit and General Express
         run: java -jar application/cli-app/build/libs/*.jar -u http://transitfeeds.com/p/siskiyou-transit-and-general-express/492/latest/download -i stge.zip -e input -o output
+      - name: Validate dataset from issue 379 -- Bay Area Rapid Transit
+        run: java -jar application/cli-app/build/libs/*.jar -u http://transitfeeds.com/p/siskiyou-transit-and-general-express/492/latest/download -i bart.zip -e input -o output
```
lionel-nj, line 55:
```suggestion
        run: java -jar application/cli-app/build/libs/*.jar -u http://www.bart.gov/dev/schedules/google_transit.zip -i bart.zip -e input -o output
```

lionel-nj (changes requested):
The link in job was not the good one, we should check with the one I suggest. However, the linked issue seems related to https://github.com/MobilityData/gtfs-validator/pull/433, where the same bug was fixed.

fabrice-v:
> The link in job was not the good one, we should check with the one I suggest. However, the linked issue seems related to #433, where the same bug was fixed.

I know. There is something weird. You should be in good shape to investigate and debug.
Thks,

lionel-nj:
>  compute Path in a separate use case (mmmm that actually was for json report path generation (name of the json report file), so not sure it applies here. I'll let you decide and let me know what you implemented/issues created/... you ended up putting or not in this PR in a convenient way for you, @lionel-nj

I will open an issue regarding this, and implement in a separate PR if that is okay with you

lionel-nj:
As https://github.com/MobilityData/gtfs-validator/runs/1344351102?check_suite_focus=true demonstrates:the error is no longer raised.

.github/workflows/end_to_end.yml
```diff
@@ -2,7 +2,7 @@ name: End to end
 
 on:
   push:
-    branches: [ master,  transport-agency-name] #<-- replace transport-agency-name by the name of the agency/publisher
+    branches: [ master,  bay-area-rapid-transit] #<-- replace transport-agency-name by the name of the agency/publisher
```
fabrice-v, line 5:
```suggestion
    branches: [ master,  transport-agency-name] #<-- replace transport-agency-name by the name of the agency/publisher
```

fabrice-v (commented):
Lgtm (can't approve as author). Thks lionel, I'm merging now

lionel-nj (approved):
LGTM

Issue #402, opened 2020-10-01
[BUG] validation error for extraneous blank rows


mcplanner:
**Describe the bug**
A file has blank rows after its content, and is rendering a validation error for "invalid row length".

**To Reproduce**
Run on this file: https://realtime.goldengate.org/gtfsstatic/GTFSTransitData.zip

**Expected behavior**
I don't expect empty rows after the body of the content to create a validation error.

**Witnessed behavior**
`Notice{filename='fare_rules.txt', level='ERROR', code='4', title='Invalid row length', description='Invalid length for row:52 -- expected:5 actual:1', extra='{actualLength=1, rowIndex=52, expectedLength=5}'}, `

**Environment used**
- OS: macOS Catalina 10.15.6
- Java version: 11.0.5
- gtfs-validator version: gtfs-validator-v1.3.0-SNAPSHOT.jar

**Additional context**
The file is not using agency_ID properly, so v1.3.0 is necessary.

mcplanner:
I believe this is also the cause of the complementary error:

`Notice{filename='fare_rules.txt', level='ERROR', code='15', title='Missing required value', description='Missing value for field:`fare_id` marked as required in entity with id:`null`', extra='{fieldName=fare_id}'}, `

fabrice-v:
After analysis, it appears the rows in the provided dataset file are not empty.

An empty row shall contain the right number of comas to define as many empty values as there are headers in the first row.
What we found in the provided dataset are malformed rows:
![image](https://user-images.githubusercontent.com/55884852/95998231-aa9be900-0e02-11eb-85c1-7c9088dd9178.png)


As such, the validator works as intended by flagging malformed CSV rows in the input data.

Note that the behavior of the validator is undefined after the first error, which is why it runs with `abort_on_error` by default.

Issue #213, opened 2020-05-11
[REQUEST] Implement CSV file is not empty (CSV rule)


maximearmstrong:
**Is your feature request related to a problem? Please describe.**
A CSV file should not be empty. This is a CSV rule implemented in Google Python validator raising an EmptyFile error.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**
Line 204 in Error support priorities https://docs.google.com/spreadsheets/d/1vqe6wq7ctqk1EhYkgtZ0_TbcQ91vccfs2daSjn20BLE/edit#gid=0

dudebrains01:
This rule is meant to flag completely empty CSV files, where no headers are provided. This is **invalid** and constitutes an **error** ([spec here](https://github.com/MobilityData/transit/blob/master/gtfs/spec/en/reference.md#file-requirements))

Many feeds have CSV files with headers but no data. This is **valid**. (For example see [frequencies.txt for Transcollines](http://transitfeeds.com/p/transcollines/1275/latest/file/frequencies.txt))

PR #465, opened 2020-10-28
False positive on missing calendar_date field

### LINKED ISSUE #401: [BUG] False positive on missing calendar date field
**Describe the bug**
The validator is turning up an error claiming that entries are missing in the calendar_dates file, but that appears to be in correct.

**To Reproduce**
Run the validator on this feed: https://www.ci.manteca.ca.us/CommunityDevelopment/MantecaTransit/googletransit/google_transit.zip

**Expected behavior**
I did not expect to see an error on this file.

**Witnessed behavior**
`Notice{filename='calendar_dates.txt', level='ERROR', code='15', title='Missing required value', description='Missing value for field:`date` marked as required in entity with id:`null`', extra='{fieldName=date}'}, `

**Environment used**
- OS: macOS Catalina 10.15.6
- Java version: 11.0.5
- gtfs validator version: gtfs-validator-v1.3.0-SNAPSHOT.jar

**Additional context**
Not that Manteca's feed is not applying the agency ID correctly.

lionel-nj:
closes #401 

**Summary:**

This PR provides support to prevent the erroneous generation of `MissingRequiredValueNotice` when a GTFS file contains blank lines.

**Expected behavior:** 

`MalformedCsvRowNotice` should be generated and added to the validation result repo if:
- [x] a GTFS file contains blank lines (1 notice per blank line)

### DIFF
--- a/adapter/exporter/src/main/java/org/mobilitydata/gtfsvalidator/exporter/JsonNoticeExporter.java
+++ b/adapter/exporter/src/main/java/org/mobilitydata/gtfsvalidator/exporter/JsonNoticeExporter.java
@@ -420,4 +420,9 @@ public class JsonNoticeExporter implements NoticeExporter {
     public void export(final ValidationProcessInfoNotice toExport) throws IOException {
         jsonGenerator.writeObject(toExport);
     }
+
+    @Override
+    public void export(MalformedCsvRowNotice toExport) throws IOException {
+        jsonGenerator.writeObject(toExport);
+    }
 }
--- a/adapter/exporter/src/main/java/org/mobilitydata/gtfsvalidator/exporter/ProtobufNoticeExporter.java
+++ b/adapter/exporter/src/main/java/org/mobilitydata/gtfsvalidator/exporter/ProtobufNoticeExporter.java
@@ -1010,6 +1010,16 @@ public class ProtobufNoticeExporter implements NoticeExporter {
                 .writeTo(streamGenerator.getStream());
     }
 
+    @Override
+    public void export(final MalformedCsvRowNotice toExport) throws IOException {
+        protoBuilder.clear()
+                .setCsvFileName(toExport.getFilename())
+                .setSeverity(GtfsValidationOutputProto.GtfsProblem.Severity.WARNING)
+                .setEntityRow((Integer) toExport.getNoticeSpecific(KEY_ROW_INDEX))
+                .build()
+                .writeTo(streamGenerator.getStream());
+    }
+
     public static class ProtobufOutputStreamGenerator {
         private final String targetPath;
         private final List<OutputStream> openedStreamCollection = new ArrayList<>();
--- a/adapter/exporter/src/test/java/org/mobilitydata/gtfsvalidator/exporter/JsonNoticeExporterTest.java
+++ b/adapter/exporter/src/test/java/org/mobilitydata/gtfsvalidator/exporter/JsonNoticeExporterTest.java
@@ -1163,4 +1163,18 @@ class JsonNoticeExporterTest {
         verify(mockGenerator, times(1)).writeObject(ArgumentMatchers.eq(toExport));
         verifyNoMoreInteractions(mockGenerator);
     }
+
+    @Test
+    void exportMalformedCsvRowNoticeShouldWriteObject() throws IOException {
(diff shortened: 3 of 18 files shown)


lionel-nj:
After analyzing the dataset provided in the linked issue, it appears that:
1. `calendar_dates.txt` might be malformed: I suspect each line of the file to miss a CRLF. Editing said file and adding a line return, allowed to get rid of the notice that were generated.

This CSV malformation should have been caught when parsing said file, not when processing the related GTFS entities.

Investigating this further.

lionel-nj:
As discussed, we will:
- [ ] investigate possibility to setup parser to take CRLF into account
- [ ] provide a minimal fix to avoid producing `ParsedEntity` wii all rows set to `null`

lionel-nj:
Turns out that I misunderstood the source of the problem. What actually happens is that for this particular dataset, `calendar_dates.txt` contains blank lines:
Capture d’écran, le 2020-10-30 à 13 59 54

Therefore, we should no longer investigate possibility to setup parser to take CRLF into account, but instead we should investigate the process we currently have to deal with such blank lines.

Will still provide a minimal fix to avoid producing `ParsedEntity` from blank lines of a `CSV` file.

lionel-nj:
Looking back to our list of closed PRs, it seems that https://github.com/MobilityData/gtfs-validator/issues/402 is related to this issue.

barbeau:
I've hit the blank line issue in both Tampa GTFS datasets I've tested (HART and USF), so I'd be in favor of treating this as a warning instead of an error.

lionel-nj:
I am not sure that we should treat that one as a warning instead of an error. My understanding is that this type of malformation can lead to generation of other notices. Since the behavior of the validator is not guaranted after the first error, treating this one as a warning would prevent users to debug their validation report (assuming `abort_on_error` was set to `true`).

However, this is just a suggestion that is totally up for change. @fabrice-v we could discuss that during our next meeting to be sure that we are all on the same page and that I am not missing some key element.

lionel-nj:
I understand also that the spec does not specify that CSV file should not contain blank lines, hence the treatment as warning you were suggesting @barbeau, right?

barbeau:
>I understand also that the spec does not specify that CSV file should not contain blank lines, hence the treatment as warning you were suggesting

Correct - to my knowledge having empty lines in a CSV file isn't against any CSV or GTFS rule, especially at the end of a file. So IMHO it seems harsh to consider this a failing error. I agree that malformed rows (e.g., that don't have enough columns based on the headers) should throw errors.

> Since the behavior of the validator is not guaranted after the first error, treating this one as a warning would prevent users to debug their validation report (assuming abort_on_error was set to true).

So the validator currently stops on warnings unless abort_on_error = false?

lionel-nj:
> So the validator currently stops on warnings unless abort_on_error = false?

The validator stops on the first error (not on the first warning) unless `abort_on_error = false`

fabrice-v:
> > I understand also that the spec does not specify that CSV file should not contain blank lines, hence the treatment as warning you were suggesting
> 
Relevant #317 , #213 
@barbeau Do the other datasets you have seen fail the validator pass the python google one? I guess they have some special magic or something, maybe @maximearmstrong will have some time to check that at some point.

Thks for the discussion, I'm still arguing the point the validator should reject those datasets

fabrice-v (changes requested):
Code hasn't been reviwed. Some issue after merging #469 

See the failing checks
image

barbeau:
>Do the other datasets you have seen fail the validator pass the python google one?

Yes, the HART dataset passes the Python validator without any errors (see below and attached)

image

[validation-results.zip](https://github.com/MobilityData/gtfs-validator/files/5476570/validation-results.zip)

Here's the HART GTFS zip file:

[google_transit_hart.zip](https://github.com/MobilityData/gtfs-validator/files/5476579/google_transit_hart.zip)

lionel-nj:
As discussed we relaunch jobs: https://github.com/MobilityData/gtfs-validator/runs/1343258317?check_suite_focus=true

fabrice-v (approved):
False alarm: CI runner had a bad network day. (Forgot to take a screen capture)
Issue rised when trying to persist the coverage report on GitHub infrastructure.

PR #478, opened 2020-11-05
Improve end to end workflow clarity

### LINKED ISSUE #473: [REQUEST] explicit long CLI options in workflow files
At present workflow files run the validator with short CLI option names. 

For the benefit of everyone and to provide easy to understand code, it is preferred to use the long CLI option names in the `end_to_end` workflow file:
- https://github.com/MobilityData/gtfs-validator/blob/f34f5ba5bf17f24a070d95c942d83206fa837c45/.github/workflows/end_to_end.yml#L47
- https://github.com/MobilityData/gtfs-validator/blob/f34f5ba5bf17f24a070d95c942d83206fa837c45/.github/workflows/end_to_end.yml#L49
- https://github.com/MobilityData/gtfs-validator/blob/f34f5ba5bf17f24a070d95c942d83206fa837c45/.github/workflows/end_to_end.yml#L51
- https://github.com/MobilityData/gtfs-validator/blob/f34f5ba5bf17f24a070d95c942d83206fa837c45/.github/workflows/end_to_end.yml#L53
- https://github.com/MobilityData/gtfs-validator/blob/f34f5ba5bf17f24a070d95c942d83206fa837c45/.github/workflows/end_to_end.yml#L55

lionel-nj:
closes #473 

**Summary:**

This PR provides support to explicit options long name instead of using options short name in end to end workflow.

**Expected behavior:** 

No change.

- ~[] Format the title like "Fix #<issue_number> - <short description of fix and changes>" (for example - "Fix #1111 - Check for null value before using field")~

### DIFF
--- a/.github/workflows/end_to_end.yml
+++ b/.github/workflows/end_to_end.yml
@@ -42,7 +42,7 @@ jobs:
           arguments: shadowJar
 
       #- name: Validate dataset from -- ACRONYM #<-- uncomment this line, replace ACRONYM by the name of the agency/publisher acronym
-      #  run: java -jar application/cli-app/build/libs/*.jar -url DATASET_PUBLIC_URL -input [[[ACRONYM]]].zip -extract pathToExtractedZip -output validationResultDirectory #<-- uncomment this line,
+      #  run: java -jar application/cli-app/build/libs/*.jar --url DATASET_PUBLIC_URL --input [[[ACRONYM]]].zip --extract pathToExtractedZipContent --output validationResultDirectory #<-- uncomment this line,
 #replace ACRONYM and [[[ACRONYM]]] by the agency/publisher acronym. Also replace DATASET_PUBLIC_URL by a public link to a GTFS Schedule zip archive
       - name: Validate dataset from -- SMART
         run: java -jar application/cli-app/build/libs/*.jar --url http://transitfeeds.com/p/sonoma-marin-area-rail-transit/1050/20200930/download --input smart.zip --extract input --output output --abort_on_error false
--- a/.github/workflows/end_to_end.yml
+++ b/.github/workflows/end_to_end.yml
@@ -42,7 +42,7 @@ jobs:
           arguments: shadowJar
 
       #- name: Validate dataset from -- ACRONYM #<-- uncomment this line, replace ACRONYM by the name of the agency/publisher acronym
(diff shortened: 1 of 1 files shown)


.github/workflows/end_to_end.yml
```diff
@@ -45,20 +45,20 @@ jobs:
       #  run: java -jar application/cli-app/build/libs/*.jar -u DATASET_PUBLIC_URL -i [[[ACRONYM]]].zip -e input -o output #<-- uncomment this line,
```
fabrice-v, line 45:
```suggestion
      #  run: java -jar application/cli-app/build/libs/*.jar -url DATASET_PUBLIC_URL -input [[[ACRONYM]]].zip -extract extractedZip -output validationResult #<-- uncomment this line,
```

fabrice-v (changes requested):
suggesting also updating the commented line as well as some renamings of folders created by the validator.

.github/workflows/end_to_end.yml
lionel-nj: Thanks for the heads up

.github/workflows/end_to_end.yml
```diff
@@ -42,23 +42,23 @@ jobs:
           arguments: shadowJar
 
       #- name: Validate dataset from -- ACRONYM #<-- uncomment this line, replace ACRONYM by the name of the agency/publisher acronym
-      #  run: java -jar application/cli-app/build/libs/*.jar -u DATASET_PUBLIC_URL -i [[[ACRONYM]]].zip -e input -o output #<-- uncomment this line,
+      #  run: java -jar application/cli-app/build/libs/*.jar -url DATASET_PUBLIC_URL -input [[[ACRONYM]]].zip -extract pathToExtractedZip -output validationResultDirectory #<-- uncomment this line,
```
maximearmstrong, line 45: Every argument is missing a dash (-). Ex. `-url` should be `--url`.

Also, does `--extract` refer to the zip file to be extracted? If yes, I would suggest modifying `pathToExtractedZip` to `pathToZipFile` or `pathToZipToExtract`.

maximearmstrong (changes requested):
Few modifications needed, but looks good other than that!

.github/workflows/end_to_end.yml
lionel-nj: Actually `--extract` is the path where the content of the zip file is extracted to. It could be `pathToExtractedZipContent`...

PR #480, opened 2020-11-05
 Implement solution to prevent worker from stopping on OutOfMemoryError

### LINKED ISSUE #474: [REQUEST] Prevent `end_to_ end` workflow from stopping in case of `OutOfMemoryException`
At present, when an OutOfMemoryException is raised, the validator ungracefully crashes.

Fail gracefully on OutOfMemoryException.

- [ ] investigate solutions to make the validator exit with a special code, so that end_to_end workflow does not stop on said exception.
 - [ ] make sure all other artifacts can be collected at the end of the workflow

lionel-nj:
closes #474

**Summary:**

This PR provides support to prevent worker from stop in end_to_end workflow when encountering OutOfMemoryException

**Expected behavior:**

Worker should skip task and continue to next job

- [ ] Run the unit tests with gradle test to make sure you didn't break anything
- ~[ ] Format the title like "Fix #<issue_number> - " (for example - "Fix #1111 - Check for null value before using field")~

### DIFF
--- a/.github/workflows/end_to_end.yml
+++ b/.github/workflows/end_to_end.yml
@@ -2,7 +2,7 @@ name: End to end
 
 on:
   push:
-    branches: [ master, transport-agency-name] #<-- replace transport-agency-name by the name of the agency/publisher
+    branches: [ master, prevent-worker-from-stoppping-on-error] #<-- replace transport-agency-name by the name of the agency/publisher
 # todo: implement #474
 jobs:
   run-on-data:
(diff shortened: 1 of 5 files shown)


lionel-nj:
As https://github.com/MobilityData/gtfs-validator/pull/480/checks?check_run_id=1363946960 demonstrates: failures in jobs implemented in `end_to_end` workflow no longer block the execution of the subsequent jobs ✨

f8full:
We'll do that differently (because this would also eat other exceptions raised by the validator and make failure hard to clearly identify on CI. We shall collect the exception ourselves, emit an error level log message (like we dont when encountering the first error when `abort_on_error` is set) with an explicit "dataset too big" or something.
Let's discuss it on Monday, @lionel-nj

fabrice-v:
Actually I remembered we now measure the archive size to generate the info notice data. We shall set a hard limit regarding the input zip size, that's an option. It would work reasonably well paired with aborting on error (as a significant part of memory consumption comes from Notices the validator creates in memory to construct the report), not just the original gtfs data
Another option would be for the validator to measure memory consumptiom while executing and stopping itself when getting over a user configurable threashold (in percents of total memory available).

fabrice-v:
- [ ] super extra nice: on memory too high, dump the content of the result repo to a file (maybe another file for starter) and flush it before resuming. Maybe that could generate an info notice explaining it happened and that there are multiple files or something. Or fuse everything at the end in one unified report as nw.

lionel-nj:
As discussed, this is actually not directly related to having too many notices in memory: here the validator crashes when it there are only 2 warning notices in memory. the problem lies in the fact that the `stop_times.txt` file is too large to be entirely processed by [`ProcessParsedStopTime`](https://github.com/MobilityData/gtfs-validator/blob/a1e19043809888a1bf707d18f44e3aa3acfc535d/usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ProcessParsedStopTime.java#L60). 

A few solutions are considered:
- [ ] measure heap space, and gracefully stop the validator when a certain % is occupied. Consider generate a notice and log the following information: "Running low on memory, one could restart the validator with `-Xms`(value to be bumped) option
   1. This might introduce variability from one user to another one
   
- [ ] measure heap space, and activate the garbage collector when a certain % is occupied
   1. The behavior of the garbage collector is not fully predictable

PR #487, opened 2020-11-18
Update .gitignore file

aababilov:
One of the main changes is to allow git to store files called `input`. There will be a Java package called `org.mobilitydata.gtfsvalidator.input`, so we need to store its directory `input`.


.gitignore
```diff
@@ -1,33 +1,75 @@
...
+# since they will be recreated, and may cause churn.  Uncomment if using
+# auto-import.
+.idea/artifacts
+.idea/compiler.xml
+.idea/jarRepositories.xml
+.idea/modules.xml
+.idea/*.iml
+.idea/modules
 *.iml
-.idea/
```
fabrice-v, line 2: OMG how do you know all those paths to exclude or not? I feel I did this a million times and this is the best I could come up with :/

fabrice-v (approved):
Awesome thks for all those path. I do not know if they have them there https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore

LGTM!

.gitignore
barbeau: In my experience with IntelliJ I always ignore the entire `.idea/` directory - otherwise other files end up popping up in version control.

.gitignore
```diff
@@ -1,33 +1,75 @@
...
+
+# Mongo Explorer plugin
+.idea/**/mongoSettings.xml
+
+# File-based project format
+*.iws
+
+# IntelliJ
 out/
-output/
```
barbeau, line 6: Should `output` still be ignored? IIRC that's the cache where txt files got written after extraction - although that could certainly change with the new architecture.

barbeau (commented):
Thanks @aababilov! A few comments in-line and below.

>There will be a Java package called org.mobilitydata.gtfsvalidator.input, so we need to store its directory input.

I'm not sure I understand this. What information is in the `input` directory that needs to be versioned?

.gitignore
aababilov: > I'm not sure I understand this. What information is in the `input` directory that needs to be versioned?
This directory is called `core/src/main/java/org/mobilitydata/gtfsvalidator/input/` and it contains Java files.

The original .gitignore file says that any directory called `input` should be ignored by git, even if it is a nested directory. This way I have lost all the Java files that I wrote and I was very happy to write them again :) So, let's not put such common names to .gitignore - we risk to discard hours of someone's work.

If we have a `ouptut` directory at the top level of the repo, we should write `/output` instead of just `output`. However, the new annotation-based validator will not create `output` directory.

aababilov:
> Awesome thks for all those path. I do not know if they have them there https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore
> 
> LGTM!

My pleasure!

.gitignore
aababilov: Oh, I just decided to find the best practices for .gitignore!

I would agree if you ask me to remove the whole `.idea` from version control. Right now, I have only two files under git control (misc.xml and vcs.xml) and both look to be useless.

Let me do this: I can first ignore the whole `.idea` and we can change that later in another pull request.

My primary goal is to allow us having the `input` subdirectory at `core/src/main/java/org/mobilitydata/gtfsvalidator/input/`

aababilov:
It's either me who is really wrong or there is a GitHub bug, but I cannot add a commit to this pull request.

Here is my [branch `gitignore`](https://www.site-shot.com/D8K1RinzEeupcQJCrBEABA) where you see a new commit called "Ignore the whole .idea directory..."

But this commit is not present in the [current pull request](https://www.site-shot.com/bhFusinzEeuoOAJCrBEABA).

aababilov:
My last commit is now present in the pull request. I guess that GitHub was waiting for tests to pass.

.gitignore
barbeau: @aababilov Yikes! Not good. I agree with you - so this can stay as-is.

.gitignore
barbeau: 👍

barbeau:
LGTM, thanks @aababilov!

aababilov:
Thanks for the review!

PR #497, opened 2020-12-01
ci: Check if PR titles are conventional

lionel-nj:
**Summary:**

This PR provides support to check if PR titles are conventional

**Expected behavior:** 

**Conventional commit message/PR title in short has this below format**
```
<type>[optional scope]: <description>

[optional body]

[optional footer]
```

**Examples for valid PR titles:**

> fix: Correct typo.
> feat: Add support for Node 12.
> refactor!: Drop support for Node 6.

**Conventional commit specification:** https://www.conventionalcommits.org/en/v1.0.0/

- [ ] ❌ Non conventional PR titles should not pass (`"check if PR titles are conventional"`)

- [x] Conventional PR titles should pass (`"feat: check if PR titles are conventional"`)

Capture d’écran, le 2020-12-02 à 09 04 38


barbeau:
@lionel-nj Instead of using this 3rd party action, can you just configure the Probot here to do the checks? https://github.com/apps/semantic-pull-requests

It's based on this tool from a GitHub employee:
https://github.com/zeke/semantic-pull-requests

lionel-nj:
Hi @barbeau! looking back at #497, I notice that even with a configuration file, the bot seems to validate PRs where the title is not conventional if at least one of the commit is conventional (that is why I switched to a third party GitHub action that seems to strictly enforce conventional PR title regardless of the commit messages). To my understanding this is not the behavior that we would like, but I might be wrong. Could you confirm this please?

barbeau:
Correct, the behavior we want is that the PR title should always be semantic, as that will be the default squashed commit title when we merge a PR. Looks like this is configurable with the bot, though, via `semantic.yml` in the `.github` directory? https://github.com/zeke/semantic-pull-requests#configuration

lionel-nj:
> Correct, the behavior we want is that the PR title should always be semantic, as that will be the default squashed commit title when we merge a PR. Looks like this is configurable with the bot, though, via `semantic.yml` in the `.github` directory? https://github.com/zeke/semantic-pull-requests#configuration

https://github.com/MobilityData/gtfs-validator/pull/497/commits/94026ebedba0efc2ff070f373e2e5884a83ec938 moves the config file to `.github` directory. As you can see: https://github.com/MobilityData/gtfs-validator/blob/94026ebedba0efc2ff070f373e2e5884a83ec938/.github/semantic.yml#L6
 the config file is theoretically set to only validate PRs title. However, this PR title ("check if titles are conventional") is not conventional and but bot still considers it as valid. @barbeau

barbeau:
@lionel-nj Let's try merging with the `titleOnly` config - the config file may need to be on the master branch to take effect. Then we can open another PR to test, and if that doesn't work, we can delete the config.

barbeau (approved):

lionel-nj:
As you suggested, this PR had to be merged to take effect. Thank you @barbeau!
https://github.com/MobilityData/gtfs-validator/pull/498

PR #502, opened 2020-12-03
test: Add unit tests for empty values in CSV

aababilov:
There is no difference between the following lines in CSV:
1) f1,"",f3
2) f1,,f3

Univocity parser returns {"f1", null, "f3"} for both rows. There is no way to tell between a missing value and an explicit empty string "".

We covered this situation with a unit test and also added an explicit code that always returns null instead of "" in case we change the CSV parser.


core/src/main/java/org/mobilitydata/gtfsvalidator/parsing/RowParser.java
```diff
@@ -416,7 +416,7 @@ public ValueParser(String formatName) {
 
         final T parseField(int columnIndex, boolean required) {
             String s = asString(columnIndex, required);
-            if (s == null || s.isEmpty()) {
```
f8full, line 419: Could Guava [isNullOrEmpty](https://guava.dev/releases/27.1-jre/api/docs/com/google/common/base/Strings.html#isNullOrEmpty-java.lang.String-) be used?

core/src/main/java/org/mobilitydata/gtfsvalidator/parsing/CsvRow.java
```diff
@@ -85,6 +85,12 @@ public String asString(int columnIndex) {
         if (columnIndex < 0 || columnIndex >= columnValues.length) {
             return null;
         }
-        return columnValues[columnIndex];
+        String s = columnValues[columnIndex];
+        // Univocity CSV parser already returns null for no explicit value and for an explicit empty string "".
+        // Here we just want to be sure that we always return null and never "".
+        if (s == null || s.isEmpty()) {
```
f8full, line 91: Could Guava [isNullOrEmpty](https://guava.dev/releases/27.1-jre/api/docs/com/google/common/base/Strings.html#isNullOrEmpty-java.lang.String-) be used?

core/src/main/java/org/mobilitydata/gtfsvalidator/parsing/RowParser.java
aababilov: We never have an empty string anymore, so I just compare it with `null`.

core/src/main/java/org/mobilitydata/gtfsvalidator/parsing/CsvRow.java
aababilov: Done. Nice!

barbeau (approved):
Thanks! 👍

PR #503, opened 2020-12-03
fix: Return builder instance from setCsvRowNumber() method

aababilov:
This is the common behaviour for all setters that allows to chain them:

new GtfsStop.Builder().setCsvRowNumber(1).setStopId("s1").build()


barbeau (approved):
Thanks @aababilov! LGTM.

PR #505, opened 2020-12-03
refactor: Do not use import wildcards

aababilov:


barbeau:
Closing and re-opening so hopefully the semantic bot updates...

lionel-nj:
The list of commit types supported by the bot is available [here](https://github.com/commitizen/conventional-commit-types/blob/master/index.json).
I believe that the closest to "cleanup" would be "chore". Otherwise, we can always configure the bot the accept "cleanup".

barbeau:
@lionel-nj Ah, yes, you're right. I'll change this to "chore:".

barbeau (commented):
@aababilov Just curious what your motivation is in changing this?

Something we should consider - IntelliJ, which is likely the most common IDE that developers will use with this project, by default lists imports individually until there are 5 from the same package, and at that point it will convert them to wildcards:
https://www.jetbrains.com/help/idea/creating-and-optimizing-imports.html#disable-wildcard-imports

I think Eclipse has a similar default, but it's been a while since I've used it.

It looks like all the imports you've expanded here fit this category (5 per package). So if someone modifies one of these classes that impacts the imports, their IntelliJ will automatically collapse them into a wildcard again. We could include a custom code template in the project to change this, but then we need to get all contributors to install the template in their IDE. And from my past experience few developers do this before contributing code.

So unless there is a clear benefit avoiding wildcards I'd be inclined to stick with the default IntelliJ behavior to avoid shuffling the imports in future changes and having to manage code templates.

aababilov:
I started integrating the new validator code at Google. Unfortunately, wildcard imports are forbidden by [Google Java style guide](https://google.github.io/styleguide/javaguide.html#s3.3.1-wildcard-imports). To be honest, sometimes I personally disagree with Google style guides, but I am forced to follow them.

Speaking of IDE: I configured IDEA to [never use wildcard imports](https://stackoverflow.com/questions/3348816/intellij-never-use-wildcard-imports), so it will not collapse them. Actually, this commit was generated by IDEA automatically.

aababilov:
> @lionel-nj Ah, yes, you're right. I'll change this to "chore:".

I think that `refactor` can be even more accurate in this case. Thanks for the link to the commit types list!

barbeau (approved):
>Unfortunately, wildcard imports are forbidden by Google Java style guide. To be honest, sometimes I personally disagree with Google style guides, but I am forced to follow them.

I was wondering that. Ok - I propose that we merge this PR and adopt the [Google code style template for IntelliJ](https://github.com/google/styleguide/blob/gh-pages/intellij-java-google-style.xml) as the official code template for the project.

PR #540, opened 2020-12-03
refactor: Fix style problems detected by Google linter

aababilov:
* Pattern is a deeply immutable type, so this field should be in UPPER_SNAKE_CASE (HHMMCC_PATTER).
* Methods and non-static variables should be named in lowerCamelCase.
* Using static imports for types is unnecessary
* Field name is CONSTANT_CASE, but field is not static and final

### DIFF
--- a/processor/src/main/java/org/mobilitydata/gtfsvalidator/processor/EnumGenerator.java
+++ b/processor/src/main/java/org/mobilitydata/gtfsvalidator/processor/EnumGenerator.java
@@ -24,6 +24,7 @@ import org.mobilitydata.gtfsvalidator.table.GtfsEnum;
 
 import javax.lang.model.element.Modifier;
 
+import static java.lang.Math.min;
 import static org.mobilitydata.gtfsvalidator.processor.GtfsEntityClasses.TABLE_PACKAGE_NAME;
 
 /**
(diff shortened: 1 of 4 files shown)


processor/src/main/java/org/mobilitydata/gtfsvalidator/processor/EnumGenerator.java
```diff
@@ -42,6 +43,15 @@ public static String createEnumName(String interfaceName) {
         return interfaceName.substring(0, interfaceName.length() - ENUM_SUFFIX.length());
     }
 
+    private static MethodSpec getNumberMethod() {
```
fabrice-v, line 46: this piece of code was just moved around?

fabrice-v:
Would it be possible to add a screenshot and/or some kind of output from the linter as to make it clearer why the changes have to happen?

processor/src/main/java/org/mobilitydata/gtfsvalidator/processor/EnumGenerator.java
aababilov: No, it became static. It was moved by IDEA automatically because it rearranges static methods.

aababilov:
> Would it be possible to add a screenshot and/or some kind of output from the linter as to make it clearer why the changes have to happen?

With pleasure! Done.

processor/src/main/java/org/mobilitydata/gtfsvalidator/processor/EnumGenerator.java
fabrice-v: oh, alright. Thks for the details!

fabrice-v (approved):

barbeau (approved):

Issue #398, opened 2020-09-30
[BUG] Unzipping Error


mcplanner:
**Describe the bug**
The validator was unable to process a feed and gave an error.

**To Reproduce**
Run the validator on the following dataset: http://transitfeeds.com/p/orange-county-transportation-authority/197/latest

**Expected behavior**
No notice should be generated in this case.

**Witnessed behavior**
Upon running the validator, I received this message before the operation ended.
```
[INFO ] 2020-09-30 13:06:31.639 [main] Main - Exporting validation repo content:[
Notice{filename='/Users/ruthmiller/Downloads/Validator/OCTA.zip', level='ERROR', code='8', title='Unzipping error', description='An error occurred while trying to unzip archive: /Users/ruthmiller/Downloads/Validator/OCTA.zip', extra='{}'}, 
Notice{filename='/Users/ruthmiller/Downloads/Validator/OCTA.zip', level='ERROR', code='8', title='Unzipping error', description='An error occurred while trying to unzip archive: /Users/ruthmiller/Downloads/Validator/OCTA.zip', extra='{}'}, 
Notice{filename='/Users/ruthmiller/Downloads/Validator/OCTA.zip', level='ERROR', code='8', title='Unzipping error', description='An error occurred while trying to unzip archive: /Users/ruthmiller/Downloads/Validator/OCTA.zip', extra='{}'}, 
Notice{filename='/Users/ruthmiller/Downloads/Validator/OCTA.zip', level='ERROR', code='8', title='Unzipping error', description='An error occurred while trying to unzip archive: /Users/ruthmiller/Downloads/Validator/OCTA.zip', extra='{}'}, 
Notice{filename='/Users/ruthmiller/Downloads/Validator/OCTA.zip', level='ERROR', code='8', title='Unzipping error', description='An error occurred while trying to unzip archive: /Users/ruthmiller/Downloads/Validator/OCTA.zip', extra='{}'}, 
Notice{filename='/Users/ruthmiller/Downloads/Validator/OCTA.zip', level='ERROR', code='8', title='Unzipping error', description='An error occurred while trying to unzip archive: /Users/ruthmiller/Downloads/Validator/OCTA.zip', extra='{}'}, 
[...]

PR #541, opened 2020-12-04
feat: Add command-line option to download GTFS file

lionel-nj:
closes #541 
**Summary:**

This PR provides support to enable download of archive from network

**Expected behavior:** 

- [x] add CLI parameter `-u` / `--url` 
- [x] validate that `-u` and `-i` are not provided at the same time
- [x] validate that exactly one of the following CLI paramters is provided:
  - `-i`
  - `-u`
- [x] download archive from network
- [x] log download information 

101217205-5a4d4480-3657-11eb-9700-e854d6cf952d

- [x] define path of the location where archive is downloaded: `-s` / `--storage_directory`
- [x] implement unit tests
  - [x] draft addiitional unit tests
- [x] add javadoc
- [x] update `.md` files to match the latest development
- [x] re-integrate end-to-end workflow
- [x] check license headers


core/src/main/java/org/mobilitydata/gtfsvalidator/input/GtfsInput.java
```diff
@@ -36,6 +36,7 @@
 public interface GtfsInput {
     int HTTP_TEMP_REDIRECT = 307;
     int HTTP_PERM_REDIRECT = 308;
+
```
fabrice-v, line 39: 👁️

README.md
```diff
@@ -2,6 +2,7 @@
 
 A GTFS Schedule (static) [General Transit Feed Specification (GTFS)](https://gtfs.mobilitydata.org/spec/gtfs-schedule) feed validator
 
+
```
f8full, line 5:
```suggestion
```

Empty line was added as a touch to trigger end to end workflow

core/src/main/java/org/mobilitydata/gtfsvalidator/input/GtfsInput.java
```diff
@@ -44,6 +54,78 @@ static GtfsInput createFromPath(String path) throws IOException {
...
+     * @param sourceUrl
+     * @param targetPathAsString
+     * @return
+     * @throws IOException
+     */
+    static GtfsInput createFromUrl(URL sourceUrl, String targetPathAsString) throws IOException {
+        InputStream inputStream;
+
+        try {
+            final HttpURLConnection httpConnection = (HttpURLConnection) sourceUrl.openConnection();
```
aababilov, line 69: How about using a higher-level Apache library for HTTP?

It has `org.apache.http.client.methods.HttpGet` class that we can use here. The default redirect strategy handles redirects for GET and HEAD, so that is what we need.

main/src/main/java/org/mobilitydata/gtfsvalidator/cli/Main.java
```diff
@@ -54,8 +62,21 @@ public static void main(String[] argv) {
...
+                        feedName,
+                        validatorLoader,
+                        noticeContainer);
+            } else {
+                feedContainer = feedLoader.loadAndValidate(
+                        GtfsInput.createFromPath(args.input),
+                        feedName,
+                        validatorLoader,
+                        noticeContainer);
+            }
```
aababilov, line 79: Let's write it shorter.

GtfsInput gtfsInput = args.input == null ? GtfsInput.createFromUrl(
                                new URL(args.url),
                                args.storageDirectory)
  : GtfsInput.createFromPath(args.input);

feedContainer = feedLoader.loadAndValidate(
                        gtfsInput,
                        feedName,
                        validatorLoader,
                        noticeContainer);

core/src/main/java/org/mobilitydata/gtfsvalidator/input/GtfsInput.java
```diff
@@ -44,6 +54,78 @@ static GtfsInput createFromPath(String path) throws IOException {
...
+     * @param toCleanOrCreate
+     * @return
+     */
+    static Path createPath(final String toCleanOrCreate) {
+        // to empty any already existing directory
+        Path pathToCleanOrCreate = Paths.get(toCleanOrCreate);
+        if (Files.exists(pathToCleanOrCreate)) {
+            try {
+                //noinspection ResultOfMethodCallIgnored -- we ignore if deletion went well or not
+                Files.walk(pathToCleanOrCreate).sorted(Comparator.reverseOrder()).map(Path::toFile).forEach(File::delete);
```
aababilov, line 104: I suggest org.apache.commons.io.FileUtils.deleteDirectory that deletes directories recursively.

core/src/main/java/org/mobilitydata/gtfsvalidator/input/GtfsInput.java
```diff
@@ -44,6 +54,78 @@ static GtfsInput createFromPath(String path) throws IOException {
...
+                Files.createDirectory(pathToCleanOrCreate);
+            } catch (IOException e) {
+                e.printStackTrace();
```
aababilov, line 107: Should we propagate all exceptions caught in this function? Is it really safe to silently ignore them all?

main/src/main/java/org/mobilitydata/gtfsvalidator/cli/Main.java
f8full, line 79: The following is just a try at a multiline suggestion in the GitHub interface with @aababilov [comment ](https://github.com/MobilityData/gtfs-validator/pull/541/files/2cf8c83c24bad5cae84d73c204ccfb369ba14ea1#r537161917)'s content

```suggestion
            GtfsInput gtfsInput = args.input == null ? GtfsInput.createFromUrl(
                                new URL(args.url),
                                args.storageDirectory)
  : GtfsInput.createFromPath(args.input);

feedContainer = feedLoader.loadAndValidate(
                        gtfsInput,
                        feedName,
                        validatorLoader,
                        noticeContainer);
```

f8full: I'm not sure about our stance on the ternary operator. @LeoFrachet ?

f8full:
```suggestion
            GtfsInput gtfsInput = args.input == null ? GtfsInput.createFromUrl(
                                new URL(args.url),
                                args.storageDirectory)
  : GtfsInput.createFromPath(args.input);

feedContainer = feedLoader.loadAndValidate(
                        gtfsInput,
                        feedName,
                        validatorLoader,
                        noticeContainer);
```

core/src/main/java/org/mobilitydata/gtfsvalidator/input/GtfsInput.java
f8full: 🙏

core/src/main/java/org/mobilitydata/gtfsvalidator/input/GtfsInput.java
f8full: Good catch. Most probably it isn't.

main/src/main/java/org/mobilitydata/gtfsvalidator/cli/Main.java
LeoFrachet: When I was software developer, I was open to use the ternary operator only if it fits in one line of code, aka if you could see the whole thing with just one glance, and we weren't allowing it when it was spanning on multiple lines.

lionel-nj: Thanks @LeoFrachet!

core/src/main/java/org/mobilitydata/gtfsvalidator/input/GtfsInput.java
lionel-nj: I actually realize that this piece of code needs to be re-worked: we just have to create/delete the `.zip` file that will contain the zipped dataset after download. If you agree I will use `org.apache.commons.io.FileUtils.deleteQuietly` which deletes files instead of directories.

core/src/main/java/org/mobilitydata/gtfsvalidator/input/GtfsInput.java
lionel-nj: Good point, it has been modified.

lionel-nj: @aababilov do you htink tht we should adopt a similar approach to what was done in `v1.4.0` which consists in: generating an notice that will be added to the `NoticeContainer`, after addition of said notice the validator would abort and export the validation report?

main/src/main/java/org/mobilitydata/gtfsvalidator/cli/Main.java
fabrice-v: I would agree with @LeoFrachet on that one. For readability, accessibility and maintenance.

main/src/main/java/org/mobilitydata/gtfsvalidator/cli/ValidateCliParameters.java
```diff
@@ -0,0 +1,28 @@
...
+
+    ValidateCliParameters(Arguments args) {
+        this.args = args;
+    }
+
+    /**
+     * Validate requirement on CLI parameters
+     * @throws IOException
+     */
+    public void validate() throws IOException {
```
barbeau, line 19: IMHO this really isn't an appropriate use of `IOException`. The validate method is intended to validate the input, and should return a value with the result of validation instead of throwing an exception to indicate the result of validate.

main/src/test/java/org/mobilitydata/gtfsvalidator/cli/ValidateCliParametersTest.java
```diff
@@ -0,0 +1,19 @@
+package org.mobilitydata.gtfsvalidator.cli;
+
+import org.junit.Test;
+
+public class ValidateCliParametersTest {
+
+    @Test
+    public void provideUrlAndInputCliParametersShouldThrowException() {
```
barbeau, line 8: For consistency should be `{}`, but I assume this will be populated :).

barbeau (commented):
@lionel-nj A few thoughts in-line.

core/build.gradle
```diff
@@ -12,6 +12,8 @@ repositories {
 }
 
 dependencies {
+    compile group: 'commons-io', name: 'commons-io', version: '2.8.0'
```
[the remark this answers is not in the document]
fabrice-v: shall we also extract the version numbers?
https://github.com/f8full/herdr/blob/5470f3a67280f447059be0cf78b4edede7161a40/app/build.gradle.kts#L149

core/src/test/java/org/mobilitydata/gtfsvalidator/input/GtfsInputTest.java
```diff
@@ -66,4 +69,24 @@ public void zipInput() throws IOException {
...
+        File toDelete = new File("storage");
+        assertTrue(toDelete.delete());
+
+    }
+
+    @Test
+    public void createFromInvalidUrlShouldThrowException() {
+        assertThrows(
+                IOException.class, () -> GtfsInput.createFromUrl(
+                        new URL("https://octa.net/current/gowkdjhfiouwehfogle_transit.zip"),
```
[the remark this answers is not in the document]
fabrice-v: good point, care to ellaborate a little? :)

core/src/test/java/org/mobilitydata/gtfsvalidator/input/GtfsInputTest.java
```diff
@@ -66,4 +69,24 @@ public void zipInput() throws IOException {
         GtfsInput gtfsInput = GtfsInput.createFromPath(zipFile.getAbsolutePath());
         assertThat(gtfsInput.getFilenames()).containsExactly("stops.txt");
     }
+
+    @Test
+    public void createFromValidUrlShouldNotThrowException() throws IOException, URISyntaxException, InterruptedException {
+        GtfsInput underTest = GtfsInput.createFromUrl(
+                new URL("https://octa.net/current/google_transit.zip"),
```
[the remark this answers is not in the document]
fabrice-v: Sure. Why though ❔

main/src/main/java/org/mobilitydata/gtfsvalidator/cli/ValidateCliParameters.java
fabrice-v: Is that always true? I've seen some multithreading patterns were the runnable would either return a value OR an exception. Some always return an exception with a Result object in it, containing either a result or an error.
What would be the reason to not do something similar here and stay bare?

core/build.gradle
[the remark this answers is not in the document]
lionel-nj: Good to know, thanks!

core/src/main/java/org/mobilitydata/gtfsvalidator/input/GtfsInput.java
```diff
@@ -44,6 +52,70 @@ static GtfsInput createFromPath(String path) throws IOException {
         return new GtfsZipFileInput(file);
     }
 
+    /**
+     * Creates an specific GtfsInput to read data from the given URL.
+     *
+     * @param sourceUrl
+     * @param targetPathAsString
```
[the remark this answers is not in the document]
lionel-nj: Indeed, update of Javadocs is imminent.

core/build.gradle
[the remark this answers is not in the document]
barbeau: >shall we also extract the version numbers?

I'm not a big fan of that unless version numbers must match across multiple dependencies from the same developer.

core/src/test/java/org/mobilitydata/gtfsvalidator/input/GtfsInputTest.java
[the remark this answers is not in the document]
lionel-nj: I picked this one: https://github.com/MobilityData/gtfs-validator/blob/v1.4.0/usecase/src/test/resources/valid_zip_sample.zip

main/src/test/java/org/mobilitydata/gtfsvalidator/cli/ValidateCliParametersTest.java
lionel-nj: Indeed, tests are coming soon :)

main/src/main/java/org/mobilitydata/gtfsvalidator/cli/ValidateCliParameters.java
barbeau: One general reason to avoid it is that throwing exceptions is expensive - see https://www.baeldung.com/java-exceptions-performance.

core/src/test/java/org/mobilitydata/gtfsvalidator/input/GtfsInputTest.java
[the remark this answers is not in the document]
barbeau: >Sure. Why though ❔

Because OCTA could change their URL and break the test. If it's under our control it's less likely to be broken randomly.

>I picked this one: https://github.com/MobilityData/gtfs-validator/blob/v1.4.0/usecase/src/test/resources/valid_zip_sample.zip

👍  Note you need to use the "raw" version of the URL though:
https://github.com/MobilityData/gtfs-validator/raw/v1.4.0/usecase/src/test/resources/valid_zip_sample.zip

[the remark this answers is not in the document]
barbeau: You can get the "raw" version by right-clicking on "Download" button in the GitHub UI and copying the link.

core/src/test/java/org/mobilitydata/gtfsvalidator/input/GtfsInputTest.java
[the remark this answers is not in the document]
barbeau: >good point, care to ellaborate a little? :)

Generally you want to use URLs under your control in tests so they don't break randomly if the owner of the URL decides to change something. Granted, in this case it's unlikely that OCTA decides to publish a valid GTFS named `gowkdjhfiouwehfogle_transit.zip` that would break this particular test, but for consistency with other tests that depend on live URLs IMHO it would be better if they are all under domains that we have control of.

main/src/main/java/org/mobilitydata/gtfsvalidator/cli/ValidateCliParameters.java
barbeau: To elaborate further, IMHO exceptions should be used for unexpected things that happen during program execution, not for expected results. In this case, we have clear expectations for what is valid and what is not, and as a result it should be handled in normal control flow and not an exception.

I can't speak to whether this is always true - there are probably exceptions :).

fabrice-v: thks for the clarifications. I agree with you then :)

core/build.gradle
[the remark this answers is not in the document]
f8full: ah. That is super valid, thks!

For reference :)
https://github.com/f8full/herdr/blob/master/buildSrc/src/main/java/Dependencies.kt

core/src/test/java/org/mobilitydata/gtfsvalidator/input/GtfsInputTest.java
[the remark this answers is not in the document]
lionel-nj: Thanks!

core/src/main/java/org/mobilitydata/gtfsvalidator/input/GtfsInput.java
```diff
@@ -44,6 +56,62 @@ static GtfsInput createFromPath(String path) throws IOException {
...
+     * @throws URISyntaxException  if URL is malformed
+     * @throws InterruptedException when a thread is waiting, sleeping, or otherwise occupied, and the thread is
+     * interrupted, either before or during the activity
+     */
+    static GtfsInput createFromUrl(URL sourceUrl, String targetPathAsString) throws IOException, URISyntaxException,
+            InterruptedException {
+        CloseableHttpClient httpClient = HttpClients.createDefault();
+        HttpGet httpGet = new HttpGet(sourceUrl.toURI());
+        CloseableHttpResponse httpResponse = httpClient.execute(httpGet);
+        InputStream inputStream = httpResponse.getEntity().getContent();
```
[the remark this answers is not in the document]
barbeau: Looking at https://hc.apache.org/httpcomponents-client-ga/tutorial/pdf/httpclient-tutorial.pdf, in section 1.1.5 it says:

>In order to ensure proper release of system resources one must close either the content stream associated
with the entity or the response itself
>...
>The difference between closing the content stream and closing the response is that the former will
attempt to keep the underlying connection alive by consuming the entity content while the latter
immediately shuts down and discards the connection.
>
>Please note that the HttpEntity#writeTo(OutputStream) method is also required to ensure proper
release of system resources once the entity has been fully written out. If this method obtains an instance
of java.io.InputStream by calling HttpEntity#getContent(), it is also expected to close the stream
in a finally clause.

It seems safer to use the `HttpEntity.writeTo()` method instead of `getContent()` as the `OutputStream` is managed by the `HttpEntity` (also please be sure to use a BufferedOutputStream). I believe that would also allow us to avoid `try/catch/finally` block within this method if you want to throw the exception to the calling code.

barbeau (commented):

core/build.gradle
```diff
@@ -12,6 +12,8 @@ repositories {
 }
 
 dependencies {
+    implementation 'commons-io:commons-io:2.8.0'
+    implementation 'org.apache.httpcomponents:httpclient:4.5.13'
```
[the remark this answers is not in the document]
barbeau: Hmm, looks like 4.5.13 is the latest in Maven Central but docs for 5.0 are published:
https://hc.apache.org/httpcomponents-client-5.0.x/index.html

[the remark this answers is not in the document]
lionel-nj: I based my research on Mavel Central which only pointed to `4.5.13` . Will update to `5.0`. Thanks!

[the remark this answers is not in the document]
barbeau: Ah, yes, looks like the publisher name changed for v5:

```groovy
implementation 'org.apache.httpcomponents.client5:httpclient5:5.0.3'
```

core/src/main/java/org/mobilitydata/gtfsvalidator/input/GtfsInput.java
```diff
@@ -44,6 +56,62 @@ static GtfsInput createFromPath(String path) throws IOException {
...
+     * @throws IOException  if no file could not be found at the specified location
+     * @throws URISyntaxException  if URL is malformed
+     * @throws InterruptedException when a thread is waiting, sleeping, or otherwise occupied, and the thread is
+     * interrupted, either before or during the activity
+     */
+    static GtfsInput createFromUrl(URL sourceUrl, String targetPathAsString) throws IOException, URISyntaxException,
+            InterruptedException {
+        CloseableHttpClient httpClient = HttpClients.createDefault();
+        HttpGet httpGet = new HttpGet(sourceUrl.toURI());
+        CloseableHttpResponse httpResponse = httpClient.execute(httpGet);
```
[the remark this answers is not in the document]
lionel-nj: It does, the execution of the end-to-end workflow shows it: https://github.com/MobilityData/gtfs-validator/runs/1518239795?check_suite_focus=true
`Validate dataset from issue #398 [...]`. 
But yes - I agree to adding tests to make sure of the behavior! :)

core/src/main/java/org/mobilitydata/gtfsvalidator/input/GtfsInput.java
aababilov: This is a great question. I would suggest to generate a separate report for system errors (IO error, out of memory, exception raised in a validation etc). 

Rationale.
------------
1. Program that runs the validator needs to know why it crashed and a JSON report is more convenient than parsing stdout.
2. There may be bugs in individual validations and we need information about them.
3. We should not leak information about system errors to clients, so system errors should not be placed to the same report as validation errors.

Implementation
--------------------
We can have a special subclass of `Notice` (`SystemError`). NoticeContainer called on SystemError should keep it in a separate place.

main/src/main/java/org/mobilitydata/gtfsvalidator/cli/Main.java
aababilov: I think that there was a misunderstanding. My point was to avoid the code duplication, not to necessarily introduce ternary operation. If you do not like ?:, you can write it this way:

```
GtfsInput gtfsInput;
if  (args.input == null) {
   gtfsInput = GtfsInput.createFromUrl(
                                new URL(args.url),
                                args.storageDirectory);
} else {
 gtfsInput =  GtfsInput.createFromPath(args.input);
}
feedContainer = feedLoader.loadAndValidate(
                        gtfsInput,
                        feedName,
                        validatorLoader,
                        noticeContainer);

File: core/src/main/java/org/mobilitydata/gtfsvalidator/input/GtfsInput.java (hunk 6 above)
Comment (lionel-nj) [reply core/src/main/java/org/mobilitydata/gtfsvalidator/input/GtfsInput.java]: Opening an issue on this subject so that we can proceed to implementation in a separate PR.

File: core/build.gradle [hunk 16]
```diff
@@ -12,13 +28,17 @@ repositories {
...
+    implementation 'org.apache.httpcomponents.client5:httpclient5:5.0.3'
+    implementation 'commons-io:commons-io:2.8.0'
     implementation 'com.univocity:univocity-parsers:2.9.0'
     implementation 'com.fasterxml.jackson.core:jackson-databind:2.10.0'
     implementation 'com.google.guava:guava:29.0-jre'
     implementation 'commons-validator:commons-validator:1.6'
     implementation 'com.googlecode.libphonenumber:libphonenumber:8.12.13'
     testImplementation group: 'junit', name: 'junit', version: '4.13'
     testImplementation "com.google.truth:truth:1.0.1"
+    testImplementation 'org.hamcrest:hamcrest:2.2'
```
barbeau, line 41: What is hamcrest used for? I don't see another reference to it.

lionel-nj: It was used for matchers, but should be removed since I ended up not using it. Will remove.

barbeau (approved):
Thanks @lionel-nj, LGTM!

core/src/main/java/org/mobilitydata/gtfsvalidator/input/GtfsInput.java
aababilov: I see that you have chosen `org.apache.hc.client5.http.classic.methods.HttpGet` instead but it is not available at Google internally, unfortunately. Please pay attention to the packages that I suggest because we have to stick to those libraries and versions that Google has internally. Otherwise the Java validator cannot run at Google.

main/src/main/java/org/mobilitydata/gtfsvalidator/cli/Main.java
aababilov: This comment was not resolved :(

lionel-nj: Indeed, thanks @aababilov for flagging this. I open an issue so that we keep track of that and modify.

PR #544, opened 2020-12-08
chore: Increase memory size up to 8 GiB

aababilov:
8 GiB is enough for validation of the largest feeds that contain 25 M stop times.


barbeau (approved):
👍

aababilov:
I implemented caching, so we do not need to increase memory size as much:

https://github.com/MobilityData/gtfs-validator/pull/546

PR #545, opened 2020-12-09
feat: Measure the time spent for validation

aababilov:


main/src/main/java/org/mobilitydata/gtfsvalidator/cli/Main.java
```diff
@@ -69,6 +70,9 @@ public static void main(String[] argv) {
         } catch (IOException exception) {
             exception.printStackTrace();
         }
+
+        final long endMillis = System.currentTimeMillis();
+        System.out.println(String.format("Validation took %.3f seconds", (endMillis - startMillis) / 1000.0));
```
lionel-nj, line 75: What do you think about leveraging `java.util.concurrent.TimeUnit`? The maths here are relatively simple but we could use this library to improve clarity (conversion from milliseconds to seconds)
```suggestion
        System.out.println(String.format("Validation took %.3f seconds", TimeUnit.MILLISECONDS.toseconds(endMillis - startMillis));
```

main/src/main/java/org/mobilitydata/gtfsvalidator/cli/Main.java
```diff
@@ -49,6 +49,7 @@ public static void main(String[] argv) {
         System.out.println("Validators:");
         System.out.println(validatorLoader.listValidators());
 
+        final long startMillis = System.currentTimeMillis();
```
barbeau, line 52: Measuring elapsed time on Java should always use `System.nanoTime()`:
https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/System.html#nanoTime()

For example see https://www.baeldung.com/java-measure-elapsed-time.

barbeau (changes requested):
This is very useful! One comment in-line on the method being used.

main/src/main/java/org/mobilitydata/gtfsvalidator/cli/Main.java
barbeau: Agree with ☝️ , although the units should be in nanos (see my other comment).

aababilov: Unfortunately, these converters in `TimeUnit` are not as good as they should be. They return `long` instead of `double` which means that we would get 0 seconds for most of the small feeds and this measurement would be pointless.

I agree to use `System.nanoTime()`, absolutely, but I insist on keeping division by double here.

main/src/main/java/org/mobilitydata/gtfsvalidator/cli/Main.java
aababilov: That's a great point! Done.

main/src/main/java/org/mobilitydata/gtfsvalidator/cli/Main.java
aababilov: We have a wonderful function absl::FDivDuration() that would do what we need here, but this is for C++ :)

barbeau: >Unfortunately, these converters in TimeUnit are not as good as they should be. They return long instead of double which means that we would get 0 seconds for most of the small feeds and this measurement would be pointless.

Ah, yes, you're right, I had forgotten about this. We did the same thing in the GTFS-RT validator.

barbeau (approved):
👍

PR #547, opened 2020-12-09
fix: Specify field type for stop_times.trip_id as ID

aababilov:
This PR provides support to specify field type for stop_times.trip_id as ID


lionel-nj (approved):
LGTM

barbeau (approved):

PR #550, opened 2020-12-09
fix: Add ID annotation to translations.txt fields

barbeau:
**Summary:**

It appears that two fields within translations.txt are missing the `ID` annotation:
* record_id
* record_sub_id

See https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md#translationstxt.

@aababilov I know translations.txt is a strange table that can contain fields from many other tables. Is there any reason not to annotate these fields with `@FieldType(FieldTypeEnum.ID)`?

**Expected behavior:** 

Fields within translations.txt that are of type ID in the GTFS spec are now treated as IDs within the validator


aababilov:
This is a good question!

Indeed, `record_id` always has some ID, according to the current [GTFS `reference.md`](https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md#translationstxt).

However, `record_sub_id` is not necessarily an ID (the `reference.md` specifies a *wrong* column type) because it may be `stop_sequence` (int), `date` (date), `start_time` (time) etc. So, having `record_sub_id` as a usual text is better because it will not invoke possible validations for IDs in future (e.g., check for certain characters).

Who knows: perhaps `record_id` would have some non-ID in the future? I do not see an immediate value in marking it as ID.

barbeau:
>However, record_sub_id is not necessarily an ID (the reference.md specifies a wrong column type) because it may be stop_sequence (int), date (date), start_time (time) etc. So, having record_sub_id as a usual text is better because it will not invoke possible validations for IDs in future (e.g., check for certain characters).

Oh, good catch, I missed that. I was going off the listed data type.

>Who knows: perhaps record_id would have some non-ID in the future? I do not see an immediate value in marking it as ID.

I was thinking of this in context of https://github.com/MobilityData/gtfs-validator/pull/546. If it's annotated as an ID then it would be cached, if not, then it would not. And in some datasets translations.txt could potentially be huge depending on the number of translations provided.

What do you think?

aababilov:
This is a very fair point that caching makes perfect sense for both fields. That caching can be enabled with `@CachedField` annotation.

One popular use case: translations for `stop_headsign` in `stop_times.txt`. These translations will look like:

```
table_name,field_name,language,translation,record_id,record_sub_id
stop_times,stop_headsign,en,ABC,TRIP_ID1,1
stop_times,stop_headsign,ko,CDE,TRIP_ID1,1
stop_times,stop_headsign,ja,FGH,TRIP_ID1,1
stop_times,stop_headsign,en,ABC,TRIP_ID1,2
stop_times,stop_headsign,ko,CDE,TRIP_ID1,2
stop_times,stop_headsign,ja,FGH,TRIP_ID1,2
```

aababilov:
Caching should be enabled for `table_name`, `field_name` and `language` as well.

Caching for `field_value` also makes sense because translations can be given for the same string into multiple languages.

aababilov (commented):
Thanks for the PR!

barbeau:
Ok, sounds good - I'll close this PR then as caching is implemented via `@CachedField` in https://github.com/MobilityData/gtfs-validator/pull/546/files#diff-34cae0bb3c3fd0369c80001b1556ee7168b89756ebee0d5216af62f718b7572dR47. We can revisit the ID issue if needed in the future.

PR #553, opened 2020-12-09
ci: Don't run GitHub Action build on documentation changes

barbeau:
**Summary:**

It's frustrating to sit around and wait for CI builds to finish when only documentation (markdown files) has changed.

**Expected behavior:** 

This PR adds configuration to GitHub Actions to ignore any changes to documentation:
* If the change is to a .md file and it was started via a push or pull request, then don't run the GitHub Action CI build
* Release builds will still be executed even if the only changes are to .md files

PR #554, opened 2020-12-10
ci: Remove Circle CI builds

barbeau:
**Summary:**

Circle CI was primarily used to troubleshoot GitHub Actions CI build issues. Now that GitHub CI seems to be working we don't need it, and if GitHub CI stops working in the future we can always restore these files at that point.

Also, Circle CI doesn't seem to support ignoring changes to markdown files (similar to https://github.com/MobilityData/gtfs-validator/pull/553 for GitHub Actions). It will save us time if we don't need to sit around and wait for the Circle CI build to pass every time we change a markdown file.

**Expected behavior:** 

Circle CI will no longer build the project on each push and PR. GitHub Actions will remain as the primary CI build platform.


barbeau:
I didn't turn off Circle CI via their portal until after I opened this PR, so I'm closing and re-opening to see if that removes Circle CI from the checks list.

barbeau:
Re-opening didn't trigger the removal, but I pushed more commits and that seems to have triggered it. So looks good now 👍 .

lionel-nj (approved):
lgtm

Issue #559, opened 2020-12-10
implement tests for existing validation rules


lionel-nj:
**Is your feature request related to a problem? Please describe.**
At present the following validation rules are not tested. 
- AgencyConsistencyValidator
- RouteColorConstrastValidator
- GtfsTripServiceIdForeignKeyValidator
- ShapeUsageValidator
- CalendarServiceDateValidator
- StopTimeArrivalAndDepartureTimeValidator
- FeedServiceDateValidator
- ShapeIncreasingDistanceValidator
- TripAgencyIdValidator

**Describe the solution you'd like**
Add unit tests

Issue #549, opened 2020-12-09
Distinguish between errors and warnings in notices


barbeau:
**Is your feature request related to a problem? Please describe.**
v1.x of the validator distinguishes between errors and warnings.

The current master branch after the annotation refactor does not.

For example, see this comment for `PlatformWithoutParentStationNotice` which should be distinguished as a warning and not an error:
https://github.com/MobilityData/gtfs-validator/pull/500#discussion_r538869906

**Describe the solution you'd like**
We should update the annotation-based validator with the concept of errors vs. warnings.

**Describe alternatives you've considered**
N/A

carlfredl:
I'm going to review with MobilityData spec team:
- Define methodology for deciding error vs. warning. (In-spec = error / Open to interpretation = warning.) What does in-spec mean?
- Example: [Shape distance travelled must always increase](https://github.com/MobilityData/gtfs-validator/issues/527)

aababilov:
This issue was not fully resolved. We still do not export info about severity and we do not make decisions based on that.

I volunteer to take this issue.

aababilov:
Actually, that was my issue from the beginning :)

barbeau:
The MobilityData spec team has decided to define ERRORS vs. WARNINGs as the following:
* If the [GTFS reference spec](https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md) says something is required ("must" or "must not" be done), then violation of this is an ERROR
* Otherwise, violations of any other language ("should", "should not") in the GTFS reference, or violation of anything in the [GTFS Best Practices](https://gtfs.org/best-practices/) is a WARNING

(The above choices for "must" vs. "should" follow [RFC 2119](https://tools.ietf.org/html/rfc2119) - with the realization that the GTFS spec will need to be modified to come into full compliance with RFC 2119 based on current GTFS industry practice)

Note that this doesn't necessarily align with the current definitions of `SeverityLevel` in the validator, which is:
* `INFO` - is used for things that do not affect the feed's quality, such as unknown files or unknown fields
* `WARNING` - is used for things that affects the feed's quality (e.g., insufficient color contrast) but the feed remains functional
* `ERROR` - is used when the feed cannot function (e.g., broken foreign key reference)

Should we reconsider the enumerations for `SeverityLevel` to be something like `HIGH`, `MEDIUM`, and `LOW`, which is possibly orthogonal to whether something is declared an ERROR or WARNING based on the spec?

[...]

Issue #417, opened 2020-10-13
Errors vs warnings - Incorrect rule about shapes.txt


mcplanner:
**Describe the bug**
I'm given an error that I believe incorrectly interprets the GTFS reference document on shapes.txt. The error claims that all shapes defined must be used, which I don't believe is true.

**To Reproduce**
I'm running this on Tuolumne's October GTFS file, which is not available online, but attached here:
[tuolumne.zip](https://github.com/MobilityData/gtfs-validator/files/5372929/tuolumne.zip)

**Expected behavior**
I did not expect to see this error message because I don't see this requirement in Google's reference document.

**Witnessed behavior**
`Notice{filename='shapes.txt', level='ERROR', code='38', title='Unused shape', description='All records defined by GTFS `shapes.txt` should be used in `trips.txt`. File `shapes.txt`defines entity with id: `23`that is not used.', extra='{fieldName=shape_id}'}]`

**Environment used**
- OS: macOS Catalina 10.15.6
- Java version: 11.0.5
- gtfs-validator version: v1.3.0-SNAPSHOT

e-lo:
To add further to this issue, there is currently a hard line distinguishing a warning from an error in the code, but this should be defined in a config such that you can run the validator based on:

1.  strict treatment of all warnings as errors
2. all GTFS spec and codifyiable best practices treated as errors
3. a user-defined configuration of what is an error vs warning

I believe this will require:
- treatment of the *definitions of* errors vs warning as the same type of entity
- either each error/warning having a boolean property associated with "spec definition" and "best practice" OR a canonical list of warnings/errors which represent each.

barbeau:
Similar issue with extra records in trips.txt, as mentioned in https://github.com/MobilityData/gtfs-validator/issues/519

barbeau:
Related issue for v2 architecture in terms of re-introducing the concept of errors vs warnings - https://github.com/MobilityData/gtfs-validator/issues/549.

barbeau:
Similar issue (unused records in fare_rules.txt) for future rule planned with https://github.com/MobilityData/gtfs-validator/issues/237

lionel-nj:
Will be closed by https://github.com/MobilityData/gtfs-validator/pull/726. This has been set to be a `WARNING` instead of an `ERROR`.

PR #560, opened 2020-12-10
test: implement unit tests for existing validation rules

lionel-nj:
closes #559 
**Summary:**

This PR adds unit tests for existing validation rules:

- [x] AgencyConsistencyValidator https://github.com/MobilityData/gtfs-validator/pull/560/commits/096e5420a26e6a3f84b0837ba4d16f6db7b2e0a6
- [x] RouteColorConstrastValidator: https://github.com/MobilityData/gtfs-validator/pull/560/commits/8c54ad26cfff24fe84fa69152916ce90dd1b41e3
- [x] GtfsTripServiceIdForeignKeyValidator https://github.com/MobilityData/gtfs-validator/pull/560/commits/dc3732727e7cdee499b1219d4e8cc562d00b7072
- [x] ShapeUsageValidator https://github.com/MobilityData/gtfs-validator/pull/560/commits/b80c49a9f3fa11918122ca420c7fe6c613fe62ad
- [x] CalendarServiceDateValidator https://github.com/MobilityData/gtfs-validator/pull/560/commits/1f55cb4d701e4c784846516fea36af5b960ac130
- [x] StopTimeArrivalAndDepartureTimeValidator https://github.com/MobilityData/gtfs-validator/pull/560/commits/6f8b7802d8a12beb9297774009aa60e167336b1b
- [x] FeedServiceDateValidator https://github.com/MobilityData/gtfs-validator/pull/560/commits/5c40995b8e580ea5dae2c706308923358960e9fe
- [x] ShapeIncreasingDistanceValidator https://github.com/MobilityData/gtfs-validator/pull/560/commits/d9ace8c4c806a5657e29feda7e7e1906552f25d6
- [x] RouteAgencyIdValidator: https://github.com/MobilityData/gtfs-validator/pull/560/commits/a48aa40c83702c8a8fea0f944289fb08bb7420b2

**Expected behavior:** 

No code change, tests should pass

- [x] Check copyright headers
- [x] Check no wildcards in import statements


lionel-nj:
@barbeau Just one thing: I believe that there should be a small code change
https://github.com/MobilityData/gtfs-validator/blob/23438807fd8ac53dcd6818fbb9869a56d79db310/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/AgencyConsistencyValidator.java#L64

The specification requires all agencies to have the same timezone, but does not state the same thing regarding agency languages. 
`v1.4.0` only checked the first case (https://github.com/MobilityData/gtfs-validator/blob/v1.4.0/usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateAgenciesHaveSameAgencyTimezone.java)

barbeau:
>The specification requires all agencies to have the same timezone, but does not state the same thing regarding agency languages.

@lionel-nj Yes, you're right, that's not canonical so we can't flag it as an error. Let's change this to a warning instead of an error. Can you please add a `FIXME` comment there with a reference to https://github.com/MobilityData/gtfs-validator/issues/549, and add a note on the issue at https://github.com/MobilityData/gtfs-validator/issues/549 that this should be a warning?

lionel-nj:
Sure! But should we flag this at all? Unless I am missing something, [the GTFS best practices ](http://gtfs.org/best-practices/#agencytxt)do not have recommendations for this field.

barbeau:
Sure, you can raise the question with the spec team. I'm believe that multiple languages across agencies are allowed, though.

[translations.txt](https://gtfs.org/reference/static/#translationstxt) says:

>In regions that have multiple official languages, transit agencies/operators typically have language-specific names and web pages. In order to best serve riders in those regions, it is useful for the dataset to include these language-dependent values.

I believe translations.txt was an expansion of the spec to allow agencies to provide their own translations, rather than leaving it up to the consumers. But it's worth asking the question.

lionel-nj:
Hi @timMillet and @scmcca!
Is there any best practices regarding agencies' languages? Is it correct to consider that multiple languages across agencies are allowed? If not, we think that such a situation should be flagged as a warning by the validator.

scmcca:
This seems highly context-specific and it would be difficult to say whether or not the use of multiple languages in a dataset is more or less meaningful to the area it concerns. I suspect @timMillet will have some good insight/examples!

timMillet:
It is ok to have multiple languages across one dataset, or even across one agency within a dataset. This situation is normal so it's not worth an error.

e.g: the Swiss railway agency SBB CFF FFS use multiple languages, depending on the dominant language where the transit service is operated.

See Example section of the field definition of `feed_info.feed_lang` for more details.

lionel-nj:
Thanks! @timMillet  @scmcca

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/ShapeIncreasingDistanceValidatorTest.java
```diff
@@ -0,0 +1,208 @@
...
+    final GtfsShapeTableContainer mockShapeTable = mock(GtfsShapeTableContainer.class);
+    @InjectMocks
+    final ShapeIncreasingDistanceValidator underTest = new ShapeIncreasingDistanceValidator();
+
+    @Before
+    public void openMocks() {
+        MockitoAnnotations.openMocks(this);
+    }
+
+    @Test
```
barbeau, line 51: Need a test for when two sequential shape distances are the same (should generate notice)

barbeau: Oh, I think there is a bug in the current implementation. Spec says they should increase, but current rule allows them to be the same. Please double-check the spec and behavior to make sure I have this right.

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/ShapeUsageValidatorTest.java
```diff
@@ -0,0 +1,109 @@
...
+
+import static com.google.common.truth.Truth.assertThat;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.verifyNoInteractions;
+import static org.mockito.Mockito.verifyNoMoreInteractions;
+import static org.mockito.Mockito.when;
+
+public class ShapeUsageValidatorTest {
```
barbeau, line 43: This is another rule that I believe, to my knowledge, the spec doesn't explicitly enforce (all shapes in shapes.txt must be referenced in trips.txt). But it's a good idea to remove unused shapes from a GTFS file to reduce the file size and processing by consumers. So maybe warning instead of error? It depends how we want to define warnings.

@timMillet @scmcca Thoughts?

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/StopTimeArrivalAndDepartureTimeValidatorTest.java
```diff
@@ -0,0 +1,353 @@
...
+    @InjectMocks
+    final StopTimeArrivalAndDepartureTimeValidator underTest = new StopTimeArrivalAndDepartureTimeValidator();
+
+    @Before
+    public void openMocks() {
+        MockitoAnnotations.openMocks(this);
+    }
+
+    @Test
+    public void departureTimeAndArrivalTimeNotProvidedShouldNotGenerateNotice() {
```
barbeau, line 56: My interpretation of the spec, given the history from when before the `timepoint` field existed, is that both `arrival_time` and `departure_time` can be blank only if this record is NOT a timepoint. 

Which means that `timepoint` field would need to be empty/missing. Note that the spec says:

>1 or empty - Times are considered exact.

...but to my knowledge this only applies IFF `arrival_time` and `departure_time` are populated - if `arrival_time` and `departure_time` are empty and `timepoint` is empty then the record is NOT a timepoint.

However, I'm not sure the Python validator ever enforced because for a time before `timepoint` existed it became common for producers to provide values for `arrival_time` and `departure_time` for non-timepoint fields even though it was against the spec. @lionel-nj Do you recall the Python validator behavior? This may be the case where if we follow the spec strictly we might be flagging errors on datasets that passed the Python validator.

@timMillet @scmcca Thoughts? The spec could definitely use clarifying in this area for the allowed states of the `arrival_time`, `departure_time`, and `timepoint` fields, including explanation of practice before and after `timepoint` was adopted.

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/StopTimeArrivalAndDepartureTimeValidatorTest.java
```diff
@@ -0,0 +1,353 @@
...
+        NoticeContainer mockNoticeContainer = mock(NoticeContainer.class);
+        ListMultimap<String, GtfsStopTime> stopTimeCollection = ArrayListMultimap.create();
+        GtfsStopTime mockStopTime = mock(GtfsStopTime.class);
+        when(mockStopTime.hasArrivalTime()).thenReturn(true);
+        GtfsTime mockArrivalTime = mock(GtfsTime.class);
+        when(mockStopTime.arrivalTime()).thenReturn(mockArrivalTime);
+        when(mockStopTime.hasDepartureTime()).thenReturn(true);
+        GtfsTime mockDepartureTime = mock(GtfsTime.class);
+        when(mockStopTime.departureTime()).thenReturn(mockDepartureTime);
+        when(mockDepartureTime.isBefore(mockArrivalTime)).thenReturn(false);
```
barbeau, line 82: Is the behavior of `isBefore()` tested anywhere? If not we should have separate tests for this, or instead of mocking `GtfsTime` we can instantiate a real object with real values. Same for `isAfter()`.

barbeau (commented):
Thanks @lionel-nj! Some thoughts in-line.

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/StopTimeArrivalAndDepartureTimeValidatorTest.java
lionel-nj: `isBefore()` is not tested yet but will be.

lionel-nj: These test will be included in https://github.com/MobilityData/gtfs-validator/pull/568.

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/ShapeIncreasingDistanceValidatorTest.java
lionel-nj: Indeed, the specification states that: 

> Values must increase along with shape_pt_sequence;


Plus, `v1.4.0` used to enforce this: https://github.com/MobilityData/gtfs-validator/blob/76970f6ab56c885ef3fad951a6bff2531ae18a3e/usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateShapeIncreasingDistance.java#L52

So I conclude that having strictly increasing values for `shape_dist_traveled` is the correct behavior. Will check the spec team thought.

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/ShapeUsageValidatorTest.java
lionel-nj: Similar issue https://github.com/MobilityData/gtfs-validator/issues/417

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/ShapeIncreasingDistanceValidatorTest.java
lionel-nj: Per discussion with the spec team: 
Two shape points can have the same `shapes.shape_dist_traveled` value but in that occasion, a notice could be could be issued.

The type of notice (warning or error)  would depend on the distance between the two shape points:
- less than 10 meters: a warning
- more dans 10 meters: an error.

This distance threshold could be user configurable.

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/CalendarServiceDateValidatorTest.java
```diff
@@ -0,0 +1,176 @@
...
+        verify(mockCalendar, times(2)).endDate();
+        verify(mockCalendar, times(2)).startDate();
+        verify(mockStartDate, times(1)).isAfter(mockEndDate);
+        verify(mockStartDate, times(1)).toYYYYMMDD();
+        verify(mockEndDate, times(1)).toYYYYMMDD();
+        verifyNoMoreInteractions(mockCalendar, mockEndDate, mockStartDate);
+    }
+
+    @Test
+    public void noStartDateShouldNotGenerateNotice() {
```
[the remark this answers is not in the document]
lionel-nj: Changing this name would require us to change the validator implementation since it actually does not throw exception when `calendar.start_date` is null. So because of that I am not sure that is out best option.

However, yes - adding comment in the validator implementation would improve the code clarity.

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/CalendarServiceDateValidatorTest.java
```diff
@@ -0,0 +1,176 @@
...
+
+        underTest.validate(mockNoticeContainer);
+
+        verifyNoInteractions(mockNoticeContainer);
+        verify(mockCalendar, times(1)).hasStartDate();
+        verifyNoMoreInteractions(mockCalendar, mockEndDate, mockStartDate);
+    }
+
+    @Test
+    public void noEndDateShouldNotGenerateNotice() {
```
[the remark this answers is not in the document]
lionel-nj: https://github.com/MobilityData/gtfs-validator/pull/560#discussion_r544426717

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/CalendarServiceDateValidatorTest.java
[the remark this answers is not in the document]
barbeau: >Changing this name would require us to change the validator implementation since it actually does not throw exception when calendar.start_date is null..so I am not sure that is out best option.

Or maybe `...ShouldRunWithoutError()`? It's still testing that the case where the field is missing doesn't crash the validator.

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/StopTimeArrivalAndDepartureTimeValidatorTest.java
lionel-nj: IIRC, the python validator only checked if the first and last stop times of a trip had both `arrival_time` and `departure_time` non `null`

barbeau: >IIRC, the python validator only checked if the first and last stop times of a trip had both arrival_time and departure_time non null

Oh, yes, that's something else to consider as well - first and last records always need to have both arrival_time and departure_time.

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/RouteColorContrastValidatorTest.java
```diff
@@ -0,0 +1,131 @@
...
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.verifyNoInteractions;
+import static org.mockito.Mockito.verifyNoMoreInteractions;
+import static org.mockito.Mockito.when;
+
+public class RouteColorContrastValidatorTest {
+
+    @Test
+    public void noRouteColorShouldNotGenerateNotice() {
```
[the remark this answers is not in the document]
timMillet: I agree. If we go strictly by the spec, the color contrast should be tested with the default colors when one of the 2 color fields is empty.

This notice should not be severe though (only a warning?). I imagine situations where this error could be counterproductive with the GTFS Grading Scheme. The notice should invite producers to provide the missing color, rather than update the color provided.

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/ShapeUsageValidatorTest.java
timMillet: I agree with a warning. All unnecessary information should not lead to an error, as the dataset and all the information contained in it may still be valid. 
This situation should have the same impact as when producers are providing extra fields or extra files.

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/StopTimeArrivalAndDepartureTimeValidatorTest.java
timMillet: I read the comment several times but I'm not sure to understand accurately the issue. I don't know the practices before the adoption of `timepoint`. Here is what I understand from the spec relatively to `arrival_time` and `departure_time` according to `timepoint`:
- if the `timepoint` field is not provided, all `arrival_time` and `departure_time` values must be provided.
- if the `timepoint` field is provided:
  - if `timepoint=1` on a stop_time record, then the `arrival_time` and `departure_time` values of that record must be provided.
  - if `timepoint=0` or is empty on a stop_time record, then the `arrival_time` and `departure_time` values of that record may be provided, i.e one of them or both can be empty.
  -  first stop_time record and last stop_time record of a trip must have `timepoint=1`, i.e `arrival_time` and `departure_time`.

Don't hesitate to tell me if you want more clarification!

barbeau: @timMillet see https://github.com/google/transit/issues/61 thread for more context on the history of timepoint and some of the current issues.

aababilov:
Thanks for you heroic deeds! Do you mind to create separate PRs for different validators? That will highly simplify the review and we will submit your code faster.

Also, if there is a pattern that we want to change, it is better to polish it on a single file rather than updating hundred files at a time.

aababilov:
Please could you choose any validator and create a separate PR for it?

lionel-nj:
> Thanks for you heroic deeds! Do you mind to create separate PRs for different validators? That will highly simplify the review and we will submit your code faster.


Sure, it's been done in :

#606, #607, #608, #609, #610, #611, #612, #613, and #614

lionel-nj:
Closing: this PR has been split into smaller PRs to ease code review: #606, #607, #608, #609, #610, #611, #612, #613, and #614

PR #561, opened 2020-12-11
fix: Use org.apache.httpcomponents:httpclient:4.5.2 instead of org.apache.httpcomponents.client5:httpclient5:5.0.3

aababilov:
Use org.apache.httpcomponents:httpclient:4.5.2 instead of org.apache.httpcomponents.client5:httpclient5:5.0.3
    
Switch to the library version supported by Google.

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -29,7 +29,7 @@ repositories {
 
 dependencies {
     implementation 'org.apache.logging.log4j:log4j-slf4j-impl:2.14.0'
-    implementation 'org.apache.httpcomponents.client5:httpclient5:5.0.3'
+    implementation 'org.apache.httpcomponents:httpclient:4.5.2'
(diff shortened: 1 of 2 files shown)


lionel-nj (approved):
lgtm, thanks!

barbeau:
@aababilov Is there a lint tool we could use to catch issues like this that we could run as a bot on pull requests that would pass/fail each contribution? 

It's going to be hard to ensure that all contributions follow all the rules, especially if MobilityData doesn't have visibility into all of them. Something like https://github.com/MobilityData/gtfs-validator/pull/561, but for dependencies too. I'm not sure if this bot would catch the default branch issue in https://github.com/MobilityData/gtfs-validator/pull/563.

barbeau (approved):

PR #562, opened 2020-12-11
chore: Switch from log4j to Flogger

aababilov:
Flogger is the recommended logger at Google. Flogger is the result of an attempt to create a unified logging API, suitable for the vast majority of Java projects in Google.

### DIFF
--- a/main/src/main/java/org/mobilitydata/gtfsvalidator/cli/Main.java
+++ b/main/src/main/java/org/mobilitydata/gtfsvalidator/cli/Main.java
@@ -17,8 +17,6 @@
 package org.mobilitydata.gtfsvalidator.cli;
 
 import com.beust.jcommander.JCommander;
-import org.apache.logging.log4j.LogManager;
-import org.apache.logging.log4j.Logger;
 import org.mobilitydata.gtfsvalidator.input.GtfsFeedName;
 import org.mobilitydata.gtfsvalidator.input.GtfsInput;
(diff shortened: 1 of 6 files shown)


main/src/main/java/org/mobilitydata/gtfsvalidator/cli/Main.java
```diff
@@ -40,9 +38,8 @@
 public class Main {
 
     public static void main(String[] argv) {
-        Logger logger = LogManager.getLogger();
         Arguments args = new Arguments();
-        CliParametersAnalyzer cliParametersAnalyzer = new CliParametersAnalyzer(logger);
+        CliParametersAnalyzer cliParametersAnalyzer = new CliParametersAnalyzer();
```
lionel-nj, line 42: Just reading Flogger's [docs best practices](https://google.github.io/flogger/best_practice) 👀📖

> Don’t pass loggers between classes
> Only log via a single static logger instance that’s created inside the same source file as the log statement. There is generally no reason to pass a logger from one class to another.
> 
> Why do people think that loggers should be passed between classes?
> I think the primary reason is that people misunderstand where the class name in log output comes from. It is not based on the name of the logger; it’s the location of the log statement, regardless of the name of the logger used.
> 
> The logger instance does affect the configuration of the logger, but if all loggers use the same configuration (which is very common) then different logger instances will behave identically and there no reason to pass them around.

Thanks Alexej, it's good to know!

lionel-nj (approved):
lgtm

PR #566, opened 2020-12-14
feat: validate frequencies.end_time is after frequencies.start_time

### LINKED ISSUE #518: `end_time` after `start_time` in `frequencies.txt`
Validate `end_time` after `start_time` in `frequencies.txt` 

**Additional context**
`v1.4.0` implementation: 
- https://github.com/MobilityData/gtfs-validator/blob/v1.4.0/usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateFrequencyStartTimeBeforeEndTime.java
- https://github.com/MobilityData/gtfs-validator/blob/v1.4.0/usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateFrequencyStartTimeBeforeEndTimeTest.java

lionel-nj:
closes #518 

**Summary:**

This PR provides an addition `SingleEntityValidator` to verify:  `frequencies.end_time` is after `frequencies.start_time`

**Expected behavior:** 

Notice should be generated if:
- `frequencies.end_time` is before `frequencies.start_time`
Capture d’écran, le 2020-12-14 à 13 51 05

### DIFF
--- a/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/FrequencyTimeInOrderValidator.java
+++ b/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/FrequencyTimeInOrderValidator.java
@@ -34,7 +34,6 @@ import static org.mobilitydata.gtfsvalidator.table.GtfsFrequencyTableLoader.FILE
 public class FrequencyTimeInOrderValidator extends SingleEntityValidator<GtfsFrequency> {
     @Override
     public void validate(GtfsFrequency frequency, NoticeContainer noticeContainer) {
-        // startTime and endTime are assumed to be not null: therefore, we do not check hasStartTime and hasEndTime.
         // validate() will only be called if startTime and endTime have been populated for this frequency
         GtfsTime startTime = frequency.startTime();
         GtfsTime endTime = frequency.endTime();
--- a/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/FrequencyTimeInOrderValidator.java
+++ b/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/FrequencyTimeInOrderValidator.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2020 Google LLC, MobilityData IO
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.mobilitydata.gtfsvalidator.validator;
+
+import org.mobilitydata.gtfsvalidator.annotation.GtfsValidator;
+import org.mobilitydata.gtfsvalidator.notice.NoticeContainer;
+import org.mobilitydata.gtfsvalidator.notice.StartAndEndTimeOutOfOrderNotice;
+import org.mobilitydata.gtfsvalidator.table.GtfsFrequency;
+
+import static org.mobilitydata.gtfsvalidator.table.GtfsFrequencyTableLoader.FILENAME;
+
+/**
+ * Validates `frequencies.start_time` is before `frequencies.end_time` for a single entity.
+ * <p>
+ * Generated notices:
+ * * StartAndEndTimeOutOfOrder
+ */
+@GtfsValidator
+public class FrequencyTimeInOrderValidator extends SingleEntityValidator<GtfsFrequency> {
+    @Override
+    public void validate(GtfsFrequency frequency, NoticeContainer noticeContainer) {
+        // startTime and endTime are assumed to be not null: therefore, we do not check hasStartTime and hasEndTime.
+        if (frequency.startTime().isAfter(frequency.endTime())) {
+            noticeContainer.addNotice(
+                    new StartAndEndTimeOutOfOrderNotice(
+                            FILENAME,
+                            frequency.tripId(),
+                            frequency.csvRowNumber(),
+                            frequency.startTime(),
+                            frequency.endTime()
+                    )
+            );
+        }
+    }
+}
+
+
--- a/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/FrequencyTimeInOrderValidator.java
+++ b/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/FrequencyTimeInOrderValidator.java
@@ -37,9 +37,6 @@ public class FrequencyTimeInOrderValidator extends SingleEntityValidator<GtfsFre
         // startTime and endTime are assumed to be not null: therefore, we do not check hasStartTime and hasEndTime.
         GtfsTime startTime = frequency.startTime();
         GtfsTime endTime = frequency.endTime();
-        if (startTime.equals(endTime)) {
-            return;
-        }
         if (startTime.isAfter(endTime)) {
             noticeContainer.addNotice(
                     new StartAndEndTimeOutOfOrderNotice(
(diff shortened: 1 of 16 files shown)


main/src/main/java/org/mobilitydata/gtfsvalidator/validator/FrequencyTimeInOrderValidator.java
```diff
@@ -0,0 +1,51 @@
...
+
+import org.mobilitydata.gtfsvalidator.annotation.GtfsValidator;
+import org.mobilitydata.gtfsvalidator.notice.NoticeContainer;
+import org.mobilitydata.gtfsvalidator.notice.StartAndEndTimeOutOfOrderNotice;
+import org.mobilitydata.gtfsvalidator.table.GtfsFrequency;
+
+import static org.mobilitydata.gtfsvalidator.table.GtfsFrequencyTableLoader.FILENAME;
+
+/**
+ * Validates `frequencies.start_time` is before `frequencies.end_time` for a single entity.
```
barbeau, line 27:
```suggestion
 * Validates `frequencies.start_time` is before or equal to `frequencies.end_time` for a single entity.
```

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/FrequencyTimeInOrderValidatorTest.java
```diff
@@ -0,0 +1,102 @@
...
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.verifyNoInteractions;
+import static org.mockito.Mockito.verifyNoMoreInteractions;
+import static org.mockito.Mockito.when;
+
+@RunWith(JUnit4.class)
+public class FrequencyTimeInOrderValidatorTest {
+
+    @Test
+    public void startTimeBeforeEndTimeShouldNotGenerateNotice() {
```
barbeau, line 41: I would also include a test validating that start time equal to end time does not generate a notice.

barbeau (commented):
Thanks @lionel-nj! A few comments in-line.

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/FrequencyTimeInOrderValidatorTest.java
lionel-nj: Good catch, thanks!

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/FrequencyTimeInOrderValidator.java
```diff
@@ -0,0 +1,56 @@
...
+ * * StartAndEndTimeOutOfOrder
+ */
+@GtfsValidator
+public class FrequencyTimeInOrderValidator extends SingleEntityValidator<GtfsFrequency> {
+    @Override
+    public void validate(GtfsFrequency frequency, NoticeContainer noticeContainer) {
+        // startTime and endTime are assumed to be not null: therefore, we do not check hasStartTime and hasEndTime.
+        GtfsTime startTime = frequency.startTime();
+        GtfsTime endTime = frequency.endTime();
+        if (startTime.equals(endTime)) {
```
barbeau, line 40: This `.equals()` check is unnecessary, right? If so I would remove.

barbeau (commented):
@lionel-nj One more comment. Also, this PR is still in "draft" state - is it ready for merge?

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/FrequencyTimeInOrderValidator.java
lionel-nj: Indeed, thanks!

lionel-nj:
@barbeau my bad, it is not officially ready for merge.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/FrequencyTimeInOrderValidator.java
```diff
@@ -0,0 +1,53 @@
...
+ * * StartAndEndTimeOutOfOrder
+ */
+@GtfsValidator
+public class FrequencyTimeInOrderValidator extends SingleEntityValidator<GtfsFrequency> {
+    @Override
+    public void validate(GtfsFrequency frequency, NoticeContainer noticeContainer) {
+        // startTime and endTime are assumed to be not null: therefore, we do not check hasStartTime and hasEndTime.
+        GtfsTime startTime = frequency.startTime();
+        GtfsTime endTime = frequency.endTime();
+        if (startTime.isAfter(endTime)) {
```
barbeau, line 40: Sorry, I missed this when I looked at it previously - so what is the behavior when these values are `null` in a real dataset? Do we stop earlier in the validation process, or will this line produce an NPE?

If the `validate()` method executes and doesn't produce a notice that's fine, but if it crashes then you do need to check for hasStartTime and hasEndTime.

lionel-nj: If these values are `null`, a `MissingRequireValueFieldError` is generated and added to the notice container, plus the GtfsEntity is not added to its `GtfsTableContainer`.
So, when executing `validate()` it is assumed that `hasStarTime()`  and `hasEndTime()` will return `true` if these methods are called.
I can add details to the comment to clarify this if needed.

I think this test cover such situations: https://github.com/MobilityData/gtfs-validator/blob/89d069a0a5022bf950c058c9f070e2aaacda02bf/main/src/test/java/org/mobilitydata/gtfsvalidator/table/GtfsFrequencyTableLoaderTest.java#L66

barbeau: >I can add details to the comment to clarify this if needed.

Yes, that would help. If I'm understanding correctly, maybe this would be better:

>validate() will only be called if startTime and endTime have been populated for this frequency

lionel-nj:
@barbeau I could not re-request a review, but I made some changes to clarify the code.

barbeau (approved):
Thanks @lionel-nj!

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/FrequencyTimeInOrderValidator.java
```diff
@@ -0,0 +1,53 @@
...
+ * * StartAndEndTimeOutOfOrder
+ */
+@GtfsValidator
+public class FrequencyTimeInOrderValidator extends SingleEntityValidator<GtfsFrequency> {
+    @Override
+    public void validate(GtfsFrequency frequency, NoticeContainer noticeContainer) {
+        // validate() will only be called if startTime and endTime have been populated for this frequency
+        GtfsTime startTime = frequency.startTime();
+        GtfsTime endTime = frequency.endTime();
+        if (startTime.isAfter(endTime)) {
```
aababilov, line 40: We at Google require `start_time < end_time`, so they may not be equal. I do not see a use case for `start_time == end_time` (covered by a unit test here) - providers should use non-frequency-based trips in this case.

barbeau: Thanks @aababilov, that's interesting to know. That's what I originally thought, but IIRC others said that having these values as equal is a potentially valid case if you want a vehicle to circulate once. What you say makes sense to me, but we need to update the spec to reflect this.

@MobilityData/transit-specs another spec clarification issue.

timMillet: IMHO, if `start_time == end_time`, it is definitively a wrong use of frequencies.txt, but it should not make a dataset invalid. As Sean said, it's a potentially valid case if you want a vehicle to circulate once. This should then be treated as a warning.

lionel-nj: Currently, the validator does not generate any notice if `start_time == end_time`

lionel-nj: Opening an issue on this matter so that we can revisit this https://github.com/MobilityData/gtfs-validator/issues/753

PR #573, opened 2020-12-14
docs: Remove "[REQUEST]" prefix from issue template

barbeau:
**Summary:**

Currently `[REQUEST]` is prepended to all issue titles. It doesn't seem to serve any purpose so let's remove it.

**Expected behavior:** 

When you open a feature request issue, `[REQUEST]` should no longer be autopopulated in the issue title.

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -1,7 +1,7 @@
 ---
 name: Feature request
 about: Suggest an idea for this project
-title: "[REQUEST]"
+title: ""
 labels: ''
 assignees: 'Cristhian-HA'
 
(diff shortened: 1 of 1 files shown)


barbeau:
GitHub Action CI also shouldn't run on this because it's a documentation change (see https://github.com/MobilityData/gtfs-validator/pull/553), so this will be a good test for that too.

lionel-nj (approved):
It seems like GitHub action did not run: Capture d’écran, le 2020-12-14 à 18 22 02  But the PR interface seems to be waiting for some checks to pass:
Capture d’écran, le 2020-12-14 à 18 22 54. I'll check and see if there is a way to avoid that. 

Other than that LGTM

lionel-nj:
@barbeau to prevent these "Waiting for status to be reported" it seems like we have to disable the actions, but I do not think that is advisable: is "simply" ignoring them on documentation changes a reasonable option?

https://github.community/t/github-actions-pull-requests-always-include-some-outdated-checks/16157

barbeau:
Ugh. I'm testing out a new setting, going to close and re-open to see if it changes anything.

PR #575, opened 2020-12-22
refactor: Switch to Gson library for JSON serialization

aababilov:
Jackson2 has a terrible security record, with new remote code execution 
vulnerabilities being published once every other month:
https://medium.com/@cowtowncoder/on-jackson-cves-dont-panic-here-is-what-you-need-to-know-54cd0d6e8062

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -31,7 +31,7 @@ dependencies {
     implementation 'org.apache.httpcomponents:httpclient:4.5.2'
     implementation 'commons-io:commons-io:2.8.0'
     implementation 'com.univocity:univocity-parsers:2.9.0'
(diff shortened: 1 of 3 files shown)


lionel-nj:
Thanks @aababilov! I'll be waiting for @barbeau's review before merging 😀

lionel-nj (approved):
Thanks @aababilov! I'll be waiting for @barbeau's review to merge 😀

barbeau (approved):
LGTM 👍. FWIW, there is likely a negative performance tradeoff here based on based comparisons I've seen of Gson vs Jackson, but I wouldn't expect it to have a large impact given that in most cases we should be serializing a relatively small number of notices. If serialization performance becomes an issue in the future we can always revisit.

Issue #574, opened 2020-12-21
Prevent code duplication


lionel-nj:
**Is your feature request related to a problem? Please describe.**

There is some duplicate code.

**Describe the solution you'd like**
 We could refactor to avoid said duplication:

**Additional context**

I think that there was a misunderstanding. My point was to avoid the code duplication, not to necessarily introduce ternary operation. If you do not like ?:, you can write it this way:

```
GtfsInput gtfsInput;
if  (args.input == null) {
   gtfsInput = GtfsInput.createFromUrl(
                                new URL(args.url),
                                args.storageDirectory);
} else {
 gtfsInput =  GtfsInput.createFromPath(args.input);
}
feedContainer = feedLoader.loadAndValidate(
                        gtfsInput,
                        feedName,
                        validatorLoader,
                        noticeContainer);

_Originally posted by @aababilov in https://github.com/MobilityData/gtfs-validator/pull/541#discussion_r539006608_

Issue #342, opened 2020-08-15
[REQUEST] Generate tests report artifact when tests fail


fabrice-v:
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when I have to scroll the log output for a failed build to find the failing test. This is due to no artifact being saved by the GitHub workflow when a step fails.

**Describe the solution you'd like**
I'd like step [`Persist test report`](https://github.com/MobilityData/gtfs-validator/blob/87e86709b2f43d1a436fbe5a5d975994038efd1d/.github/workflows/gradle.yml#L33) to always run so that a nice html test report with the failing test(s) is available as an artifact.

**Describe alternatives you've considered**
Doing nothing. Scrolling and searching the log output is a workable solution. As well as syncing locally and run the tests in an IDE on a Java 11 enabled machine.

**Additional context**
https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#job-status-check-functions

PR #576, opened 2020-12-22
feat: Use java.nio.file.Path and support loading of ZIP files to memory

aababilov:
java.nio.file.Path is a more flexible way to retrieve files and
directories.

ZIP file can be loaded to memory instead of being saved to a temporary
directory. This is useful when the ZIP is placed on a non-local file
system or has to be downloaded from the given URL.

### DIFF
--- a/core/src/main/java/org/mobilitydata/gtfsvalidator/input/GtfsInput.java
+++ b/core/src/main/java/org/mobilitydata/gtfsvalidator/input/GtfsInput.java
@@ -44,7 +44,7 @@ import java.util.Set;
  */
 public abstract class GtfsInput {
     /**
-     * Creates an specific GtfsInput to read data from the given path.
+     * Creates a specific GtfsInput to read data from the given path.
      *
      * @param path the path to the resource
      * @return the {@code GtfsInput} created after processing the GTFS archive
@@ -98,12 +98,12 @@ public abstract class GtfsInput {
     }
 
     /**
-     * Downloads data from network
+     * Downloads data from network.
      *
-     * @param sourceUrl     the fully qualified URL
-     * @param outputStream  the output stream
-     * @throws IOException         if no file could not be found at the specified location
-     * @throws URISyntaxException  if URL is malformed
+     * @param sourceUrl    the fully qualified URL
+     * @param outputStream the output stream
+     * @throws IOException        if no file could not be found at the specified location
+     * @throws URISyntaxException if URL is malformed
      */
     private static void loadFromUrl(URL sourceUrl, OutputStream outputStream)
             throws IOException, URISyntaxException {
@@ -117,7 +117,7 @@ public abstract class GtfsInput {
     }
 
     /**
-     * Creates a path from a given string or cleans it if the path already exists
+     * Creates a path from a given string or cleans it if the path already exists.
      *
      * @param toCleanOrCreate the path to clean or create as string
      * @return the created @code{Path}
--- a/core/src/main/java/org/mobilitydata/gtfsvalidator/input/GtfsInput.java
+++ b/core/src/main/java/org/mobilitydata/gtfsvalidator/input/GtfsInput.java
@@ -23,14 +23,17 @@ import org.apache.http.impl.client.CloseableHttpClient;
 import org.apache.http.impl.client.HttpClients;
 
 import java.io.BufferedOutputStream;
+import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.OutputStream;
 import java.net.URISyntaxException;
 import java.net.URL;
 import java.nio.file.AccessDeniedException;
+import java.nio.file.FileSystems;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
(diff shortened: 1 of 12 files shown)


aababilov:
Test loading from URL without a temporary directory:

``` % ./gradlew run  --args="-u https://github.com/MobilityData/gtfs-validator/raw/v1.4.0/usecase/src/test/resources/valid_zip_sample.zip -o $HOME/gtfs-from-url -t 4 --feed_name ca-test-feed"
```

core/src/main/java/org/mobilitydata/gtfsvalidator/input/GtfsZipInMemoryInput.java
```diff
@@ -0,0 +1,78 @@
...
+        ZipEntry entry = zipInputStream.getNextEntry();
+        while (entry != null) {
+            if (!insideZipDirectory(entry)) {
+                filenames.add(entry.getName());
+            }
+            entry = zipInputStream.getNextEntry();
+        }
+    }
+
+    private boolean insideZipDirectory(ZipEntry entry) {
```
lionel-nj, line 50: What do you think about renaming this method for clarity as:
```suggestion
    private boolean isInsideZipDirectory(ZipEntry entry) {
```

or as this:
```suggestion
    private boolean isNested(ZipEntry entry) {
```

core/src/main/java/org/mobilitydata/gtfsvalidator/input/GtfsZipInMemoryInput.java
```diff
@@ -0,0 +1,78 @@
...
+        // We do not use File.separator because the .zip file specification states:
```
lionel-nj, line 51: 👍🏾

main/src/main/java/org/mobilitydata/gtfsvalidator/cli/Main.java
```diff
@@ -63,26 +64,22 @@ public static void main(String[] argv) {
...
-                        feedName,
-                        validatorLoader,
-                        noticeContainer);
+                gtfsInput = GtfsInput.createFromPath(Paths.get(args.getInput()));
             }
         } catch (IOException | URISyntaxException | InterruptedException e) {
             e.printStackTrace();
             return;
         }
+        feedContainer = feedLoader.loadAndValidate(gtfsInput, feedName, validatorLoader, noticeContainer);
```
lionel-nj, line 82: Thanks, I did not have the time yet to remove the code duplication #574

lionel-nj (changes requested):
Thanks @aababilov just one comment in line.

main/src/main/java/org/mobilitydata/gtfsvalidator/cli/Main.java
aababilov: You are welcome! Now we have a clear case when code duplication should be avoided since there are three different ways to create GtfsInput.

aababilov:
Updated.

lionel-nj (approved):
Thanks @aababilov, I will wait for @barbeau's final approval to merge this PR.

core/src/main/java/org/mobilitydata/gtfsvalidator/input/GtfsInput.java
```diff
@@ -39,57 +42,87 @@
 /**
  * GtfsInput provides a common interface for reading GTFS data, either from a ZIP archive or from a directory.
```
barbeau, line 43: Javadoc needs to be updated for abstract class vs. interface

barbeau (commented):
@aababilov Nice, LGTM! One comment on documentation. Also, the main README.md and USAGE.md files could use updating to reflect this change too (the README.md doesn't currently even list the `-u` or `-s` parameters).

barbeau:
Actually, looks like there is one [failing test](https://github.com/MobilityData/gtfs-validator/pull/576/checks?check_run_id=1657793723) too after the master branch was merged from another PR:

>org.mobilitydata.gtfsvalidator.cli.CliParametersAnalyzerTest > provideStorageDirectoryCliParameterWithoutSpecifyingUrlShouldReturnFalse FAILED
    org.mockito.exceptions.verification.WantedButNotInvoked at CliParametersAnalyzerTest.java:113

barbeau:
@aababilov I'd also suggest turning off CircleCI builds via the CircleCI console for your fork of this repo. We removed the CircleCI config in this repo and turned off builds here because it was slowing down approvals without any benefits. I'm still getting notifications of failed CircleCI builds on your fork, though, and I believe you can turn these off by disabling CircleCI builds for your fork in the console.

core/src/main/java/org/mobilitydata/gtfsvalidator/input/GtfsInput.java
aababilov: An abstract class is still providing an interface, so I would like to keep the existing phasing..

aababilov:
> @aababilov Nice, LGTM! One comment on documentation. Also, the main README.md and USAGE.md files could use updating to reflect this change too (the README.md doesn't currently even list the `-u` or `-s` parameters).

Good point! I would prefer to make it in a separate pull request since options `-u` and `-s` already existed before my commit.

aababilov:
`Test Package Document / test` is failing on `provideStorageDirectoryCliParameterWithoutSpecifyingUrlShouldReturnFalse` which passes locally on my machine. The [test report](https://github.com/MobilityData/gtfs-validator/pull/576/checks?check_run_id=1657793723) suggests me to open file:///home/runner/work/gtfs-validator/gtfs-validator/main/build/reports/tests/test/index.html which is obviously not available.

Can we make the automatic testing platform more easy for debugging?

aababilov:
> @aababilov I'd also suggest turning off CircleCI builds via the CircleCI console for your fork of this repo. We removed the CircleCI config in this repo and turned off builds here because it was slowing down approvals without any benefits. I'm still getting notifications of failed CircleCI builds on your fork, though, and I believe you can turn these off by disabling CircleCI builds for your fork in the console.

Thanks, done!

barbeau:
>Can we make the automatic testing platform more easy for debugging?

@aababilov Agreed that it's not always easy to dig in detail here. I believe @lionel-nj and @fabrice-v set up a packaging of the test results to help with this, although it requires several clicks to download the test report. Here's what I was able to find clicking through some of the commits. @lionel-nj may be able to share more when he's back.

image

f8full:
> a packaging of the test results to help with this,

Yep, see #342 and #455

PR #577, opened 2020-12-22
feat: Avoid race condition for notice containers

aababilov:
Notice container is not thread-safe, otherwise it will slow down adding
new notices. Each thread now has its own instance of notice container.
When the thread ends, its container is added to the global one.

### DIFF
--- a/core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NoticeContainer.java
+++ b/core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NoticeContainer.java
@@ -30,7 +30,8 @@ import java.util.List;
 /**
  * Container for validation notices (errors and warnings).
  * <p>
- * This class is not thread-safe.
+ * This class is not intentionally not thread-safe to increase performance. Each thread
+ * has it's own NoticeContainer, and after execution is complete the results are merged.
  */
 public class NoticeContainer {
     private static final int MAX_EXPORTS_PER_NOTICE_TYPE = 100000;
--- a/core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NoticeContainer.java
+++ b/core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NoticeContainer.java
@@ -27,6 +27,11 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 
+/**
+ * Container for validation notices (errors and warnings).
+ * <p>
+ * This class is not thread-safe.
+ */
 public class NoticeContainer {
     private static final int MAX_EXPORTS_PER_NOTICE_TYPE = 100000;
     private static final Gson DEFAULT_GSON = new GsonBuilder().serializeNulls().create();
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/mobilitydata/gtfsvalidator/table/GtfsFeedLoader.java
```diff
@@ -116,10 +119,11 @@ public GtfsFeedContainer loadAndValidate(GtfsInput gtfsInput, GtfsFeedName feedN
         }
         try {
             try {
-                List<Future<GtfsTableContainer>> results = exec.invokeAll(loaderCallables);
-                results.forEach(f -> {
+                exec.invokeAll(loaderCallables).forEach(f -> {
```
lionel-nj, line 122: I think the code would be more clear if `f` was renamed, maybe as `tableAndNoticeContainer` or `container`.

lionel-nj, line 122:
```suggestion
                exec.invokeAll(loaderCallables).forEach(container -> {
```

lionel-nj (changes requested):
Thanks @aababilov! Just one suggestion :)

core/src/main/java/org/mobilitydata/gtfsvalidator/table/GtfsFeedLoader.java
aababilov: This "f" is for future. We start from 

```
TableAndNoticeContainers containers = f.get();
```

and do not use it after that.

aababilov: D'accord!

lionel-nj: Oh okay! I missed that one, thanks.

lionel-nj (approved):
Thanks @aababilov! Will wait for @barbeau's final approval to merge.

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NoticeContainer.java
```diff
@@ -75,4 +75,16 @@ public String exportJson() {
         }
         return noticesByType;
     }
+
+    /**
+     * Adds all notices from another container.
+     * <p>
+     * This is useful for multithreaded validation: each thread has its own notice container which is merged
```
barbeau, line 82: I suggest adding a comment at the top of the class explaining the general lack of thread safety in this class, so later someone doesn't see the word "multithreaded" and start synchronizing everything.

barbeau (commented):
@aababilov LGTM, although one comment about adding more Javadoc explanation about lack of thread safety in `NoticeContainer`

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NoticeContainer.java
aababilov: Done.

barbeau (approved):
Thanks @aababilov!

Issue #539, opened 2020-12-03
Make -f an optional command-line parameter


barbeau:
**Is your feature request related to a problem? Please describe.**
IMHO it's a bit awkward to have `-f` or `--feed_name` be a required command-line parameter, as it forces you to have a priori knowledge about the dataset you're validating. If given a directory of GTFS zip files with no metadata, it's not a trivial task to find the country code for each.

**Describe the solution you'd like**
Make `-f` an optional command-line parameter, where rules that require `-f` are skipped if `-f` isn't provided.

IIRC right now the feed name is only used for the country code to validate phone numbers.

@aababilov Thoughts on this? Is this something you could work on?

**Describe alternatives you've considered**
Clarify how `-f` parameter should be used in documentation and what happens if you provide a wrong country code, and what the valid values for the name is.

maximearmstrong:
@barbeau I agree with you on this. @annaelleduff and me noticed this while implementing the Validator 2.0 in the MobilityDatabase pipeline. Most of the time, we won't have access to the country code prior to validation.

At the moment, for the harvesting of OpenMobilityData, we managed to obtain the country code using [Nominatim](nominatim.openstreetmap.org). This is possible thanks to the latitude and longitude of a dataset given by the API of OpenMobilityData. However, this information might not always be accessible prior to validation of a dataset.

Could the Validator use the latitude and longitude pairs found inside the dataset to get the country code?
Also, it would remove potential human errors from modifying the validation report.
cc: @lionel-nj @aababilov

barbeau:
There are also embedded libraries that we could leverage to do lookup of country code from lat/long without having to hit an external API, like https://github.com/bencampion/reverse-country-code.

lionel-nj:
It appears that `GtfsFeedName` is used to check the phone number formatting. A possible implementation could consist in an additional multiple file `validator` that would check phone numbers formatting (in agency.txt, and attributions.txt). This use case would be executed if `-f` is not provided. 

@barbeau @aababilov  However I am afraid that this approach might not scale efficiently: if the validator was to be extended (addition of another table which contains a `phone number` field) the implementor would have to modify said use case to validate the extension.

aababilov:
I can propose the following option.

Add a flag for country code. It may be used to validate phone numbers and it may have other use cases in the future. The full feed name is not needed, so that flag may be removed.

Reverse country code is a good option for some use cases but we cannot use it for phone validation :(

[...]

PR #578, opened 2020-12-22
feat: Support UK as a synonym for GB country code

aababilov:
ISO 3166-1 Alpha 2 country code for the United Kingdom is "GB" but feed
names use "UK" instead.

### DIFF
--- a/core/src/main/java/org/mobilitydata/gtfsvalidator/input/GtfsFeedName.java
+++ b/core/src/main/java/org/mobilitydata/gtfsvalidator/input/GtfsFeedName.java
@@ -36,6 +36,10 @@ public class GtfsFeedName {
 
     private static final String INVALID_FEED_NAME_MESSAGE = "Feed name must follow format `cc-abc-...', where cc is an ISO Alpha 2 country code";
 
+    // ISO 3166-1 Alpha 2 country code for the United Kingdom is "GB" but feed names may use "UK".
+    private static final String UK = "UK";
+    private static final String GB = "GB";
+
     private final String countryFirstName;
 
     private GtfsFeedName(String countryFirstName) {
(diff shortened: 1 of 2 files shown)


lionel-nj (approved):
Good catch, thanks @aababilov. Waiting for @barbeau's approval to merge.

barbeau:
@aababilov Which feed names are you referring to, and where are they named? Just trying to understand why this changed is needed, if the feed name is something defined by the person inputting the feed into the validator (as opposed to part of the GTFS feed itself).

aababilov:
All feeds for United Kingdom at Google start from the `uk-` prefix instead of `gb-`. I would like to support them out of the box.

dbabramov:
I am wondering whether deriving the locale from the feed name is necessarily the most robust way to go about this.
That requires all consumers to name their feeds in the same way, and break in strange and obscure ways if the feed naming pattern is not complied with.
Would it be a better idea to pass the locale explicitly?

barbeau:
>Would it be a better idea to pass the locale explicitly?

I agree with @dbabramov on this ☝️ - it's a more standardized approach. My concern with the solution in this PR is that we'll be getting all kinds of aliasing and formatting requests to support different naming conventions.

It would also be nice if we could make at least the locale (and maybe name?) parameters optional to simplify things for users. I could see needing to know the locale of feeds upfront potentially problematic for some - I opened an issue for this at https://github.com/MobilityData/gtfs-validator/issues/539.

barbeau:
Actually, looking into this more for `uk`, it looks like ISO does recognize this officially as an alias for `gb`:
https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2#Exceptional_reservations

So we'd still be within the ISO spec for this PR implementation.

I propose we merge this PR and then separate the country code from the feed name in a new PR.

@aababilov @dbabramov thoughts?

aababilov:
I agree to pass the country code to the validator because feed names are too Google-specific. Let's merge this PR for now to unblock me and I will prepare another one.

barbeau:
Hmmm, I'm not sure why the CI build is failing again:

>org.mobilitydata.gtfsvalidator.cli.CliParametersAnalyzerTest > provideStorageDirectoryCliParameterWithoutSpecifyingUrlShouldReturnFalse FAILED
    org.mockito.exceptions.verification.WantedButNotInvoked at CliParametersAnalyzerTest.java:113

It passed locally for me via gradle command line and within IntelliJ.

I'm going to go ahead and merge because it's unrelated to this PR. We'll see if it passes after that and if not open an issue.

barbeau (approved):

barbeau:
Huh...passes fine in the master branch 🤷:
https://github.com/MobilityData/gtfs-validator/runs/1665054412

PR #581, opened 2021-01-07
feat: Validate that trip frequencies do not overlap

### LINKED ISSUE #522: Validate trip frequencies do not overlap
Validate trip frequencies do not overlap

**Additional context**
`v1.4.0` implementation:
- https://github.com/MobilityData/gtfs-validator/blob/v1.4.0/usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateTripFrequenciesOverlap.java
- https://github.com/MobilityData/gtfs-validator/blob/v1.4.0/usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateTripFrequenciesOverlapTest.java

aababilov:
closes #522

**Summary:**
This PR validates that trip frequencies do not overlap. 

Two entries X and Y in frequencies.txt are considered to directly overlap if X.start_time <= Y.start_time and Y.start_time < X.end_time.

### DIFF
--- a/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/OverlappingFrequencyValidatorTest.java
+++ b/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/OverlappingFrequencyValidatorTest.java
@@ -1,0 +1,114 @@
+package org.mobilitydata.gtfsvalidator.validator;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+import org.mobilitydata.gtfsvalidator.notice.Notice;
+import org.mobilitydata.gtfsvalidator.notice.NoticeContainer;
+import org.mobilitydata.gtfsvalidator.notice.OverlappingFrequencyNotice;
+import org.mobilitydata.gtfsvalidator.table.GtfsFrequency;
+import org.mobilitydata.gtfsvalidator.table.GtfsFrequencyTableContainer;
+import org.mobilitydata.gtfsvalidator.type.GtfsTime;
+
+import java.util.Arrays;
+import java.util.List;
+
+import static com.google.common.truth.Truth.assertThat;
+
+@RunWith(JUnit4.class)
+public class OverlappingFrequencyValidatorTest {
+    private GtfsFrequency createFrequency(
+            long csvRowNumber, String tripId, String startTime, String endTime, int headwaySecs) {
+        return new GtfsFrequency.Builder()
+                .setCsvRowNumber(csvRowNumber)
+                .setTripId(tripId)
+                .setStartTime(GtfsTime.fromString(startTime))
+                .setEndTime(GtfsTime.fromString(endTime))
+                .setHeadwaySecs(headwaySecs).build();
+    }
+
+    private List<Notice> validateFrequencies(GtfsFrequency... frequencies) {
+        NoticeContainer noticeContainer = new NoticeContainer();
+        OverlappingFrequencyValidator validator = new OverlappingFrequencyValidator();
+        validator.table = GtfsFrequencyTableContainer.forEntities(Arrays.asList(frequencies), noticeContainer);
+        validator.validate(noticeContainer);
+        return noticeContainer.getNotices();
+    }
+
+    @Test
+    public void validSequentialInOrder() {
+        assertThat(validateFrequencies(
+                createFrequency(2, "t0", "05:00:00", "07:00:00", 600),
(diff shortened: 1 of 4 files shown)


main/src/test/java/org/mobilitydata/gtfsvalidator/validator/OverlappingFrequencyValidatorTest.java
```diff
@@ -0,0 +1,114 @@
...
+    @Test
+    public void overlappingIncludedSameEnd() {
+        assertThat(validateFrequencies(
+                createFrequency(2, "t0", "05:00:00", "07:00:00", 600),
+                createFrequency(3, "t0", "06:30:00", "07:00:00", 300)))
+                .containsExactly(new OverlappingFrequencyNotice(
+                        2, GtfsTime.fromString("07:00:00"),
+                        3, GtfsTime.fromString("06:30:00"), "t0"));
+    }
+
```
barbeau, line 99: I would suggest adding a test case for when one trip is fully contained within the other and doesn't share start or end times, which should generate a notice:

~~~
     * a |--------|            07:00 am to 12:00 am
     * b   |----|              08:00 am to 11:00 am
~~~

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/OverlappingFrequencyValidatorTest.java
```diff
@@ -0,0 +1,114 @@
...
+                createFrequency(4, "t0", "05:20:00", "05:40:00", 300)))
+                .containsExactly(
+                        new OverlappingFrequencyNotice(
+                                3, GtfsTime.fromString("05:15:00"),
+                                2, GtfsTime.fromString("05:00:00"), "t0"),
+                        new OverlappingFrequencyNotice(
+                                2, GtfsTime.fromString("05:25:00"),
+                                4, GtfsTime.fromString("05:20:00"), "t0"));
+    }
+}
```
barbeau, line 114: Since we're testing stop times here, do you think we should include at a least one test for when times are greater than 24:00:00? Seems like a corner case (at least by conventional time calculations) that may be worth covering.

barbeau (commented):
Thanks @aababilov, this looks good! Two thoughts in-line.

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/OverlappingFrequencyValidatorTest.java
aababilov: I do not think that it is worth testing. That is not a corner case for this validator. If we really want to check that 24:00:00 is properly compared to 01:00:00, then we should place this to `GtfsTimeTest.java`.

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/OverlappingFrequencyValidatorTest.java
aababilov: Done.

barbeau (approved):
Thanks @aababilov, LGTM!

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/OverlappingFrequencyValidatorTest.java
derhuerst: As someone who's stumbled over the slightly unusual semantics of GTFS `Time` before and seen many, many wrong implementations, I think it would be cool to very that this project handles them properly. During DST switches, [it is especially unintuitive](https://gist.github.com/derhuerst/574edc94981a21ef0ce90713f1cff7f6).

aababilov: DST is not related to frequency validation. The GTFS time abstracts out all these details.

Yes, representing time is a hard task but this validation is much simpler and it does not require tests for:

* DST
* Gregorian vs Julian calendar
* Leap seconds
* Time after 24:00
* Timezones
* 12 clock format vs 24 clock
* AM vs PM
* Year 2000 problem
* End of Maya calendar
* Leap years

derhuerst: Ah, so you're purely validating based on the "nominal" value, not on it's "effects" on the referenced trips?

After thinking about it, that makes sense, because one `frequency.txt` entry references one trip, which usually has *many* service days.

PR #582, opened 2021-01-07
refactor: Migrate to java.time.ZoneId from java.util.TimeZone

aababilov:
One should avoid using the date/time APIs in java.util.
This includes java.util.Date (and its subclass, java.sql.Timestamp),
java.util.Calendar, java.util.TimeZone, etc. They are bad APIs [1].
Even Oracle now acknowledges that these classes have "several
drawbacks" [2].

java.time is the best date/time library that Java offers. Compared to
JodaTime, it has better precision, better APIs, and is built directly
into the language.

[1] https://dzone.com/articles/game-over-jdks-date-and-time-c
[2] https://docs.oracle.com/javase/tutorial/datetime/iso/legacy.html


barbeau (approved):
👍  Thanks @aababilov!

lionel-nj:
Thank you @aababilov

Issue #584, opened 2021-01-07
Test fails due to inability to delete file


barbeau:
**Summary**
When I run `GtfsInputTest.createFromRedirectedUrlShouldNotThrowException()` locally I get:

```
> Configure project :
System.getenv("CI"): null - System.getenv("versionTag"): null
> Task :core:compileJava UP-TO-DATE
> Task :core:processResources NO-SOURCE
> Task :core:classes UP-TO-DATE
> Task :core:cleanTest
> Task :core:compileTestJava UP-TO-DATE
> Task :core:processTestResources NO-SOURCE
> Task :core:testClasses UP-TO-DATE
> Task :core:test

java.lang.AssertionError
	at org.junit.Assert.fail(Assert.java:87)
	at org.junit.Assert.assertTrue(Assert.java:42)
	at org.junit.Assert.assertTrue(Assert.java:53)
	at org.mobilitydata.gtfsvalidator.input.GtfsInputTest.createFromRedirectedUrlShouldNotThrowException(GtfsInputTest.java:97)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.rules.ExternalResource$1.evaluate(ExternalResource.java:54)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:100)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:103)
[...]

PR #585, opened 2021-01-07
fix: Don't fail test on failure to delete temp file

barbeau:
**Summary:**

Closes https://github.com/MobilityData/gtfs-validator/issues/584.

Delete file failures can happen due to local conditions like file locks and aren't related to validator performance, so this PR removes the assertions that check if the file has been successfully deleted.


aababilov:
LGTM!

Thank you for fixing this hurting problem!

PR #586, opened 2021-01-07
fix: Use try-with-resources for Files.list

aababilov:
The returned stream encapsulates a DirectoryStream. If timely disposal
of file system resources is required, the try-with-resources construct
should be used to ensure that the stream's close method is invoked after
the stream operations are completed.


barbeau (approved):
👍

PR #587, opened 2021-01-08
feat: Add indexing of trips on block_id

aababilov:
This will allow to quickly find all trips that belong to a given block.

This also fixes a bug in the code generator. There will be tests for the
processor in future but now the existing GTFS schema serves as the
integration test.


barbeau (approved):
Thanks @aababilov, LGTM!

processor/src/main/java/org/mobilitydata/gtfsvalidator/processor/TableContainerGenerator.java
```diff
@@ -294,8 +294,8 @@ private MethodSpec generateSetupIndicesMethod() {
                         "$L.put(entity.$L(), entity)",
                         byKeyMapName(indexField.name()),
                         indexField.name());
-                method.endControlFlow();
             }
+            method.endControlFlow();
```
f8full, line 298: Nice catch!

PR #588, opened 2021-01-11
fix: Make the generated getSingleEntity() method public

aababilov:
This method is generated for single-entity file containers such as
GtfsFeedInfoTableContainer.


barbeau (approved):
👍

Issue #583, opened 2021-01-07
IndexOutOfBoundsException when combining noticeContainers


barbeau:
**Summary**

When running the master branch at https://github.com/MobilityData/gtfs-validator/commit/4d3ebb3ef9cbdf8d69f611d7bed620d7af028c22 with the following arguments:

`-i hart_gtfs.zip -o output -f us-fl-hart`

on this dataset:

[hart_gtfs.zip](https://github.com/MobilityData/gtfs-validator/files/5782983/hart_gtfs.zip)

I get this error:

```
...
Cache for trips.txt block_id: size = 506, lookup count = 12161, hits = 95.84%, misses = 4.16%
Cache for trips.txt shape_id: size = 154, lookup count = 12161, hits = 98.73%, misses = 1.27%
Cache for shapes.txt shape_id: size = 154, lookup count = 39127, hits = 99.61%, misses = 0.39%
Reading stop_times.txt, row 200000
Reading stop_times.txt, row 400000
Cache for stop_times.txt trip_id: size = 12161, lookup count = 400959, hits = 96.97%, misses = 3.03%
Cache for stop_times.txt arrival_time: size = 60885, lookup count = 801918, hits = 92.41%, misses = 7.59%
Cache for stop_times.txt departure_time: size = 60885, lookup count = 801918, hits = 92.41%, misses = 7.59%
Cache for stop_times.txt stop_id: size = 2164, lookup count = 400959, hits = 99.46%, misses = 0.54%
Cache for stop_times.txt stop_headsign: size = 0, lookup count = 400959, hits = 100.00%, misses = 0.00%
java.util.concurrent.ExecutionException: java.lang.IndexOutOfBoundsException: Index 0 out of bounds for length 0
	at java.base/java.util.concurrent.FutureTask.report(FutureTask.java:122)
	at java.base/java.util.concurrent.FutureTask.get(FutureTask.java:191)
	at org.mobilitydata.gtfsvalidator.table.GtfsFeedLoader.lambda$loadAndValidate$3(GtfsFeedLoader.java:146)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1510)
	at org.mobilitydata.gtfsvalidator.table.GtfsFeedLoader.loadAndValidate(GtfsFeedLoader.java:144)
	at org.mobilitydata.gtfsvalidator.cli.Main.main(Main.java:82)
Caused by: java.lang.IndexOutOfBoundsException: Index 0 out of bounds for length 0
[...]

PR #589, opened 2021-01-11
fix: Do not crash on 0 entities in feed_info.txt

aababilov:
The previously used method isEmptyFile() tells if the whole text file is
empty. That crashed the Validator for a non-empty feed_info.txt file
that contains only the headers.

This commit adds documentation to isEmptyFile() and other methods in
GtfsTableContainer.

The unit test was changed to use the real TableContainer classes with
their real behaviour.

Closes https://github.com/MobilityData/gtfs-validator/issues/583

### DIFF
--- a/core/src/main/java/org/mobilitydata/gtfsvalidator/parsing/CsvFile.java
+++ b/core/src/main/java/org/mobilitydata/gtfsvalidator/parsing/CsvFile.java
@@ -58,7 +58,7 @@ public class CsvFile implements Iterable<CsvRow> {
     /**
      * Tells if the file is empty, i.e. it has no rows and even no headers.
      *
-     * @return @code true} is the file is empty, {@code false} otherwise
+     * @return true is the file is empty, false otherwise
      */
     public boolean isEmpty() {
         return isEmpty;
@@ -80,7 +80,7 @@ public class CsvFile implements Iterable<CsvRow> {
     /**
      * Advances to the next row.
      *
-     * @return the next @code CsvRow} or null if end of file was reached.
+     * @return the next {@link CsvRow} or null if end of file was reached.
      */
     @Nullable
     private CsvRow nextResult() {
--- a/core/src/main/java/org/mobilitydata/gtfsvalidator/parsing/CsvFile.java
+++ b/core/src/main/java/org/mobilitydata/gtfsvalidator/parsing/CsvFile.java
@@ -58,7 +58,7 @@ public class CsvFile implements Iterable<CsvRow> {
     /**
      * Tells if the file is empty, i.e. it has no rows and even no headers.
      *
-     * @return true is the file is empty, false otherwise
+     * @return true if the file is empty, false otherwise
      */
     public boolean isEmpty() {
         return isEmpty;
--- a/core/src/main/java/org/mobilitydata/gtfsvalidator/table/GtfsTableContainer.java
+++ b/core/src/main/java/org/mobilitydata/gtfsvalidator/table/GtfsTableContainer.java
@@ -57,7 +57,7 @@ public abstract class GtfsTableContainer<T extends GtfsEntity> {
     /**
      * Tells if the file is missing.
      *
-     * @return true is the file is missing, false otherwise
+     * @return true if the file is missing, false otherwise
      */
     public boolean isMissingFile() {
         return missingFile;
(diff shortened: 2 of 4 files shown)


aababilov:
Hi Sean,

Did you really need to merge the `master` branch here? I think it's still possible to squash this commit and merge it with `master` since there are no conflicts.

barbeau:
@aababilov This repo has had the GitHub "Require branches to be up to date before merging" setting turned on, which requires clicking "Update" to merge from master into the PR before the PR can be merged into master (it says "to ensure the tests include all recent changes to code").

But I agree, that's not really necessary, and it just slows everything down. Any failures that don't include conflicts will be rare, and we'll catch them when they are merged. So I just turned it off, and we shouldn't have to do that anymore (unless there are conflicts with master).

core/src/main/java/org/mobilitydata/gtfsvalidator/parsing/CsvFile.java
```diff
@@ -58,7 +58,7 @@ public CsvFile(Reader reader, String filename) {
     /**
      * Tells if the file is empty, i.e. it has no rows and even no headers.
      *
-     * @return @code true} is the file is empty, {@code false} otherwise
+     * @return true is the file is empty, false otherwise
```
barbeau, line 61:
```suggestion
     * @return true if the file is empty, false otherwise
```

core/src/main/java/org/mobilitydata/gtfsvalidator/table/GtfsTableContainer.java
```diff
@@ -41,6 +41,11 @@ public int entityCount() {
 
     public abstract String gtfsFilename();
 
+    /**
+     * Tells if the file is empty, i.e. it has no rows and even no headers.
+     *
+     * @return true is the file is empty, false otherwise
```
barbeau, line 47:
```suggestion
     * @return true if the file is empty, false otherwise
```

core/src/main/java/org/mobilitydata/gtfsvalidator/table/GtfsTableContainer.java
```diff
@@ -49,6 +54,11 @@ public void setEmptyFile(boolean emptyFile) {
         this.emptyFile = emptyFile;
     }
 
+    /**
+     * Tells if the file is missing.
+     *
+     * @return true is the file is missing, false otherwise
```
barbeau, line 60:
```suggestion
     * @return true if the file is missing, false otherwise
```

core/src/main/java/org/mobilitydata/gtfsvalidator/table/GtfsTableContainer.java
```diff
@@ -61,6 +71,13 @@ public boolean hasInvalidHeaders() {
         return invalidHeaders;
     }
 
+    /**
+     * Tells if the file is invalid headers, e.g., some required headers are missing.
+     * <p>
+     * Note that unknown headers are not considered invalid.
+     *
+     * @return true is the file has invalid headers, false otherwise
```
barbeau, line 79:
```suggestion
     * @return true if the file has invalid headers, false otherwise
```

barbeau (approved):
Thanks, LGTM! There are a few small typos that I'll fix here and then merge.

core/src/main/java/org/mobilitydata/gtfsvalidator/table/GtfsTableContainer.java
```diff
@@ -61,6 +71,13 @@ public boolean hasInvalidHeaders() {
         return invalidHeaders;
     }
 
+    /**
+     * Tells if the file is invalid headers, e.g., some required headers are missing.
+     * <p>
+     * Note that unknown headers are not considered invalid.
+     *
+     * @return true if the file has invalid headers, false otherwise
+     */
```
f8full, line 80: Shouldn't that be `setMissingRequiredHeaders` ?
Honnestly I was confused only reading the documentation as I couldn't figure what could be invalid if unkown is ok. Then I reread.

aababilov: There may be other errors related to headers, e.g. the same column name used multiple times. Such files are rejected and further validation is not performed.

PR #590, opened 2021-01-12
chore: Use FluentLogger instead of stderr and stdout

aababilov:
FluentLogger is more flexible and professional than stdout and stderr.

Note that Main.java keeps printing to stdout because its information
should be visible to the user on screen.


barbeau:
Still need to click "Update branch" here to merge master, presumably because this PR was opened before I changed the GitHub setting...

barbeau (approved):
Thanks @aababilov!

f8full:
> the GitHub setting...

What setting? :)

f8full:
Got it https://github.com/MobilityData/gtfs-validator/pull/589#issuecomment-758813234

PR #595, opened 2021-01-13
chore: Upgrade to Java 11

### LINKED ISSUE #593: CI environment Java runtime doesn't match source code
**Summary**
In our `build.gradle` we currently have:

```groovy
sourceCompatibility = JavaVersion.VERSION_1_8
```

...which indicates that the project can be compiled and run on JRE 8 or higher.

However, our CI configuration, for example:
https://github.com/MobilityData/gtfs-validator/blob/master/.github/workflows/test_pack_doc.yml#L35

...runs on JDK 11. As an end result we could have code sneak in that's not compatible with Java 8.

**What I expect**

CI Java runtime should match project code - if we say the project runs on Java 8, we need to test on Java 8. It's ok to test on 8 and 11 both (not sure if that's possible with GitHub Actions).

**What I see**

Code says it runs on Java 8, but we're only testing on JDK 11, which may pass on new variables that were added to Java after Java 8 (e.g., `LocaDate.EPOCH`), but result in errors when running on Java 8.

aababilov:
closes #593

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -21,7 +21,7 @@ plugins {
 group 'org.mobilitydata'
 version "gtfs-validator-${System.getenv("versionTag")}"
 
-sourceCompatibility = JavaVersion.VERSION_1_8
+sourceCompatibility = JavaVersion.VERSION_11
 
 repositories {
     mavenCentral()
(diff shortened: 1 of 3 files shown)


barbeau (approved):
Thanks! 👍

PR #596, opened 2021-01-14
feat: Introduce a concept of system errors

aababilov:
closes #548 

The base class Notice has direct two subclasses:
* ValidationNotice - for validation errors and warnings. This is the
  parent class for the most of notices, such as DuplicatedColumnNotice,
  NumberOutOfRangeError or RouteColorContrastNotice;
* SystemError - for all notices related to retrieving and processing a
  GTFS feed, but not related to the content of the feed itself (out of
  memory, failure to download etc.).

Rationale.
* Program that runs the validator needs to know why it crashed and a JSON report is more convenient than parsing stdout.
* There may be bugs in individual validations and we need information about them.
* We should not leak information about system errors to clients, so system errors should not be placed to the same report as validation errors.

This is the report generated for IndexOutOfBoundsException introduced in #552.

```
{
    "notices": [
        {
            "code": "runtime_exception_in_validator",
            "notices": [
                {
                    "exception": "java.lang.IndexOutOfBoundsException",
                    "message": "Index 0 out of bounds for length 0",
                    "validator": "org.mobilitydata.gtfsvalidator.validator.MatchingFeedAndAgencyLangValidator"
                }
            ],
            "totalNotices": 1
        }
    ]
}
```

PR #598, opened 2021-01-14
chore: Add a unit test for trips with 0 or 1 stop

aababilov:
Test that BlockTripsWithOverlappingStopTimesValidator can handle trips
with 0 or 1 stop gracefully.


barbeau (approved):
👍

aababilov:
Thanks for the fast review! Now I can update #596 because the current PR added a new unit test.

PR #600, opened 2021-01-18
fix: Make children of GtfsTableLoader and GtfsTableContainer final

aababilov:
Remove the temptation to inherit from automatically generated classes.

Generated children of GtfsEntity and their builders are already marked
as final.

This is a common pattern The AutoValue library also makes its generated
classes final.


barbeau (approved):
👍

aababilov:
Thank you!

Issue #138, opened 2020-04-29
[REQUEST] Implement feed start and expiration date (GTFS rule)


maximearmstrong:
**Is your feature request related to a problem? Please describe.**
Feed must have a valid start date and expiration date. This is a GTFS rule implemented in Google Python validator and featured in Google Type Error as TYPE_FEED_EXPIRATION_DATE, TYPE_FEED_FUTURE_SERVICE, TYPE_FEED_HAS_VERY_SHORT_SERVICE  and TYPE_EXPIRED_FEED_HAS_VERY_SHORT_SERVICE.

**Describe the solution you'd like**
Actual Google GTFS validator behaviour : verifies if feed does not start in the future, or does not expire within 60 days.

**Describe alternatives you've considered**

**Additional context**
Line 49, 50, 54 and 55 in Error support priorities https://docs.google.com/spreadsheets/d/1vqe6wq7ctqk1EhYkgtZ0_TbcQ91vccfs2daSjn20BLE/edit#gid=0

barbeau:
@maximearmstrong This is looking at feed_info.txt `feed_start_date` and `feed_end_date`, right?

Does the Python tool put any parameters on what a "valid" date is? For example, does it need to be in the future from now? Or within the next 10 years? Or just it just validate the YYYYMMDD format?

maximearmstrong:
@barbeau Yes. In the Python validator, feed_start_date must be in the future, and feed_end_date must be in more than 60 days.

barbeau:
@maximearmstrong Thanks, that's interesting. 

@timMillet this is another rule we'll need to look at in terms of what's considered current GTFS best practice.

timMillet:
@barbeau I agree. AFAIK, there are no such rules anywhere in the spec or in the Best Practices. The only things specified in the Best Practices are:
- "At any time, the published GTFS dataset should be valid for at least the next 7 days, and ideally for as long as the operator is confident that the schedule will continue to be operated." and
- "If possible, the GTFS dataset should cover at least the next 30 days of service."

Why the Google Validator was issuing errors when `feed_start_date` was not in the future, and `feed_end_date` was not in more than 60 days?

barbeau:
>Why the Google Validator was issuing errors when feed_start_date was not in the future, and feed_end_date was not in more than 60 days?

I'm guessing this rule goes way back to the early days of GTFS when producers were consistently publishing a GTFS dataset every 3-4 months, with no overlap between feed service periods. 

[...]

PR #675, opened 2021-01-26
feat: validate requirement on `feed_info.start_date` and `feed_info.e…

### LINKED ISSUE #531: `feed_end_date` should be provided if `feed_start_date` is provided
Validate `feed_end_date` is provided if `feed_start_date` is provided

**Additional context**
`v1.4.0` implementation:
- https://github.com/MobilityData/gtfs-validator/blob/v1.4.0/usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateFeedInfoFeedEndDateIsPresent.java
- tests: https://github.com/MobilityData/gtfs-validator/blob/v1.4.0/usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateFeedInfoFeedEndDateIsPresentTest.java

### LINKED ISSUE #532: `feed_start_date` should be provided if `feed_end_date` is provided
Validate `feed_start_date` is provided if `feed_end_date` is provided

**Additional context**
`v1.4.0` implementation:
- https://github.com/MobilityData/gtfs-validator/blob/v1.4.0/usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateFeedInfoFeedStartDateIsPresent.java
- tests: https://github.com/MobilityData/gtfs-validator/blob/v1.4.0/usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateFeedInfoFeedStartDateIsPresentTest.java

lionel-nj:
closes #531 
closes #532 

**Summary:**

This PR provides support to check the requirements on `feed_info.start_date` and `feed_info.end_date.
- (1) Both fields can be blank
- (2) if one of said fields is provided, the other one should also be provided.

**Expected behavior:** 

In situation (2), a `MissingRequiredFieldNotice` should be generated and added to the `noticeContainer`.
In situation (1), no notice should be generated. 


main/src/main/java/org/mobilitydata/gtfsvalidator/validator/FeedServiceDateValidator.java
```diff
@@ -18,13 +18,16 @@
...
 import org.mobilitydata.gtfsvalidator.annotation.Inject;
+import org.mobilitydata.gtfsvalidator.notice.MissingRequiredFieldError;
 import org.mobilitydata.gtfsvalidator.notice.NoticeContainer;
 import org.mobilitydata.gtfsvalidator.notice.StartAndEndDateOutOfOrderNotice;
 import org.mobilitydata.gtfsvalidator.table.GtfsFeedInfo;
 import org.mobilitydata.gtfsvalidator.table.GtfsFeedInfoTableContainer;
 
 /**
- * Validates that start_date <= end_date for all rows in "feed_info.txt".
+ * Validates : * start_date <= end_date for all rows in "feed_info.txt" * feed_info.start_date is
```
barbeau, line 28: If these are supposed to be bulleted it looks like the format isn't allowing this, and as a result it's hard to read. I'd use numbers instead of bullets or just write it out.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/FeedServiceDateValidator.java
```diff
@@ -35,9 +38,19 @@
   @Override
   public void validate(NoticeContainer noticeContainer) {
     for (GtfsFeedInfo feedInfo : feedInfoTable.getEntities()) {
-      if (feedInfo.hasFeedStartDate()
-          && feedInfo.hasFeedEndDate()
-          && feedInfo.feedStartDate().isAfter(feedInfo.feedEndDate())) {
+      if (!feedInfo.hasFeedStartDate() || !feedInfo.hasFeedEndDate()) {
```
barbeau, line 41: I think this first IF statement is redundant? I believe it can be removed.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/FeedServiceDateValidator.java
```diff
@@ -35,9 +38,19 @@
...
+              new MissingRequiredFieldError(
+                  feedInfoTable.gtfsFilename(), feedInfo.csvRowNumber(), "feed_end_date"));
+        } else if (!feedInfo.hasFeedStartDate() && feedInfo.hasFeedEndDate()) {
+          noticeContainer.addValidationNotice(
+              new MissingRequiredFieldError(
+                  feedInfoTable.gtfsFilename(), feedInfo.csvRowNumber(), "feed_start_date"));
+        }
+        return;
+      }
+      if (feedInfo.feedStartDate().isAfter(feedInfo.feedEndDate())) {
```
barbeau, line 53: Hmmm, maybe you added the above IF statement to avoid having the `hasXXX()` methods here? IMHO it's easier to read and makes more sense if you keep this IF statement as it was originally written and remove the redundant IF above for that case.

barbeau (commented):
Thanks @lionel-nj! A few comments in-line.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/FeedServiceDateValidator.java
lionel-nj: Thanks.

barbeau (approved):
👍

aababilov:
> (2) if one of said fields is provided, the other one should also be provided.

This is an interesting requirement. I cannot find it in the [spec](https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md#feed_infotxt) nor in [best practices](https://gtfs.org/best-practices/#feed_infotxt). Moreover, there is a remarkable amount of production feeds at Google that provide only one field and we accept them.

Would you agree to remove that requirement or at least relax it to warning? It should be probably another notice type so that we can disable it easily.

aababilov:
However, we may make clear in the GTFS spec that both fields must be either provided or not provided, and then we at Google can work with feed providers to fix their data.

lionel-nj:
> This is an interesting requirement. I cannot find it in the spec nor in best practices. Moreover, there is a remarkable amount of production feeds at Google that provide only one field and we accept them.

That rule comes from reaching parity with the Python validator (https://github.com/MobilityData/gtfs-validator/issues/138#issuecomment-641551704). I agree that since the specification does not explicitly require this, then we should use another type of notice (with severity being set to `WARNING`.


> However, we may make clear in the GTFS spec that both fields must be either provided or not provided, and then we at Google can work with feed providers to fix their data.

@MobilityData/transit-specs

PR #689, opened 2021-01-26
feat: verify unicity of routes.route_long_name and routes.route_short_name

### LINKED ISSUE #533: Duplicate `routes.route_short_name`
Validate `routes.route_short_name` unicity

**Additional context**
`v1.4.0` implementation:
- https://github.com/MobilityData/gtfs-validator/blob/v1.4.0/usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateRouteShortNameAreUnique.java
- tests: https://github.com/MobilityData/gtfs-validator/blob/v1.4.0/usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateRouteShortNameAreUniqueTest.java

### LINKED ISSUE #534: Duplicate `routes.route_long_name`
Validate unicity of `routes.route_long_name`

**Additional context**
`v1.4.0` implementation: 
- https://github.com/MobilityData/gtfs-validator/blob/v1.4.0/usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateRouteLongNameAreUnique.java
- tests: https://github.com/MobilityData/gtfs-validator/blob/v1.4.0/usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateRouteLongNameAreUniqueTest.java

lionel-nj:
closes #533
closes #534

**Summary:**

This PR provides code do implement the following validation rule:
- All routes should have different `routes.route_long_name`. If routes have the same `routes.route_long_name`, they must be different routes serving different areas; and must not be different trips of the same route or different directions of the same route.
Note that two routes can have the same `routes.route_long_name` if they do not belong to the same agency.

- All routes should have different `rouytes.route_short_name`. If routes have the same `routes.route_short_name`, they must be different routes serving different areas; and must not be different trips of the same route or different directions of the same route. 
Note that two routes can have the same `routes.route_short_name` if they do not belong to the same agency.

**Expected behavior:** 

- a notice should be generated for each route entity with duplicate `routes.route_long_name`
- a notice should be generated for each route entity with duplicate `routes.route_short_name`
- 

### DIFF
--- a/RULES.md
+++ b/RULES.md
@@ -71,8 +71,8 @@ Note that the notice ID naming conventions changed in `v2` to make contributions
 |                           | [W009](https://github.com/MobilityData/gtfs-validator/blob/v1.4.0/RULES.md#W009)        	| Dataset should cover at least the next 30 days of service                       	|
 | [`MissingRequiredFieldError`](#MissingRequiredFieldError) | [W010](https://github.com/MobilityData/gtfs-validator/blob/v1.4.0/RULES.md#W010)        	| `feed_end_date` should be provided if `feed_start_date` is provided             	|
 | [`MissingRequiredFieldError`](#MissingRequiredFieldError) | [W011](https://github.com/MobilityData/gtfs-validator/blob/v1.4.0/RULES.md#W011)        	| `feed_start_date` should be provided if `feed_end_date` is provided             	|
-|                           | [W014](https://github.com/MobilityData/gtfs-validator/blob/v1.4.0/RULES.md#W014)        	| Duplicate `routes.route_long_name`                                              	|
-|                           | [W015](https://github.com/MobilityData/gtfs-validator/blob/v1.4.0/RULES.md#W015)        	| Duplicate `routes.route_short_name`                                             	|
+|[`DuplicateRouteNameNotice`](#DuplicateRouteNameNotice)| [W014](https://github.com/MobilityData/gtfs-validator/blob/v1.4.0/RULES.md#W014)        	| Duplicate `routes.route_long_name`                                              	|
+|[`DuplicateRouteNameNotice`](#DuplicateRouteNameNotice)| [W015](https://github.com/MobilityData/gtfs-validator/blob/v1.4.0/RULES.md#W015)        	| Duplicate `routes.route_short_name`                                             	|
 |                           | [W016](https://github.com/MobilityData/gtfs-validator/blob/v1.4.0/RULES.md#W016)        	| Duplicate combination of fields `route_long_name` and `routes.route_short_name` 	|
 
 ## Notices
@@ -274,3 +274,14 @@ Field `parent_station` must be empty when `location_type` is 2.
 
 #### References:
 * [stops.txt specification](http://gtfs.org/reference/static/#stopstxt)
+
+### DuplicateRouteNameNotice
+
+All routes should have different `routes.route_long_name`. If routes have the same `routes.route_long_name`, they must be different routes serving different areas; and must not be different trips of the same route or different directions of the same route.
+Note that two routes can have the same `routes.route_long_name` if they do not belong to the same agency.
+
+All routes should have different `rouytes.route_short_name`. If routes have the same `routes.route_short_name`, they must be different routes serving different areas; and must not be different trips of the same route or different directions of the same route. 
+Note that two routes can have the same `routes.route_short_name` if they do not belong to the same agency.
+
+#### References:
+* [routes.txt specification](http://gtfs.org/reference/static/#routestxt)
--- a/RULES.md
+++ b/RULES.md
@@ -277,11 +277,13 @@ Field `parent_station` must be empty when `location_type` is 2.
 
 ### DuplicateRouteNameNotice
 
-All routes should have different `routes.route_long_name`. If routes have the same `routes.route_long_name`, they must be different routes serving different areas; and must not be different trips of the same route or different directions of the same route.
-Note that two routes can have the same `routes.route_long_name` if they do not belong to the same agency.
+All routes should have different routes.route_long_name - if two routes.route_long_name are the same, and the two routes belong to the same agency, a notice is generated.
 
-All routes should have different `rouytes.route_short_name`. If routes have the same `routes.route_short_name`, they must be different routes serving different areas; and must not be different trips of the same route or different directions of the same route. 
-Note that two routes can have the same `routes.route_short_name` if they do not belong to the same agency.
+Note that there may be valid cases where routes may have the same routes.route_long_name and this notice can be ignored. For example, routes may have the same routes.route_long_name if they serve difference areas. However, they must not be different trips of the same route or different directions of the same route - these cases should always have unique routes.route_long_name.
+
+All routes should have different routes.route_short_name - if two routes.route_short_name are the same, and the two routes belong to the same agency, a notice is generated.
+
+Note that there may be valid cases where routes may have the same routes.route_short_name and this notice can be ignored. For example, routes may have the same routes.route_short_name if they serve difference areas. However, they must not be different trips of the same route or different directions of the same route - these cases should always have unique routes.route_short_name.
 
 #### References:
 * [routes.txt specification](http://gtfs.org/reference/static/#routestxt)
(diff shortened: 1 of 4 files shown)


RULES.md
```diff
@@ -274,3 +274,14 @@ Field `parent_station` must be empty when `location_type` is 2.
 
 #### References:
 * [stops.txt specification](http://gtfs.org/reference/static/#stopstxt)
+
+### DuplicateRouteNameNotice
+
+All routes should have different `routes.route_long_name`. If routes have the same `routes.route_long_name`, they must be different routes serving different areas; and must not be different trips of the same route or different directions of the same route.
```
barbeau, line 280: I'd suggest being clearer here for the behavior of the validator vs. a caveat for when the notice may be detected in valid cases.

For example:

>All routes should have different `routes.route_long_name` - if two `routes.route_long_name` are the same, and the two routes belong to the same agency, a notice is generated.
>
>Note that there may be valid cases where routes may have the same `routes.route_long_name` and this notice can be ignored. For example, routes may have the same `routes.route_long_name` if they serve difference areas. However,  they must not be different trips of the same route or different directions of the same route - these cases should always have unique `routes.route_long_name`.

barbeau (commented):
Thanks @lionel-nj! Looks good - a few comments in-line.

RULES.md
lionel-nj: > Note that there may be valid cases where routes may have the same routes.route_long_name and this notice can be ignored.

The code is actually written in a way that such a situation does not generate notice. Does  that mean that we should generate the notice anyway? @barbeau

barbeau: @lionel-nj Which case does not generate a notice? We aren't checking if routes serve different areas...

lionel-nj: My bad -  I got confused, please disregard my previous comment @barbeau.

barbeau (approved):
👍  LGTM!

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateRouteNameValidator.java
```diff
@@ -0,0 +1,88 @@
...
+  /**
+   * Determines if two routes are from the same agency
+   *
+   * @param routeAgencyId first agency_id
+   * @param otherRouteAgencyId second agency_id
+   * @return true if both agency ids are equals returns false otherwise.
+   */
+  private boolean areRoutesFromSameAgency(
+      final String routeAgencyId, final String otherRouteAgencyId) {
+    return routeAgencyId.equalsIgnoreCase(otherRouteAgencyId);
```
aababilov, line 86: IDs are case-sensitive. Also, it may be useful to show warnings if several agencies have the same ID ignoring case.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateRouteNameValidator.java
```diff
@@ -0,0 +1,88 @@
...
+    final Map<String, GtfsRoute> routeByShortName = new HashMap<>(routeTable.entityCount());
+    routeTable
+        .getEntities()
+        .forEach(
+            route -> {
+              if (route.hasRouteLongName()) {
+                if (routeByRouteLongName.containsKey(route.routeLongName())) {
+                  if (areRoutesFromSameAgency(
+                      route.agencyId(),
+                      routeByRouteLongName.get(route.routeLongName()).agencyId())) {
```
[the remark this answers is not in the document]
lionel-nj: Do you mean just using `.get()` that will return `null` if the given key a key of the HashMap?

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateRouteNameValidator.java
barbeau: I remember debating if we should treat IDs as case-sensitive. So I assume Google treats them as case-sensitive?

In the spec, the only mention I could find is:

>All field names are case-sensitive.

...but it doesn't say anything about field values.

@MobilityData/transit-specs another issue that needs clarification in the spec.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateRouteNameValidator.java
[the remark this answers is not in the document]
barbeau: I think @aababilov is saying that `.containsKey()` and `.get()` both require a hash table lookup. 

`.containsKey()` can be functionally equivalent as checking if `.get()` returns `null`, because `.get()` returns `null` if the value isn't in the hash table. 

So you can reduce this to a single lookup by calling `.get()` initially and checking if it's null, and if not then using the value that was already returned by `.get()`.

So something like:

```java
                GtfsRoute cachedRoute = routeByLongName.get(route.routeLongName());
                if (cachedRoute != null) {
                  if (areRoutesFromSameAgency(
                      route.agencyId(), cachedRoute.agencyId())) {
```

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateRouteNameValidator.java
```diff
@@ -0,0 +1,88 @@
...
+                  if (areRoutesFromSameAgency(
+                      route.agencyId(), routeByShortName.get(route.routeShortName()).agencyId())) {
+                    noticeContainer.addValidationNotice(
+                        new DuplicateRouteNameNotice(
+                            "route_short_name", route.csvRowNumber(), route.routeId()));
+                  }
+                } else {
+                  routeByShortName.put(route.routeShortName(), route);
+                }
+              }
```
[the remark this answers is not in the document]
barbeau: @aababilov https://github.com/MobilityData/gtfs-validator/pull/726/files#diff-301d42f01987bbe9b969f2b4201f80e1176f141c40568b2aadbc7decdbe78377R32 is proposing to make this a WARNING anyway for now due to the spec language.

@lionel-nj Could you please open an issue on this and we can revisit how to improve this?

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateRouteNameValidator.java
timMillet: The spec says: "An ID field value [...] is a sequence of any UTF-8 characters." ALL-CAP letters and small-cap letters are different UTF-8 characters. So IMHO, ID values are implicitly case sensitive. Is there an ERROR or a WARNING issued in the validator on this matter?

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateRouteNameValidator.java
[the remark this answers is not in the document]
lionel-nj: Thanks @barbeau !

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateRouteNameValidator.java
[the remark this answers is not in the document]
lionel-nj: Opened https://github.com/MobilityData/gtfs-validator/issues/754 on this matter.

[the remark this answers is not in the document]
lionel-nj: This notice has been made a `WARNING`, see https://github.com/MobilityData/gtfs-validator/blob/a5155d58b1a2be6cf996be4c8bc1b666cb845d14/main/src/main/java/org/mobilitydata/gtfsvalidator/notice/DuplicateRouteNameNotice.java#L25.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateRouteNameValidator.java
lionel-nj: @timMillet as discussed, at present the validator does not issue any notice if two agencies have the same value for `agency.agency_id`. As you mentioned, `route id VALUE` and `route ID vaLUe` are two different strings and should not be considered as equal. I'm opening an issue on this matter to improve de validator's behavior.

lionel-nj:
Closing as:

> we can work on more targeted solutions within validators (see some thoughts in https://github.com/MobilityData/gtfs-validator/pull/765#issuecomment-785200442) when we know that bad data could influence a validator's output.

PR #604, opened 2021-01-21
chore: Apply google-java-format to all files for consistency

aababilov:
See https://github.com/google/google-java-format.

$ google-java-format -i $(find . -name '*java')

PR #690, opened 2021-01-27
fix: Support trivial enum that have a single value (#605)

aababilov:
Google has use cases for that.

This restores https://github.com/MobilityData/gtfs-validator/commit/dcbe2f33371a319ed1d7e0df7565e769928ffa46
that was partially reverted in https://github.com/MobilityData/gtfs-validator/pull/604


barbeau (approved):
Thanks for fixing this!

The bot for apply the Google formatter is failing on forks, which we thought may happen.

@aababilov Just curious - did you allow the option for maintainers to edit PRs on GitHub? I'm guessing so, but the bot just doesn't have correct permissions. In which case we'll need to move it to formatting on merge to master.

PR #691, opened 2021-01-27
fix: Change enum name to GtfsStopTimeTimepointEnum

aababilov:
Old name: GtfsStopTimesTimepointEnum
New name: GtfsStopTimeTimepointEnum

By convention, all table names should be in singular in schema
definition.


barbeau (approved):
LGTM, thanks @aababilov!

PR #692, opened 2021-01-27
feat: Support CSV files that start with UTF-8 BOM

aababilov:
GTFS spec says:

Files should be encoded in UTF-8 to support all Unicode characters.
Files that include the Unicode byte-order mark (BOM) character are
acceptable. See http://unicode.org/faq/utf_bom.html#BOM for more
information on the BOM character and UTF-8.

The Unicode Standard permits the BOM in UTF-8, but does not require
or recommend its use. Byte order has no meaning in UTF-8, so its only
use in UTF-8 is to signal at the start that the text stream is
encoded in UTF-8, or that it was converted to UTF-8 from a stream that
contained an optional BOM.


aababilov:
Why do we run tests if they just fail for valid production code?

barbeau:
Closing and re-opening to re-trigger checks...

barbeau (approved):
👍  Thanks @aababilov!

PR #694, opened 2021-01-27
feat: check feed expiration date

### LINKED ISSUE #515: Dataset should cover at least the next 7 days after `feed_info.start_date`
Validate a dataset is valid for at least the next 7 days after `feed_info.start_date`

**Additional context**
`v1.4.0` implementation:
- https://github.com/MobilityData/gtfs-validator/blob/v1.4.0/usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateFeedCoversTheNext7ServiceDays.java
- tests: https://github.com/MobilityData/gtfs-validator/blob/v1.4.0/usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateFeedCoversTheNext7ServiceDaysTest.java

### LINKED ISSUE #530: Dataset should cover at least the next 30 days of service
Validate dataset covers at least the next 30 days of service

**Additional context**
`v1.4.0` implementation: 
- https://github.com/MobilityData/gtfs-validator/blob/v1.4.0/usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateFeedCoversTheNext30ServiceDays.java
- tests: https://github.com/MobilityData/gtfs-validator/blob/v1.4.0/usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateFeedCoversTheNext30ServiceDaysTest.java

lionel-nj:
closes #530 
closes #515

**Summary:**

This PR implements the following validation rules:
1. At any time, the published GTFS dataset should be valid for at least the next 7 days, 
2. If possible, the GTFS dataset should cover at least the next 30 days of service.

**Expected behavior:** 

A `FeedExpirationDateNotice` should be generated when:
- a feed expires less than 7 days from now,
- a feed expires in 7 exactly days from now,
- a feed expires between 7 and 30 days from now.

The information contained in the notice changes according to the expiry date.


RULES.md
```diff
@@ -68,7 +68,7 @@ Note that the notice ID naming conventions changed in `v2` to make contributions
 |                           | [W006](https://github.com/MobilityData/gtfs-validator/blob/v1.4.0/RULES.md#W006)        	| Missing route short name                                                        	|
 |                           | [W007](https://github.com/MobilityData/gtfs-validator/blob/v1.4.0/RULES.md#W007)        	| Missing route long name                                                         	|
 |                           | [W008](https://github.com/MobilityData/gtfs-validator/blob/v1.4.0/RULES.md#W008)        	| Route long name contains short name                                             	|
-|                           | [W009](https://github.com/MobilityData/gtfs-validator/blob/v1.4.0/RULES.md#W009)        	| Dataset should cover at least the next 30 days of service                       	|
+|[`FeedExpirationDateNotice`](#FeedExpirationDateNotice)| [W009](https://github.com/MobilityData/gtfs-validator/blob/v1.4.0/RULES.md#W009)        	| Dataset should cover at least the next 30 days of service                       	|
```
barbeau, line 71: For multiple rules in v1 that are being condensed under v2, I'd only list one row in the table that references both rules. IMHO our goal should be to have a compact table for v2.

So something like this:

| Error ID (v2.0+) | Error ID (v1.x) 	| Error Title                                                             	|
|--------------------------	|--------------------	|-------------------------------------------------------------------------	|
|[`FeedExpirationDateNotice`](#FeedExpirationDateNotice)| [W009](https://github.com/MobilityData/gtfs-validator/blob/v1.4.0/RULES.md#W009), [E040](https://github.com/MobilityData/gtfs-validator/blob/v1.4.0/RULES.md#E040)        	| Dataset should cover at least the next 30 days of service. Dataset should be valid for at least the next 7 days.

main/src/main/java/org/mobilitydata/gtfsvalidator/notice/FeedExpirationDateNotice.java
```diff
@@ -0,0 +1,41 @@
+/*
+ * Copyright 2020 Google LLC, MobilityData IO
```
barbeau, line 2: If Google didn't contribute any code to this file, the copyright should just be MobilityData.

main/src/main/java/org/mobilitydata/gtfsvalidator/notice/FeedExpirationDateNotice.java
```diff
@@ -0,0 +1,41 @@
...
+      long csvRowNumber, GtfsDate currentDate, GtfsDate feedEndDate, GtfsDate suggestedExpiryDate) {
+    super(
+        ImmutableMap.of(
+            "csvRowNumber",
+            csvRowNumber,
+            "currentDate",
+            currentDate.toYYYYMMDD(),
+            "feedEndDate",
+            feedEndDate.toYYYYMMDD(),
+            "suggestedExpiryDate",
```
barbeau, line 33: I'd use `suggestedExpirationDate` instead, unless that's too long

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/FeedExpirationDateValidator.java
```diff
@@ -0,0 +1,67 @@
...
+      GtfsDate currentDate = GtfsDate.fromLocalDate(LocalDate.now());
+      GtfsDate currentDatePlusSevenDays = GtfsDate.fromLocalDate(LocalDate.now().plusDays(7));
+      GtfsDate currentDatePlusThirtyDays = GtfsDate.fromLocalDate(LocalDate.now().plusDays(30));
+      if (entity.feedEndDate().isBefore(currentDatePlusSevenDays)) {
+        noticeContainer.addValidationNotice(
+            new FeedExpirationDateNotice(
+                entity.csvRowNumber(),
+                currentDate,
+                entity.feedEndDate(),
+                currentDatePlusSevenDays));
```
barbeau, line 52: The way this notice is structured now, I think it would make the code easier to read if you `return;` here, and then you don't have to check for `.equals()` and `isAfter()` the 7 day variable below. Because if it's before 7 days, it's definitely before 30 days.

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/FeedExpirationDateValidatorTest.java
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  @Test
+  public void feedExpiringInFiveDaysFromNowShouldGenerateNotice() {
+    assertThat(
+            validateFeedInfo(createFeedInfo(GtfsDate.fromLocalDate(LocalDate.now().plusDays(3)))))
+        .containsExactly(
+            new FeedExpirationDateNotice(
+                1,
+                GtfsDate.fromLocalDate(LocalDate.now()),
```
barbeau, line 57: Instead of using now and `plusDays()`, which mirrors the implementation, I would instead use exact dates. IMHO this way the test is always executing on the exact same dates, and the intent on whether the 7 and 30 days is inclusive or not is clearer. You're also more likely to catch bugs if you don't mirror the original implementation here.

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/FeedExpirationDateValidatorTest.java
```diff
@@ -0,0 +1,104 @@
...
+    FeedExpirationDateValidator validator = new FeedExpirationDateValidator();
+    validator.validate(feedInfo, container);
+    return container.getValidationNotices();
+  }
+
+  private GtfsFeedInfo createFeedInfo(GtfsDate feedEndDate) {
+    return new GtfsFeedInfo.Builder()
+        .setCsvRowNumber(1)
+        .setFeedPublisherName("feed publisher name value")
+        .setFeedPublisherUrl("www.mobilitydata.org")
```
barbeau, line 44: Should the URL be fully qualified with `https://`?

barbeau (commented):

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/FeedExpirationDateValidator.java
lionel-nj: True! Modifying this, thanks!

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/FeedExpirationDateValidatorTest.java
lionel-nj: Yes, modifying it.

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/FeedExpirationDateValidatorTest.java
lionel-nj: After discussion with @barbeau, we agreed on keeping the tests the way they are written because: `LocalDate.now()` is used in the use case implementation, and the current implementation does not seem to allow us to easily inject `GtfsDate` as a parameter of the use case. We will be revisiting this after the release of `v2.0`.

barbeau:
@lionel-nj Looks like this PR still has conflicts - could you resolve?

lionel-nj:
> @lionel-nj Looks like this PR still has conflicts - could you resolve?

Just noticed that. Done✅

barbeau (approved):
👍

aababilov:
Unfortunately, this code fails to compile at Google.

```
main/src/main/java/org/mobilitydata/gtfsvalidator/validator/FeedExpirationDateValidator.java:43: error: [JavaTimeDefaultTimeZone] LocalDate.now() is not allowed because it silently uses the system default time-zone. You must pass an explicit time-zone (e.g., ZoneIds.googleZoneId()) to this method.
      GtfsDate currentDate = GtfsDate.fromLocalDate(LocalDate.now());
                                                                 ^
main/src/main/java/org/mobilitydata/gtfsvalidator/validator/FeedExpirationDateValidator.java:44: error: [JavaTimeDefaultTimeZone] LocalDate.now() is not allowed because it silently uses the system default time-zone. You must pass an explicit time-zone (e.g., ZoneIds.googleZoneId()) to this method.
      GtfsDate currentDatePlusSevenDays = GtfsDate.fromLocalDate(LocalDate.now().plusDays(7));
                                                                              ^
main/src/main/java/org/mobilitydata/gtfsvalidator/validator/FeedExpirationDateValidator.java:45: error: [JavaTimeDefaultTimeZone] LocalDate.now() is not allowed because it silently uses the system default time-zone. You must pass an explicit time-zone (e.g., ZoneIds.googleZoneId()) to this method.
      GtfsDate currentDatePlusThirtyDays = GtfsDate.fromLocalDate(LocalDate.now().plusDays(30));
                                                                               ^
```

barbeau:
@aababilov I supposed we could instantiate `LocalDate` using agency.txt agency_timezone to fix this?

It's a required field so it should be in every GTFS dataset.

aababilov:
This is a situation when nobody was unable to predict that problem: neither Lionel nor me. The only way to find this compilation error was to pull the fresh code to the Google repository and compile it there. We at Google may setup an automated process of pulling and compiling every fresh commit from GitHub to detect problems earlier - or we may find the pivot using binary search.

aababilov:
Now let's discuss how we can make this compile.

We should pass the "now" as a parameter to validators - it must be a part of a certain validation context. That context may also include country code (and feed name). It should be possible to override "now" from the command-line flag. And this "now" may default to the local timezone (I will find a portable way to find that date).

So, we need a certain ValidationContext object or so with read-only information regarding the feed. This object will be passed as a parameter to the `validate()` method - or it may be injected on demand since it is rarely used.

Alternative 1.

```
public void validate(GtfsFeedInfo entity, ValidationContext context, NoticeContainer noticeContainer) {
```

Alternative 2.
```
@Inject ValidationContext context;
```

I personally vote for the second alternative. Thoughts?

lionel-nj:
I would also opt for the second option also as it would induce less code refactor.

barbeau:
>We should pass the "now" as a parameter to validators - it must be a part of a certain validation context. That context may also include country code (and feed name). It should be possible to override "now" from the command-line flag. And this "now" may default to the local timezone (I will find a portable way to find that date).

👍   @lionel-nj and I briefly discussed this approach - it's similar to how we handled this case in the GTFS RT validator

>`@Inject ValidationContext context;`

I prefer this as well 👍

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/FeedExpirationDateValidator.java
```diff
@@ -0,0 +1,66 @@
...
+        return;
+      }
+      if (entity.feedEndDate().equals(currentDatePlusThirtyDays)
+          || entity.feedEndDate().isBefore(currentDatePlusThirtyDays)) {
```
aababilov, line 56: You do not have to compare the dates twice - just use the `compareTo()` method.

Analogy: you have
```
if (a == b || a < b) 
```

I am asking you to write:
```
if (a <= b)
```

Issue #591, opened 2021-01-12
CLI tests fail randomly on CI


barbeau:
**Summary**

We're seeing the `provideStorageDirectoryCliParameterWithoutSpecifyingUrlShouldReturnFalse()` test randomly fail on CI. For example:
https://github.com/MobilityData/gtfs-validator/runs/1685201080

```
org.mobilitydata.gtfsvalidator.cli.CliParametersAnalyzerTest > provideStorageDirectoryCliParameterWithoutSpecifyingUrlShouldReturnFalse FAILED
    org.mockito.exceptions.verification.WantedButNotInvoked at CliParametersAnalyzerTest.java:113
```

This passes for me locally in IntelliJ.

**What I expect**

CI should be predictable and always pass unless there are bugs in the code

**What I see**

CI fails seemingly randomly

barbeau:
Rerunning the test, and now it passes...

https://github.com/MobilityData/gtfs-validator/runs/1689871565

barbeau:
The [line](https://github.com/MobilityData/gtfs-validator/blob/bf9d681e6361e6fe3cbcacd4fc41475035917cdc/main/src/test/java/org/mobilitydata/gtfsvalidator/cli/CliParametersAnalyzerTest.java#L113) that is failing here is:

```java
        verify(mockHandler).publish(logRecordCaptor.capture());
```

...which, to my understanding, is just grabbing the output log. So maybe the log isn't being output synchronously and occasionally takes longer to output, resulting in occasional failures here?

@aababilov Any ideas of why this would occasionally fail?

barbeau:
It looks like we're having a similar issue with this test now as well:

`provideUrlAndInputCliParametersShouldReturnFalse()`

See what seems like a random failure here - https://github.com/MobilityData/gtfs-validator/runs/1780031307

PR #700, opened 2021-01-27
fix: Comment out flaky provideUrlAndInputCliParametersShouldReturnFalse()

barbeau:
**Summary:**

It currently randomly fails on CI. See https://github.com/MobilityData/gtfs-validator/issues/591.

**Expected behavior:** 

Tests shouldn't randomly fail on CI


lionel-nj (approved):
Thanks @barbeau, I will be working on fixing these tests shortly.

PR #705, opened 2021-02-01
feat: Do not skip a row if an optional field is invalid

aababilov:
GTFS validator tries to recover after parsing errors, so that it can
demonstrate more validation notices.

The old behaviour.
If any field fails to parse (e.g., invalid integer etc.), then the
whole row is skipped (e.g., we skip the whole trip entity).
This leads to a cascade of foreign key errors since other tables may
reference this invalid row (e.g., stop_times may reference this trip).

The new behaviour.
If an OPTIONAL field fails to parse, we emit a validation error
(e.g., FieldParsingError), treat this field as missing and still accept
this row. That allows to keep validating the feed.
However, if a REQUIRED field fails to parse, then we still skip the
whole row since we do not want to read an invalid entity from it.


aababilov:
I think I have a better solution.

PR #706, opened 2021-02-01
feat: Abort validation if some files failed to parse

aababilov:
Results of file validations and cross-file validations require all feed
data to be parsed and loaded to memory. This is possible that some
required files are missing or some rows or headers failed to parse,
e.g., a required field is missing or a string cannot be parsed as integer.

This change is aborting validation process if some files failed to parse
to avoid misleading validation notices. The change also removes amount
of boilerplate regarding GTFS table status (empty file, missing file
etc.).

Example.

Consider we failed to parse a row trip.txt but there is another row
in stop_times.txt that references a trip. Then foreign key validator
may notify about a missing trip_id which would be very misleading for
the feed provider.


barbeau (approved):
Thanks @aababilov, LGTM!

aababilov:
Thanks for you kind and prompt review, Sean!

PR #707, opened 2021-02-01
feat: Inject a context object to validators

aababilov:
ValidationContext is a read-only object that describes the feed as
a whole and also some system properties relevant for validation.

Any validator may request to inject ValidationContext using the usual
@Inject annotation. We chose that instead of modifying the signature of
validate() method because most validators do not need this context and
we want to avoid a huge refactoring.

### DIFF
--- a/core/src/main/java/org/mobilitydata/gtfsvalidator/validator/ValidationContext.java
+++ b/core/src/main/java/org/mobilitydata/gtfsvalidator/validator/ValidationContext.java
@@ -1,0 +1,41 @@
+package org.mobilitydata.gtfsvalidator.validator;
+
+import com.google.auto.value.AutoValue;
+import java.time.ZonedDateTime;
+import org.mobilitydata.gtfsvalidator.input.GtfsFeedName;
+
+/**
+ * A read-only context passed to particular validator objects. It gives information relevant for
+ * validation: properties of the feed as a whole, system properties (current time) etc.
+ */
+@AutoValue
+public abstract class ValidationContext {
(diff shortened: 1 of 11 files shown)


barbeau:
This addresses the issue first mentioned at https://github.com/MobilityData/gtfs-validator/pull/694#issuecomment-768625303 for a failed compilation at Google due to not passing an explicit timezone.

barbeau (approved):
@aababilov Thanks, this LGTM! Could you please resolve the conflicts and I'll merge?

aababilov:
Thank you, Sean! I resolved the conflicts and uploaded the fresh code.

barbeau:
👍

aababilov:
Thanks a lot!

core/src/main/java/org/mobilitydata/gtfsvalidator/validator/ValidationContext.java
```diff
@@ -0,0 +1,41 @@
...
+  public abstract GtfsFeedName feedName();
+
+  /**
+   * The time when validation started.
+   *
+   * <p>Validator code should use this property instead of calling LocalDate.now() etc. for the
+   * following reasons:
+   *
+   * <ul>
+   *   <li>current date and time is changing but it should not randomly affect validation notices;
```
f8full, line 27: Super valid points.
🗨️  Reminds me 3D rendering loops having multiple methods to serve time
https://docs.unity3d.com/Manual/TimeFrameManagement.html 🗨️

PR #708, opened 2021-02-01
feat:  Stop too far from trip shape

### LINKED ISSUE #521: Validate Stop not too far from trip shape
Validate stops are not too far from their trip shape

**Additional context**
`v1.4.0` implementation:
- https://github.com/MobilityData/gtfs-validator/blob/v1.4.0/usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateStopTooFarFromTripShape.java
- https://github.com/MobilityData/gtfs-validator/blob/v1.4.0/usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateStopTooFarFromTripShapeNoticeTest.java

lionel-nj:
closes #521 

**Summary:**

> This PR implements the rule to check when a stop is too far from the trip shape.
> 
> [GTFS Best Practices](https://gtfs.org/best-practices/#shapestxt) say:
> 
> >Route alignments (in shapes.txt) should be within 100 meters of stop locations which a trip serves.
> 
> Data used in unit tests is available as GeoJSON on GitHub here (GitHub provides a toggle between text and map view):
> https://gist.github.com/barbeau/d9c0b90a26a3e2ba105cae5f0e8aec4a
> 
> image
> 
> If desired, you can visualize the buffer by exporting a JTS-version of the buffer in WKT format using code at https://gist.github.com/barbeau/d9c0b90a26a3e2ba105cae5f0e8aec4a#gistcomment-3425554 and then visualizing the WKT using https://arthur-e.github.io/Wicket/sandbox-gmaps3.html.
> 
> image
> 
> The spatial4j version of the buffer can't easily be visualized using GeoJSON or WKT because it uses a LineString and a proprietary "buffer" extension to GeoJSON and WKT, which most tools don't support.
> 

from #340 

New notice: `StopTooFarFromTripShapeNotice`
New util class: `GeospatialUtil`

### DIFF
--- a/main/src/main/java/org/mobilitydata/gtfsvalidator/util/GeospatialUtil.java
+++ b/main/src/main/java/org/mobilitydata/gtfsvalidator/util/GeospatialUtil.java
@@ -1,0 +1,161 @@
+/*
+ * Copyright 2021 MobilityData IO
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.mobilitydata.gtfsvalidator.util;
+
+import static org.locationtech.spatial4j.context.SpatialContext.GEO;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Set;
+import org.locationtech.spatial4j.distance.DistanceCalculator;
+import org.locationtech.spatial4j.distance.DistanceUtils;
+import org.locationtech.spatial4j.shape.Point;
+import org.locationtech.spatial4j.shape.Shape;
+import org.locationtech.spatial4j.shape.ShapeFactory;
+import org.locationtech.spatial4j.shape.SpatialRelation;
+import org.mobilitydata.gtfsvalidator.notice.StopTooFarFromTripShapeNotice;
+import org.mobilitydata.gtfsvalidator.table.GtfsLocationType;
+import org.mobilitydata.gtfsvalidator.table.GtfsShape;
+import org.mobilitydata.gtfsvalidator.table.GtfsStop;
+import org.mobilitydata.gtfsvalidator.table.GtfsStopTableContainer;
+import org.mobilitydata.gtfsvalidator.table.GtfsStopTime;
+import org.mobilitydata.gtfsvalidator.table.GtfsTrip;
+
+public class GeospatialUtil {
+  static final int KILOMETER_TO_METER_CONVERSION_FACTOR =
+      1000; // conversion factor from kilometers to meters
+  static final double TRIP_BUFFER_METERS =
+      100; // Per GTFS Best Practices (https://gtfs.org/best-practices/#shapestxt)
+  public static final double TRIP_BUFFER_DEGREES =
+      DistanceUtils.KM_TO_DEG * (TRIP_BUFFER_METERS / 1000.0d);
+
+  private GeospatialUtil() {}
+
+  /**
+   * Method returning a {@code ShapeFactory}
+   *
+   * @return a {@link ShapeFactory}
+   */
+  private static ShapeFactory getShapeFactory() {
+    return GEO.getShapeFactory();
+  }
+
+  /**
+   * Method returning a {@code DistanceCalculator}
+   *
+   * @return a {@link DistanceCalculator}
+   */
+  private static DistanceCalculator getDistanceCalculator() {
+    return GEO.getDistCalc();
+  }
+
+  /**
+   * This method calculates distance between two sets of coordinates. Result is expressed in meters.
+   *
+   * @param fromLat latitude of the first coordinates
+   * @param fromLng longitude of the first coordinates
+   * @param toLat latitude of the second coordinates
+   * @param toLng longitude of the second coordinates
+   * @return the calculation result in meters
+   */
+  public static int distanceBetweenMeter(
+      double fromLat, double fromLng, double toLat, double toLng) {
+    final ShapeFactory shapeFactory = getShapeFactory();
+    final DistanceCalculator distanceCalculator = getDistanceCalculator();
+    final Point origin = shapeFactory.pointXY(fromLng, fromLat);
+    final Point destination = shapeFactory.pointXY(toLng, toLat);
+    return (int)
+        (DistanceUtils.DEG_TO_KM
+            * distanceCalculator.distance(origin, destination)
+            * KILOMETER_TO_METER_CONVERSION_FACTOR);
+  }
+
+  /**
+   * Returns a list of E052 errors for the given input, one for each stop that is too far from the
+   * trip shapePoints
+   *
+   * @param trip Trip for this GTFS trip
+   * @param stopTimes a map of StopTimes for a trip, sorted by stop_sequence
+   * @param shapePoints a map of ShapePoints for a trip, sorted by shape_pt_sequence
+   * @param stopTable a map of all stops (keyed on stop_id), needed to obtain the latitude and
+   *     longitude for each stop
+   * @param testedCache a cache for previously tested shape_id and stop_id pairs (keyed on
+   *     shape_id+stop_id). If the combination of shape_id and stop_id appears in this set, we
+   *     shouldn't test it again. Shapes and stops tested in this method execution will be added to
+   *     this testedCache.
+   * @return a list of E052 errors, one for each stop that is too far from the trip shapePoints
+   */
+  public static List<StopTooFarFromTripShapeNotice> checkStopsWithinTripShape(
+      final GtfsTrip trip,
+      final List<GtfsStopTime> stopTimes,
+      final List<GtfsShape> shapePoints,
+      final GtfsStopTableContainer stopTable,
+      final Set<String> testedCache) {
+    List<StopTooFarFromTripShapeNotice> errors = new ArrayList<>();
+    if (trip == null
+        || stopTimes == null
+        || stopTimes.isEmpty()
+        || shapePoints == null
+        || shapePoints.isEmpty()) {
+      // Nothing to do - return empty list
+      return errors;
+    }
+
+    // Create a polyline from the GTFS shapes data
+    ShapeFactory.LineStringBuilder lineBuilder = getShapeFactory().lineString();
+    shapePoints.forEach(
+        (shapePoint -> lineBuilder.pointXY(shapePoint.shapePtLon(), shapePoint.shapePtLat())));
+    Shape shapeLine = lineBuilder.build();
+
+    // Create the buffered version of the trip as a polygon
+    Shape shapeBuffer = shapeLine.getBuffered(TRIP_BUFFER_DEGREES, shapeLine.getContext());
+
+    // Check if each stop is within the buffer polygon
+    stopTimes.forEach(
+        stopTime -> {
+          GtfsStop stop = stopTable.byStopId(stopTime.stopId());
+          if (stop == null || stop.hasStopLat() || stop.hasStopLon()) {
+            // Lat/lon are optional for location_type 4 - skip to the next stop if they aren't
+            // provided
+            return;
+          }
+          if (!(stop.locationType() == GtfsLocationType.STOP)
+              && !(stop.locationType() == GtfsLocationType.BOARDING_AREA)) {
+            // This rule only applies to stops of location_type 0 and 4 - skip to next stop
+            return;
+          }
+          if (testedCache.contains(trip.shapeId() + stop.stopId())) {
+            // We've already tested this combination of shape ID and stop ID - skip to next stop
+            return;
+          }
+          testedCache.add(trip.shapeId() + stop.stopId());
+          org.locationtech.spatial4j.shape.Point p =
+              getShapeFactory().pointXY(stop.stopLon(), stop.stopLat());
+          if (!shapeBuffer.relate(p).equals(SpatialRelation.CONTAINS)) {
+            errors.add(
+                new StopTooFarFromTripShapeNotice(
+                    stopTime.stopId(),
+                    stopTime.stopSequence(),
+                    trip.tripId(),
+                    trip.shapeId(),
+                    TRIP_BUFFER_METERS));
+          }
+        });
+
+    return errors;
+  }
+}
--- a/main/src/main/java/org/mobilitydata/gtfsvalidator/util/GeospatialUtil.java
+++ b/main/src/main/java/org/mobilitydata/gtfsvalidator/util/GeospatialUtil.java
@@ -38,10 +38,13 @@ import org.mobilitydata.gtfsvalidator.table.GtfsTrip;
 public class GeospatialUtil {
   static final int KILOMETER_TO_METER_CONVERSION_FACTOR =
       1000; // conversion factor from kilometers to meters
+
+  static final double METER_TO_KILOMETER_CONVERSION_FACTOR =
+      1 / 1000; // conversion factor from kilometers to meters
   static final double TRIP_BUFFER_METERS =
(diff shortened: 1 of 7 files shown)


main/src/test/java/org/mobilitydata/gtfsvalidator/util/GeospatialUtilTest.java
```diff
@@ -0,0 +1,650 @@
...
+            createShapePoint(5, "shape id", 28.05724310653972D, -82.41350776611507D, 1, 400f),
+            createShapePoint(6, "shape id", 28.05746701492806D, -82.41493135129478D, 2, 400f),
+            createShapePoint(7, "shape id", 28.05800068503469f, -82.4159394137605D, 3, 400f),
+            createShapePoint(8, "shape id", 28.05808869825447D, -82.41648754043338D, 4, 400f),
+            createShapePoint(9, "shape id", 28.05809979887893D, -82.41773971025437D, 5, 400f));
+
+    GtfsStopTableContainer stopTable =
+        GtfsStopTableContainer.forEntities(
+            ImmutableList.of(
+                createStop(2, "1001", 28.05811731042478D, -82.41616877502503D, 2),
```
maximearmstrong, line 131: Should not the stops be of `location_type` 0 or 4?

main/src/test/java/org/mobilitydata/gtfsvalidator/util/GeospatialUtilTest.java
```diff
@@ -0,0 +1,650 @@
...
+    List<GtfsStopTime> stopTimes =
+        ImmutableList.of(createStopTime(5, "t1", "1001", 1), createStopTime(8, "t1", "1002", 2));
+
+    // No shapes.txt data
+    List<GtfsShape> shapePoints = null;
+
+    GtfsStopTableContainer stopTable =
+        GtfsStopTableContainer.forEntities(
+            ImmutableList.of(
+                createStop(2, "1001", 28.05811731042478D, -82.41616877502503D, 2),
```
maximearmstrong, line 278: Same as above regarding `location_type`

main/src/test/java/org/mobilitydata/gtfsvalidator/util/GeospatialUtilTest.java
```diff
@@ -0,0 +1,650 @@
...
+            createShapePoint(6, "shape id", 28.05746701492806D, -82.41493135129478D, 2, 400f),
+            createShapePoint(7, "shape id", 28.05800068503469f, -82.4159394137605D, 3, 400f),
+            createShapePoint(8, "shape id", 28.05808869825447D, -82.41648754043338D, 4, 400f),
+            createShapePoint(9, "shape id", 28.05809979887893D, -82.41773971025437D, 5, 400f));
+
+    GtfsStopTableContainer stopTable =
+        GtfsStopTableContainer.forEntities(
+            ImmutableList.of(
+                // this location is inside buffer
+                createStop(4, "1002", 28.05812364854794D, -82.41617370439423D, 2)),
```
maximearmstrong, line 423: Same as above

maximearmstrong (changes requested):

main/src/test/java/org/mobilitydata/gtfsvalidator/util/GeospatialUtilTest.java
lionel-nj: Indeed, thanks!

maximearmstrong (approved):

maximearmstrong:
LGTM

aababilov:
Thank you for that great contribution, Lionel!

I see that you added a dependency on spatial4j which is not available on Google, so I cannot compile the fresh validator. I will evaluate the alternatives.

Please may I ask you to let me know when you are adding a new dependency before you merge you commits? The thing is that when I cannot compile Validator at Google, that is a blocker for me.

aababilov:
We had a similar issue with https://github.com/MobilityData/gtfs-validator/pull/541 that added org.apache.httpcomponents.client5 which is banned at Google.

lionel-nj:
Oh, apologies @aababilov. 

> Please may I ask you to let me know when you are adding a new dependency before you merge you commits? The thing is that when I cannot compile Validator at Google, that is a blocker for me.

Definitely yes! We will proceed this way in the future if that is okay with @barbeau and @carlfredl.

main/src/main/java/org/mobilitydata/gtfsvalidator/util/GeospatialUtil.java
```diff
@@ -0,0 +1,161 @@
...
+  /**
+   * This method calculates distance between two sets of coordinates. Result is expressed in meters.
+   *
+   * @param fromLat latitude of the first coordinates
+   * @param fromLng longitude of the first coordinates
+   * @param toLat latitude of the second coordinates
+   * @param toLng longitude of the second coordinates
+   * @return the calculation result in meters
+   */
+  public static int distanceBetweenMeter(
```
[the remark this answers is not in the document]
f8full: If I may suggest so, there's some soft convention to put units at the end. It's definitively something that just requires consistency across the whole code base.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
```diff
@@ -0,0 +1,78 @@
...
+    List<StopTooFarFromTripShapeNotice> notices = new ArrayList<>();
+
+    // Cache for previously tested shape_id and stop_id pairs - we don't need to test them more than
+    // once
+    final Set<String> testedCache = new HashSet<>();
+
+    stopTimeTable
+        .byTripIdMap()
+        .forEach(
+            (tripId, tripStopTimes) -> {
```
aababilov, line 59: Here you are checking every trip as many times as you have it in stop_times.txt. That is a bug in the algorithm and a serious performance issue.

What was the size of the largest feed that you used for testing? I used a medium-size feed with 1 million stop_times.txt and this code runs for several minutes on it. If we add a large enough feed to the test set, we can catch such bugs earlier.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
```diff
@@ -0,0 +1,78 @@
...
+              List<StopTooFarFromTripShapeNotice> noticesForTrip =
+                  GeospatialUtil.checkStopsWithinTripShape(
+                      trip,
+                      stopTimeTable.byTripId(tripId),
+                      shapeTable.byShapeId(trip.shapeId()),
+                      stopTable,
+                      testedCache);
+              notices.addAll(noticesForTrip);
+            });
+    notices.forEach(noticeContainer::addValidationNotice);
```
aababilov, line 76: What is the rationale of placing all notices to a single container instead of emitting them inside the lambda function?

main/src/main/java/org/mobilitydata/gtfsvalidator/util/GeospatialUtil.java
```diff
@@ -0,0 +1,163 @@
...
+import org.mobilitydata.gtfsvalidator.table.GtfsStop;
+import org.mobilitydata.gtfsvalidator.table.GtfsStopTableContainer;
+import org.mobilitydata.gtfsvalidator.table.GtfsStopTime;
+import org.mobilitydata.gtfsvalidator.table.GtfsTrip;
+
+public class GeospatialUtil {
+  static final int KILOMETER_TO_METER_CONVERSION_FACTOR =
+      1000; // conversion factor from kilometers to meters
+  static final float METER_TO_KILOMETER_CONVERSION_FACTOR =
+      1 / 1000; // conversion factor from kilometers to meters
```
aababilov, line 42: You are dividing an integer by integer, so the result is [rounded](https://stackoverflow.com/questions/37795248/integer-division-in-java) and it is 0 (zero). You probably wanted to write `1.0 / 1000` instead.

When I executed your code on a real feed, it produced 8521 errors for 68157 trips, so that looked suspicious.

What was the size and diversity of the feeds that you use? Do you mind to cover that bug with unit tests?

Also, why are you using `float` instead of `double` here?

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
aababilov: Another way to find this bug earlier was to see that you are producing the same notice multiple times. Can you add unit tests for that?

f8full: > If we add a large enough feed to the test set, we can catch such bugs earlier.

I think that might explains what happens with Buenos Aires
image
image
https://github.com/MobilityData/gtfs-validator/runs/1881156285?check_suite_focus=true

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
lionel-nj: It is true that this could be changed to emit notices in the lambda function. Thanks for flagging that @aababilov.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
lionel-nj: This piece of code origins from: https://github.com/MobilityData/gtfs-validator/pull/340. 
@barbeau could you mind taking a look at this please?

main/src/main/java/org/mobilitydata/gtfsvalidator/util/GeospatialUtil.java
lionel-nj: @aababilov we originally tested the validator against 7 datasets. The biggest dataset in terms of `stop_times.txt` is from STM (https://openmobilitydata.org/p/societe-de-transport-de-montreal/39/latest/download) and contains +/- 5.6 millions rows in `stop_times.txt`.

lionel-nj: The datasets the validator was tested against are listed in the `end_to_end` workflow : https://github.com/MobilityData/gtfs-validator/blob/v1.4.0/.github/workflows/end_to_end.yml

aababilov: Unfortunately, this feed for Montreal was not testing StopTooFarFromTripShapeValidator - its code was never called. The thing is that the Montreal feed has `field_parsing_error` for stop_url field (it uses non-ASCII characters), so parsing failed, the validation process was stopped and StopTooFarFromTripShapeValidator was never invoked.

You can check the generated report.json to find the parsing errors.

```JSON
        {
            "code": "field_parsing_error",
            "notices": [
                {
                    "csvRowNumber": 163,
                    "fieldName": "stop_url",
                    "fieldType": "URL",
                    "fieldValue": "http://www.stm.info/fr/infos/reseaux/metro/square-victoria\u2013oaci",
                    "filename": "stops.txt"
                },
                {
                    "csvRowNumber": 232,
                    "fieldName": "stop_url",
                    "fieldType": "URL",
                    "fieldValue": "http://www.stm.info/fr/infos/reseaux/metro/longueuil\u2013universite-de-sherbrooke",
                    "filename": "stops.txt"
                }
            ],
            "severity": "ERROR",
            "totalNotices": 2
        },
```

aababilov: I wrote https://github.com/MobilityData/gtfs-validator/pull/742 to make clear that parsing failed.

f8full: In v1.4, the workflows would surface those kind of situation in the GitHub interface by outputting an error, which would in turn lead to all subsequent feeds to be skipped. I wonder if such a process could be replicated.

barbeau: >You are dividing an integer by integer, so the result is rounded and it is 0 (zero). You probably wanted to write 1.0 / 1000 instead.
>Also, why are you using float instead of double here?

Good catch - originally this used a `double` and divided by `1000.0d` as implemented in https://github.com/MobilityData/gtfs-validator/pull/340/files#diff-fd338ec09e9005136186658756b593cc6e7e9917476a0b92a59f27071a1c3314R45. So it should be changed back.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
barbeau: I need to look at this a little closer, but I think this is a translation issue from the original implementation due to the differences in data structures between v1 and v2. 

In the original implementation of v1, the in-memory GTFS data repository held a Map data structure where each trip_id appeared once as the key and the value was a SortedMap of StopTimes for that trip_id. This rule looped through each of those entries and therefore only processed each trip once.

But the similarly named variable in v2 actually contains the entire contents of stop_times.txt. So it looks like this implementation is fetching all StopTimes and then getting the trip_id for each StopTime and then getting all the StopTimes for that trip_id, which blows up the processing time significantly.

I'll take a closer look tomorrow.

barbeau: @aababilov Are you working on re-implementing this with a separate spatial library? Or should MobilityData work on fixing the implementation issues with this rule using Spatial4j?

aababilov: > This piece of code origins from: #340.

It actually does not origin from the Sean's commit. Sean wrote another logic that iterated on each trip only once. The new code, however, iterates on each pair of <trip_id, GtfsStopTime>. IDEA can spell out the types for you:

```
    stopTimeTable
        .byTripIdMap()
        .forEach(
            (String tripId, GtfsStopTime tripStopTimes) -> {
                ....
```

Also, the new code does not use the `tripStopTimes` variable - instead, it calls `stopTimeTable.byTripId(tripId)` that really returns `List<GtfsStopTime>`. If you use your `tripStopTimes` in that place, that won't compile and you would catch this bug.

aababilov: Overall, I think I can suggest you to spell out the types and also prefer plain cycles to lambdas. That will help to avoid such kind of bugs in the future.

aababilov: > @aababilov Are you working on re-implementing this with a separate spatial library? Or should MobilityData work on fixing the implementation issues with this rule using Spatial4j?

This validation rule is not the top priority for me right now. What I did was to temporarily remove that code at Google.

I think that we should put as a priority that our MobilityData upstream code works as expected for a large set of feeds (such as several thousands that we have at Google). I do not think that we need to rush implementing as many validations as possible without proper testing since that puts too much technical debt for the future maintainers.

I am always testing my code on all Google feeds and I check the results, so that gives a certain confidence. I see that MobilityData is not actually testing the code on a considerable feed set. Even more, the test results may be misinterpreted (as it happened for the Montreal feed), so that reduces positive impact of such testing.

@barbeau @lionel-nj Are you able to setup a reliable testing procedure on your side and run the Validator on a large set of feeds (hundreds or better thousands)? Without such testing, we always risk to accumulate bugs in the upstream code. May we hold on all code contributions from MobilityData before such testing procedure is established?

lionel-nj: @aababilov could you provide the url of the dataset you were looking at please?

lionel-nj: > I used a medium-size feed with 1 million stop_times.txt and this code runs for several minutes on it.


@aababilov could you provide the url of the dataset you were looking at please?

PR #710, opened 2021-02-01
feat: Too fast trip travel speed

### LINKED ISSUE #517: Fast travel between stops in `stop_times.txt`
Validate travel between two stop times is done within a reasonable time range.

**Additional context**
`v1.4.0` implementation:
- https://github.com/MobilityData/gtfs-validator/blob/v1.4.0/usecase/src/main/java/org/mobilitydata/gtfsvalidator/usecase/ValidateTripTravelSpeed.java
- tests: https://github.com/MobilityData/gtfs-validator/blob/v1.4.0/usecase/src/test/java/org/mobilitydata/gtfsvalidator/usecase/ValidateTripTravelSpeedTest.java

lionel-nj:
closes #517
### Please review after #708

**Summary:**

This PR implements a new validation rule: check trip travel is reasonable.

**Expected behavior:** 

- a notice should be generated if at some point the mean speed of a trip (or part or it) exceed 150 km/h

- ~[ ] Include screenshot(s) showing how this pull request works and fixes the issue(s)`


lionel-nj:
This PR relies on a dependency that is not supported by Google, as a result, the implementation of this new rule is postponed until @aababilov evaluates possibilities https://github.com/MobilityData/gtfs-validator/pull/708#issuecomment-773786217

https://github.com/MobilityData/gtfs-validator/pull/708

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TripTravelSpeedValidator.java
```diff
@@ -0,0 +1,131 @@
...
+ * not above 150 km/h
+ *
+ * <p>Time complexity: <i>O(n * p)</i>, where <i>n</i> is the number of records in
+ * <i>stop_times.txt</i> and <i>p</i> is the number of points in a trip shape.
+ *
+ * <p>Generated notice: {@link FastTravelBetweenStopsNotice}.
+ */
+@GtfsValidator
+public class TripTravelSpeedValidator extends FileValidator {
+  private static final double METER_PER_SECOND_TO_KMH_CONVERSION_FACTOR = 3.6D;
```
barbeau, line 47: If you're going to add the explicit designation for double I'd use `d` as I think that is more readable than `D`.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TripTravelSpeedValidator.java
```diff
@@ -0,0 +1,131 @@
...
+        Multimaps.asMap(stopTimeTable.byTripIdMap()).entrySet()) {
+      List<FastTravelBetweenStopsNotice> noticesForTrip = checkSpeedAlongTrip(entry);
+      for (FastTravelBetweenStopsNotice notice : noticesForTrip) {
+        noticeContainer.addValidationNotice(notice);
+      }
+    }
+  }
+
+  /**
+   * Calculates the travel speed between {@code GtfsStopTime}. Generates a {@code
```
barbeau, line 64: I would add "for a single trip" in the top Javadocs here just for clarity. Something like:

```
Calculates the travel speed between {@code GtfsStopTimes} for a single trip. The key for the provided map {@code Entry} is the trip_id as a string, with the value being the list of {@code GtfsStopTimes} for that trip.
```

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TripTravelSpeedValidator.java
```diff
@@ -0,0 +1,131 @@
...
+   * 150 km/h.
+   *
+   * @param tripStopTimes the list of {@code GtfsStopTimes} sorted by stop_times.stop_sequence for a
+   *     {@code GtfsTrip}
+   * @return the list of {@code FastTravelBetweenStopsNotice} generated when checking speed travel
+   *     along the given trip
+   */
+  private List<FastTravelBetweenStopsNotice> checkSpeedAlongTrip(
+      Entry<String, List<GtfsStopTime>> tripStopTimes) {
+    var previousStopsData =
```
barbeau, line 75: Please don't use `var`, as it's new to Java in JDK 10, and we're trying to revert back to 8 - see PR https://github.com/MobilityData/gtfs-validator/pull/759.

Also, IMHO it really messes with your head if you're used to reading/writing Kotlin as well.

Also, personally I'd prefer to move all these variables out of this object and into individual variables in the method. IMHO it's easier to read and avoid another object instantiation.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TripTravelSpeedValidator.java
barbeau: Thinking about this more, I think the method would be much easier to read if it just took in two parameters - `(String tripId, List<GtfsStopTime> stopTimes)` - vs. passing in the `Entry` directly.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TripTravelSpeedValidator.java
```diff
@@ -0,0 +1,131 @@
...
+          // times
+          int accumulatedDistanceMeter = 0;
+        };
+    List<FastTravelBetweenStopsNotice> notices = new ArrayList<>();
+    for (GtfsStopTime stopTime : tripStopTimes.getValue()) { // prepare data for current iteration
+      GtfsStop currentStop = stopTable.byStopId(stopTime.stopId());
+      double currentStopLat = currentStop.stopLat();
+      double currentStopLon = currentStop.stopLon();
+      GtfsTime currentArrivalTime = stopTime.arrivalTime();
+      double distanceFromPreviousStopMeter =
```
barbeau, line 91: `distanceFromPreviousStopMeter` -> `distanceFromPreviousStopMeters`

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TripTravelSpeedValidator.java
```diff
@@ -0,0 +1,131 @@
...
+      if (previousStopsData.departureTime != null && currentArrivalTime != null) {
+        sameArrivalAndDeparture = currentArrivalTime.equals(previousStopsData.departureTime);
+        if (!sameArrivalAndDeparture) {
+          int durationSecond =
+              currentArrivalTime.getSecondsSinceMidnight()
+                  - previousStopsData.departureTime.getSecondsSinceMidnight();
+          double distanceMeter =
+              distanceFromPreviousStopMeter + previousStopsData.accumulatedDistanceMeter;
+          double speedMeterPerSecond = distanceMeter / durationSecond;
+          if (speedMeterPerSecond > 42) { // roughly 150 km per hour. Put it in default parameters
```
barbeau, line 107: `42` should be pulled out into it's own static final constant at the top of this class. I'd make it a `double` too, and add in a comment what the value is in mph as well.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TripTravelSpeedValidator.java
```diff
@@ -0,0 +1,131 @@
...
+              previousStopsData.latitude,
+              previousStopsData.longitude,
+              currentStopLat,
+              currentStopLon);
+      boolean sameArrivalAndDeparture = false;
+      if (previousStopsData.departureTime != null && currentArrivalTime != null) {
+        sameArrivalAndDeparture = currentArrivalTime.equals(previousStopsData.departureTime);
+        if (!sameArrivalAndDeparture) {
+          int durationSecond =
+              currentArrivalTime.getSecondsSinceMidnight()
```
barbeau, line 102: If there is a `StopTimeWithArrivalBeforePreviousDepartureTimeNotice` for this trip this `TripTravelSpeedValidator` will produce some strange results. You may want to detect this here and bail on this trip if so.

I'm also curious as to what the real-world performance benefits would be if we consolidated this and some of the other rules like `StopTimeWithArrivalBeforePreviousDepartureTimeNotice` that loop through all of stop_times.txt, as we're currently doing this loop multiple times for each `*Validator` class, and stop_times.txt can be very large. Looks like `StopTimeWithDepartureBeforeArrivalTimeNotice` and `StopTimeWithArrivalBeforePreviousDepartureTimeNotice` are already consolidated within `StopTimeArrivalAndDepartureTimeValidator`.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TripTravelSpeedValidator.java
```diff
@@ -0,0 +1,131 @@
...
+          double distanceMeter =
+              distanceFromPreviousStopMeter + previousStopsData.accumulatedDistanceMeter;
+          double speedMeterPerSecond = distanceMeter / durationSecond;
+          if (speedMeterPerSecond > 42) { // roughly 150 km per hour. Put it in default parameters
+            previousStopsData.accumulatedStopSequence.add(stopTime.stopSequence());
+            notices.add(
+                new FastTravelBetweenStopsNotice(
+                    tripStopTimes.getKey(),
+                    speedMeterPerSecond * METER_PER_SECOND_TO_KMH_CONVERSION_FACTOR,
+                    new ArrayList<>(previousStopsData.accumulatedStopSequence)));
```
barbeau, line 113: I don't think the stop sequence list really provides value here? Unless I'm missing something. I'd be inclined to omit this to avoid another ArrayList allocation.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TripTravelSpeedValidator.java
```diff
@@ -0,0 +1,131 @@
...
+      for (FastTravelBetweenStopsNotice notice : noticesForTrip) {
+        noticeContainer.addValidationNotice(notice);
+      }
+    }
+  }
+
+  /**
+   * Calculates the travel speed between {@code GtfsStopTime}. Generates a {@code
+   * FastTravelBetweenStopsNotice} if the travel speed between {@code GtfsStopTimes} is greater than
+   * 150 km/h.
```
barbeau, line 66: Could you add a high level description of the algorithm being used here too? For example, looks like it allows for blank arrival/departure records (e.g., non-timepoints) and skips over them, but still accumulates distance.

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/TripTravelSpeedValidatorTest.java
```diff
@@ -0,0 +1,238 @@
...
+      NoticeContainer noticeContainer, List<GtfsStopTime> entities) {
+    return GtfsStopTimeTableContainer.forEntities(entities, noticeContainer);
+  }
+
+  private static GtfsTripTableContainer createTripTable(
+      NoticeContainer noticeContainer, List<GtfsTrip> entities) {
+    return GtfsTripTableContainer.forEntities(entities, noticeContainer);
+  }
+
+  public static GtfsTrip createTrip(
```
barbeau, line 62: Should be `private`?

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/TripTravelSpeedValidatorTest.java
```diff
@@ -0,0 +1,238 @@
...
+import org.mobilitydata.gtfsvalidator.table.GtfsStop;
+import org.mobilitydata.gtfsvalidator.table.GtfsStopTableContainer;
+import org.mobilitydata.gtfsvalidator.table.GtfsStopTime;
+import org.mobilitydata.gtfsvalidator.table.GtfsStopTimeTableContainer;
+import org.mobilitydata.gtfsvalidator.table.GtfsTrip;
+import org.mobilitydata.gtfsvalidator.table.GtfsTripTableContainer;
+import org.mobilitydata.gtfsvalidator.type.GtfsTime;
+
+public class TripTravelSpeedValidatorTest {
+  public static GtfsStopTime createStopTime(
```
barbeau, line 35: should be `private`?

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/TripTravelSpeedValidatorTest.java
```diff
@@ -0,0 +1,238 @@
...
+    underTest.stopTimeTable =
+        createStopTimeTable(
+            noticeContainer,
+            ImmutableList.of(
+                createStopTime(
+                    4,
+                    "t1",
+                    "s0",
+                    0,
+                    GtfsTime.fromSecondsSinceMidnight(130),
```
barbeau, line 217: Could you please add tests with empty arrival and departure times as well (e.g., non-timepoints)?

barbeau (changes requested):

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TripTravelSpeedValidator.java
```diff
@@ -0,0 +1,131 @@
...
+import org.mobilitydata.gtfsvalidator.table.GtfsTripTableContainer;
+import org.mobilitydata.gtfsvalidator.type.GtfsTime;
+import org.mobilitydata.gtfsvalidator.util.GeospatialUtil;
+
+/**
+ * Validates: all records of `trips.txt` refer to a collection of {@code GtfsStopTime} from file
+ * `stop_times.txt` of which -when sorted by `stop_sequence`- the travel speed between each stop is
+ * not above 150 km/h
+ *
+ * <p>Time complexity: <i>O(n * p)</i>, where <i>n</i> is the number of records in
```
barbeau, line 40: I believe this time complexity is wrong? If it's just measuring stop to stop distance and scanning through stop_times.txt once, then it should be O(n), where n is the number of records in stop_times.txt.

RULES.md
```diff
@@ -408,3 +408,7 @@ The combination of `fare_rules.route_id`, `fare_rules.origin_id`, `fare_rules.co
 ### MissingFeedInfoDateNotice
 
 Even though `feed_info.start_date` and `feed_info.end_date` are optional, if one field is provided the second one should also be provided.
+
+### FastTravelBetweenStopsNotice
+
+Calculated speed between stops is too fast (>150 kmh).
```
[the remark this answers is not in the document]
lionel-nj: The SI unit for speed id km/h as you mentioned, will add mph to ease the reading.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TripTravelSpeedValidator.java
lionel-nj: This stop sequence lets us know between which stop times the travel is too fast. If we omit it, we would just know that  -for example- for trip `t0` the speed is incorrect which would not allow us to traceback the wrong stop times.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TripTravelSpeedValidator.java
lionel-nj: Thanks for flagging this, I modified the javadocs. But it seems to me that the complexity would be O(n*p), where 
- n: the number of trips in `trips.txt`
- p: the number of stop times referred to by a trip. 
For each trip we calculate the speed between each pair of stop times.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TripTravelSpeedValidator.java
barbeau: Couldn't you do the same with one or two stop sequence values without the list though? I'm not seeing why you would need more than two values at the most (the starting and ending stop sequence) to define where the travel is too fast. So you could say "Too fast travel starting at X" or "Too fast travel between X and Y", but I don't see why you'd need a third value.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TripTravelSpeedValidator.java
barbeau: Well, yes, that's another way to phrase it but when possible I think it's more useful to give a tighter bound with a single metric based on a single file, especially in relationship to the files that end up being huge like stop_times.txt and shapes.txt. Each stop time can only belong to one trip, so you're only processing each stop_times.txt record once. And each speed measurement between pair of stop times is constant. So you can really forget about the trips table, it's only the size of stop_times.txt that matters for big O notation here.

lionel-nj: Alright, `stop_times.txt` takes precedence over `trips.txt` because of the size of the files!  Will modify :)

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TripTravelSpeedValidator.java
lionel-nj: Totally true! Did not realize it that way. Modifying this to reflect your thoughts. Thanks @barbeau  :)

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/TripTravelSpeedValidatorTest.java
```diff
@@ -0,0 +1,443 @@
...
+                    GtfsTime.fromSecondsSinceMidnight(770000),
+                    GtfsTime.fromSecondsSinceMidnight(390000))));
+
+    underTest.validate(noticeContainer);
+
+    assertThat(noticeContainer.getValidationNotices()).isEmpty();
+  }
+
+  @Test
+  public void fastTravelMissingDepartureTimeShouldGenerateNotice() {
```
barbeau, line 312: We need another two tests where both arrival and departure time aren't provided in between stop times that have arrival and departure times, both for the case where there is and is not too fast travel. The below is valid legacy GTFS if the "timepoint" field isn't provided.

|  stop_sequence | arrival_time | departure_time |
|  ------ | ------ | ------ | 
1 | 00:00:00 | 00:00:00  |  
2 |  |  |  
3 |  |  |  
4 | 00:10:00 | 00:10:00  |  
5 |  |  |  
6 | 00:20:00 | 00:20:00  |

barbeau (changes requested):
Thanks @lionel-nj, looks pretty good! A few more items in-line.

maximearmstrong (approved):
LGTM

main/src/main/java/org/mobilitydata/gtfsvalidator/notice/FastTravelBetweenStopsNotice.java
```diff
@@ -0,0 +1,44 @@
...
+            .put("speedkmh", speedkmh)
+            .put("firstStopSequence", firstStopSequence)
+            .put("lastStopSequence", lastStopSequence)
+            .build(),
+        SeverityLevel.WARNING);
+  }
+
+  @Override
+  public String getCode() {
+    return "too_fast_trip_travel_speed";
```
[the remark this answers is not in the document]
lionel-nj: Changed to `ExcessiveTravelSpeedValidator`, `ExcessiveTripTravelSpeedNotice` and `excessive_trip_travel_speed`

lionel-nj:
@barbeau reading https://github.com/MobilityData/gtfs-validator/pull/765#issuecomment-785200442: it seems like I implemented the solution 2 you suggested.  

> Add checks within separate TripTravelSpeedValidator for valid times (but don't generate notice for invalid times) - Pros are that this ensures the output of FastTravelBetweenStopsNotice is valid and it doesn't add complexity to a single validator class for this and other notices. Cons are duplicating some code for the time checks and another loop through stop_times.txt for this separate validator.

lionel-nj:
> We need another two tests where both arrival and departure time aren't provided in between stop times that have arrival and departure times, both for the case where there is and is not too fast travel

Done. I also reworked the tests to ease clarity and reading. @barbeau

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/TooFastTravelValidatorTest.java
```diff
@@ -0,0 +1,535 @@
...
+    underTest.validate(noticeContainer);
+
+    assertThat(noticeContainer.getValidationNotices())
+        .containsExactly(new TooFastTravelNotice("t1", 200.15114352186376d, 0, 1));
+  }
+
+  @Test
+  public void fastTravelFarStopsShouldGenerateNotice() {
+    // sample table descriptions available at:
+    // https://gist.github.com/lionel-nj/d21cc8b1519d4f44f9ebd53858cda78d
```
barbeau, line 134: 💯  On the Gist with the table

barbeau (approved):
LGTM, thanks @lionel-nj

PR #711, opened 2021-02-02
feat: Accept the existing feeds in feed-and-agency-lang validator

aababilov:
To tell the truth, agency_lang field is misleading and it would be great
to deprecate it in favour of feed_lang. That would be nice to encourage
using feed_lang and omitting agency_lang.

1. We add more context to the generated error. Now it is clear which
   agency on which line is breaking the validation rule.
2. There are a lot of feeds that specify feed_lang and no agency_lang.
   They are accepted after that change.
3. The existing http://gtfs.org/reference/static/#feed_infotxt does
   not forbid having feed_lang="mul" and a single agency.
   Consider a feed that covers Biel/Bienne, sets feed_lang=mul,
   agency_lang='' and provides translations.txt for German and French.

### DIFF
--- a/main/src/main/java/org/mobilitydata/gtfsvalidator/notice/FeedInfoLangAndAgencyLangMismatchNotice.java
+++ b/main/src/main/java/org/mobilitydata/gtfsvalidator/notice/FeedInfoLangAndAgencyLangMismatchNotice.java
@@ -20,12 +20,14 @@ import com.google.common.collect.ImmutableMap;
 import java.util.Set;
 
 public class FeedInfoLangAndAgencyLangMismatchNotice extends ValidationNotice {
-  public FeedInfoLangAndAgencyLangMismatchNotice(
-      String feedInfoLang, Set<String> agencyLangCollection) {
+  public FeedInfoLangAndAgencyLangMismatchNotice(long csvRowNumber, String agencyId, String agencyName, String agencyLang, String feedLang) {
     super(
         ImmutableMap.of(
-            "feedInfoLang", feedInfoLang,
-            "agencyLangCollection", agencyLangCollection));
+            "csvRowNumber", csvRowNumber,
+            "agencyId", agencyId,
+            "agencyName", agencyName,
+            "agencyLang", agencyLang,
+            "feedLang", feedLang));
   }
 
   @Override
--- a/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/MatchingFeedAndAgencyLangValidator.java
+++ b/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/MatchingFeedAndAgencyLangValidator.java
@@ -16,79 +16,62 @@
 
 package org.mobilitydata.gtfsvalidator.validator;
 
-import java.util.HashSet;
 import java.util.Locale;
-import java.util.Set;
 import org.mobilitydata.gtfsvalidator.annotation.GtfsValidator;
 import org.mobilitydata.gtfsvalidator.annotation.Inject;
 import org.mobilitydata.gtfsvalidator.notice.FeedInfoLangAndAgencyLangMismatchNotice;
 import org.mobilitydata.gtfsvalidator.notice.NoticeContainer;
+import org.mobilitydata.gtfsvalidator.table.GtfsAgency;
 import org.mobilitydata.gtfsvalidator.table.GtfsAgencyTableContainer;
 import org.mobilitydata.gtfsvalidator.table.GtfsFeedInfoTableContainer;
 
 /**
- * Files `agency.txt` and `feed_info.txt` must define matching `agency.agency_lang` and
- * `feed_info.feed_lang`. The default language may be multilingual for datasets with the original
- * text in multiple languages. In such cases, the feed_lang field should contain the language code
- * mul defined by the norm ISO 639-2. If `feed_lang` is not `mul` and does not match with
- * `agency_lang`, that's an error If there is more than one `agency_lang` and `feed_lang` isn't
- * `mul`, that's an error If `feed_lang` is `mul` and there isn't more than one `agency_lang`,
- * that's an error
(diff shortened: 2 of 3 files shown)


barbeau:
@aababilov Thanks for this. I generally agree with the above points, although  I'm curious how you interpret this part of the `feed_lang` definition:

>If all the original text in the dataset is in the same language, then mul should not be used.

To deal with the feed that covers Biel/Bienne, should we generate a notice if `feed_lang` is `mul` and all the `agency_lang` values are the same (or there is only one agency), and translations.txt is not provided?

aababilov:
> If all the original text in the dataset is in the same language, then mul should not be used.

I interpret it that way: if all names in stops.txt, routes.txt etc etc use the same language, then it should be used as feed_lang instead of `mul`.

> To deal with the feed that covers Biel/Bienne, should we generate a notice if feed_lang is mul and all the agency_lang values are the same (or there is only one agency), and translations.txt is not provided?

That sounds acceptable to me: `feed_lang = agency_lang = mul` and translations because the names in the feed just match what you see on the signboards.

tl;dr Indeed, what is that agency_lang? 

1. Is that the language of stop_names that are served by that agency? This can't be supported since stops may be covered by multiple routes.
2. Is that the language of route_names that belong to the agency? We do not have real feeds that use that feature and this is still quite confusing.
3. Is that the language that you may use if you call agency_phone? Hmm, sounds strange :)

So, the only legitimate way to interpret agency_lang that I see is to use it in the same sense as feed_lang. That is why all agencies must have the same agency_lang (we have AgencyConsistencyValidator for that).

LeoFrachet:
IMHO the biggest issue is that `feed_lang` is not required because the file `feed_info.txt` isn't required. I would personally be in favor of requiring the file `feed_info.txt`. I know it would be not "backward compatible" in the meaning that older dataset would become non valid (but this wouldn't break the existing tools, so it's a "light" non-backward-compatible), but we should at least ask the community to enforce it in the future.

For example, changing the spec to say that "any dataset published after 2023-01-01 must contain `feed_info.txt`. It would be giving 2y in the industry to change, and anyway more producer are already producing that file.

By doing that change, we would always have `feed_lang` defined and we could deprecate `agency_lang` which is so badly defined.

timMillet:
I agree with Leo's strategy as it will simplify language management within datasets. In the meantime, or if that non-backward-compatible change got rejected for any reason, we can figure out an alternate strategy.

This strategy could be making explicit the relationship between `agency_lang` and `feed_lang`, according to the current rules of GTFS, by defining any possible use cases. Then these precisions could be added to the official GTFS without requiring non-backward-compatible changes.

Example of use-case definitions:
- `feed_lang`
  - If `feed_lang` is provided, it is the default language for the dataset.
  - `feed_lang` can be overridden by `agency_lang` for the text fields in agency.txt, routes.txt, trips.txt, and stop_times.txt that are related to the agency.
- `agency_lang`
  - If `feed_lang` is not provided, and if all `agency_lang` values are provided and are the same: `agency_lang` is the default language for the dataset.
  -  If `feed_lang` is not provided, and if some `agency_lang` values are not provided or are not the same:
      - `agency_lang` is the language for the text fields in agency.txt, routes.txt, trips.txt, and stop_times.txt that are related to the agency. If `agency_lang` is not provided, its value can be considered as `und` (undetermined, according to IETF BCP 47 language code).
      - The language of other text fields in the dataset can be considered as `und`.
  - If `feed_lang` and `agency_lang` are provided: `agency_lang` is the language for the text fields in agency.txt, routes.txt, trips.txt, and stop_times.txt that are related to the agency. `feed_lang` is the language for the other text fields.
  - If `feed_lang` is provided and `agency_lang` is not provided: `feed_lang` is the language for the text fields in agency.txt, routes.txt, trips.txt, and stop_times.txt that are related to the agency, and is the language for the other text fields.

aababilov:
I propose to agree with the de-facto usage of `agency_lang` and `feed_lang` that we have at several thousands production feeds at Google. All those feeds agree with the validation requirement that I have implemented in this change: if `feed_lang` is provided, then all `agency_lang` which are provided must be the same as `feed_lang`. Note that `agency_lang` is optional, so it is possible to give `feed_lang` only and omit one or all `agency_lang`.

Tim, I would like to warn you against treating `agency_lang` as the language for the text fields in agency.txt, routes.txt and other files related to the agency.

1. There are no use cases for that in thousands of feeds that Google has.
2. This interpretation sounds quite complicated since now agency.txt, routes.txt etc may contain mixtures of different languages while all stops.txt and many other file must use the same `feed_lang`.
3. If you want to simply merge two GTFS feeds which use different languages into a single feed and you just merge their files, that would not work since all text fields in stops.txt must have the same language.

timMillet:
Thank you for the heads up Alexej!

1. I had a quick look at GTFS datasets of transit networks in multilingual regions, and each one of them had a different approach to language management. I guess there is an issue with language normalization in current datasets.

2. This interpretation may sound complex but it was an example that can be improved. My concern with the interpretation you suggest is that it forces data producers to provide only one language per dataset. This can be an issue for merging datasets, as the language information from the original datasets can be lost.
  E.g: the current dataset from [SBB CFF FFS](https://openmobilitydata.org/p/sbb-cff-ffs/793) has `feed_lang=de` with all `agency_lang=de`. However, the dataset contains agencies with related information in german, french, and italian.

3. If datasets with different `agency_lang` are merged, then `feed_lang` should be `mul`. In this case, all text fields in stops.txt would technically have the same language.

IMHO, I don't think there will be a perfect solution to this issue. We can go forward with the validation requirement that you have implemented in this change, as long as it issues a warning message and these rules are added in the Best Practices.

barbeau (approved):
LGTM, thanks @aababilov!

aababilov:
Thanks everyone!

In the future, we may revisit and relax that requirement.

f8full:
Clearly the kind of conversation this tool aimed to make happen. I'm glad to witness it today 🥳 🎆 🚬

PR #712, opened 2021-02-02
feat: non ascii character in ID

### LINKED ISSUE #529: Non ascii or non printable char in `id`
Validate `id` fields are ascii or printable characters

**Additional context**
`v1.4.0` implementation: 
- https://github.com/MobilityData/gtfs-validator/blob/76970f6ab56c885ef3fad951a6bff2531ae18a3e/adapter/validator/src/main/java/org/mobilitydata/gtfsvalidator/validator/GtfsTypeValidator.java#L151

lionel-nj:
closes #529
**Summary:**

This PR implements a new validation rule: no non-ascii character in IDs.

**Expected behavior:** 

- a notice should be generated if an ID contains non ascii characters

### DIFF
--- a/RULES.md
+++ b/RULES.md
@@ -65,7 +65,7 @@ Note that the notice ID naming conventions changed in `v2` to make contributions
 |                          	| [E061](https://github.com/MobilityData/gtfs-validator/blob/v1.4.0/RULES.md#E061)      	| Out of memory                                                           	|
 |                           | [W001](https://github.com/MobilityData/gtfs-validator/blob/v1.4.0/RULES.md#W001)          | Input zip archive contains folder | 
 |                           | [W002](https://github.com/MobilityData/gtfs-validator/blob/v1.4.0/RULES.md#W002)          | Non standard field name | 
-|                           | [W003](https://github.com/MobilityData/gtfs-validator/blob/v1.4.0/RULES.md#W003)        	| Non ascii or non printable char in `id`                                         	|
+|[`NonAsciiOrNonPrintableCharNotice`](#NonAsciiOrNonPrintableCharNotice)| [W003](https://github.com/MobilityData/gtfs-validator/blob/v1.4.0/RULES.md#W003)        	| Non ascii or non printable char in `id`                                         	|
 |                           | [W006](https://github.com/MobilityData/gtfs-validator/blob/v1.4.0/RULES.md#W006)        	| Missing route short name                                                        	|
 |                           | [W007](https://github.com/MobilityData/gtfs-validator/blob/v1.4.0/RULES.md#W007)        	| Missing route long name                                                         	|
 |                           | [W008](https://github.com/MobilityData/gtfs-validator/blob/v1.4.0/RULES.md#W008)        	| Route long name contains short name                                             	|
@@ -319,3 +319,10 @@ Both files calendar_dates.txt and calendar.txt are missing from the GTFS archive
 ### DecreasingStopTimeDistanceNotice
 
 Stop times in a trip should have increasing distance.
+
+### NonAsciiOrNonPrintableCharNotice
+
+A value of filed with type `id contains non ASCII or non printable characters. This is not recommended.
+
+#### References:
+* [Field type recommendations](http://gtfs.org/reference/static/#field-types)
--- a/RULES.md
+++ b/RULES.md
@@ -322,7 +322,7 @@ Stop times in a trip should have increasing distance.
 
 ### NonAsciiOrNonPrintableCharNotice
 
-A value of filed with type `id contains non ASCII or non printable characters. This is not recommended.
+A value of filed with type `id` contains non ASCII or non printable characters. This is not recommended.
 
 #### References:
 * [Field type recommendations](http://gtfs.org/reference/static/#field-types)
--- a/core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NonAsciiOrNonPrintableCharNotice.java
+++ b/core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NonAsciiOrNonPrintableCharNotice.java
@@ -1,0 +1,36 @@
+/*
+ * Copyright 2021 MobilityData IO
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.mobilitydata.gtfsvalidator.notice;
+
+import com.google.common.collect.ImmutableMap;
+
+/** A row in the input file has a different number of values than specified by the CSV header. */
+public class NonAsciiOrNonPrintableCharNotice extends ValidationNotice {
+  public NonAsciiOrNonPrintableCharNotice(
+      String filename, long csvRowNumber, String columnName) {
+    super(
+        ImmutableMap.of(
+            "filename", filename,
+            "csvRowNumber", csvRowNumber,
+            "columnName", columnName));
+  }
+
+  @Override
+  public String getCode() {
+    return "invalid_row_length";
+  }
+}
--- a/core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NonAsciiOrNonPrintableCharNotice.java
+++ b/core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NonAsciiOrNonPrintableCharNotice.java
@@ -31,6 +31,6 @@ public class NonAsciiOrNonPrintableCharNotice extends ValidationNotice {
 
   @Override
   public String getCode() {
-    return "invalid_row_length";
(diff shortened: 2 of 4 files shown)


RULES.md
```diff
@@ -319,3 +319,10 @@ Both files calendar_dates.txt and calendar.txt are missing from the GTFS archive
 ### DecreasingStopTimeDistanceNotice
 
 Stop times in a trip should have increasing distance.
+
+### NonAsciiOrNonPrintableCharNotice
+
+A value of filed with type `id contains non ASCII or non printable characters. This is not recommended.
```
barbeau, line 325: Typo in formatting

core/src/main/java/org/mobilitydata/gtfsvalidator/parsing/RowParser.java
```diff
@@ -216,6 +217,18 @@ public String asText(int columnIndex, boolean required) {
 
   @Nullable
   public String asId(int columnIndex, boolean required) {
+    String value = row.asString(columnIndex);
+    if(value == null) {
```
barbeau, line 221:
```suggestion
    if (value == null) {
```

barbeau (approved):
Thanks @lionel-nj, mostly LGTM! Two formatting items in-line.

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NonAsciiOrNonPrintableCharNotice.java
```diff
@@ -0,0 +1,36 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.mobilitydata.gtfsvalidator.notice;
+
+import com.google.common.collect.ImmutableMap;
+
+/** A row in the input file has a different number of values than specified by the CSV header. */
```
f8full, line 21: ❗

core/src/main/java/org/mobilitydata/gtfsvalidator/parsing/RowParser.java
```diff
@@ -216,6 +217,18 @@ public String asText(int columnIndex, boolean required) {
 
   @Nullable
   public String asId(int columnIndex, boolean required) {
+    String value = row.asString(columnIndex);
+    if (value == null) {
+      return asString(columnIndex, required);
+    }
+    for (char ch : value.toCharArray()) {
```
f8full, line 224: Maybe a short comment to link to the issue being checked? Doesn't seem that obvious in the context of a RowParser, IMHO

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NonAsciiOrNonPrintableCharNotice.java
lionel-nj: Thanks @f8full, wil modify!

core/src/main/java/org/mobilitydata/gtfsvalidator/parsing/RowParser.java
aababilov, line 224: This makes a copy of the string that is a small performance issue. It is better to avoid them from the beginning rather than fixing them later. Please use a usual cycle here:

for(int i = 0, n = value.length() ; i < n ; i++) { 
 ...

core/src/main/java/org/mobilitydata/gtfsvalidator/parsing/RowParser.java
```diff
@@ -216,6 +217,18 @@ public String asText(int columnIndex, boolean required) {
...
+      if (!(ch >= 32 && ch < 127)) {
+        addErrorInRow(
+            new NonAsciiOrNonPrintableCharNotice(
+                row.getFileName(), row.getRowNumber(), row.getColumnName(columnIndex)));
+        break;
+      }
+    }
```
aababilov, line 231: Why are you calling asString once again at the end if you already have the value to return?

core/src/main/java/org/mobilitydata/gtfsvalidator/parsing/RowParser.java
```diff
@@ -216,6 +217,18 @@ public String asText(int columnIndex, boolean required) {
 
   @Nullable
   public String asId(int columnIndex, boolean required) {
+    String value = row.asString(columnIndex);
+    if (value == null) {
+      return asString(columnIndex, required);
```
aababilov, line 222: Why are you calling asString which is calling row.asString again if you already have the value?

core/src/main/java/org/mobilitydata/gtfsvalidator/parsing/RowParser.java
```diff
@@ -216,6 +217,18 @@ public String asText(int columnIndex, boolean required) {
...
+    }
+    for (char ch : value.toCharArray()) {
+      if (!(ch >= 32 && ch < 127)) {
+        addErrorInRow(
+            new NonAsciiOrNonPrintableCharNotice(
```
aababilov, line 227: That would be great to include the invalid value as well.

aababilov (commented):
Unfortunately, that change treats many real production feeds as invalid since they use non-ASCII characters and therefore parsing stops for them. So, that becomes a blocker for me since I cannot update to the latest code from GitHub. This change ought to be submitted after support for warnings was added. This new NonAsciiOrNonPrintableCharNotice should be a warnings rather than an error because [GTFS spec](https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md) says:


> An ID field value is an internal ID, not intended to be shown to riders, and is a sequence of any UTF-8 characters. Using only printable ASCII characters is recommended.

core/src/main/java/org/mobilitydata/gtfsvalidator/parsing/RowParser.java
f8full: Good piece of advice. Strings are full of hidden quirks.
🗨️ https://www.baeldung.com/java-string-performance 🗨️

lionel-nj:
🙏🏾 Thanks @aababilov for your review. As you suggest, the changes will be provided after support for warnings was added.

lionel-nj:
@aababilov My bad, I just noticed: https://github.com/MobilityData/gtfs-validator/pull/728/files#diff-930a6cafc085554f210fb236918cea34a463932cbe3b7ea6738cc2aa6311f4bd

aababilov:
No worries, things happen!

Can we extend the amount of feeds that you use to test your commits? A large amount of feeds would probably show that valid data is rejected with errors instead of warnings or infos.

f8full:
> A large amount of feeds would probably show that valid data is rejected with errors instead of warnings or infos.

I can propose to extend the end to end workflow here before submitting it so that @lionel-nj  could merge it ?https://github.com/f8full/gtfs-validator/pull/1

lionel-nj:
> Can we extend the amount of feeds that you use to test your commits? A large amount of feeds would probably show that valid data is rejected with errors instead of warnings or infos.

Sure! Do you have recommendations on a set of feeds that we could test the validator against? @aababilov @MobilityData/transit-specs

f8full:
I could take said list and implement the workflow extension? :)

timMillet:
- Suggestions of feeds using non-ASCII characters in IDs:
I checked 10+ feeds in countries where the primary language is not transcribed in the Latin alphabet, but I haven't found any datasets with non-ASCII characters in IDs.
- Suggestion of feeds with "good data", to extend the pool of the feeds tested, available in OpenMobilityData:
  - AMT (Genova, Italy)
  - Bibus (Brest, France)
  - Metro (Christchurch, New Zealand)
  - Ruter (Oslo, Norway)
  - TAG (Grenoble, France)
  - Translink (Vancouver, Canada)
  - VAG (Freiburg, Germany)

f8full:
Thks @timMillet , will implement those tonight.

aababilov:
The more feeds we use to test the validator, the better. It increases the chance to find a problem.

At Google, I just run the validator against all prod feeds without exception. That allows me to find many hidden bugs.

How many open-data feeds you have?

f8full:
Maxime would now the answer @maximearmstrong

maximearmstrong:
@aababilov We have access to over 2000 datasets (static and real-time) through OpenMobilityData. A first step could be sampling a part of the collection to integrate it to our end-to-end workflow. This should be discussed with @carlfredl, @barbeau and @lionel-nj in order to find the best strategy and amount of feeds to integrate.

aababilov:
2000 feeds is what we need! Do you have technologies to run validation in
parallel on them? I use a proprietary Google library for that, so I can't
publish my code that validates several thousands of our feeds.

Den ons 10 feb. 2021 08:30Maxime Armstrong <notifications@github.com> skrev:

lionel-nj:
As @maximearmstrong suggested we will evaluate our possibilities. Meanwhile I would be curious to know your process to check all the validation reports that are generated @aababilov, we could get inspiration from that.

f8full:
@aababilov I guess the end to end workflow could be split in like, 50 or a 100 feeds per workflow file? They would run in parallel I think. 

Excellent point by @lionel-nj.

f8full:
@aababilov @lionel-nj #736 is ready for comments. Lemme know and I'll complete it to a 100 feeds.

f8full:
Increasing heap size in the run command fixed the IDFM feed but it seems there is an issue with Bueno Aires even with the increased heap size. I'll comment it until I get to a 100 feed at which point I'll switch the PR status to ready and let you decide.

Issue #709, opened 2021-02-01
list validator's errors and warnings


lionel-nj:

lionel-nj:
### [WIP]

### Notice list
DuplicatedColumnNotice
DuplicateKeyError
EmptyFileNotice
FieldParsingError
ForeignKeyError
InvalidRowLength
IOError
MissingRequiredColumnError
MissingRequiredFieldError
MissingRequiredFileError
MoreThanOneEntityNotice
NonAsciiOrNonPrintableCharNotice
NumberOutOfRangeError
RuntimeExceptionInLoaderError
RuntimeExceptionInValidatorError
ThreadInterruptedError
UnexpectedEnumValueError
UnknownColumnNotice
URISyntaxError
BlockTripsWithOverlappingStopTimesNotice
DecreasingShapeDistanceNotice
DecreasingStopTimeDistanceNotice
DuplicateRoteNameNotice
FeedExpirationDateNotice
FeedInfoLangAndAgencyLangMismatchNotice
MissingCalendarAndCalendarDatesFileNotice
MissingTripEdgeNotice
OverlappingFrequencyNotice
PlatformWithoutParentNotice
RouteBothShortAndLongNameMissingNotice
RouteColorContrastNotice
RouteShortAndLongNameEqualNotice
RouteShortNameTooLongNotice
SameNameAndDescriptionForRouteNotice
StartAndEndDateTimeOutOfOrderNotice
StartAndEndTimeOutOfOrderNotice
StationWithParentStationNotice
StopTimeWithArrivalBeforePreviousDepartureTimeNotice
StopTimeWithDepartureBeforeArrivalTimeNotice
UnusableTripNotice
UnusedTripNotice
WrongParentLocationTypeNotice

lionel-nj:
@barbeau please take a look at this categorization and let me know if something should be changed/clarified 😀

PR #716, opened 2021-02-02
refactor: distinguish between info, warning, and error notices

### LINKED ISSUE #549: Distinguish between errors and warnings in notices

lionel-nj:
closes #549 

**Summary:**

This PR is a refactor proposal aimed at implementing the distinction between `INFO`, `WARNING`, and `ERROR` notices.

**Expected behavior:** 

- by default: `SystemError` severity level is set to `SeverityLevel.ERROR`
- by default: `ValidationNotice` severity level is set to `SeverityLevel.ERROR`

### DIFF
--- a/core/src/main/java/org/mobilitydata/gtfsvalidator/notice/Notice.java
+++ b/core/src/main/java/org/mobilitydata/gtfsvalidator/notice/Notice.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020 Google LLC
+ * Copyright 2020 Google LLC, MobilityData IO
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -24,6 +24,7 @@ import java.util.Objects;
 /** Base class for all notices produced by GTFS validator. */
 public abstract class Notice {
   private Map<String, Object> context;
+  private SeverityLevel severityLevel;
 
   public Notice(Map<String, Object> context) {
     this.context = context;
@@ -31,6 +32,14 @@ public abstract class Notice {
 
   public Map<String, Object> getContext() {
     return Collections.unmodifiableMap(context);
+  }
+
+  public SeverityLevel getSeverityLevel() {
+    return this.severityLevel;
+  }
+
+  public void setSeverityLevel(SeverityLevel severityLevel) {
+    this.severityLevel = severityLevel;
   }
 
   /**
--- a/core/src/main/java/org/mobilitydata/gtfsvalidator/notice/Notice.java
+++ b/core/src/main/java/org/mobilitydata/gtfsvalidator/notice/Notice.java
@@ -26,8 +26,9 @@ public abstract class Notice {
   private Map<String, Object> context;
   private SeverityLevel severityLevel;
 
-  public Notice(Map<String, Object> context) {
+  public Notice(Map<String, Object> context, SeverityLevel severityLevel) {
     this.context = context;
+    this.severityLevel = severityLevel;
   }
 
   public Map<String, Object> getContext() {
@@ -36,10 +37,6 @@ public abstract class Notice {
 
   public SeverityLevel getSeverityLevel() {
     return this.severityLevel;
-  }
-
-  public void setSeverityLevel(SeverityLevel severityLevel) {
-    this.severityLevel = severityLevel;
   }
 
   /**
@@ -55,18 +52,23 @@ public abstract class Notice {
       return true;
     }
     if (other instanceof Notice) {
-      return context.equals(((Notice) other).context);
+      return context.equals(((Notice) other).context)
+          && severityLevel.equals(((Notice) other).severityLevel);
     }
     return false;
   }
 
   @Override
   public String toString() {
-    return getCode() + " " + Joiner.on(",").withKeyValueSeparator("=").join(context);
+    return getCode()
+        + " "
+        + Joiner.on(",").withKeyValueSeparator("=").join(context)
+        + " "
+        + getSeverityLevel().toString();
   }
 
   @Override
   public int hashCode() {
-    return Objects.hash(getCode(), getContext());
+    return Objects.hash(getCode(), getContext(), getSeverityLevel());
   }
 }
--- a/core/src/main/java/org/mobilitydata/gtfsvalidator/notice/SeverityLevel.java
+++ b/core/src/main/java/org/mobilitydata/gtfsvalidator/notice/SeverityLevel.java
@@ -1,0 +1,25 @@
+/*
+ * Copyright 2021 MobilityData IO
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.mobilitydata.gtfsvalidator.notice;
+
+public enum SeverityLevel {
+  WARNING,
+
+  ERROR,
+
+  SYSTEM_ERROR;
+}
--- a/core/src/main/java/org/mobilitydata/gtfsvalidator/notice/SeverityLevel.java
+++ b/core/src/main/java/org/mobilitydata/gtfsvalidator/notice/SeverityLevel.java
@@ -16,16 +16,20 @@
 
 package org.mobilitydata.gtfsvalidator.notice;
 
-/**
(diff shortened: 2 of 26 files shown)


lionel-nj:
[WIP] Listing of errors/warnings/system errors: https://github.com/MobilityData/gtfs-validator/issues/709

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/SeverityLevel.java
```diff
@@ -0,0 +1,31 @@
...
+ * of quality e.g {@code UnusedTripNotice} 2) ERROR: the GTFS dataset does not comply to an explicit
+ * rule e.g. {@code MissingCalendarAndCalendarDateFilesNotice} 3) SYSTEM_ERROR: level of severity to
+ * represent an internal error
+ */
+public enum SeverityLevel {
+  WARNING,
+
+  ERROR,
+
+  SYSTEM_ERROR;
```
aababilov, line 30: I suggest the following constants:

* INFO - for things that do not affect feed quality, such as unknown files or unknown fields;
* WARNING - affect feed quality (e.g., insufficient color contrast) but the feed remains functional;
* ERROR - the feed cannot function (e.g., broken foreign key reference);

We may also add CRITICAL since many logging frameworks have that concept but I do not see how that differs from ERROR in our case.

System errors are all ERRORs by definition. A system error is an internal error in the validator and this means that the feed was not processed successfully.

aababilov: Please leave comments at each of the constants.

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/Notice.java
```diff
@@ -33,6 +34,14 @@ public Notice(Map<String, Object> context) {
     return Collections.unmodifiableMap(context);
   }
 
+  public SeverityLevel getSeverityLevel() {
+    return this.severityLevel;
+  }
+
+  public void setSeverityLevel(SeverityLevel severityLevel) {
+    this.severityLevel = severityLevel;
+  }
```
aababilov, line 43: I am unsure if we should make Notices mutable. Let's just have two constructors:

* public Notice(Map<String, Object> context, SeverityLevel severityLevel)
* public Notice(Map<String, Object> context) - defaults to ERROR. This constructor may be removed later during refactoring.

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/ValidationNotice.java
```diff
@@ -29,5 +29,7 @@
 public abstract class ValidationNotice extends Notice {
   public ValidationNotice(Map<String, Object> context) {
     super(context);
+    // by default severityLevel is set to SeverityLevel.WARNING
+    this.setSeverityLevel(SeverityLevel.WARNING);
```
aababilov, line 33: Pass severity to the super() constructor.

Let's have this ERROR by default and leave WARNING to the next refactoring.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/RouteNameValidator.java
```diff
@@ -42,8 +47,10 @@ public void validate(GtfsRoute entity, NoticeContainer noticeContainer) {
     final boolean hasShortName = entity.hasRouteShortName();
 
     if (!hasLongName && !hasShortName) {
-      noticeContainer.addValidationNotice(
-          new RouteBothShortAndLongNameMissingNotice(entity.routeId(), entity.csvRowNumber()));
+      RouteBothShortAndLongNameMissingNotice notice =
+          new RouteBothShortAndLongNameMissingNotice(entity.routeId(), entity.csvRowNumber());
```
aababilov, line 51: This is an ERROR by default, so please do not set it here explicitly.

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/Notice.java
```diff
@@ -24,6 +24,7 @@
 /** Base class for all notices produced by GTFS validator. */
 public abstract class Notice {
   private Map<String, Object> context;
+  private SeverityLevel severityLevel;
```
aababilov, line 27: severityLevel must be mentioned in equals(), toString() and hashCode().

aababilov (changes requested):
This is a good start! Thank you, Lionel.

lionel-nj:
Thanks @aababilov for your review! I modified the code and provided sample usage for `SeverityLevel.INFO` and `SeverityLevel.WARNING`.

lionel-nj:
I noticed that some notices can be used in different context (as `ERROR` or `WARNING`) e.g. `MissingTripEdgeNotice` which can be an error if `stop_times.arrival_time` is missing, or a warning if `stop_times.departure_times` is missing.
> An arrival time must be specified for the first and the last stop in a trip.
http://gtfs.org/reference/static#stop_timestxt

Is providing two constructors for this kind of notice the right approach to this situation (as done in https://github.com/MobilityData/gtfs-validator/pull/716/commits/156ccf543f92bae5a1cdfda6752bbe0c33b7fb9d)?

aababilov:
https://github.com/MobilityData/gtfs-validator/commit/156ccf543f92bae5a1cdfda6752bbe0c33b7fb9d is very nice. We actually need just a single constructor that requires to provide severity explicitly, so that the caller never forgets to set it.

aababilov:
Very good, shit it! And thanks!

aababilov (approved):

lionel-nj:
Thanks @aababilov!

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/ValidationNotice.java
```diff
@@ -27,7 +27,12 @@
  * notices.
  */
 public abstract class ValidationNotice extends Notice {
+  // default constructor: might be removed later during refactor.
   public ValidationNotice(Map<String, Object> context) {
-    super(context);
+    // by default severityLevel is set to SeverityLevel.WARNING
```
barbeau, line 32:
```suggestion
    // by default severityLevel is set to SeverityLevel.ERROR
```

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/SystemError.java
```diff
@@ -33,6 +33,7 @@
  */
 public abstract class SystemError extends Notice {
   public SystemError(Map<String, Object> context) {
-    super(context);
+    // by default SystemError.severity is set to SeverityLevel.EEROR
```
barbeau, line 36:
```suggestion
    // by default SystemError.severity is set to SeverityLevel.ERROR
```

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/MissingTripEdgeValidator.java
```diff
@@ -54,31 +55,35 @@ public void validate(NoticeContainer noticeContainer) {
...
               }
               if (!tripFirstStop.hasDepartureTime()) {
                 noticeContainer.addValidationNotice(
                     new MissingTripEdgeNotice(
                         tripFirstStop.csvRowNumber(),
                         tripFirstStop.stopSequence(),
                         trip.tripId(),
-                        "departure_time"));
+                        "departure_time",
+                        SeverityLevel.WARNING));
```
barbeau, line 68: Overall the design here for rules that could be warnings or errors in different context is good, although IMHO MissingTripEdgeValidator isn't one of them.

http://gtfs.org/reference/static/#stop_timestxt says:

...but earlier it also says:

So in all cases all trips must have both arrival and departures times for the first and last record, or it's an ERROR.

barbeau (commented):
@lionel-nj Thanks for this, it mostly LGTM! A few corrections to typos and a comment in-line.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/MissingTripEdgeValidator.java
lionel-nj: I removed these changes from this PR so that we can focus on the design rather than making the distinction for each particular notice. I'll open a new PR to discuss this. 

So in all cases all trips must have both arrival and departures times for the first and last record, or it's an ERROR.

> >If there are not separate times for arrival and departure at a stop, enter the same value for arrival_time and departure_time.
> 
> So in all cases all trips must have both arrival and departures times for the first and last record, or it's an ERROR.
> 
> 

Yes - I missed that one, I agree it should be an ERROR in this case. Thanks :)

barbeau (approved):
👍

PR #726, opened 2021-02-05
docs: Define severity for each notice

### LINKED ISSUE #417: Errors vs warnings - Incorrect rule about shapes.txt

### LINKED ISSUE #472: Error 55 (language) should be changed from ERROR to WARNING
**Describe the bug**
Error code 55 appears to require that the language in agency.agency_lang matches feed_info.feed_lang. Agency_lang is optional, so this appears to be an erroneous error.

**To Reproduce**
Run the validator on the following dataset: http://api.511.org/transit/datafeeds?api_key=[key]&operator_id=BA

**Expected behavior**

**Witnessed behavior**
This error message was included in the validator results.
```
Notice{filename='feed_info.txt', level='ERROR', code='55', title='Mismatching feed and agency language fields', description='Agency `agency_name`: `Bay Area Rapid Transit` has mismatching `agency.agency_lang`: `null` and feed_info.feed_lang`: `en`.', extra='{feedInfoFeedLang=en, agencyAgencyLang=null, agencyName=Bay Area Rapid Transit}'}]
```

**Environment used**
- OS: macOS Catalina 10.15.6
- Java version: 11.0.5
- gtfs-validator version: v1.2.2

**Additional context**
None

### LINKED ISSUE #709: list validator's errors and warnings

lionel-nj:
closes #709 
closes #472 
closes #417 

**Summary:**

This PR suggests a category for each specific notice.

**Expected behavior:** 

Tests should pass. 


lionel-nj:
@barbeau  
I did this first categorization of notices, what do you think about that?


Also:
> 🚩Tests are broken for now. They will be fixed once we agree on the category of each notice.

aababilov:
That would increase the amount of job for me. I see that you set severity of DuplicateKeyError and DuplicatedColumnNotice to INFO while they are ERRORs. That will lead to incorrect validation reports and to errors in the core code. Please let me set the priorities because I know what they should be and what we actually have in the production feeds in the world.

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/DuplicateKeyError.java
```diff
@@ -18,7 +18,11 @@
 
 import com.google.common.collect.ImmutableMap;
 
-/** The values of the given key and rows are duplicates. */
+/**
+ * The values of the given key and rows are duplicates.
+ *
+ * <p>Severity: {@code SeverityLevel.INFO}
```
barbeau, line 24: Should be an ERROR? You can't have multiple identical primary keys.

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/DuplicatedColumnNotice.java
```diff
@@ -18,7 +18,11 @@
 
 import com.google.common.collect.ImmutableMap;
 
-/** The input file CSV header has the same column name repeated. */
+/**
+ * The input file CSV header has the same column name repeated.
+ *
+ * <p>Severity: {@code SeverityLevel.INFO}
```
barbeau, line 24: Should be an WARNING? The spec doesn't explicitly require this, although it would be a big headache for consumers, as they wouldn't know which column to use.

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/DuplicateKeyError.java
barbeau: Also, this is listed as DuplicatedKeyError in the RULES.md (extra `d`)

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NonAsciiOrNonPrintableCharNotice.java
```diff
@@ -18,7 +18,11 @@
 
 import com.google.common.collect.ImmutableMap;
 
-/** A row in the input file has a different number of values than specified by the CSV header. */
+/**
```
barbeau, line 21: This header doesn't match this notice. NonAsciiOrNonPrintableCharNotice should be a WARNING as discussed in https://github.com/MobilityData/gtfs-validator/pull/728

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/UnexpectedEnumValueError.java
```diff
@@ -18,7 +18,11 @@
 
 import com.google.common.collect.ImmutableMap;
 
-/** An enum has an unexpected value. */
+/**
+ * An enum has an unexpected value.
+ *
+ * <p>Severity: {@code SeverityLevel.ERROR}
```
barbeau, line 24: I think this has to be a WARNING? There are a lot of examples where producers have extended enumeration values in their own feeds with values that don't exist in the spec. And to my knowledge the spec doesn't explicitly say you can't use non-standard enum values.

main/src/main/java/org/mobilitydata/gtfsvalidator/notice/BlockTripsWithOverlappingStopTimesNotice.java
```diff
@@ -3,7 +3,11 @@
 import com.google.common.collect.ImmutableMap;
 import org.mobilitydata.gtfsvalidator.type.GtfsDate;
 
-/** Describes two trips with the same block id that have overlapping stop times. */
+/**
+ * Describes two trips with the same block id that have overlapping stop times.
+ *
+ * <p>Severity: {@code SeverityLevel.WARNING}
```
barbeau, line 9: The spec doesn't explicitly say you can't do this, although it logically doesn't make sense. So for the canonical validator, I believe this should still be a WARNING?

@MobilityData/transit-specs?

Also, this is listed as `BlockTripsWithOverlappingStopTimes` (no `Notice` on the end) in the RULES.md table.

main/src/main/java/org/mobilitydata/gtfsvalidator/notice/DecreasingShapeDistanceNotice.java
```diff
@@ -18,6 +18,15 @@
...
 import com.google.common.collect.ImmutableMap;
 
+/**
+ * When sorted on `shapes.shape_pt_sequence` key, shape points should have increasing values for
+ * `shapes.shape_dist_traveled`
+ *
+ * <p>"Values must increase along with shape_pt_sequence."
+ * (http://gtfs.org/reference/static/#shapestxt)
+ *
+ * <p>Severity: {@code SeverityLevel.WARNING}
```
barbeau, line 28: Needs to be an ERROR - spec says `must`:

>Values must increase along with shape_pt_sequence.

main/src/main/java/org/mobilitydata/gtfsvalidator/notice/DecreasingStopTimeDistanceNotice.java
```diff
@@ -18,7 +18,15 @@
...
 
-/** Stop times of a trip have increasing distance (stops.shape_dist_traveled) */
+/**
+ * When sorted on `stops.stop_sequence` key, stop times should have increasing values for
+ * `stops.shape_dist_traveled`
+ *
+ * <p>"Values used for shape_dist_traveled must increase along with stop_sequence"
+ * (http://gtfs.org/reference/static/#stoptimestxt)
+ *
+ * <p>Severity: {@code SeverityLevel.WARNING}
```
barbeau, line 28: Needs to be ERROR - uses `must`

main/src/main/java/org/mobilitydata/gtfsvalidator/notice/OverlappingFrequencyNotice.java
```diff
@@ -22,7 +22,8 @@ public OverlappingFrequencyNotice(
             "prevEndTime", prevEndTime.toHHMMSS(),
             "currCsvRowNumber", currCsvRowNumber,
             "currStartTime", currStartTime.toHHMMSS(),
-            "tripId", tripId));
+            "tripId", tripId),
+        SeverityLevel.WARNING);
```
barbeau, line 26: The spec uses `may`, so WARNING is correct here per RFC2119. 👍 

Although this notice is missing the severity designation in the Javadoc.

main/src/main/java/org/mobilitydata/gtfsvalidator/notice/PlatformWithoutParentStationNotice.java
```diff
@@ -22,6 +22,8 @@
  * A platform has no `parent_station` field set.
  *
  * <p>This is different from {@code LocationWithoutParentStationNotice} since it is less severe.
+ *
+ * <p>Severity: {@code SeverityLevel.ERROR}
```
barbeau, line 26: This needs to be a WARNING. The spec says parent_station is optional for platforms.

main/src/main/java/org/mobilitydata/gtfsvalidator/notice/RouteColorContrastNotice.java
```diff
@@ -19,6 +19,12 @@
 import com.google.common.collect.ImmutableMap;
 import org.mobilitydata.gtfsvalidator.type.GtfsColor;
 
+/**
+ * "The color difference between route_color and route_text_color should provide sufficient contrast
+ * when viewed on a black and white screen." (http://gtfs.org/best-practices/#shapestxt)
```
barbeau, line 24: The URL here is wrong - it should be http://gtfs.org/reference/static/#routestxt

main/src/main/java/org/mobilitydata/gtfsvalidator/notice/SameNameAndDescriptionForRouteNotice.java
```diff
@@ -20,6 +20,12 @@
 
 import com.google.common.collect.ImmutableMap;
 
+/** A {@code GtfsRoute] has identical value for `routes.route_desc` and
+ * `routes.route_long_name`/`routes/route_short_name.
+ * <p>"Do not simply duplicate the name of the location." (http://gtfs.org/reference/static#routestxt)
+ *
+ * <p>Severity: {@code SeverityLevel.WARNING}
```
barbeau, line 27: I think we can safely call this an ERROR. The spec says `do not`.

main/src/main/java/org/mobilitydata/gtfsvalidator/notice/StartAndEndDateOutOfOrderNotice.java
```diff
@@ -19,6 +19,11 @@
 import com.google.common.collect.ImmutableMap;
 import org.mobilitydata.gtfsvalidator.type.GtfsDate;
 
+/**
+ * Two {@code GtfsDate} are out of order
+ *
+ * <p>Severity: {@code SeverityLevel.WARNING}
```
barbeau, line 25: I think we can safely call this an ERROR. feed_info.txt says:

> The feed_end_date date must not precede the feed_start_date date if both are given. 

It's not explicitly listed for calendar.txt, but it doesn't make sense to be to split off separate WARNINGs for that file.

@MobilityData/transit-specs?

main/src/main/java/org/mobilitydata/gtfsvalidator/notice/StartAndEndTimeOutOfOrderNotice.java
```diff
@@ -19,6 +19,11 @@
 import com.google.common.collect.ImmutableMap;
 import org.mobilitydata.gtfsvalidator.type.GtfsTime;
 
+/**
+ * Two {@code GtfsTime} are out of order
+ *
+ * <p>Severity: {@code SeverityLevel.WARNING}
```
barbeau, line 25: Hmmm...seems like this should be an obvious error for frequencies.txt start_time and end_time, but the spec doesn't explicitly say it.

@MobilityData/transit-specs?

main/src/main/java/org/mobilitydata/gtfsvalidator/notice/StopTimeWithArrivalBeforePreviousDepartureTimeNotice.java
```diff
@@ -19,6 +19,11 @@
 import com.google.common.collect.ImmutableMap;
 import org.mobilitydata.gtfsvalidator.type.GtfsTime;
 
+/**
+ * Two {@code GtfsTime} are out of order
+ *
+ * <p>Severity: {@code SeverityLevel.WARNING}
```
barbeau, line 25: Hmmm...seems like this should be an obvious error for stop_times.txt arrival_time and departure_time, but the spec doesn't explicitly say it.

@MobilityData/transit-specs?

main/src/main/java/org/mobilitydata/gtfsvalidator/notice/StopTimeWithDepartureBeforeArrivalTimeNotice.java
```diff
@@ -19,6 +19,11 @@
 import com.google.common.collect.ImmutableMap;
 import org.mobilitydata.gtfsvalidator.type.GtfsTime;
 
+/**
+ * Two {@code GtfsTime} are out of order
+ *
+ * <p>Severity: {@code SeverityLevel.WARNING}
```
barbeau, line 25: Hmmm...seems like this should be an obvious error for stop_times.txt arrival_time and departure_time, but the spec doesn't explicitly say it.

@MobilityData/transit-specs?

barbeau (commented):
Thanks @lionel-nj! My comments on errors vs. warnings are in-line, following the approach of the canonical validator where only items that are explicitly required in the spec (based on [RFC 2119](https://tools.ietf.org/html/rfc2119) language definitions) can be errors. I believe some require the feedback of @MobilityData/transit-specs.

main/src/main/java/org/mobilitydata/gtfsvalidator/notice/RouteColorContrastNotice.java
f8full: https://gtfs.mobilitydata.org/spec/gtfs-schedule#h.elenes7l8gmy

lionel-nj:
Thanks @barbeau just updated this PR. I will wait for @MobilityData/transit-specs to confirm the comments I left unresolved.

main/src/main/java/org/mobilitydata/gtfsvalidator/notice/BlockTripsWithOverlappingStopTimesNotice.java
aababilov: This is a data bug, that must be an ERROR.

I would like to repeat: we should not be prescriptionists. Forget the GTFS spec. It was written ages ago and it not describing the real feeds. The source of truth are the existing use cases. We may and we should update GTFS reference when somethings that it says does not make sense.

main/src/main/java/org/mobilitydata/gtfsvalidator/notice/OverlappingFrequencyNotice.java
aababilov: This should be an ERROR.

Again, taking the GTFS reference and writing validator strictly based on it is like taking the Webster's *A Compendious Dictionary of the English Language* published in 1806 and saying that all words that don't exist in it are not proper English.

main/src/main/java/org/mobilitydata/gtfsvalidator/notice/BlockTripsWithOverlappingStopTimesNotice.java
f8full: I agree with @aababilov on the need to follow the current data that is already live in the field. Problematic -negative- habits could be rooted out in a spec update indeed later down the line.
That being said, some kind of process is then needed to know what's in the current data, I certainly don't have much expertise myself in that regard. I'll be watching the conversation 👀

timMillet: I agree, it doesn't logically make sense. This can be treated as an ERROR.
Registered in Discrepancies Specification vs Validator https://github.com/MobilityData/gtfs-tasks/issues/87

main/src/main/java/org/mobilitydata/gtfsvalidator/notice/StartAndEndDateOutOfOrderNotice.java
timMillet: I agree, it doesn't logically make sense. This can be treated as an ERROR.
Registered in Discrepancies Specification vs Validator https://github.com/MobilityData/gtfs-tasks/issues/87

main/src/main/java/org/mobilitydata/gtfsvalidator/notice/StartAndEndTimeOutOfOrderNotice.java
timMillet: I agree, it doesn't logically make sense. This can be treated as an ERROR.
Registered in Discrepancies Specification vs Validator https://github.com/MobilityData/gtfs-tasks/issues/87

main/src/main/java/org/mobilitydata/gtfsvalidator/notice/StopTimeWithArrivalBeforePreviousDepartureTimeNotice.java
timMillet: I agree, it doesn't logically make sense. This can be treated as an ERROR.
Registered in Discrepancies Specification vs Validator https://github.com/MobilityData/gtfs-tasks/issues/87

main/src/main/java/org/mobilitydata/gtfsvalidator/notice/StopTimeWithDepartureBeforeArrivalTimeNotice.java
timMillet: I agree, it doesn't logically make sense. This can be treated as an ERROR.
Registered in Discrepancies Specification vs Validator https://github.com/MobilityData/gtfs-tasks/issues/87

lionel-nj:
@barbeau this PR is now ready for review. Thanks @timMillet !

main/src/main/java/org/mobilitydata/gtfsvalidator/notice/OverlappingFrequencyNotice.java
timMillet: The spec says "Multiple headways for the same trip are allowed, but may not overlap". "May not" should be considered as "must not", and treated as an ERROR.

lionel-nj: This notice has been changed to `ERROR` in https://github.com/MobilityData/gtfs-validator/pull/726/commits/a5155d58b1a2be6cf996be4c8bc1b666cb845d14.

barbeau (commented):
@lionel-nj This looks pretty good! I found two things that needs to be changed, plus another I'd like feedback on from @MobilityData/transit-specs.

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/DuplicatedColumnNotice.java
```diff
@@ -18,7 +18,11 @@
 
 import com.google.common.collect.ImmutableMap;
 
-/** The input file CSV header has the same column name repeated. */
+/**
+ * The input file CSV header has the same column name repeated.
+ *
+ * <p>Severity: {@code SeverityLevel.WARNING}
```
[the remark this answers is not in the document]
timMillet: IMHO, this is an ERROR. The consumer would not know which column to rely on, and this would make the dataset unusable.
Is there a similar rule for files that get provided twice or more?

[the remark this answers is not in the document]
f8full: @timmillet you link to non public ressources, is that wanted? :)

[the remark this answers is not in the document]
lionel-nj: @timMillet 

> Is there a similar rule for files that get provided twice or more?

We actually do not have this rule implemented. @barbeau should we open an issue in the validator now on this matter?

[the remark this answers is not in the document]
lionel-nj: @timMillet we do not have this rule implemented. But from my understanding having such a situation is not possible (@barbeau  correctly me if I'm wrong please). One cannot have two .txt files having the same name on the same repository path: one of the files would be name "agency-copy.txt", which would be flagged as UnknownFileError.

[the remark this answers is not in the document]
barbeau: >Is there a similar rule for files that get provided twice or more?
> One cannot have two .txt files having the same name on the same repository path: one of the files would be name "agency-copy.txt", which would be flagged as UnknownFileError.

From some quick Googling it seems like it might be possible to intentionally include two files with the same file name within a zip file, but it would take some intentional programming jujitsu as most normal zip tools won't allow it. I'd be inclined to ignore this case until we encounter it in practice for GTFS, given we'd need to programmatically examine the zip file to detect it (i.e., we couldn't extract the files otherwise they'd collide and overwrite one another).

barbeau:
@lionel-nj Could you resolve the conflicts here, and change DuplicatedColumnNotice to ERROR?

If you do that, and update the notices merged in https://github.com/MobilityData/gtfs-validator/pull/756 (just the Javadocs, I think, to match those in this PR), then I think this should be ready to merge!

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/DuplicatedColumnNotice.java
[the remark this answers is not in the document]
aababilov: Please leave it as an ERROR. 

1. This is a de facto requirement met by thousands of feeds at Google.
2. If column names are duplicated, which column should we use then? How the feed provider may know that Validator uses a certain column? Consider ambiguous error messages that say that you have an invalid value in column `xyz` when there are multiple columns with that name.

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/EmptyFileNotice.java
```diff
@@ -18,7 +18,11 @@
 
 import com.google.common.collect.ImmutableMap;
 
-/** A CSV file is empty. */
+/**
+ * A CSV file is empty.
+ *
+ * <p>Severity: {@code SeverityLevel.ERROR}
```
aababilov, line 24: We do not enforce that check at Google, so this should be rather a warning. This is not a blocker for using feed data if, e.g., there is an empty file `pathways.txt`. An empty file is practically equivalent to a missing file. 

We have 3 situations that may be treated as equivalent:
* a file is missing
* a file has only headers
* a file is empty

GTFS marks certain files as required. I think that we should amend the GTFS reference and say that a required file must contain at least one record (not counting the headers).

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/UnexpectedEnumValueError.java
```diff
@@ -18,7 +18,11 @@
 
 import com.google.common.collect.ImmutableMap;
 
-/** An enum has an unexpected value. */
+/**
+ * An enum has an unexpected value.
+ *
+ * <p>Severity: {@code SeverityLevel.WARNING}
```
aababilov, line 24: We treat it as an ERROR at Google but it probably may be a WARNING in the upstream.

main/src/main/java/org/mobilitydata/gtfsvalidator/notice/BlockTripsWithOverlappingStopTimesNotice.java
```diff
@@ -24,7 +28,8 @@ public BlockTripsWithOverlappingStopTimesNotice(
             .put("serviceIdB", serviceIdB)
             .put("blockId", blockId)
             .put("intersection", intersection.toYYYYMMDD())
-            .build());
+            .build(),
+        SeverityLevel.WARNING);
```
aababilov, line 32: It is documented as ERROR at the top. Google enforces that as an ERROR, so let's make it an ERROR.

main/src/main/java/org/mobilitydata/gtfsvalidator/notice/FeedExpirationDateNotice.java
```diff
@@ -19,6 +19,14 @@
 import com.google.common.collect.ImmutableMap;
 import org.mobilitydata.gtfsvalidator.type.GtfsDate;
 
+/**
+ * "At any time, the published GTFS dataset should be valid for at least the next 7 days, and
```
aababilov, line 23: You can remove quotes here.

main/src/main/java/org/mobilitydata/gtfsvalidator/notice/FeedInfoLangAndAgencyLangMismatchNotice.java
```diff
@@ -18,6 +18,11 @@
 
 import com.google.common.collect.ImmutableMap;
 
+/**
+ * {@code agency.agency_lang} and {@code feed_info.feed_lang} do not match
+ *
+ * <p>Severity: {@code SeverityLevel.WARNING}
```
aababilov, line 24: Let's make it an ERROR. If feed gives different agency_lang and feed_lang, the intended meaning of that is undefined. Google rejects such feeds.

main/src/main/java/org/mobilitydata/gtfsvalidator/notice/SameNameAndDescriptionForRouteNotice.java
```diff
@@ -20,6 +20,12 @@
 
 import com.google.common.collect.ImmutableMap;
 
+/** A {@code GtfsRoute} has identical value for `routes.route_desc` and
+ * `routes.route_long_name`/`routes/route_short_name.
```
aababilov, line 24: Backticks do not work in Javadoc, so you have to rewrite that as {@code }

main/src/main/java/org/mobilitydata/gtfsvalidator/notice/SameNameAndDescriptionForRouteNotice.java
```diff
@@ -20,6 +20,12 @@
...
+ * <p>"Do not simply duplicate the name of the location." (http://gtfs.org/reference/static#routestxt)
+ *
+ * <p>Severity: {@code SeverityLevel.ERROR}
```
aababilov, line 27: This should be a WARNING. This does not make the feed non-functional.

aababilov:
That is a big contribution, Lionel. Thank you!

I left some comments mostly based on the real production feeds.

aababilov (changes requested):

main/src/main/java/org/mobilitydata/gtfsvalidator/notice/BlockTripsWithOverlappingStopTimesNotice.java
lionel-nj: Thanks for flagging that.

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/UnexpectedEnumValueError.java
lionel-nj: @barbeau  suggested to make that one a `WARNING` 

> I think this has to be a WARNING? There are a lot of examples where producers have extended enumeration values in their own feeds with values that don't exist in the spec. And to my knowledge the spec doesn't explicitly say you can't use non-standard enum values.

from https://github.com/MobilityData/gtfs-validator/pull/726#discussion_r574691932

main/src/main/java/org/mobilitydata/gtfsvalidator/notice/SameNameAndDescriptionForRouteNotice.java
lionel-nj: > I think we can safely call this an ERROR. The spec says do not.

from https://github.com/MobilityData/gtfs-validator/pull/726#discussion_r574703383

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/EmptyFileNotice.java
barbeau: The Javadocs for `CsvFile.isEmpty()` currently say:

>Tells if the file is empty, i.e. it has no rows and even no headers.

And GTFS says:

>The first line of each file must contain field names. 

...so this seems to align as an ERROR to me. Let's discuss in a new issue/PR if needed so we can get this merged - it's getting hard to manage this PR due to length.

main/src/main/java/org/mobilitydata/gtfsvalidator/notice/FeedInfoLangAndAgencyLangMismatchNotice.java
barbeau: Let's discuss in a new issue/PR if needed so we can get this merged - it's getting hard to manage this PR due to length.

lionel-nj: New issue on this matter: https://github.com/MobilityData/gtfs-validator/issues/762

aababilov:
Hi @lionel-nj 

Please can you remove FieldParsingError.java file that resurrected in this PR?

Thanks!

lionel-nj:
Sure @aababilov! It's done in https://github.com/MobilityData/gtfs-validator/pull/775/

PR #727, opened 2021-02-08
feat: Export severity in JSON

aababilov:
Issue #549

Notices of different severities are grouped independently. This we make
sure that they the MAX_EXPORTS_PER_NOTICE_TYPE limit is applied to
errors and warnings independently.

We chose "severity" instead of "severityLevel" in the JSON format
because the former is shorter.

We also sort the exported notices alphabetically by code and severity.
This makes the report more readable (earlier we sorted them by hashes
that is stable but randomized).


barbeau (approved):
Thanks @aababilov, LGTM!

PR #728, opened 2021-02-08
feat: Treat NonAsciiOrNonPrintableCharNotice as a warning instead of error

aababilov:
Issue #549

An error in RowParser stops validation for the whole feed. However, if
ID has non-printable or non-ASCII characters, the feed may be still
successfully parsed and served to the end user.

We are lowering priority of NonAsciiOrNonPrintableCharNotice to WARNING
and modify RowParser so that it checks severity level of its notices.

We also simplify the code of asId function and reduce amount of calls to
row.asString from 3 to 1.

### DIFF
--- a/core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NonAsciiOrNonPrintableCharNotice.java
+++ b/core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NonAsciiOrNonPrintableCharNotice.java
@@ -18,14 +18,15 @@ package org.mobilitydata.gtfsvalidator.notice;
 
 import com.google.common.collect.ImmutableMap;
 
-/** A row in the input file has a different number of values than specified by the CSV header. */
+/** ID value contains something different from printable ASCII characters. */
 public class NonAsciiOrNonPrintableCharNotice extends ValidationNotice {
   public NonAsciiOrNonPrintableCharNotice(String filename, long csvRowNumber, String columnName) {
     super(
         ImmutableMap.of(
             "filename", filename,
             "csvRowNumber", csvRowNumber,
-            "columnName", columnName));
+            "columnName", columnName),
+        SeverityLevel.WARNING);
   }
 
   @Override
--- a/core/src/main/java/org/mobilitydata/gtfsvalidator/parsing/RowParser.java
+++ b/core/src/main/java/org/mobilitydata/gtfsvalidator/parsing/RowParser.java
@@ -31,6 +31,7 @@ import org.mobilitydata.gtfsvalidator.notice.MissingRequiredFieldError;
 import org.mobilitydata.gtfsvalidator.notice.NonAsciiOrNonPrintableCharNotice;
 import org.mobilitydata.gtfsvalidator.notice.NoticeContainer;
 import org.mobilitydata.gtfsvalidator.notice.NumberOutOfRangeError;
+import org.mobilitydata.gtfsvalidator.notice.SeverityLevel;
 import org.mobilitydata.gtfsvalidator.notice.UnexpectedEnumValueError;
 import org.mobilitydata.gtfsvalidator.notice.ValidationNotice;
 import org.mobilitydata.gtfsvalidator.type.GtfsColor;
@@ -45,6 +46,7 @@ import org.mobilitydata.gtfsvalidator.type.GtfsTime;
  * throw an exception.
  */
 public class RowParser {
+
   public static final boolean REQUIRED = true;
   public static final boolean OPTIONAL = false;
   private final NoticeContainer noticeContainer;
@@ -190,7 +192,7 @@ public class RowParser {
 
   public void checkRowColumnCount(CsvFile csvFile) {
     if (row.getColumnCount() != csvFile.getColumnCount()) {
-      addErrorInRow(
+      addNoticeInRow(
           new InvalidRowLengthError(
               csvFile.getFileName(),
               row.getRowNumber(),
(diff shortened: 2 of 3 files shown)


core/src/main/java/org/mobilitydata/gtfsvalidator/parsing/RowParser.java
```diff
@@ -432,13 +447,13 @@ public GtfsColor asColor(int columnIndex, boolean required) {
     try {
       i = Integer.parseInt(s);
     } catch (Exception ex) {
-      addErrorInRow(
+      addNoticeInRow(
```
aimiebell, line 450: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Use of identifier that is a keyword in later versions of Java (NM_FUTURE_KEYWORD_USED_AS_IDENTIFIER)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-use-of-identifier-that-is-a-keyword-in-later-versions-of-java-nm-future-keyword-used-as-identifier).
The identifier is a word that is reserved as a keyword in later versions of Java, and your code will need to be changed in order to compile it in later versions of Java.

aababilov: I am unsure if that is what you wanted to say. addNoticeInRow is not reserved as a keyword in later versions of Java, as far as I know. Please correct me if I am wrong.

aababilov:
Friendly ping here! That's a blocker because right now our upstream validator rejects valid feeds that have non-ASCII characters in IDs.

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/LeadingOrTrailingWhitespacesNotice.java
```diff
@@ -0,0 +1,58 @@
+/*
```
barbeau, line 1: Note that this has a dependency on https://github.com/MobilityData/gtfs-validator/pull/729 as both commits and included here (although I believe the new test for NonAsciiOrNonPrintableCharNotice is missing). Can that dependency be removed so we can merge this separately?

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NonAsciiOrNonPrintableCharNotice.java
```diff
@@ -18,14 +18,15 @@
...
+/** ID value contains something different from printable ASCII characters. */
 public class NonAsciiOrNonPrintableCharNotice extends ValidationNotice {
   public NonAsciiOrNonPrintableCharNotice(String filename, long csvRowNumber, String columnName) {
     super(
         ImmutableMap.of(
             "filename", filename,
             "csvRowNumber", csvRowNumber,
-            "columnName", columnName));
+            "columnName", columnName),
+        SeverityLevel.INFO);
```
barbeau, line 29: Shouldn't this be `SeverityLevel.WARNING`?

core/src/main/java/org/mobilitydata/gtfsvalidator/parsing/RowParser.java
```diff
@@ -203,10 +206,19 @@ public void checkRowColumnCount(CsvFile csvFile) {
   public String asString(int columnIndex, boolean required) {
     String s = row.asString(columnIndex);
     if (required && s == null) {
-      addErrorInRow(
+      addNoticeInRow(
           new MissingRequiredFieldError(
               row.getFileName(), row.getRowNumber(), row.getColumnName(columnIndex)));
     }
+    if (s != null) {
+      final String trimmed = s.strip();
```
barbeau, line 214: Nit - given that there is also a Java method `trim()`, this variable is probably better named `stripped`.

barbeau (commented):
Thanks @aababilov! Converting this to a warning LGTM, as [the spec](https://gtfs.org/reference/static#field-types) says it's recommended (but not required) to use printable ASCII:

>Using only printable ASCII characters is recommended.

A few comments in-line

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NonAsciiOrNonPrintableCharNotice.java
aababilov: OK, done.

core/src/main/java/org/mobilitydata/gtfsvalidator/parsing/RowParser.java
aababilov: Done.

By the way, can you tell the difference between "strip" and "trim" as a native English speaker? To me, these words are interchangeable and one must check the documentation to understand the difference :)

aababilov: Actually, that change belongs to another pull request. Fixed.

https://github.com/MobilityData/gtfs-validator/pull/729

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/LeadingOrTrailingWhitespacesNotice.java
aababilov: Done. My bad, GitHub is a bit challenging when you have dependent pull requests :)

core/src/main/java/org/mobilitydata/gtfsvalidator/parsing/RowParser.java
barbeau: >By the way, can you tell the difference between "strip" and "trim" as a native English speaker? 

Not in the context of Java :). In general use "strip" has somewhat of a more aggressive and coarse connotation than "trim". "trim" implies a more precise and neat action.

barbeau (approved):
Thanks @aababilov!

aababilov:
Thank you for review!

PR #730, opened 2021-02-08
feat: Allow agencies to omit agency_lang without errors

aababilov:
Some real-world feeds provide agency_lang only for some agencies and
omit it for the rest since this field is optional. Those feeds are
accepted at Google, that is why we relax the existing validation rule.

In general, agency_lang field should be discouraged and feed_lang should
be used instead.

### DIFF
--- a/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/AgencyConsistencyValidator.java
+++ b/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/AgencyConsistencyValidator.java
@@ -41,6 +41,7 @@ import org.mobilitydata.gtfsvalidator.table.GtfsAgencyTableLoader;
  */
 @GtfsValidator
 public class AgencyConsistencyValidator extends FileValidator {
+
   @Inject GtfsAgencyTableContainer agencyTable;
 
   @Override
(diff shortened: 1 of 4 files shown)


main/src/main/java/org/mobilitydata/gtfsvalidator/validator/AgencyConsistencyValidator.java
```diff
@@ -75,18 +74,26 @@ public void validate(NoticeContainer noticeContainer) {
...
       }
-      // FIXME: this should be a warning. See
-      // https://github.com/MobilityData/gtfs-validator/issues/549.
-      if (hasLanguage != agency.hasAgencyLang()
-          || (hasLanguage
-              && agency.hasAgencyLang()
-              && !commonLanguage.equals(agency.agencyLang()))) {
+    }
+
+    // agency_lang field is optional. All provided values must be the same for all agencies.
```
barbeau, line 79: From what I can tell the spec doesn't require all the agencies to have the same `agency_lang`? This isn't called out in the definition like `agency_timezone`.

If so this comment needs to be revised and the notice should be downgraded from ERROR to WARNING.

barbeau (commented):
Thanks @aababilov! One comment in-line

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/AgencyConsistencyValidator.java
aababilov: This is a de-facto requirement for thousands of feeds at Google. I agree to change that to WARNING in this pull request but let's revisit that later.

A feed that provides different agency_lang is almost certainly making the wrong thing - we discussed that on https://github.com/MobilityData/gtfs-validator/pull/711 Having different agency_lang is ambiguous and there is no officially recommended way on how to treat that (is that a language for `agency_name`? for `route_long_name`? for `stop_name`? nobody knows!)

aababilov: I have introduced different types of notices: InconsistentAgencyLangNotice and InconsistentAgencyTimezoneNotice.

I agree that they have different severity and they should be treated differently.

barbeau (approved):
LGTM, thanks @aababilov!

aababilov:
Thanks a lot!

PR #731, opened 2021-02-09
feat: Put @Generated annotation on enums

aababilov:
This is a common Java recommendation that all generated classes should
have @Generated annotation.


barbeau (approved):
👍

PR #735, opened 2021-02-09
fix: remove requirement on feed_info.start_date and feed_info.end_date

### LINKED ISSUE #733: [BUGI] feed_info_start_date and feed_info_end_date requirements
At present, the validation rule generates a `MissingRequiredFieldError` when one of `feed_info_start_date` and `feed_info_end_date` fields is missing.

This should not happen since the specification does not explicitly require these fields. However, such a situation should be flagged as a `WARNING`. 

https://github.com/MobilityData/gtfs-validator/blob/feff1ec4ca3a1edb53f2b3369e66c0639e05a7f0/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/FeedServiceDateValidator.java#L47

lionel-nj:
closes #733 
**Summary:**

This PR removes the requirements on feed_info.start_date and feed_info.end_date

**Expected behavior:** 

No notice should be generated if  `feed_info.start_date` and `feed_info.end_date` is missing Also, both fields can be blank.

### DIFF
--- a/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/FeedServiceDateValidator.java
+++ b/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/FeedServiceDateValidator.java
@@ -18,7 +18,7 @@ package org.mobilitydata.gtfsvalidator.validator;
 
 import org.mobilitydata.gtfsvalidator.annotation.GtfsValidator;
 import org.mobilitydata.gtfsvalidator.annotation.Inject;
-import org.mobilitydata.gtfsvalidator.notice.MissingRequiredFieldError;
+import org.mobilitydata.gtfsvalidator.notice.MissingFeedInfoDateNotice;
 import org.mobilitydata.gtfsvalidator.notice.NoticeContainer;
 import org.mobilitydata.gtfsvalidator.notice.StartAndEndDateOutOfOrderNotice;
 import org.mobilitydata.gtfsvalidator.table.GtfsFeedInfo;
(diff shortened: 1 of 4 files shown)


main/src/main/java/org/mobilitydata/gtfsvalidator/validator/FeedServiceDateValidator.java
```diff
@@ -38,15 +35,6 @@
   @Override
   public void validate(NoticeContainer noticeContainer) {
     for (GtfsFeedInfo feedInfo : feedInfoTable.getEntities()) {
-      if (feedInfo.hasFeedStartDate() && !feedInfo.hasFeedEndDate()) {
```
barbeau, line 41: Instead of removing the rule logic entirely, shouldn't we just downgrade the existing notice to a WARNING?

barbeau (commented):
@lionel-nj Thanks - one question below

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/FeedServiceDateValidator.java
aababilov: That's a great idea, let's have a warning.

lionel-nj: Done in https://github.com/MobilityData/gtfs-validator/pull/735/commits/167a4cd757507539d4cdfb3ae98e686a90363625

maximearmstrong (approved):
LGTM

PR #741, opened 2021-02-11
feat: Report errors for new lines and carriage returns in field values

aababilov:
GTFS reference requires that field values do not contain new lines or
carriage returns (\n or \r).

RFC4180 CSV allows to spread a value between multiple lines if it is
surrounded by double quotes. But on practice, the feeds that have
multiline values are actually omitting the closing quote. That makes
different parsers interpret the file in different ways, so we need to
detect such ambiguous data in the validator.

### DIFF
--- a/core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NewLineInValueNotice.java
+++ b/core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NewLineInValueNotice.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.mobilitydata.gtfsvalidator.notice;
+
+import com.google.common.collect.ImmutableMap;
+
+/**
+ * A value in CSV file has a new line or carriage return.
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NewLineInValueNotice.java
```diff
@@ -0,0 +1,56 @@
...
+        ImmutableMap.of(
+            "filename",
+            filename,
+            "csvRowNumber",
+            csvRowNumber,
+            "fieldName",
+            fieldName,
+            "fieldValue",
+            fieldValue),
+        SeverityLevel.ERROR);
```
barbeau, line 49: Ugh - so unfortunately the GTFS reference says:

>Field values may not contain tabs, carriage returns or new lines.

[RFC 2119](https://tools.ietf.org/html/rfc2119) defines "MAY" as optional, but does not address "MAY NOT". 

Microsoft apparently treats "MAY NOT" as a prohibited behavior in their [open office spec](https://docs.microsoft.com/en-us/openspecs/office_standards/ms-oe376/cc313153-88b6-4a89-9fdd-8f21e8d1ffb3), and I think most English speakers would agree that this is the intent here in GTFS. So I think it's ok to treat this is as an ERROR and not a WARNING.

@MobilityData/transit-specs?

To align with RFC 2119 the language in the spec should be updated to use `must` instead of `may`.

barbeau (commented):
@aababilov This LGTM, although question for @MobilityData/transit-specs in-line.

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NewLineInValueNotice.java
aababilov: I am open to any point of view.

Sean, if you wish, I can make it WARNING right now and we can change it to ERROR later. I can always patch it to ERROR at Google since we really suffer from such feeds.

timMillet: I agree: `may not` should be treated as `must not`.

barbeau (approved):
👍

Issue #734, opened 2021-02-09
Write higher level acceptance tests


lionel-nj:
At present, most implemented tests are unit test. These make sure that the individual units / components of are tested. 

Higher level test should be written to verify the execution of multiple modules.

lionel-nj:
After discussion with @barbeau we think that executing the validator on sample GTFS datasets, and analyzing the output of the generated report could be a way to write such tests.
We could start by elaborating GTFS datasets that will lead to the generation of different system errors, and extend the same process to validation errors.

Any thoughts/advise? @aababilov?

derhuerst:
From the related Issue https://github.com/public-transport/ideas/issues/17, which in turn quotes  https://github.com/georust/transitfeed/issues/5#issuecomment-649095015:

> > ## reusability
> > Answering quite basic (but very relevant in practical usage of GTFS) questions like _When does any vehicle depart at a bus stop?_ is surprisingly much work: [GTFS Time values are inherently timezone-dependent](https://gist.github.com/derhuerst/574edc94981a21ef0ce90713f1cff7f6), `frequencies.txt` with `exact_times=1` defines "stop times" as well, etc. With the ever-growing number of optional parts and extensions, **doing GTFS processing right is a lot of work, so we should make the implementation in this project as reusable/flexible as possible.**
> > Also, a **project- and language-independent test suite**, i.e. a set of fixtures per "question"/operation, would be very helpful for this. Those have been very successful in other areas, e.g. [for WebSocket implementations](https://github.com/crossbario/autobahn-testsuite) or [for HTML5 parsers](https://github.com/html5lib/html5lib-tests).
> 
> > ## validation
> > There are a bazillion validation (i.e. "semantic checks on the actual data") cases; The [best practices page](https://gtfs.org/best-practices/) is long, and the [GTFS issue tracker](https://github.com/google/transit/labels/GTFS%20%28static%29) and [mailing lists](https://gtfs.org/getting-started/#mailing-lists) are full of edge cases. There are at east 20 libs across languages doing some form of validation, but none of them cover all the issues that we see with GTFS feeds in the wild.
[...]

PR #742, opened 2021-02-11
feat: Show a big error when parsing failed

aababilov:
If the feed failed to parse, then most validators are not invoked
because when they operate on an incomplete feed, they may produce
misleading messages. That is why we need to show a big error box to
attract user's attention to that.

```
 -----------------------------------------
|       !!!    PARSING FAILED    !!!      |
|   Most validators were never invoked.   |
|   Please see report.json for details.   |
 -----------------------------------------
```

### DIFF
--- a/main/src/main/java/org/mobilitydata/gtfsvalidator/cli/Main.java
+++ b/main/src/main/java/org/mobilitydata/gtfsvalidator/cli/Main.java
@@ -105,6 +105,13 @@ public class Main {
     // Output
     exportReport(args.getOutputBase(), noticeContainer);
     final long endNanos = System.nanoTime();
+    if (!feedContainer.isParsedSuccessfully()) {
+      System.out.println(" ----------------------------------------- ");
+      System.out.println("|       !!!    PARSING FAILED    !!!      |");
+      System.out.println("|   Most validators were never invoked.   |");
+      System.out.println("|   Please see report.json for details.   |");
+      System.out.println(" ----------------------------------------- ");
+    }
     System.out.printf("Validation took %.3f seconds%n", (endNanos - startNanos) / 1e9);
     System.out.println(feedContainer.tableTotals());
   }


main/src/main/java/org/mobilitydata/gtfsvalidator/cli/Main.java
```diff
@@ -105,6 +105,13 @@ public static void main(String[] argv) {
     // Output
     exportReport(args.getOutputBase(), noticeContainer);
     final long endNanos = System.nanoTime();
+    if (!feedContainer.isParsedSuccessfully()) {
```
barbeau, line 108: Should we move this to the end of the method and invoke `System.exit()` as well with an nonzero error code?

https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/System.html#exit(int)

I believe this would ensure that the CI builds on the GitHub workflow show up as failed as well.

barbeau (approved):
Thanks @aababilov, this is a very good idea. One comment below.

main/src/main/java/org/mobilitydata/gtfsvalidator/cli/Main.java
aababilov: No, please let's not do that. Right now, the Validator always exits with 0 and all errors are stored in the JSON files, even runtime exceptions. I agree to revisit that behaviour and have non-zero exit codes but let's have a separate discussion for that and a separate PR.

barbeau: Sounds good - let's discuss separately 👍

barbeau: Actually, this fits well into https://github.com/MobilityData/gtfs-validator/issues/734.

PR #750, opened 2021-02-15
fix: Check StopTooFarFromTripShape once per trip stoptimes

### LINKED ISSUE #747: StopTooFarFromTripShapeValidator processes trips multiple times
**Describe the bug**
First mentioned in https://github.com/MobilityData/gtfs-validator/pull/708/files#r574728828.

In the original implementation of https://github.com/MobilityData/gtfs-validator/pull/708 in v1 (https://github.com/MobilityData/gtfs-validator/pull/340), the in-memory GTFS data repository held a Map data structure where each trip_id appeared once as the key and the value was a SortedMap of StopTimes for that trip_id. StopTooFarFromTripShapeValidator  looped through each of those entries and therefore only processed each trip once.

But the similarly named variable in v2 actually contains the entire contents of stop_times.txt. So it looks like this implementation is fetching all StopTimes and then getting the trip_id for each StopTime and then getting all the StopTimes for that trip_id, which blows up the processing time significantly.

There are also some issues with the definition of constants `METER_TO_KILOMETER_CONVERSION_FACTOR ` and `KILOMETER_TO_METER_CONVERSION_FACTOR ` as specified in https://github.com/MobilityData/gtfs-validator/pull/708/files#r574738652.

**How we reproduce the bug**
1. Run StopTooFarFromTripShapeValidator

**Expected behaviour**
It processes each trip and it's stop_times once

**Observed behaviour**
Multiple duplicate notices are produced because trips are processed multiple times

**Environment versions**
- validator version: Master branch at https://github.com/MobilityData/gtfs-validator/commit/90166bf3772775a314c7673cba7378390c4921fc
- Java version: JDK 11
- OS versions: Windows 10

barbeau:
**Summary:**

Change to check StopTooFarFromTripShape once per trip stoptimes, plus a number of other improvements:

* Only process each trip once
* Refactor forEach to simple loop to clarify behavior
* Refactor distance variables and method to use double instead of float and int
* Refactor utility method signature to remove need for trip object
* Cleanup docs - remove references to "error"
* Add docs for method of distance calculation
* Fix existing tests

Closes #747

@lionel-nj Please review and also add the additional unit tests we discussed. Thanks!

**Expected behavior:** 

Only check each trip once, and only check the stop_times.txt for that trip once


lionel-nj (changes requested):
Just one comment in-line

main/src/main/java/org/mobilitydata/gtfsvalidator/util/GeospatialUtil.java
```diff
@@ -66,98 +65,94 @@ private static DistanceCalculator getDistanceCalculator() {
...
+   * @param stopTable a container for all stops (keyed on stop_id), needed to obtain the latitude
+   *     and longitude for each stop on the trip
    * @param testedCache a cache for previously tested shape_id and stop_id pairs (keyed on
    *     shape_id+stop_id). If the combination of shape_id and stop_id appears in this set, we
    *     shouldn't test it again. Shapes and stops tested in this method execution will be added to
    *     this testedCache.
-   * @return a list of E052 errors, one for each stop that is too far from the trip shapePoints
+   * @return a list of notices, one for each stop that is too far from the trip shape
    */
   public static List<StopTooFarFromTripShapeNotice> checkStopsWithinTripShape(
```
barbeau, line 107: @lionel-nj As discussed let's refactor `checkStopsWithinTripShape()` into `StopTooFarFromTripShapeValidator`. It was originally put into the GeospatialUtil class in v1 in an attempt to adhere to the clean architecture principles (because it depends on an outside library). We aren't using clean architecture now, and given this returns a list of `StopTooFarFromTripShapeNotice` it can't be used in another validation rule, so it more naturally fits in the `StopTooFarFromTripShapeValidator` class. Also, this will allow us to more easily mock it to count the number of invocations to ensure we aren't accidentally processing trips more than once. We can't rely on the notice counts in unit tests for this rule to detect accidental executions because it uses a cache to avoid spamming the user with a large number of notices for the same stop and shape combination (e.g., if multiple trips re-use the same shape), and also as an optimization to avoid checking distance on these same combinations as well. But this means that if you process the same trip twice, you won't see more than the expected number of notices for processing the trip once - hence the need to mock the method to count invocations.

lionel-nj:
@barbeau I drafted an additional unit to verify the number of invocation on `checkStopsWithinTripShape` but it does not pass although  think that they should. Do you mind taking a look at the two last assertions of `StopTooFarFromTripShapeValidatorTest.eachTripShouldOnlyBeProcessedOnce` please? I might have missed something

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
```diff
@@ -36,7 +47,7 @@
 /**
  * Validates: a {@link GtfsStop} is within a distance threshold for a trip shape.
  *
- * <p>Generated notice: {@link StartAndEndDateOutOfOrderNotice}.
+ * <p>Generated notice: {@link StopTooFarFromTripShapeNotice}.
```
aababilov, line 50: Could you also give time complexity of this algoritm?

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
```diff
@@ -47,32 +58,100 @@
...
+    // Check if each stop is within the buffer polygon
+    for (GtfsStopTime stopTime : stopTimes) {
+      GtfsStop stop = stopTable.byStopId(stopTime.stopId());
+      if (stop == null || !stop.hasStopLat() || !stop.hasStopLon()) {
+        // Lat/lon are optional for location_type 4 - skip to the next stop if they aren't
+        // provided
+        continue;
+      }
+      if (!(stop.locationType() == GtfsLocationType.STOP)
+          && !(stop.locationType() == GtfsLocationType.BOARDING_AREA)) {
```
aababilov, line 138: This can be written simpler:

if (stop.locationType() != GtfsLocationType.STOP &&
stop.locationType() != GtfsLocationType.BOARDING_AREA)

However, there is already a requirement that a trip must visit only stops or boarding areas, so I think you can omit that extra check.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
```diff
@@ -47,32 +58,100 @@
...
+    Shape shapeLine = lineBuilder.build();
+
+    // Create the buffered version of the trip as a polygon
+    Shape shapeBuffer = shapeLine.getBuffered(TRIP_BUFFER_DEGREES, shapeLine.getContext());
+
+    // Check if each stop is within the buffer polygon
+    for (GtfsStopTime stopTime : stopTimes) {
+      GtfsStop stop = stopTable.byStopId(stopTime.stopId());
+      if (stop == null || !stop.hasStopLat() || !stop.hasStopLon()) {
+        // Lat/lon are optional for location_type 4 - skip to the next stop if they aren't
```
aababilov, line 133: We should use the coordinates of the parent location (i.e., platform) if stop_lat or stop_lon are missing.

aababilov (changes requested):
Thank you!

barbeau:
>Here, I think that this set should be empty since it is the first call to checkStopsWithinTripShape.
>Here I think that this set should be [shape11002, shape11001, shape11003] since it's the second (and last) time checkStopsWithinTripShape is called.

@lionel-nj Your expectations are right, and match actual execution when you set a debugger breakpoint within `checkStopsWithinTripShape()` to see the `Set` contents when executing `eachTripShouldOnlyBeProcessedOnce()`. So it seems something is wrong with the Mockito setup?

barbeau:
>So it seems something is wrong with the Mockito setup?

So it seems that Mockito captures parameters for each method invocation by reference, so as a result after execution when we check the contents of the cache it's the final contents after all execution finishes, not the contents after each method invocation.  I pushed a change in https://github.com/MobilityData/gtfs-validator/pull/750/commits/70213ac95db889fc5df72410ced69596bb093ba1 that allows us to test each invocation separately without testing the cache, and then have a final cache test with a note in comments on the above. This fixes the test.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
barbeau: Any suggestions on how to report that the parent stop has been checked in the notice? Should we have a separate `ParentStopTooFarFromTripShapeNotice`? Or a new field in the notice map?

lionel-nj:
Wow, good to know! Thanks @barbeau!

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
lionel-nj: Don't we risk to generate twice the same notice? One once the parent is checked, and the second one when the children is checked?

barbeau: >Don't we risk to generate twice the same notice: one once the parent is checked, and the second one when the child is checked?

Good point - we would if we didn't track that we generate a notice for this parent and we encounter multiple children that don't have lat/longs for the same parent. But we could do something similar to what we do now, and track via the cache that we already generated a notice for this parent & shape combination.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
barbeau: >However, there is already a requirement that a trip must visit only stops or boarding areas, so I think you can omit that extra check.

@aababilov Is this checked in the existing validator? I'm not seeing it, but could be missing it.

main/src/main/java/org/mobilitydata/gtfsvalidator/util/GeospatialUtil.java
```diff
@@ -66,98 +66,101 @@ private static DistanceCalculator getDistanceCalculator() {
...
         || stopTimes.isEmpty()
         || shapePoints == null
         || shapePoints.isEmpty()) {
       // Nothing to do - return empty list
-      return errors;
+      return notices;
     }
 
     // Create a polyline from the GTFS shapes data
     ShapeFactory.LineStringBuilder lineBuilder = getShapeFactory().lineString();
```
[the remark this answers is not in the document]
barbeau: Good idea, but unfortunately no :(. I looked at the source of spatial4j, and it has a comment in the builder class:

>// TODO add dimensionality hint method?

Looks like it's backed by an ArrayList, so a lot of re-sizing will happen here :(

https://github.com/locationtech/spatial4j/blob/1f6e2047f0574a430fc711cf2cd5adf141a8bda9/src/main/java/org/locationtech/spatial4j/shape/jts/JtsShapeFactory.java#L312

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
barbeau: I went ahead and removed the check here in https://github.com/MobilityData/gtfs-validator/pull/750/commits/b6c414917461c0d1951ac7223691555b71e7e8c7 as I agree that it should be checked elsewhere.

main/src/main/java/org/mobilitydata/gtfsvalidator/util/GeospatialUtil.java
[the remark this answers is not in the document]
barbeau: In good news, in looking at the library source I realized that you can specify a buffered line directly in the `lineBuilder`, without having to build the line first and then generating the buffer. So that saves memory and execution time for each trip shape.

However, I tested the change and initially got some confusing results. The unit test (using an original shape to line distance of around 140m - the threshold is 100m) passed with the change, but I ran it against MBTA's dataset with the old and new code, and unexpectedly got different results - directly buffering the line in the builder gave me 8 notices, while creating the second instance (original code) gave me 6 notices.

After more poking around in spatial4j code, it looks like the difference is that when buffering within `ShapeFactory.LineStringBuilder` directly, it always returns a non-geodesic (straight-line) buffer. Conversely, when you first build and then separately buffer the line, it always returns a geodesic buffer (i.e., one that follows the surface of the earth).

The differences between the two are slight, but enough to cause differences in results on real world datasets. I manually measured the distance between one of the stops and the shape in the newly generated notice using Google Earth Pro and it measures around 106 meters.

Another consideration - as you might expect, I noticed that execution time on Euclidean buffers seems to be consistently less than on geodesic buffers. Here are some measurements running on my laptop from within IntelliJ.

Execution time - Euclidean (seconds):
* 15
* 11
* 12
* 13
* 11
* 12

Execution time - Geodesic (seconds):
* 21
* 18
* 17
* 16
* 17
* 17
* 16

Given what seems to be a decent performance increase in execution time alone (memory performance should be better too, but I didn't measure it), I'm inclined to go with the Euclidean measurements. For our purposes I think the small error doesn't really matter - our intent is to draw attention to potential GTFS errors with stops far from the trip shape, and if we flag one that's a few meters off it shouldn't be problematic.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
barbeau: I've added notes on time complexity in https://github.com/MobilityData/gtfs-validator/pull/750/commits/e47ee19d3730fa1281782e89a75a156524f65641. I actually had to think about this a bit, and hopefully I've expressed it in a way that's understandable. Improvements welcome!

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
barbeau: I've opened this issue to add checks for location_types of stop_times.stop_id stops here - https://github.com/MobilityData/gtfs-validator/issues/763.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
barbeau: So it looks like the spec was clarified since I first wrote this rule (https://github.com/google/transit/pull/258). It now states:

>Referenced locations must be stops/platforms, i.e. their stops.location_type value must be 0 or empty.

And given that `stop_lat` and `stop_lon` are required for `location_type=0` stops, this is actually a non-issue - we should never encounter stops in stop_times.txt without locations, and if we do we can ignore them.

lionel-nj (approved):
That LGTM @barbeau, thanks!

aababilov:
This still can be significantly optimized because we need to validate each line variant only once (line variant is effectively a trip without time information). However, we can submit it as-is and improve later.

barbeau:
@aababilov I agree - I opened https://github.com/MobilityData/gtfs-validator/issues/766 to discuss the line variant optimizations.

PR #756, opened 2021-02-18
feat: Split FieldParsingError into different classes

aababilov:
Format requirements for different fields are different, so that probably
does not make sense to group them all at the same error class.

Separate error classes may be placed in different sections of a
validation report that add details on the expected format.

Users of GTFS Validator may want to reduce the severity of certain
errors, e.g., validation of URLs and phone numbers.


aababilov:
@barbeau @lionel-nj 
How do you like this idea?

lionel-nj (changes requested):
Great idea! That will clarify the validation reports. Just one comment inline. Thanks @aababilov!

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/InvalidTimeNotice.java
```diff
@@ -0,0 +1,49 @@
...
+ */
+
+package org.mobilitydata.gtfsvalidator.notice;
+
+import com.google.common.collect.ImmutableMap;
+
+/**
+ * A field cannot be parsed as time.
+ *
+ * <p>Time must be in the {@code HH:MM:SS} or {@code H:MM:SS} format.
```
barbeau, line 24: We also support `HHH:MM:SS` in `GtfsTime`

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NonAsciiOrNonPrintableCharNotice.java
```diff
@@ -20,12 +20,14 @@
 
 /** ID value contains something different from printable ASCII characters. */
```
barbeau, line 21: Should have a similar comment header as other notices with more description?

core/src/main/java/org/mobilitydata/gtfsvalidator/parsing/RowParser.java
```diff
@@ -268,218 +163,181 @@ static boolean hasOnlyPrintableAscii(String s) {
...
+        columnIndex,
+        required,
+        s -> PhoneNumberUtil.getInstance().isPossibleNumber(s, feedName.getISOAlpha2CountryCode()),
+        InvalidPhoneNumberNotice::new);
   }
 
   @Nullable
   public Locale asLanguageCode(int columnIndex, boolean required) {
-    return languageCodeParser.parseField(columnIndex, required);
+    // FIXME: Enhance checks for IETF BCP 47 language code.
```
barbeau, line 199: @aababilov Could you open an issue on GitHub for this with a better description of how this should be enhanced?

barbeau (changes requested):
Thanks @aababilov! I agree that this is a good idea. Some comments in-line.

core/src/test/java/org/mobilitydata/gtfsvalidator/parsing/RowParserTest.java
```diff
@@ -173,8 +189,15 @@ public void asPhoneNumber() {
...
+  }
 
-    assertThat(createParser("au-feed", "invalid").asPhoneNumber(0, true)).isNull();
+  @Test
+  public void asPhoneInvalid() {
+    RowParser parser = createParser("us-feed", "invalid");
+    assertThat(parser.asPhoneNumber(0, true)).isNull();
+    assertThat(parser.hasParseErrorsInRow()).isTrue();
+    assertThat(parser.getNoticeContainer().getValidationNotices())
+        .containsExactly(new InvalidEmailNotice(TEST_FILENAME, 8, "column name", "invalid"));
```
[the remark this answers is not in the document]
aababilov: That was a bug in `Notice.equals()` that did not take `getCode()` into account. Fixed.

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/InvalidColorNotice.java
```diff
@@ -0,0 +1,50 @@
...
+        ImmutableMap.of(
+            "filename",
+            filename,
+            "csvRowNumber",
+            csvRowNumber,
+            "fieldName",
+            fieldName,
+            "fieldValue",
+            fieldValue),
+        SeverityLevel.ERROR);
```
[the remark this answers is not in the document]
aababilov: This is my plan: severity should be mentioned explicitly and the overloaded constructor should be removed. That makes code more readable and explicit. The reader does not need to know that `ERROR` is the default priority.

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/InvalidEmailNotice.java
```diff
@@ -0,0 +1,52 @@
...
+package org.mobilitydata.gtfsvalidator.notice;
+
+import com.google.common.collect.ImmutableMap;
+
+/**
+ * A field contains a malformed email address.
+ *
+ * <p>Definitions for valid emails are quite vague. We perform strict validation in the upstream
+ * using the Apache Commons EmailValidator but we allow GTFS consumers to patch the severity level
+ * of that validation.
```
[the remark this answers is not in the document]
aababilov: Done.

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/InvalidTimeNotice.java
aababilov: Indeed! We should fix the documentation.

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NonAsciiOrNonPrintableCharNotice.java
aababilov: What description do you want me to put here?

Please note that many notices (such as EmptyFileNotice) have one-line comments.

barbeau: To match the other data type errors, I'd add this from the "ID" definition in GTFS:

>An ID field value is an internal ID, not intended to be shown to riders, and is a sequence of any UTF-8 characters. Using only printable ASCII characters is recommended.

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/InvalidTimeNotice.java
barbeau: @MobilityData/transit-specs To flag this for updating GTFS reference.

barbeau: @aababilov To make the validator internally consistent, I'd recommend updating these Javadocs to what the `GtfsTime` field supports (i.e., add HHH:MM:SS).

core/src/main/java/org/mobilitydata/gtfsvalidator/parsing/RowParser.java
aababilov: We don't actually need them, as far as I can tell. `Locale` fully supports IETF BCP 47. I added some unit tests to verify that.

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/InvalidTimeNotice.java
aababilov: Done for InvalidTimeNotice.

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NonAsciiOrNonPrintableCharNotice.java
aababilov: Good point. Done.

aababilov:
I am glad that you liked that idea!

barbeau (approved):
LGTM, thanks @aababilov!

aababilov:
Thanks for the review!

PR #758, opened 2021-02-18
fix: Use String.trim instead of String.strip

aababilov:
String.strip is available only from Java 11 while String.trim is more
portable.

The difference is tiny and we can neglect it in our context:
* trim() removes all characters that are less than or equal to 'U+0020'
  (the space character)
* strip() uses Character.isWhitespace function, so it removes fewer
  characters

### DIFF
--- a/core/src/main/java/org/mobilitydata/gtfsvalidator/parsing/RowParser.java
+++ b/core/src/main/java/org/mobilitydata/gtfsvalidator/parsing/RowParser.java
@@ -136,12 +136,12 @@ public class RowParser {
             new NewLineInValueNotice(
                 row.getFileName(), row.getRowNumber(), row.getColumnName(columnIndex), s));
       }
-      final String stripped = s.strip();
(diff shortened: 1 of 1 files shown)


aababilov:
`String.strip` is philosophically better but we don't have it at Google (ugh, that old Java).

@barbeau Do you find acceptable if we make our upstream compatible with older Java 8?

barbeau:
>Do you find acceptable if we make our upstream compatible with older Java 8?

That's fine by me to revert back to `String.trim()`, although to guarantee Java 8 compatibility I believe we should then change the `sourceCompatibility` in `build.gradle`s from 11 to 8, and change the JDK that the CI is running on from 11 to 8 as well (see https://github.com/MobilityData/gtfs-validator/issues/593).

barbeau (approved):
👍

aababilov:
https://github.com/MobilityData/gtfs-validator/pull/759 to downgrade the rest of the code.

PR #759, opened 2021-02-18
chore: Downgrade to Java 8

aababilov:
Google is still using the old Java.

The updated code was tested on OpenJDK 1.8.


barbeau (commented):
Thanks @aababilov! Could you also change the *.yml files in `.github/workflows` where JDK 1.11 is referenced to use JDK 1.8 instead?

This will make sure the build and tests pass in the CI environment on Java 8 as well.

aababilov:
> Thanks @aababilov! Could you also change the *.yml files in `.github/workflows` where JDK 1.11 is referenced to use JDK 1.8 instead?

Sure, done!

barbeau (approved):
LGTM, thanks @aababilov!

aababilov:
Thanks for review!

PR #776, opened 2021-02-26
feat: Emit a warning for empty column name

aababilov:
Consider a header row:
stop_id,stop_name,

Its third column does not have a name. We ignore it in the same way as
we ignore unknown columns but we need to emit a clear warning for that.


barbeau (approved):
Thanks @aababilov, good catch.

PR #777, opened 2021-02-26
feat: Use javax.inject.Inject instead of a custom annotation

aababilov:
Also support injection to private fields since that supported by
popular injection frameworks like Guice. This will allow injection to
third-party validators located in custom Java packages.


barbeau (approved):
👍  Thanks @aababilov!

PR #778, opened 2021-02-26
fix: trips.txt is the child table

aababilov:
trips.txt has service_id column that references calendar.txt or
calendar_dates.txt.


barbeau (approved):
Thanks for the fix @aababilov!

PR #779, opened 2021-02-26
fix: route name validator

### LINKED ISSUE #754: Revisit DuplicateRouteNameValidator
## **Feature request**
At present, `routes.route_type` is not taken into consideration when evaluating duplicate route names. Considering this field would improve the logic og this validation rule.

**Proposed solution**
Consider `routes.route_type` in DuplicateRouteNameValidator`.

**Additional context**

> Those rules are too strict and they reject valid data. E.g., in my home city of Harkov, we have trolleybus route 2 and tram route 2 which are operated by the same agency. You would need to make route type a part of your key.

https://github.com/MobilityData/gtfs-validator/pull/689#discussion_r575051421

lionel-nj:
closes #754 

**Summary:**

This PR provides changes to rework rue implemented in `DuplicateRouteNameValidator`.

**Expected behavior:** 

- `route.route_type` should be taken into consideration when checking for duplicate route names
- the validator is case sensitive when it comes to check different agency IDs. `agenCY id` and `AgencY id` are different following https://github.com/MobilityData/gtfs-validator/pull/689#discussion_r577701672

### DIFF
--- a/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateRouteNameValidator.java
+++ b/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateRouteNameValidator.java
@@ -26,8 +26,7 @@ import org.mobilitydata.gtfsvalidator.table.GtfsRoute;
 import org.mobilitydata.gtfsvalidator.table.GtfsRouteTableContainer;
 
 /**
- * Validates unicity of short and long name for all routes. A {@code GtfsRoute}'s `route_short_name`
- * and `route_long_name` should be unique.
+ * Validates unicity of short and long name for all routes.
  *
  * <p>When a {@code GtfsRoute} names are found to be duplicate a {@code DuplicateRouteNameNotice} is
  * generated and added to the {@code NoticeContainer} except if routes are from the same agency
--- a/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateRouteNameValidator.java
+++ b/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateRouteNameValidator.java
@@ -109,7 +109,7 @@ public class DuplicateRouteNameValidator extends FileValidator {
   }
 
   /**
-   * Generate a key used to store {@code GtfsRoute} by `routes.route_long_name`
+   * Generate a key used to store {@code GtfsRoute} by `routes.route_long_name`.
    *
    * @param route the {@code GtfsRoute} to generate the key from
    * @return `routes.route_long_name`+`route.routeType`
@@ -119,7 +119,7 @@ public class DuplicateRouteNameValidator extends FileValidator {
   }
 
   /**
-   * Generate a key used to store {@code GtfsRoute} by `routes.route_short_name`
+   * Generate a key used to store {@code GtfsRoute} by `routes.route_short_name`.
    *
    * @param route the {@code GtfsRoute} to generate the key from
    * @return `routes.route_short_name`+`route.routeType`
@@ -130,7 +130,7 @@ public class DuplicateRouteNameValidator extends FileValidator {
 
   /**
    * Generate a key used to store {@code GtfsRoute} by both `routes.route_short_name` and
-   * `routes.route_long_name`
+   * `routes.route_long_name`.
    *
    * @param route the {@code GtfsRoute} to generate the key from
    * @return `routes.route_short_name`+`routes.route_long_name`+`route.routeType`
--- a/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateRouteNameValidator.java
+++ b/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateRouteNameValidator.java
@@ -49,7 +49,7 @@ public class DuplicateRouteNameValidator extends FileValidator {
             route -> {
               if (route.hasRouteShortName() && route.hasRouteLongName()) {
                 if (routeByShortAndLongName.containsKey(
-                    route.routeShortName() + route.routeLongName())) {
+                    route.routeShortName() + route.routeLongName() + route.routeType())) {
                   noticeContainer.addValidationNotice(
                       new DuplicateRouteNameNotice(
                           "route_short_name and route_long_name",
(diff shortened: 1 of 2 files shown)


main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateRouteNameValidator.java
```diff
@@ -49,7 +49,7 @@ public void validate(NoticeContainer noticeContainer) {
             route -> {
               if (route.hasRouteShortName() && route.hasRouteLongName()) {
                 if (routeByShortAndLongName.containsKey(
-                    route.routeShortName() + route.routeLongName())) {
+                    route.routeShortName() + route.routeLongName() + route.routeType())) {
```
barbeau, line 52: When you have a composite key of this many values, I'd suggest creating a single private method to construct the key - that will enforce that it always has the correct components in the correct order.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateRouteNameValidator.java
```diff
@@ -98,6 +98,6 @@ public void validate(NoticeContainer noticeContainer) {
    */
   private boolean areRoutesFromSameAgency(
       final String routeAgencyId, final String otherRouteAgencyId) {
-    return routeAgencyId.equalsIgnoreCase(otherRouteAgencyId);
```
barbeau, line 101: I would add an explicit note in the Javadocs of this method that IDs are considered case-sensitive to call attention to it. I'm not sure there will be universal agreement on this.

barbeau (commented):
@lionel-nj Thanks, looks pretty good! Comments in-line.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateRouteNameValidator.java
lionel-nj: Updated :)

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateRouteNameValidator.java
```diff
@@ -48,56 +53,89 @@ public void validate(NoticeContainer noticeContainer) {
...
                 }
               }
               if (route.hasRouteShortName()) {
-                if (routeByShortName.containsKey(route.routeShortName())) {
+                if (routeByShortName.containsKey(getRouteShortNameKey(route))) {
                   if (areRoutesFromSameAgency(
-                      route.agencyId(), routeByShortName.get(route.routeShortName()).agencyId())) {
+                      route.agencyId(),
+                      routeByShortName
+                          .get(route.routeShortName() + route.routeType())
```
barbeau, line 86: I think this should use method as well?

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateRouteNameValidator.java
```diff
@@ -48,56 +53,89 @@ public void validate(NoticeContainer noticeContainer) {
...
+                  routeByShortAndLongName.put(getShortAndLongNameKey(route), route);
                 }
               }
               if (route.hasRouteLongName()) {
-                if (routeByLongName.containsKey(route.routeLongName())) {
+                if (routeByLongName.containsKey(getRouteLongNameKey(route))) {
                   if (areRoutesFromSameAgency(
-                      route.agencyId(), routeByLongName.get(route.routeLongName()).agencyId())) {
+                      route.agencyId(),
+                      routeByLongName.get(route.routeLongName() + route.routeType()).agencyId())) {
```
barbeau, line 71: I think this should use method as well?

barbeau (commented):
Thanks @lionel-nj! I think there are two places where the methods could still be used to generate the keys? If so, after fixing this please feel free to merge.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateRouteNameValidator.java
lionel-nj: Indeed.

lionel-nj:
Thanks @lionel-nj! I think there are two places where the methods could still be used to generate the keys? If so, after fixing this please feel free to merge.

Double checked the code implementation and the tests, it should be good to go.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateRouteNameValidator.java
```diff
@@ -48,56 +53,89 @@ public void validate(NoticeContainer noticeContainer) {
...
+
+  /**
+   * Generate a key used to store {@code GtfsRoute} by both `routes.route_short_name` and
+   * `routes.route_long_name`.
+   *
+   * @param route the {@code GtfsRoute} to generate the key from
+   * @return `routes.route_short_name`+`routes.route_long_name`+`route.routeType`
+   */
+  private String getShortAndLongNameKey(GtfsRoute route) {
+    return route.routeShortName() + route.routeLongName() + route.routeType();
```
aababilov, line 139: Concatenation of strings is a non-efficient operation that leads to memory fragmentation. A hashcode would suffice here.

aababilov (commented):
I think that the upstread code should avoid inefficient practices such as memory fragmentation. Thoughts?

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateRouteNameValidator.java
barbeau: Agreed, these could be replaced with `Objects.hash(...)`

lionel-nj: Thanks for the review. The changes will be linked to https://github.com/MobilityData/gtfs-validator/issues/805.

lionel-nj: Thanks for flagging that, the changes have been provided in #805 and #812

PR #783, opened 2021-03-02
refactor: notice classes names and codes

### LINKED ISSUE #784: Review notices class names and code
## **Feature request**

> We should review all rule class names and codes to make sure they follow the same naming pattern (this could be done in another PR). They should be the same, but there are a few that are different like this example - SameNameAndDescriptionForRouteNotice and same_route_name_and_description.

https://github.com/MobilityData/gtfs-validator/pull/781/files/67e5c7ad22e64c8bf00041baa2f28f21ba1c245b#discussion_r585866237

`UnexpectedEnumError` should be refactored to `UnexpectedEnumNotice`

> This notice should be refactored to UnexpectedEnumValueNotice because it's not an error, it's a warning.

https://github.com/MobilityData/gtfs-validator/pull/781#discussion_r585900910

lionel-nj:
closes #784 

**Summary:**

This PR refactors notice class names and codes so that they are all the same. It implements getCode() in the base Notice class.

**Expected behavior:** 

Notice code should be derived from the notice class name.

### DIFF
--- a/core/src/main/java/org/mobilitydata/gtfsvalidator/notice/InvalidUriSyntax.java
+++ b/core/src/main/java/org/mobilitydata/gtfsvalidator/notice/InvalidUriSyntax.java
@@ -23,14 +23,14 @@ import com.google.common.collect.ImmutableMap;
  *
  * <p>Severity: {@code SeverityLevel.ERROR}
  */
-public class URISyntaxError extends SystemError {
+public class InvalidUriSyntax extends SystemError {
 
-  public URISyntaxError(String message) {
+  public InvalidUriSyntax(String message) {
     super(ImmutableMap.of("message", message));
   }
 
   @Override
   public String getCode() {
-    return "uri_syntax_error";
+    return "invalid_uri_syntax";
   }
 }
--- a/core/src/main/java/org/mobilitydata/gtfsvalidator/notice/MissingFeedInfoDateNotice.java
+++ b/core/src/main/java/org/mobilitydata/gtfsvalidator/notice/MissingFeedInfoDateNotice.java
@@ -33,6 +33,6 @@ public class MissingFeedInfoDateNotice extends ValidationNotice {
 
   @Override
   public String getCode() {
-    return "missing_feed_info_start_date_or_end_date";
+    return "missing_feed_info_date";
   }
 }
--- a/core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NonAsciiOrNonPrintableCharNotice.java
+++ b/core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NonAsciiOrNonPrintableCharNotice.java
@@ -40,6 +40,6 @@ public class NonAsciiOrNonPrintableCharNotice extends ValidationNotice {
 
   @Override
   public String getCode() {
-    return "id_contains_non_ascii_characters";
+    return "non_ascii_or_non_printable_char";
   }
 }
--- a/core/src/main/java/org/mobilitydata/gtfsvalidator/notice/Notice.java
+++ b/core/src/main/java/org/mobilitydata/gtfsvalidator/notice/Notice.java
@@ -16,6 +16,7 @@
 
 package org.mobilitydata.gtfsvalidator.notice;
 
+import com.google.common.base.CaseFormat;
 import com.google.common.base.Joiner;
 import java.util.Collections;
 import java.util.Map;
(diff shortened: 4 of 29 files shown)


aababilov:
@lionel-nj I would be glad to discuss your ideas before you start writing code. Just drop me an email next time.

aababilov:
I would suggest to implement getCode() in the base `Notice` class and just derive it from the class name.

barbeau:
>I would suggest to implement getCode() in the base Notice class and just derive it from the class name.

👍  I was thinking along these lines too - plus, if the code is always derived from the class name, we don't even need to list the code in the table in RULES.md for https://github.com/MobilityData/gtfs-validator/pull/781

lionel-nj:
> > I would suggest to implement getCode() in the base Notice class and just derive it from the class name.
> 
> 
> 👍 I was thinking along these lines too - plus, if the code is always derived from the class name, we don't even need to list the code in the table in RULES.md for #781

Good point! Will update both PRs.

lionel-nj:
> > > I would suggest to implement getCode() in the base Notice class and just derive it from the class name.
> > 
> > 👍 I was thinking along these lines too - plus, if the code is always derived from the class name, we don't even need to list the code in the table in RULES.md for #781
> 
> Good point! Will update both PRs to meet these thoughts.

@aababilov @barbeau before going further, I'd love to have your feedback on https://github.com/MobilityData/gtfs-validator/pull/783/commits/6eb3b5c5308ce9aa21e4e32995d0d2e2e708f346 please :) 
Here is a screenshot of a test execution in debug mode: the notice's code has effectively been changed. It now the name_of_the_class_with_underscores. In order to avoid introducing a new dependency a re-used the code defined in FieldNameConverter (from the `processor` module).

lionel-nj:
@barbeau also https://github.com/MobilityData/gtfs-validator/pull/783/commits/7dd23a3469b9e54baf5ee286702c9aaeb97febbd and https://github.com/MobilityData/gtfs-validator/pull/783/commits/3df44f232d69e0f3b46ee39eed8aa9ca759d2d77 refactor some notice class names to account for https://github.com/MobilityData/gtfs-validator/pull/781#discussion_r585866237

barbeau:
>In order to avoid introducing a new dependency in the core module on processor) I re-used the code defined in FieldNameConverter (from the processor module). I am not sure this is best practice or the most effective way to get around this, I am all hears for any suggestions 😃

Since processor depends on core already we could move any relevant utility methods into core and then call them in processor. I could go either way in this case because it's a single line of code, but as a best practice maybe it would be good to start this now.

barbeau:
>Should we remove the terms "notice" or "error" (as done for most validation rule related notices)?

You mean for the notice code, right (not class name)? For the notice code, I'd suggest we keep the same pattern that we currently have - it's less verbose when looking at the error output.

lionel-nj:
> You mean for the notice code, right (not class name)? For the notice code, I'd suggest we keep the same pattern that we currently have - it's less verbose when looking at the error output.

The notice code. Alright!

aababilov:
I suggest to make changes step by step in separate PRs to facilitate review.

Step 1. Implement `Notice.getCode()` method. It should strip the "notice" substring but not strip "error" since we will carefully remove that suffix in the further steps.
Step 2. Change `Notice.equals()` and `Notice.hashCode()` to use the class instead of `getCode()`.
Step 3. Remove implementations of `getCode()` in child classes. Some classes will be renamed in accordance to their "code".
Here we can establish a convention: all children of `SystemError` end with `Error`, all children of `ValidationNotice` end with `Notice`.

lionel-nj:
> Step 1. Implement Notice.getCode() method. It should strip the "notice" substring but not strip "error" since we will carefully remove that suffix in the further steps.

Implemented in https://github.com/MobilityData/gtfs-validator/pull/788

> Step 2. Change Notice.equals() and Notice.hashCode() to use the class instead of getCode().


Implemented in https://github.com/MobilityData/gtfs-validator/pull/789

> Step 3. Remove implementations of getCode() in child classes. 

https://github.com/MobilityData/gtfs-validator/pull/790

> Some classes will be renamed in accordance to their "code".

This will be done in a separate PR in order to avoid confusion

Thanks for your feedback @barbeau @aababilov!

lionel-nj:
> Since processor depends on core already we could move any relevant utility methods into core and then call them in processor. I could go either way in this case because it's a single line of code, but as a best practice maybe it would be good to start this now.

Opening an issue for discussions.

lionel-nj:
Closing this PR since it has been split in three others: #788, #789, #790.

https://github.com/MobilityData/gtfs-validator/pull/783#issuecomment-790994961

PR #797, opened 2021-03-10
feat: warning if one the route names is missing

lionel-nj:
**Summary:**

This PR provides support to emit a `WARNING`  if one of the names fields for a row from `routes.txt` is missing or empty

**New notice**
- RouteNameMissingOrEmptyNotice

**Expected behavior:** 

a notice should be generated if: 
- `routes.route_short_name` is provided and `routes.route_long_name` is missing or empty
- `routes.route_long_name` is provided and `routes.route_short_name` is missing or empty


main/src/test/java/org/mobilitydata/gtfsvalidator/validator/RouteNameValidatorTest.java
```diff
@@ -122,4 +123,20 @@ public void equalRouteShortNameRouteLongNameAndRouteDescShouldGenerateTwoNotices
         .contains(
             new SameNameAndDescriptionForRouteNotice(1, "r1", "duplicate", "route_short_name"));
   }
+
+  @Test
+  public void nullOrEmptyRouteLongNameShouldGenerateNotice() {
+    assertThat(validateRoute(createRoute("short name", null, "desc")))
```
maximearmstrong, line 129: Isn't it the same test as in line 103?

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/RouteNameValidatorTest.java
```diff
@@ -122,4 +123,20 @@ public void equalRouteShortNameRouteLongNameAndRouteDescShouldGenerateTwoNotices
...
+        .containsExactly(new RouteNameMissingOrEmptyNotice(1, "route_long_name"));
+    assertThat(validateRoute(createRoute("short name", "", "desc")))
+        .containsExactly(new RouteNameMissingOrEmptyNotice(1, "route_long_name"));
+  }
+
+  @Test
+  public void nullOrEmptyShortNameShouldGenerateNotice() {
+    assertThat(validateRoute(createRoute(null, "long name", "desc")))
```
maximearmstrong, line 137: Same as above but for line 109

maximearmstrong (changes requested):

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/RouteNameValidatorTest.java
lionel-nj: Indeed, collapsed these tests in https://github.com/MobilityData/gtfs-validator/pull/797/commits/acda7f672e3e0bb58167d297b5f6e3ee7e9c8367.

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/RouteNameValidatorTest.java
lionel-nj: Indeed, collapsed these tests in https://github.com/MobilityData/gtfs-validator/pull/797/commits/acda7f672e3e0bb58167d297b5f6e3ee7e9c8367.

RULES.md
```diff
@@ -429,3 +430,14 @@ At least one of the fields `is_producer`, `is_operator`, or `is_authority` shoul
...
+<a name="RouteNameMissingOrEmptyNotice"/>
+
+### RouteNameMissingOrEmptyNotice
+
+One of the following fields is missing or empty: 
+- `routes.route_short_name` 
+- `routes.route_long_name`
+
+#### References:
+* [Python validator original implementation](https://github.com/google/transitfeed/blob/master/transitfeed/route.py)
```
aababilov, line 443: I am unsure if we need this check. Here is some statistics from all prod feeds at Google. 

- 45% give both names
- 35% give ONLY long name
- 20% give ONLY short name

Less than a half routes give both names and I think that makes sense. Real-world transit routes may indeed have just one name. If we issue warnings, that may push some GTFS providers to invent short or long names that never existed in the real world just to satisfy the validator.

Ceterum censeo Python validator should be taken with a grain of salt. It was never used on a wide range of feeds and before porting a validation rules from Python, we should have evidence that it's a meaningful rule

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/RouteNameValidator.java
```diff
@@ -85,4 +90,13 @@ private boolean isValidRouteDesc(String routeDesc, String routeShortOrLongName)
...
     return !routeDesc.equalsIgnoreCase(routeShortOrLongName);
   }
+
+  /**
+   * @param routeName a name of a Route
+   * @return true if this {@code GtfsRoute} name (hsort or long) is not missing or empty, false
+   *     otherwise.
+   */
+  private boolean isPresentName(final String routeName) {
+    return routeName != null && !routeName.isEmpty();
```
aababilov, line 100: Route name may never be `null`. This was a deliberate decision in Validator Core to prevent NullPointerException.

I will add the `@Nonnull` annotations to avoid such misunderstandings in future.

lionel-nj:
Closing for now. We will work in conjuction with @MobilityData/transit-specs to clarify the language and get feedback from other data consumers before re-opening the issue related to this PR.

timMillet:
I think we can remove the warning in case the route_short_name *and* the route_long_name are not *both* provided together. Spec and Best Practices say:
> potentially both if appropriate

If it’s not appropriate to provide both, we don't want to push producers to invent names, as Alexei said.

PR #798, opened 2021-03-10
feat: check timepoints time presence

lionel-nj:
**Summary:**

This PR provides changes to check: timepoint define time values in `stop_times.txt`.

**New notice**
- `StopTimeTimepointWithoutTimesNotice`

**Expected behavior:** 

A `WARNING`  should be issued if a row from `stop_times.txt`is defined as a timepoint (i.e. `timepoint == 1`) and one of the time fields (`stop_times.arrival_time` or `stop_time.departure_time` is missing)


main/src/test/java/org/mobilitydata/gtfsvalidator/validator/StopTimeArrivalAndDepartureTimeValidatorTest.java
```diff
@@ -217,12 +228,96 @@ public void missingDepartureTimeShouldGenerateNoticeIfArrivalTimeIsProvided() {
...
+                    2,
+                    0)));
 
     underTest.validate(noticeContainer);
     assertThat(noticeContainer.getValidationNotices())
         .containsExactly(
             new StopTimeWithOnlyArrivalOrDepartureTimeNotice(
                 1, "first trip id", 2, "arrival_time"));
   }
+
```
barbeau, line 240: I believe this is missing the false-positive check - timepoint with both times should not generate notice?

RULES.md
```diff
@@ -429,3 +430,9 @@ At least one of the fields `is_producer`, `is_operator`, or `is_authority` shoul
 
 #### References:
 * [attributions.txt specification](https://gtfs.org/reference/static#attributionstxt)
+
+<a name="StopTimeTimepointWithoutTimeNotice"/>
+
+### StopTimeTimepointWithoutTimeNotice
+
+Timepoints from GTFS file `stop_times.txt` should defined values for both `stop_times.arrival_time` and `stop_times.departure_time` fields.
```
barbeau, line 438: The description should include "If `timepoint` is set to 1..."

There is another case where the timepoint column is omitted, and arrival and departure times are omitted for non-timepoints and all other records with times are assumed to be timepoints - this was the original definition for timepoint before the `timepoint` field was introduced. And we're not addressing this case here.

barbeau (commented):
Thanks @lionel-nj! Comments in-line.

RULES.md
lionel-nj: > There is another case where the timepoint column is omitted, and arrival and departure times are omitted for non-timepoints and all other records with times are assumed to be timepoints - this was the original definition for timepoint before the timepoint field was introduced. And we're not addressing this case here.

Opening an issue to keep track of this.

lionel-nj:
@MobilityData/transit-specs, here the rule is implemented as a `WARNING` however we believe that this should be an `ERROR` based on the past GTFS spec definition before timepoint was added. When timepoint was added the language wasn't as descriptive as it should have been to make it clear that this is an error. Here's the related GitHub discussion for a proposal to clarify this - https://github.com/google/transit/issues/61.

@barbeau

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTimeArrivalAndDepartureTimeValidator.java
```diff
@@ -93,4 +104,8 @@ public void validate(NoticeContainer noticeContainer) {
       }
     }
   }
+
+  private boolean isTimepoint(GtfsStopTime stopTime) {
+    return stopTime.timepoint() == GtfsStopTimeTimepoint.EXACT;
```
aababilov, line 109: Please use `.equals()` method for enums - this is a requirement of Google Java style.

> Both choices above produce the correct result. However, we prefer code that is not just correct, but self-evidently correct. It's best if the reader can know the code is safe from bugs without having to look up whether the type of e1 and e2 is an enum. This makes equals the better choice

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTimeArrivalAndDepartureTimeValidator.java
```diff
@@ -53,6 +57,13 @@ public void validate(NoticeContainer noticeContainer) {
         GtfsStopTime stopTime = stopTimeList.get(i);
         final boolean hasDeparture = stopTime.hasDepartureTime();
         final boolean hasArrival = stopTime.hasArrivalTime();
+        if (isTimepoint(stopTime)) {
```
aababilov, line 60: This validation rule does not need the whole stop_times.txt table. Please make it a SingleEntityValidator instead.

aababilov (changes requested):

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTimeArrivalAndDepartureTimeValidator.java
lionel-nj: Thanks! Modified in https://github.com/MobilityData/gtfs-validator/pull/798/commits/f44acbb99a2cdcba38daf116e41034ad127b6a7a.

barbeau: I was wondering about this. Logic for `StopTimeWithDepartureBeforeArrivalTimeNotice` could potentially be moved to the same SingleEntityValidator as well, right?

Looks like there are some other FileValidators that could be moved to SingleEntityValidators, including:
* CalendarServiceDateValidator
* FeedServiceDateValidator

lionel-nj: https://github.com/MobilityData/gtfs-validator/issues/804 has been opened to keep track of this.

maximearmstrong (approved):
LGTM

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TimepointTimeValidator.java
```diff
@@ -0,0 +1,68 @@
...
+ *   <li>{@link StopTimeTimepointWithoutTimesNotice} - a timepoint does not specifies arrival_time
+ *       or departure_time
+ * </ul>
+ */
+@GtfsValidator
+public class TimepointTimeValidator extends SingleEntityValidator<GtfsStopTime> {
+
+  @Override
+  public void validate(GtfsStopTime stopTime, NoticeContainer noticeContainer) {
+    if (isTimepoint(stopTime)) {
```
aababilov, line 41: Prefer an early return here to avoid deeply nested code.

if (!...) {
  return;
}

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TimepointTimeValidator.java
```diff
@@ -0,0 +1,68 @@
...
+        noticeContainer.addValidationNotice(
+            new StopTimeTimepointWithoutTimesNotice(
+                stopTime.csvRowNumber(),
+                stopTime.tripId(),
+                stopTime.stopSequence(),
+                stopTime.hasArrivalTime()
+                    ? GtfsStopTimeTableLoader.DEPARTURE_TIME_FIELD_NAME
+                    : GtfsStopTimeTableLoader.ARRIVAL_TIME_FIELD_NAME));
+      }
+    }
```
aababilov, line 62: Let's make it simpler. Check each field individually and emit one notice per field.

```
if (!stopTime.hasArrivalTime()) {
  noticeContainer.addValidationNotice(...);
}
if (!stopTime.hasDepartureTime()) {
  noticeContainer.addValidationNotice(...);
}
```

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/TimepointTimeValidatorTest.java
```diff
@@ -0,0 +1,115 @@
...
+        .containsExactly(
+            new StopTimeTimepointWithoutTimesNotice(
+                1,
+                "first trip id",
+                2,
+                String.format(
+                    "%s and %s",
+                    GtfsStopTimeTableLoader.ARRIVAL_TIME_FIELD_NAME,
+                    GtfsStopTimeTableLoader.DEPARTURE_TIME_FIELD_NAME)));
+  }
```
aababilov, line 66: Here and below: keep the tests short. Sample:

```
private static List<ValidationNotice> generateNotices(GtfsAttribution attribution) {
    NoticeContainer noticeContainer = new NoticeContainer();
    AttributionWithoutRoleValidator validator = new AttributionWithoutRoleValidator();
    validator.validate(attribution, noticeContainer);
    return noticeContainer.getValidationNotices();
  }

  @Test
  public void attributionWithoutRoleShouldGenerateNotice() {
    assertThat(
            generateNotices(
                new GtfsAttribution.Builder()
                    .setCsvRowNumber(2)
                    .setAttributionId("attr-1")
                    .build()))
        .containsExactly(new AttributionWithoutRoleNotice(2, "attr-1"));
  }
```

aababilov (changes requested):

lionel-nj:
Thank you all for your review!
@aababilov 
@barbeau 
@maximearmstrong

main/src/main/java/org/mobilitydata/gtfsvalidator/notice/StopTimeTimepointWithoutTimesNotice.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+package org.mobilitydata.gtfsvalidator.notice;
+
+import com.google.common.collect.ImmutableMap;
+
+/**
+ * Timepoint without time
+ *
+ * <p>Severity: {@code SeverityLevel.ERROR}
```
barbeau, line 24:
```suggestion
 * <p>Severity: {@code SeverityLevel.WARNING}
```

barbeau (commented):
Thanks @lionel-nj! One documentation issue, then feel free to merge when that's fixed.

main/src/main/java/org/mobilitydata/gtfsvalidator/notice/StopTimeTimepointWithoutTimesNotice.java
lionel-nj: Thanks @barbeau!

PR #800, opened 2021-03-15
fix: Inherit AttributionWithoutRoleValidator from SingleEntityValidator

aababilov:
AttributionWithoutRoleValidator validates a single attribution. It does
not need the whole file to be scanned.

* SingleEntityValidators are invoked just after a row is parsed.
* FileValidators need the whole file to be scanned. They are skipped if
  any row failed to parse or primary keys are non-unique.

### DIFF
--- a/main/src/main/java/org/mobilitydata/gtfsvalidator/notice/AttributionWithoutRoleNotice.java
+++ b/main/src/main/java/org/mobilitydata/gtfsvalidator/notice/AttributionWithoutRoleNotice.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2021 MobilityData IO
+ * Copyright 2021 Google LLC, MobilityData IO
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


main/src/main/java/org/mobilitydata/gtfsvalidator/notice/AttributionWithoutRoleNotice.java
```diff
@@ -25,9 +25,10 @@
  * <p>Severity: {@code SeverityLevel.WARNING}
  */
 public class AttributionWithoutRoleNotice extends ValidationNotice {
-  public AttributionWithoutRoleNotice(long csvRowNumber) {
+
+  public AttributionWithoutRoleNotice(long csvRowNumber, String attributionId) {
     super(
-        new ImmutableMap.Builder<String, Object>().put("csvRowNumber", csvRowNumber).build(),
+        ImmutableMap.of("csvRowNumber", csvRowNumber, "attributionId", attributionId),
```
lionel-nj, line 31: Thanks for flagging this @aababilov

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/AttributionWithoutRoleValidatorTest.java
```diff
@@ -18,66 +18,73 @@
...
-    attributionWithoutRouteValidator.validate(noticeContainer);
-    assertThat(noticeContainer.getValidationNotices())
-        .containsExactlyElementsIn(
-            new AttributionWithoutRoleNotice[] {
-              new AttributionWithoutRoleNotice(3),
-              new AttributionWithoutRoleNotice(8),
-              new AttributionWithoutRoleNotice(13)
-            });
+  @Test
+  public void attributionWithoutRoleAndNoIdShouldGenerateNotice() {
```
lionel-nj, line 55: @aababilov @barbeau  why shouldn't we issue a notice in this case?

lionel-nj (commented):
Thanks for the changes @aababilov, one question in the comments.

lionel-nj (approved):
LGTM

PR #801, opened 2021-03-15
fix: Initialize BigDecimal to ZERO

aababilov:
Avoid null values that can lead to NullPointerException.


barbeau (approved):
👍  Thanks @aababilov! 

Not directly related - I also noticed the above TODO for when the value is present:

>// TODO: Support all types or throw an exception.

Should we open another issue for this? What still needs to be done there? It wasn't obvious for me looking at that `switch` that `DECIMAL` was being handled.

PR #802, opened 2021-03-15
feat: Add Nullable or Nonnull annotation to entity classes

aababilov:
Most of fields in entity classes never have null values: they are
initialized to "", 0 etc. This is made to avoid NullPointerException.
Users should call hasFIELD() method to check if a field is assigned.

Sample of generated code:

```
  @Override
  @Nonnull
  public ZoneId agencyTimezone() {
    return agencyTimezone;
  }

  public boolean hasAgencyTimezone() {
    return (bitField0_ & 0x8) != 0;
  }

  @Override
  @Nullable
  public Locale agencyLang() {
    return agencyLang;
  }

  public boolean hasAgencyLang() {
    return (bitField0_ & 0x10) != 0;
  }
```


barbeau (approved):
Thanks @aababilov!

PR #806, opened 2021-03-15
chore: improve efficiency of DuplicateRouteNameValidator

### LINKED ISSUE #805: Improve DuplicateRouteNameValidator
## **Feature request**

At present, `DuplicateRouteNameValidator` performs a non -efficient operation that leads to memory fragmentation. 

**Proposed solution**
Replace String concatenation by the use of an hashcode. 

**Additional context**
https://github.com/MobilityData/gtfs-validator/pull/779#discussion_r593996579

lionel-nj:
closes #805 

**Summary:**

This PR provides changes to improve `DuplicateRouteNameValidator`

**Expected behavior:** 

Routes are stored by the hashcode associated to:
- route_short_name, route_type instead of the concatenation of both fields
- route_long_name, route_type instead of the concatenation of both fields
- route_long_name, route_short_name, route_type instead of the concatenation of these 3 fields

Same tests should pass as there is no other logic change. 

### DIFF
--- a/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateRouteNameValidator.java
+++ b/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateRouteNameValidator.java
@@ -52,24 +52,21 @@ public class DuplicateRouteNameValidator extends FileValidator {
         .getEntities()
         .forEach(
             route -> {
-              int hash;
+              int hash = getLongNameAndTypeHash(route);
               GtfsRoute otherRoute;
               if (route.hasRouteLongName()) {
-                hash = getLongNameAndTypeHash(route);
-                otherRoute = routeByLongName.get(hash);
+                otherRoute = routeByLongName.putIfAbsent(hash, route);
                 if (otherRoute != null) {
                   if (areRoutesFromSameAgency(route, otherRoute)) {
                     noticeContainer.addValidationNotice(
                         new DuplicateRouteNameNotice(
                             "route_long_name", route.csvRowNumber(), route.routeId()));
                   }
-                } else {
-                  routeByLongName.put(hash, route);
                 }
               }
               if (route.hasRouteShortName()) {
                 hash = getShortNameAndTypeHash(route);
-                otherRoute = routeByShortName.get(hash);
+                otherRoute = routeByShortName.putIfAbsent(hash, route);
                 if (otherRoute != null) {
                   if (areRoutesFromSameAgency(route, otherRoute)) {
                     noticeContainer.addValidationNotice(
--- a/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateRouteNameValidator.java
+++ b/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateRouteNameValidator.java
@@ -31,8 +31,8 @@ import org.mobilitydata.gtfsvalidator.table.GtfsRouteTableContainer;
  *
  * <p>When a {@code GtfsRoute} short and/or long names are found to be duplicate a {@code
  * DuplicateRouteNameNotice} is generated and added to the {@code NoticeContainer} except if routes
- * are from the same agency (values for `route.agency_id` are case-sensitive) or routes have
- * different `routes.route_type`.
+ * are from the same agency (values for "route.agency_id" are case-sensitive) or routes have
+ * different "routes.route_type".
  *
  * <p>Generated notice:
  *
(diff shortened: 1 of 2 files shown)


maximearmstrong (approved):
LGTM

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateRouteNameValidator.java
```diff
@@ -45,50 +46,50 @@
...
+    final Map<Integer, GtfsRoute> routeByLongName = new HashMap<>(routeTable.entityCount());
+    final Map<Integer, GtfsRoute> routeByShortName = new HashMap<>(routeTable.entityCount());
+    final Map<Integer, GtfsRoute> routeByShortAndLongName = new HashMap<>(routeTable.entityCount());
     routeTable
         .getEntities()
         .forEach(
             route -> {
               if (route.hasRouteShortName() && route.hasRouteLongName()) {
-                if (routeByShortAndLongName.containsKey(getShortAndLongNameKey(route))) {
+                if (routeByShortAndLongName.containsKey(getShortAndLongNameRouteTypeHash(route))) {
```
barbeau, line 57: This doesn't check for `areRoutesFromSameAgency()` like the others - should it? It's currently being compared to all other routes, so it seems like it should be. If it needs to change, the tests should be updated with the case that should have failed here.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateRouteNameValidator.java
```diff
@@ -107,33 +108,39 @@ private boolean areRoutesFromSameAgency(
...
+   * store and retrieve routes by short name.
    *
-   * @param route the {@code GtfsRoute} to generate the key from
-   * @return `routes.route_long_name`+`route.routeType`
+   * @param route the {@code GtfsRoute} to generate the hash from
+   * @return the hash associated to `routes.route_long_name` and `routes.route_type`.
    */
-  private String getRouteLongNameKey(GtfsRoute route) {
-    return route.routeLongName() + route.routeType();
+  private int getRouteLongNameRouteTypeHash(GtfsRoute route) {
```
barbeau, line 118: These method names could be shortened to remove `Route` to help with readability - for example, `getLongNameAndTypeHash()`

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateRouteNameValidator.java
```diff
@@ -107,33 +108,39 @@ private boolean areRoutesFromSameAgency(
   }
```
barbeau, line 108: To make the code in the main validate() method more readable, I'd suggest passing the two `Route` objects into this `areRoutesFromSameAgency()` method and pulling out the agency IDs here.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateRouteNameValidator.java
```diff
@@ -45,50 +46,50 @@
...
                           route.routeId()));
                   return;
                 } else {
-                  routeByShortAndLongName.put(getShortAndLongNameKey(route), route);
+                  routeByShortAndLongName.put(getShortAndLongNameRouteTypeHash(route), route);
                 }
               }
               if (route.hasRouteLongName()) {
-                if (routeByLongName.containsKey(getRouteLongNameKey(route))) {
+                if (routeByLongName.containsKey(getRouteLongNameRouteTypeHash(route))) {
```
barbeau, line 69: I think this would be easier to read, and also avoids looking up the hash twice and computing the key twice, if you assign to local variables here, like:

```java
                int key = getRouteLongNameKey(route);
                GtfsRoute otherRoute = routeByLongName.get(key);
                if (otherRoute != null) {
                  if (areRoutesFromSameAgency(route.agencyId(), otherRoute.agencyId())) {
                        new DuplicateRouteNameNotice(
                            "route_long_name", route.csvRowNumber(), route.routeId()));
                  }
                  return;
                } else {
                  routeByLongName.put(key, route);
                }
```

Same for route short name and route short and long name IF statements.

barbeau (commented):

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateRouteNameValidator.java
lionel-nj: I actually wonder if it would be simpler if we just issued one notice per field as similarly suggested by @aababilov in https://github.com/MobilityData/gtfs-validator/pull/798#discussion_r594776483: if a route has duplicate for both names it has duplicate for route_short_name and has duplicate for route_long_name.

barbeau: ☝️  Sounds good to me

lionel-nj: Done in https://github.com/MobilityData/gtfs-validator/pull/806/commits/c8b90740ba2bd3ba5ceb290863447a38e01fcc1b

barbeau (approved):
Thanks @lionel-nj! 👍

PR #810, opened 2021-03-15
refactor: make FeedServiceDateValidator extend SingleEntityValidator

lionel-nj:
**Summary:**

This PR provides changes to make `FeedServiceDateValidator` extend `SingleFileValidator`  class instead of `FileValidator`. 

**Expected behavior:** 

Unit tests have been adapted to this refactor, but no logic has changed. Therefore the same logic for existing tests should pass. 


main/src/main/java/org/mobilitydata/gtfsvalidator/validator/FeedServiceDateValidator.java
```diff
@@ -16,43 +16,43 @@
...
 import org.mobilitydata.gtfsvalidator.notice.StartAndEndDateOutOfOrderNotice;
 import org.mobilitydata.gtfsvalidator.table.GtfsFeedInfo;
-import org.mobilitydata.gtfsvalidator.table.GtfsFeedInfoTableContainer;
+import org.mobilitydata.gtfsvalidator.table.GtfsFeedInfoTableLoader;
 
 /**
- * Validates 3 rules: 1) start_date &lt;= end_date for all rows in "feed_info.txt" 2)
- * feed_info.start_date is provided if feed_info.end_date is provided 3) feed_info.end_date is
- * provided if feed_info.start_date is provided.
+ * Validates 3 rules: 1) start_date &lt;= end_date 2) feed_info.start_date is provided if
```
aababilov, line 27: I suggest you to use an HTML list here - that will be more readable.

```
Validates {@code start_date} and {@code end_date} fields in {@code feed_info.txt}.

<p>Generated notices:
<ul>
<li>{@link StartAndEndDateOutOfOrderNotice} - ...

```

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/FeedServiceDateValidatorTest.java
```diff
@@ -44,44 +36,34 @@ public static GtfsFeedInfo createFeedInfo(
...
+    underTest.validate(
+        createFeedInfo(
+            1,
+            "name value",
+            "url value",
+            Locale.CANADA,
+            GtfsDate.fromEpochDay(340),
+            GtfsDate.fromEpochDay(450)),
+        noticeContainer);
     assertThat(noticeContainer.getValidationNotices().isEmpty());
```
aababilov, line 50: You can make the tests more concise. Please take a look at my example.

```
private static List<ValidationNotice> generateNotices(GtfsAttribution attribution) {
    NoticeContainer noticeContainer = new NoticeContainer();
    AttributionWithoutRoleValidator validator = new AttributionWithoutRoleValidator();
    validator.validate(attribution, noticeContainer);
    return noticeContainer.getValidationNotices();
  }

  @Test
  public void attributionWithoutRoleShouldGenerateNotice() {
    assertThat(
            generateNotices(
                new GtfsAttribution.Builder()
                    .setCsvRowNumber(2)
                    .setAttributionId("attr-1")
                    .build()))
        .containsExactly(new AttributionWithoutRoleNotice(2, "attr-1"));
  }
```

aababilov (changes requested):
Thank you for doing that change!

aababilov:
Here is an annotation-based solution: https://github.com/MobilityData/gtfs-validator/pull/813

barbeau:
@lionel-nj This PR is still relevant after https://github.com/MobilityData/gtfs-validator/pull/813, right? Because it tests for conditional presence of start/end date fields?

lionel-nj:
@barbeau yes!

barbeau (approved):
Thanks @lionel-nj!

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/FeedServiceDateValidator.java
```diff
@@ -30,25 +28,16 @@
...
-        noticeContainer.addValidationNotice(
-            new MissingFeedInfoDateNotice(feedInfo.csvRowNumber(), "feed_end_date"));
-      } else if (!feedInfo.hasFeedStartDate() && feedInfo.hasFeedEndDate()) {
-        noticeContainer.addValidationNotice(
-            new MissingFeedInfoDateNotice(feedInfo.csvRowNumber(), "feed_start_date"));
-      }
+  public void validate(GtfsFeedInfo feedInfo, NoticeContainer noticeContainer) {
+    if (feedInfo.hasFeedStartDate() && !feedInfo.hasFeedEndDate()) {
+      noticeContainer.addValidationNotice(
+          new MissingFeedInfoDateNotice(feedInfo.csvRowNumber(), "feed_end_date"));
```
aababilov, line 37: And now I started thinking of an annotation for conditionally required fields that depend on presence of other fields...

lionel-nj: That is a good idea! It would allow to have more straightforward code and less validation code to be written by users! @barbeau what do ou think about that?

barbeau: Yeah, I was thinking the same thing. We already have the `@ConditionallyRequired` annotation so I suppose the implementation could just extend that? Although looking at the schema now `feedStartDate` and `feedEndDate` aren't annotated with this - probably because it's a WARNING and not an ERROR (i.e., the spec doesn't require it). I suppose we'd need to allow the annotation to specify if the corresponding notice would be an ERROR or WARNING?

Here are some other fields that are already annotated with `@ConditionallyRequired`:
* agency.agency_id
* stop_times.arrival_time
* stop_times.departure_time
* calendar_dates.txt (entire schema)
* calendar.txt (entire schema)
* fare_attributes.agency_id
* routes.agency_id
* routes.short_name
* routes.long_name
* stops.name
* stops.lat
* stops.lon
* stops.zone_id
* stops.parent_station
* ...

Some of the above are conditionally required depending on the *value* not the *presence* of other fields - we would just limit this annotation to generating rules for the *presence* of fields, right?

PR #713, opened 2021-02-02
feat: no duplicate for `fare_rules.txt` zone id fields

lionel-nj:
closes #356

**Summary:**

This PR implements a new validation rule: the combination of `fare_rules.origin_id`, `fare_rules.contains_id` and `fare_rules.destination_id` fields should be unique in GTFS file `fare_rules.txt`.

**Expected behavior:** 

- a notice should be generated if two (or more rows) from GTFS files `fare_rules.txt` have duplicate value for the combination of  `fare_rules.origin_id`, `fare_rules.contains_id` and `fare_rules.destination_id` fields

Please make sure these boxes are checked before submitting your pull request - thanks!

- [x] Run the unit tests with `gradle test` to make sure you didn't break anything
- [x] Format the title like "feat: -new feature short description-" (PR title must follow the [conventional commit specification](https://www.conventionalcommits.org/en/v1.0.0/))
- [x] Linked all relevant issues
- ~[ ] Include screenshot(s) showing how this pull request works and fixes the issue(s)~

PR #811, opened 2021-03-15
chore: improve efficiency of DuplicateFareRuleZoneIdFieldsValidator

lionel-nj:
**Summary:**

This PR provides changes to improve DuplicateRouteNameValidator

**Expected behavior:** 

An hash code is used to check for duplicate instead of using String concatenation. 
Same tests should pass.

### DIFF
--- a/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateFareRuleZoneIdFieldsValidator.java
+++ b/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateFareRuleZoneIdFieldsValidator.java
@@ -63,11 +63,11 @@ public class DuplicateFareRuleZoneIdFieldsValidator extends FileValidator {
   }
 
   /**
-   * Returns the hashcode associated to the combination of `fare_rules.route_id`,
+   * Returns the hashcode associated to the combination of `fare_rules.origin_id`,
    * `fare_rules.contains_id` and `fare_rules.destination_id`.
    *
    * @param fareRule
-   * @return the hashcode associated to the combination of this {@code GtfsFareRule} `route_id`,
+   * @return the hashcode associated to the combination of this {@code GtfsFareRule} `origin_id`,
    *     `contains_id` and `destination_id`.
    */
   private int getZoneFieldsCombinationHash(GtfsFareRule fareRule) {
--- a/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateFareRuleZoneIdFieldsValidator.java
+++ b/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateFareRuleZoneIdFieldsValidator.java
@@ -18,6 +18,7 @@ package org.mobilitydata.gtfsvalidator.validator;
 
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Objects;
 import javax.inject.Inject;
 import org.mobilitydata.gtfsvalidator.annotation.GtfsValidator;
 import org.mobilitydata.gtfsvalidator.notice.DuplicateFareRuleZoneIdFieldsNotice;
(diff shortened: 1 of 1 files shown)


main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateFareRuleZoneIdFieldsValidator.java
```diff
@@ -41,27 +42,35 @@
...
         .forEach(
             fareRule -> {
-              String fieldsCombination =
-                  fareRule.routeId()
-                      + fareRule.originId()
-                      + fareRule.containsId()
-                      + fareRule.destinationId();
-              if (fareRuleByZoneIdFieldsCombination.containsKey(fieldsCombination)) {
+              int fieldsCombinationHash = getZoneFieldsCombinationHash(fareRule);
+              if (fareRuleByZoneIdFieldsCombination.containsKey(fieldsCombinationHash)) {
```
barbeau, line 52: Similar to https://github.com/MobilityData/gtfs-validator/pull/806#discussion_r595456725, if you get the value from the HashMap and assign to a local variable and check for null instead of using `containsKey()`, you'll avoid multiple lookups here for the same hash.

Generally speaking, you only want to use `containsKey()` if you don't need to reference the value. If you need to reference the value, you can avoid the duplicate lookup by using `.get()` and checking for `null` instead.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateFareRuleZoneIdFieldsValidator.java
```diff
@@ -41,27 +42,35 @@
...
         .getEntities()
         .forEach(
             fareRule -> {
-              String fieldsCombination =
-                  fareRule.routeId()
-                      + fareRule.originId()
-                      + fareRule.containsId()
-                      + fareRule.destinationId();
-              if (fareRuleByZoneIdFieldsCombination.containsKey(fieldsCombination)) {
+              int fieldsCombinationHash = getZoneFieldsCombinationHash(fareRule);
```
barbeau, line 51: `fieldsCombinationHash` could be shortened to `hash` or `key` for readability

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateFareRuleZoneIdFieldsValidator.java
```diff
@@ -41,27 +42,35 @@
...
+
+  /**
+   * Returns the hashcode associated to the combination of `fare_rules.origin_id`,
+   * `fare_rules.contains_id` and `fare_rules.destination_id`.
+   *
+   * @param fareRule
+   * @return the hashcode associated to the combination of this {@code GtfsFareRule} `origin_id`,
+   *     `contains_id` and `destination_id`.
+   */
+  private int getZoneFieldsCombinationHash(GtfsFareRule fareRule) {
```
barbeau, line 73: For code readability I think you could shorten this method name to just `getHash()`. The Javadocs and code give an explanation of what it includes.

barbeau (commented):
Thanks @lionel-nj, comments in-line.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateFareRuleZoneIdFieldsValidator.java
lionel-nj: Fixed in https://github.com/MobilityData/gtfs-validator/pull/811/commits/553ba6d945455276dbbc2a44a8dbee9baf0aa534

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateFareRuleZoneIdFieldsValidator.java
```diff
@@ -41,27 +42,36 @@
...
+                  new DuplicateFareRuleZoneIdFieldsNotice(
+                      fareRule.csvRowNumber(),
+                      fareRule.fareId(),
+                      otherFareRule.csvRowNumber(),
+                      otherFareRule.fareId()));
             });
   }
+
+  /**
+   * Returns the hashcode associated to the combination of `fare_rules.origin_id`,
```
aababilov, line 67: Backticks do not work in Javadoc, unfortunately. You can use `{@code ...}` or double quotes.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateFareRuleZoneIdFieldsValidator.java
```diff
@@ -41,27 +42,36 @@
...
-                        fareRuleByZoneIdFieldsCombination.get(fieldsCombination).csvRowNumber(),
-                        fareRuleByZoneIdFieldsCombination.get(fieldsCombination).fareId()));
-              } else {
-                fareRuleByZoneIdFieldsCombination.put(fieldsCombination, fareRule);
+              int hash = getHash(fareRule);
+              GtfsFareRule otherFareRule = fareRuleByZoneIdFieldsCombination.get(hash);
+              if (otherFareRule == null) {
+                fareRuleByZoneIdFieldsCombination.put(hash, fareRule);
+                return;
               }
```
aababilov, line 56: Use putIfAbsent instead of combination of get + put. This way you make the code shorter and avoid the temporary `hash` variable.

aababilov (changes requested):
Very nice, thank you!

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateFareRuleZoneIdFieldsValidator.java
lionel-nj: Fixed in https://github.com/MobilityData/gtfs-validator/pull/811/commits/8b867b4dc69652fc75a9065dbb1c9f7e7af3809f

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateFareRuleZoneIdFieldsValidator.java
```diff
@@ -41,27 +42,34 @@
...
         .getEntities()
         .forEach(
             fareRule -> {
-              String fieldsCombination =
-                  fareRule.routeId()
-                      + fareRule.originId()
-                      + fareRule.containsId()
-                      + fareRule.destinationId();
-              if (fareRuleByZoneIdFieldsCombination.containsKey(fieldsCombination)) {
+              int hash = getHash(fareRule);
```
aababilov, line 51: You can inline this temp variable.

aababilov (approved):

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateFareRuleZoneIdFieldsValidator.java
barbeau: @aababilov Just curious - in https://github.com/MobilityData/gtfs-validator/blob/3ceadcb1ee9626499de40e68c6cc88aa034688df/core/src/main/java/org/mobilitydata/gtfsvalidator/parsing/FieldCache.java#L60 ...you mentioned that:

>    // Benchmarks show that computeIfAbsent() is about 20% more expensive than calling get() and
    // put().

Do you know if the same applies to `putIfAbsent()`?

This post seems to think so, although the video it references is from 2009 - https://www.javaer101.com/en/article/4544460.html.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateFareRuleZoneIdFieldsValidator.java
```diff
@@ -41,27 +42,34 @@
...
+
+  /**
+   * Returns the hashcode associated to the combination of "fare_rules.origin_id",
+   * "fare_rules.contains_id" and "fare_rules.destination_id".
+   *
+   * @param fareRule the {@code GtfsFareRule} to generate the hash from
+   * @return the hashcode associated to the combination of this {@code GtfsFareRule} "origin_id",
+   *     "contains_id" and "destination_id".
+   */
+  private int getHash(GtfsFareRule fareRule) {
```
lionel-nj, line 72: As discussed `route_id` should be part of this composite key (https://github.com/MobilityData/gtfs-validator/pull/713#issuecomment-776047895). Opening an issue to include this.

PR #814, opened 2021-03-16
feat: Support injectable constructors

aababilov:
Because Java has specific, well-defined semantics for constructors
and how they relate to instance creation, constructor injection is
the simplest and most predictable form of dependency injection:

* Only one injectable constructor can exist per class
* Constructor injection must happen exactly once per instance
* Constructor injection can set final fields, which guarantees
  that a dependency has been injected. In real-world software,
  this is decidedly to your advantage.

The existing validator classes will be updated in a separate commit.

Support for injectable fields will be eventually removed.

### DIFF
--- a/core/src/main/java/org/mobilitydata/gtfsvalidator/validator/ValidatorLoader.java
+++ b/core/src/main/java/org/mobilitydata/gtfsvalidator/validator/ValidatorLoader.java
@@ -21,13 +21,16 @@ import com.google.common.collect.ListMultimap;
 import com.google.common.flogger.FluentLogger;
 import com.google.common.reflect.ClassPath;
 import java.io.IOException;
+import java.lang.reflect.Constructor;
 import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.function.Function;
 import javax.inject.Inject;
 import org.mobilitydata.gtfsvalidator.annotation.GtfsValidator;
 import org.mobilitydata.gtfsvalidator.notice.NoticeContainer;
@@ -40,6 +43,7 @@ import org.mobilitydata.gtfsvalidator.table.GtfsTableContainer;
  * annotation and provides convenient methods to invoke them on a single entity of file.
  */
 public class ValidatorLoader {
+
   private static final FluentLogger logger = FluentLogger.forEnclosingClass();
 
   private final ListMultimap<Class<? extends GtfsEntity>, Class<? extends SingleEntityValidator<?>>>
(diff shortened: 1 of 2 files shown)


barbeau:
@aababilov Thanks for looking into this! Can you elaborate on the downsides of not adopting injectable constructors?

Looking at https://github.com/MobilityData/gtfs-validator/pull/815/files#diff-e20e822423c69120a7f2767673186e0ec1381c7f1ed48b18e4ea11ebb7dbb727, it appears the definition of the injection for validators changes from:

```java
public class AgencyConsistencyValidator extends FileValidator {

  @Inject GtfsAgencyTableContainer agencyTable;
...
```

...to:

```java
public class AgencyConsistencyValidator extends FileValidator {
  @Inject
  public AgencyConsistencyValidator(GtfsAgencyTableContainer agencyTable) {
    this.agencyTable = agencyTable;
  }
  private final GtfsAgencyTableContainer agencyTable;
...
```

So it seems more complex and verbose for someone to define the injection using the new approach. 

IMHO we should consider the ease of implementation of validators for newcomers and potential for future errors when weighing this change.

aababilov:
Yes, the constructor injection is more verbose but members injection has somewhat more complicated semantics and often weaker guarantees:

* Injectable members are injected in the sequence specified in @Inject. Members injection is not (cannot be) atomic.
* It is not possible to only inject a subset of injectable members.
* The presence of an injectable member makes no guarantees about whether or not that member will be injected. Members injection can even be performed multiple times on a single instance.
* Constructor injection guarantees members injection, but the presence of an injectable constructor does not — manual invocation of an injectable constructor circumvents members injection. Thus, care should always be taken to ensure that injectable members are in the proper state before they are used.

That rationale applies to all dependency injection frameworks. I think that GTFS Validator should demonstrate the best Java practices.

* GTFS community members may perform code review before adopting the Validator. They would be praised by the best Java practices. Use of poor style would erode their trust.
* This teaches newcomers to write reliable code.

barbeau:
@aababilov Thanks for the explanation, that makes sense. It sounds like these would be the next steps to adopt injectable constructors (correct me if I'm wrong):
1. Merge this PR
2. Finish migration to injectable constructors in https://github.com/MobilityData/gtfs-validator/pull/815, merge that PR
3. Open new PR removing injectable members, merge that PR

Is finishing implementation of 2) and 3) above something you could complete in the next day or two?

MobilityData is trying to get to a v2 release, and IMHO we shouldn't change the validator definitions just after a major release (as this would be a breaking change and require a new major release), so if we want to get v2 released soon this would need to be completed quickly.

aababilov:
Challenge accepted, Sean! I will migrate the rest of validators today.

Den ons 17 mars 2021 08:25Sean Barbeau ***@***.***> skrev:

aababilov:
All validators are here: https://github.com/MobilityData/gtfs-validator/pull/815

lionel-nj (approved):
LGTM, thanks @aababilov.

barbeau:
Thanks @aababilov for completing this and #815 so quickly! I think we just need the last step now of removing injectable members?

aababilov:
Here you go: https://github.com/MobilityData/gtfs-validator/pull/819

aababilov:
I am also thinking of adopting Dagger in Validator. This may simplify our code since dependency injection is a fairly unusual thing to implement. But switching to Dagger can wait.

PR #819, opened 2021-03-17
feat: Remove support for injectable fields

aababilov:
Only injectable constructors should be used.


lionel-nj (approved):
Thanks @aababilov, LGTM 👍🏾!

barbeau (approved):
Thanks @aababilov!

PR #821, opened 2021-03-18
feat: Add indices on stop.parent_station and stop_times.stop_id

aababilov:
That will allow us to find all child stops of a station and all stop
times that belong to the same stop.

```
  stopTable.byParentStation("station1")
  stopTimeTable.byStopId("stop1")
```

Also reorder some annotations. We would like to have the following
common order:

* FieldType
* PrimaryKey, FirstKey or SequenceKey
* Index
* Required or ConditionallyRequired
* ForeignKey


barbeau (approved):
Thanks @aababilov!

PR #822, opened 2021-03-18
feat: Move notice subclasses inside the corresponding validators

aababilov:
Each notice from `main` is produced by a single validator.
Here we limit visibility of notice subclass better.


barbeau:
@aababilov Good idea! This is cleaner and makes it easier to explain how to contribute a new validation rule too (as it's all in one place). Could you refactor the remaining notices too so all changes are in the same PR?

lionel-nj:
That looks generally good to me! As @barbeau said it makes it easier to implement new rule (as a newcomer): everything related to a validation rule would be self-contained.

aababilov:
I am glad that you like that!

IDEA makes refactoring easy with drag-and-drop, so here is the full migration.

lionel-nj (approved):
🚀 LGTM ! Thanks a lot @aababilov.

PR #828, opened 2021-03-22
feat: Change Error suffix to Notice

aababilov:
Only child classes of ValidationNotice were affected.


lionel-nj (approved):
Thanks @aababilov  for providing these changes! LGTM

barbeau (approved):
Thanks @aababilov!

PR #833, opened 2021-03-24
feat: Add field names to the context of several notices

aababilov:
Affected notices:
* DuplicateKeyNotice
* StartAndEndRangeEqualNotice
* StartAndEndRangeOutOfOrderNotice

Fields that are present in the context of DuplicateKeyNotice now have
fixed names. Previously, field name depended on the actual GTFS file
that had the error. This was hard to handle in visualization. Consider
that you have a format string to convert your notice to text - then you
need to know all possible field names in advance.


lionel-nj (approved):
thanks @aababilov LGTM!

PR #839, opened 2021-03-29
feat: Validate that combination of route names and type are unique wi…

aababilov:
…thin an agency

This logic corresponds to real-world use cases validated at Google for
all its production feeds.

Previous validation produced too many false positives because it ignored
route type and validated uniqueness of short and long names separately.

### DIFF
--- a/RULES.md
+++ b/RULES.md
@@ -557,7 +557,7 @@ At least one of the fields `is_producer`, `is_operator`, or `is_authority` shoul
 
 All routes of the same `route_type` with the same `agency_id` should have unique combinations of `route_short_name` and `route_long_name`.
 
-Note that there may be valid cases where routes have the same short and long name, e.g., if they serve difference areas. However, different directions must be modeled as the same route.
+Note that there may be valid cases where routes have the same short and long name, e.g., if they serve different areas. However, different directions must be modeled as the same route.
 
 ##### References:
 * [routes.txt specification](http://gtfs.org/reference/static/#routestxt)
--- a/RULES.md
+++ b/RULES.md
@@ -555,15 +555,9 @@ At least one of the fields `is_producer`, `is_operator`, or `is_authority` shoul
 
 #### DuplicateRouteNameNotice
 
-All routes should have different `routes.route_long_name` - if two `routes.route_long_name` are the same, and the two routes belong to the same agency, a notice is generated.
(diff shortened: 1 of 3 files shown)


lionel-nj (approved):
Thanks @aababilov LGTM

lionel-nj (dismissed):
Just noticed that `README.md` should be updated to reflect the changes.

lionel-nj (changes requested):
LGTM! Just one thing: `README.md` should be updated to reflect the changes.

Thanks @aababilov

aababilov:
Done!

lionel-nj:
Thanks @aababilov ! :)

RULES.md
```diff
@@ -555,15 +555,9 @@ At least one of the fields `is_producer`, `is_operator`, or `is_authority` shoul
...
-
-All routes should have different `routes.route_short_name` - if two `routes.route_short_name` are the same, and the two routes belong to the same agency, a notice is generated.
-
-Note that there may be valid cases where routes may have the same `routes.route_short_name` and this notice can be ignored. For example, routes may have the same routes.route_short_name if they serve difference areas. However, they must not be different trips of the same route or different directions of the same route - these cases should always have unique `routes.route_short_name`.
-
-The same combination of `route_short_name` and `route_long_name` should not be used for more than one route.
+Note that there may be valid cases where routes have the same short and long name, e.g., if they serve difference areas. However, different directions must be modeled as the same route.
```
lionel-nj, line 560:
```suggestion
Note that there may be valid cases where routes have the same short and long name, e.g., if they serve different areas. However, different directions must be modeled as the same route.
```

lionel-nj, line 560:
```suggestion
Note that there may be valid cases where routes have the same short and long name, e.g., if they serve different areas. However, different directions must be modeled as the same route.
```

lionel-nj: Just one little typo here.

aababilov:
Thanks for review, @lionel-nj !

lionel-nj (changes requested):
Just one typo and then we're good to go :)

RULES.md
aababilov: Fixed. That typo has been there for weeks :)

lionel-nj (approved):
👍🏾 ✅ 🚀

RULES.md
lionel-nj: Oops! My bad, I'll keep an eye open to avoid making typos in the future. Thanks for flagging that 🙏🏾

PR #843, opened 2021-03-31
feat: Provide enum setters and clearFIELD methods

aababilov:
Entity Builder classes are used in two contexts:
* in autogenerated Loader classes;
* in unit tests.

Loader classes call setFieldName() methods that accept Integer. But for unit tests, it is more convenient to pass the enum.

```
    @Nonnull
    public GtfsPathway.Builder setPathwayMode(@Nullable Integer value) {
      if (value == null) {
        return clearPathwayMode();
      }
      pathwayMode = value;
      bitField0_ |= 0x8;
      return this;
    }

    @Nonnull
    public GtfsPathway.Builder setPathwayMode(@Nullable GtfsPathwayMode value) {
      return value == null ? clearPathwayMode() : setPathwayMode(value.getNumber());
    }

    @Nonnull
    public GtfsPathway.Builder clearPathwayMode() {
      pathwayMode = DEFAULT_PATHWAY_MODE;
      bitField0_ &= ~0x8;
      return this;
    }
```


barbeau (approved):
Thanks @aababilov!

PR #850, opened 2021-04-14
feat: Make file headers available in CSV table containers

aababilov:
Some validation rules may need to check whether a certain column is
present. E.g., this can allow them to skip further validation.

Unit tests are simplified and mocks are removed.

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -43,7 +43,6 @@ dependencies {
     implementation 'com.google.flogger:flogger:0.5.1'
     testImplementation group: 'junit', name: 'junit', version: '4.13'
     testImplementation "com.google.truth:truth:1.0.1"
-    testImplementation 'org.mockito:mockito-core:1.10.19'
 }
 
 test {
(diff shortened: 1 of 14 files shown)


aababilov:
Note that JavaDocs do not have the `@return` or `@param` section. This conforms to Google Java style that says that redundant sections are not required if the main comment already explains everything.

barbeau:
>Column names are case-insensitive. Real feeds can capitalize them.

https://gtfs.org/reference/static says under "File Requirements:"

>All field names are case-sensitive.

@MobilityData/transit-specs Thoughts on this?

lionel-nj:
Hi @aababilov Thanks for working on this. I will soon be reviewing :)

aababilov:
> > Column names are case-insensitive. Real feeds can capitalize them.
> 
> https://gtfs.org/reference/static says under "File Requirements:"
> 
> > All field names are case-sensitive.
> 
> @MobilityData/transit-specs Thoughts on this?

Fair point! Let's keep column names case-sensitive. We can revisit that topic later if needed.

I would like to avoid extra discussion for this commit, so let's keep the old behavior.

aababilov:
> Hi @aababilov Thanks for working on this. I will soon be reviewing :)

Merci beaucoup!

aababilov:
Friendly ping 😉

Den tors 15 apr. 2021 07:21Lionel Nébot Janvier ***@***.***>
skrev:

lionel-nj (approved):
Thanks @aababilov LGTM

barbeau (approved):
Thanks @aababilov! 👍

aababilov:
Thanks for review!

PR #851, opened 2021-04-14
feat!: deprecate -f 

### LINKED ISSUE #539: Make -f an optional command-line parameter

lionel-nj:
closes #539

**Summary:**

This PR provides support to make `-f` an optional argument when running the validator. This is required to run the validator programatically for:
- integration to the MobilityDatabase
- integration tests purpose

**Expected behavior:** 

if `-f`  is now optional. If provided, phone numbers will be validated according to the country code, else phone number validation is skipped.

No NPEs should be thrown.


lionel-nj:
@barbeau  Question: should we modify `-f` to be `-c` (meaning country code)?

core/src/main/java/org/mobilitydata/gtfsvalidator/input/GtfsFeedName.java
```diff
@@ -141,13 +158,16 @@ public boolean equals(Object other) {
...
-      return this.countryFirstName.equals(((GtfsFeedName) other).countryFirstName);
+      return Objects.equals(this.countryFirstName, ((GtfsFeedName) other).countryFirstName);
     }
     return false;
   }
 
   @Override
   public int hashCode() {
+    if (countryFirstName == null) {
+      return 0;
```
lionel-nj, line 169: > the hash code of a non-null argument and 0 for a null argument


According to https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#hashCode-java.lang.Object-

barbeau: That is just the behavior of that static method because given `null` it can't compute any other relevant hashcode for the given value. In this case, you would want to return `this.hashCode()` for the hashcode of the object instance so you can tell if two instances are the same.

barbeau (commented):
I support making `-f` optional but I'm not convinced that wrapping `null` within `GtfsFeedName` is the right way to approach this. Seems dangerous and prone to future NPEs.

Instead, I think a better approach would be to have a new `-c` command line parameter for country code and truly make `-f` optional, and deprecate it in future releases (unless we see other value in keeping it). If either are provided you can validate phone number, and if neither are provided you skip that rule.

aababilov:
Thanks for working in that direction!

One more thing came to my mind. Let's remove the monolithic ValidationContext and inject only direct dependencies.

"Inject only direct dependencies" is one of the main DI principles at Google. Here is what it means.

This approach is not recommended:
```
public class ShowBudgets {
   private final Account account;

   @Inject
   ShowBudgets(Customer customer) {
     account = customer.getPurchasingAccount();
   }
```

This one is preferred:

```
public class ShowBudgets {
   private final Account account;

   @Inject
   ShowBudgets(Account account) {
     this.account = account;
   }
```


So, instead of injecting ValidationContext, we should inject the CountryCode instead. 
We probably should not inject ZonedDateTime but we need a wrapper class for it (say, CurrentDateTime) because ZonedDateTime does not have a semantic that it is the current datetime.

How does that sound?

main/src/main/java/org/mobilitydata/gtfsvalidator/cli/Arguments.java
```diff
@@ -38,11 +38,11 @@
   private int numThreads = 1;
 
   @Parameter(
-      names = {"-f", "--feed_name"},
```
barbeau, line 41: If we remove `-f` completely I believe this will be a breaking change for clients/users and the next release should be v3.

core/src/main/java/org/mobilitydata/gtfsvalidator/input/CountryCode.java
```diff
@@ -0,0 +1,106 @@
...
+
+/** Represents the country code of a GTFS feed. */
+public class CountryCode {
+  private static final Set<String> ISO_COUNTRIES = ImmutableSet.copyOf(Locale.getISOCountries());
+  private static final String INVALID_COUNTRY_CODE_MESSAGE =
+      "Country code must follow format `cc`, where cc is an ISO Alpha 2 country code";
+  // ISO 3166-1 Alpha 2 country code for the United Kingdom is "GB" but feed names may use "UK".
+  private static final String UK = "UK";
+  private static final String GB = "GB";
+  private final String countryCode;
```
barbeau, line 31: I would use Java `Optional` here for `countryCode` instead of storing `null` directly to avoid potential future NPEs:
https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Optional.html

core/src/main/java/org/mobilitydata/gtfsvalidator/input/CountryCode.java
```diff
@@ -0,0 +1,106 @@
...
+   * <p>Note that this returns "GB" for feeds that use the non-standard "UK" country code.
+   *
+   * @return uppercase ISO Alpha country code
+   */
+  public String getISOAlpha2CountryCode() {
+    String countryCode = this.countryCode.toUpperCase();
+    return countryCode.equals(UK) ? GB : countryCode;
+  }
+
+  /**
```
barbeau, line 83: You can get rid of this method if you use `Optional` as I mention above.

core/src/main/java/org/mobilitydata/gtfsvalidator/input/CountryCode.java
```diff
@@ -0,0 +1,106 @@
...
+    }
+    if (other instanceof CountryCode) {
+      return this.countryCode.equals(((CountryCode) other).countryCode);
+    }
+    return false;
+  }
+
+  @Override
+  public int hashCode() {
+    return countryCode.hashCode();
```
barbeau, line 104: Looks like your previous approach was correct - `Optional.hashCode()` returns `0` if empty:
https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Optional.html#hashCode()

I would recommend using `Optional.hashCode()` directly instead though.

core/src/main/java/org/mobilitydata/gtfsvalidator/parsing/RowParser.java
```diff
@@ -464,6 +469,21 @@ private String asValidatedString(
     return s;
   }
 
+  /**
```
barbeau, line 472: To simplify things, I think you could get rid of this method and just use `return asString(columnIndex, required);` where this method is used - I believe they are equivalent, as the `if` statement below is redundant.

lionel-nj:
> How does that sound?

Good to me! 👍🏾

barbeau (commented):
Thanks @lionel-nj! Some comments in-line. The README and other docs also need to be updated.

core/src/main/java/org/mobilitydata/gtfsvalidator/input/CountryCode.java
```diff
@@ -0,0 +1,103 @@
...
+
+/** Represents the country code of a GTFS feed. */
+public class CountryCode {
+  private static final Set<String> ISO_COUNTRIES = ImmutableSet.copyOf(Locale.getISOCountries());
+  private static final String INVALID_COUNTRY_CODE_MESSAGE =
+      "Country code must follow format `cc`, where cc is an ISO Alpha 2 country code";
+  // ISO 3166-1 Alpha 2 country code for the United Kingdom is "GB" but feed names may use "UK".
+  private static final String UK = "UK";
+  private static final String GB = "GB";
+  private final Optional<String> optionalCountryCode;
```
barbeau, line 32: To keep it brief `countryCode` should be fine.

README.md
```diff
@@ -16,17 +16,17 @@ This is a command-line tool written in Java that performs the following steps:
...
 To download and validate a GTFS dataset from a URL:
 
-`java -jar gtfs-validator-v2.0.0_cli.jar -u https://www.abc.com/gtfs.zip -o output -f ca-myFeedName`
+`java -jar gtfs-validator-v2.0.0_cli.jar -u https://www.abc.com/gtfs.zip -o output -c ca`
 
 where:
 * `--input` or `-i`: the path to the GTFS file (e.g., `/myDirectory/gtfs.zip`)
 * `--url` or `-u`: the fully qualified URL to the GTFS file (e.g., `https://www.abc.com/gtfs.zip`)
 * `--output` or `-o`: the path where the validation report will be stored (e.g., `output`)
-* `--feed_name` or `-f`: the name of the feed as a valid [ISO two letter country code](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2), followed by `-`, followed by a user-defined name for the feed. (e.g., `ca-myFeedName`, `us-myFeedName`)
```
barbeau, line 29: So what is the behavior of the code in this PR when someone passes in `-f` as a parameter?

We should provide some indication that `-f` is no longer supported so it the user doesn't continue to use `-f`, not realizing that it no longer does anything. I would consider returning false from `CliParametersAnalyzer.isValid()` if `-f` is provided, saying to use `-c` instead.

core/src/main/java/org/mobilitydata/gtfsvalidator/input/CountryCode.java
```diff
@@ -0,0 +1,103 @@
...
+
+  /**
+   * Returns the uppercase ISO 3166-1 Alpha 2 country code component, e.g., "NL" or "AU".
+   *
+   * <p>Note that this returns "GB" for feeds that use the non-standard "UK" country code.
+   *
+   * @return uppercase ISO Alpha country code
+   */
+  public String getISOAlpha2CountryCode() {
+    String countryCode = optionalCountryCode.get().toUpperCase();
```
barbeau, line 84: If `optionalCountryCode` is empty `optionalCountryCode.get()` will throw an exception:
https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Optional.html#get()

You need to protect `.get()` by wrapping it in a call to `isPresent()` first.

I think a better approach is to move this logic into the constructor and store the uppercase ISO code in `optionalCountryCode` - the only reason it was in a separate method previously was because the code needed to be parsed from the name.

Then, you can get rid of this method so the only way to access the data would be `getCountryCode()`. This would then be a clear indication to the calling code that this value may not be populated (it's `Optional`).

barbeau (commented):
Thanks @lionel-nj, some thoughts in-line.

barbeau:
Oh, if this is a breaking change, the commit/PR title should include a `!` - https://www.conventionalcommits.org/en/v1.0.0/

README.md
lionel-nj: > So what is the behavior of the code in this PR when someone passes in -f as a parameter?

As of now, an exception is raised if `-f` is provided. 
```
Exception in thread "main" com.beust.jcommander.ParameterException: Unknown option: -f
	at com.beust.jcommander.JCommander.parseValues(JCommander.java:742)
	at com.beust.jcommander.JCommander.parse(JCommander.java:282)
	at com.beust.jcommander.JCommander.parse(JCommander.java:265)
	at org.mobilitydata.gtfsvalidator.cli.Main.main(Main.java:49)
```

> I would consider returning false from CliParametersAnalyzer.isValid() if -f is provided, saying to use -c instead.

Done in https://github.com/MobilityData/gtfs-validator/pull/851/commits/251cb314de0d04521bc7f74af14fa18264fc1f90

lionel-nj:
Removing mocks from `CliParametersAnalyzerTest` in #854

core/src/main/java/org/mobilitydata/gtfsvalidator/input/CountryCode.java
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  /**
+   * Parses a string as a country code name.
+   *
+   * @param countryCode country code for a GTFS dataset
+   * @return a valid GtfsCountryCode object
+   * @throws IllegalArgumentException if illegal @param feedName
+   */
+  public static CountryCode parseString(String countryCode) throws IllegalArgumentException {
```
aababilov, line 67: Let's return a ZZ for invalid countries and simplify the code by removing handling for a null parameter. Let's have:

```
public static CountryCode forStringOrUnknown(String cldrCode) {
  String upperCaseCldrCode = Ascii.toUpperCase(cldrCode);
  if (upperCaseCldrCode.equals(UK)) {
    upperCaseCldrCode = GB;
  }
  if (!ISO_COUNTRIES.contains(upperCaseCldrCode)) {
    upperCaseCldrCode = ZZ;
  }
  return new CountryCode(upperCaseCldrCode);
}

File: core/src/main/java/org/mobilitydata/gtfsvalidator/input/CountryCode.java [hunk 11]
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  /**
+   * Returns the uppercase ISO 3166-1 Alpha 2 country code component, e.g., "NL" or "AU".
+   *
+   * <p>Note that this returns "GB" for feeds that use the non-standard "UK" country code.
+   *
+   * @return uppercase ISO Alpha country code
+   */
+  public Optional<String> getCountryCode() {
```
aababilov, line 84: public String getCountryCode() {

core/src/main/java/org/mobilitydata/gtfsvalidator/input/CountryCode.java
```diff
@@ -0,0 +1,103 @@
...
+  /**
+   * Checks that the given string is a valid ISO Alpha 2 country code (case insensitive).
+   *
+   * <p>We support "UK" as an additional country code equivalent to the ISO 3166-1 Alpha 2 code
+   * "GB".
+   *
+   * @param s the given string
+   * @return true if the given string is a valid ISO Alpha 2 country code (case insensitive)
+   */
+  public static boolean isValidISOAlpha2(String s) {
```
aababilov, line 52: You can drop this.

core/src/main/java/org/mobilitydata/gtfsvalidator/input/CountryCode.java
```diff
@@ -0,0 +1,103 @@
...
+import javax.annotation.Nullable;
+
+/** Represents the country code of a GTFS feed. */
+public class CountryCode {
+  private static final Set<String> ISO_COUNTRIES = ImmutableSet.copyOf(Locale.getISOCountries());
+  private static final String INVALID_COUNTRY_CODE_MESSAGE =
+      "Country code must follow format `cc`, where cc is an ISO Alpha 2 country code";
+  // ISO 3166-1 Alpha 2 country code for the United Kingdom is "GB" but feed names may use "UK".
+  private static final String UK = "UK";
+  private static final String GB = "GB";
```
aababilov, line 32: Please add:
private static final String ZZ = "ZZ";

core/src/main/java/org/mobilitydata/gtfsvalidator/input/CountryCode.java
```diff
@@ -0,0 +1,103 @@
...
+  public boolean equals(Object other) {
+    if (this == other) {
+      return true;
+    }
+    if (other instanceof CountryCode) {
+      return this.countryCode.equals(((CountryCode) other).countryCode);
+    }
+    return false;
+  }
+
```
aababilov, line 98: Please also override toString():

```
  return countryCode;
```

core/src/main/java/org/mobilitydata/gtfsvalidator/input/CountryCode.java
```diff
@@ -0,0 +1,103 @@
...
+  @Override
+  public int hashCode() {
+    return countryCode.hashCode();
+  }
+}
```
aababilov, line 103: Please also add:

 /**
   * Returns true if this is the unknown region ("ZZ").
   * This value is returned by APIs when no suitable region can be determined,
   * which can indicate bad input or missing data.
   */
  public boolean isUnknown() {
    return countryCode.equals(ZZ);
  }

core/src/main/java/org/mobilitydata/gtfsvalidator/input/CountryCode.java
```diff
@@ -0,0 +1,103 @@
...
+  private static final String GB = "GB";
+  private Optional<String> countryCode;
+
+  private CountryCode(@Nullable String countryCode) {
+    this.countryCode = Optional.ofNullable(countryCode);
+    if (countryCode != null) {
+      this.countryCode = Optional.of(countryCode.toUpperCase().equals(UK) ? GB
+          : countryCode.toUpperCase());
+    }
+  }
```
aababilov, line 41: This may be simplified:

```
private CountryCode(String countryCode) {
    this.countryCode = countryCode;
}
```

core/src/main/java/org/mobilitydata/gtfsvalidator/input/CountryCode.java
```diff
@@ -0,0 +1,103 @@
...
+import com.google.common.collect.ImmutableSet;
+import java.util.Locale;
+import java.util.Optional;
+import java.util.Set;
+import javax.annotation.Nullable;
+
+/** Represents the country code of a GTFS feed. */
+public class CountryCode {
+  private static final Set<String> ISO_COUNTRIES = ImmutableSet.copyOf(Locale.getISOCountries());
+  private static final String INVALID_COUNTRY_CODE_MESSAGE =
```
aababilov, line 28: This may be removed.

core/src/main/java/org/mobilitydata/gtfsvalidator/parsing/RowParser.java
```diff
@@ -187,10 +197,15 @@ public String asEmail(int columnIndex, boolean required) {
 
   @Nullable
   public String asPhoneNumber(int columnIndex, boolean required) {
+    if (!countryCode.getCountryCode().isPresent()) {
+      return asString(columnIndex, required);
+    }
```
aababilov, line 202: This can be removed.

core/src/main/java/org/mobilitydata/gtfsvalidator/parsing/RowParser.java
```diff
@@ -464,6 +457,18 @@ private String asValidatedString(
...
+    NON_NEGATIVE,
+    NON_ZERO,
+  }
+
+  @FunctionalInterface
+  public interface EnumCreator<E> {
+
+    E convert(int t);
+  }
+
```
aababilov, line 471: This is a refactoring, possibly initiated by Idea when you allow it to rearrange entries. Please restore the previous position since refactoring should not be mixed with functional changes.

core/src/main/java/org/mobilitydata/gtfsvalidator/parsing/RowParser.java
```diff
@@ -66,15 +66,34 @@
...
+  /**
+   * Tells if a given notice is an {@code ERROR}.
+   *
+   * @param notice the notice to check
+   * @return true if the notice is an error, false otherwise
+   */
+  private static boolean isError(ValidationNotice notice) {
+    return notice.getSeverityLevel().ordinal() >= SeverityLevel.ERROR.ordinal();
+  }
+
```
aababilov, line 96: Please move it back.

README.md
```diff
@@ -16,17 +16,17 @@ This is a command-line tool written in Java that performs the following steps:
...
+* *(Optional)* `--country_code` or `-c`: the country code of the feed as a valid [ISO two letter country code](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2) (e.g., `ca`, `us`)
```
aababilov, line 29: Please mention that it can be either upper or lower case and give a sample, e.g., "FR"  or "GP" ;)

And mention that it is optional. If country is unknown, we still validate numbers that start from +.

aababilov (changes requested):
Nice, can't wait to see it merged!

lionel-nj:
Thanks @aababilov  for reviewing! I provided the changes you suggested in the previous commits. At the exception of this one:  https://github.com/MobilityData/gtfs-validator/pull/851#discussion_r615135074

> private static final String ZZ = "ZZ";

I made this constant `public`, so that we can access from the main method in order to avoid NPE.

main/src/test/java/org/mobilitydata/gtfsvalidator/cli/ArgumentsTest.java
```diff
@@ -83,7 +83,7 @@ public void longNameShouldInitializeArguments() {
     commandLineArgumentAsStringArray =
         new String[] {
           "--output_base", "output value",
-          "--feed_name", "feed name value",
+            "--country_code", "ca",
```
aababilov, line 86: Looks like a wrong indentation.

I usually do this to format all files:

```
google-java-format -i $(find . -name '*.java')
```

aababilov:
Beautiful! Please format the code and submit it, I can't wait to see it in upstream!

```
google-java-format -i $(find . -name '*.java')
```

lionel-nj:
Thanks @aababilov for this feedback!

barbeau (approved):
Nice work @lionel-nj! 👍

PR #859, opened 2021-04-20
perf: Use farmHashFingerprint64 instead of Objects.hash

aababilov:
farmHashFingerprint64 is better protected against collisions.

Note that we do not use concatenated strings as keys in our hash maps
since they are long, occupy much space and negatively impact
productivity.

### DIFF
--- a/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateFareRuleZoneIdFieldsValidator.java
+++ b/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateFareRuleZoneIdFieldsValidator.java
@@ -17,9 +17,10 @@
 package org.mobilitydata.gtfsvalidator.validator;
 
 import com.google.common.collect.ImmutableMap;
+import com.google.common.hash.HashFunction;
+import com.google.common.hash.Hashing;
 import java.util.HashMap;
 import java.util.Map;
-import java.util.Objects;
 import javax.inject.Inject;
 import org.mobilitydata.gtfsvalidator.annotation.GtfsValidator;
 import org.mobilitydata.gtfsvalidator.notice.NoticeContainer;
@@ -40,7 +41,7 @@ import org.mobilitydata.gtfsvalidator.table.GtfsFareRuleTableContainer;
  */
 @GtfsValidator
 public class DuplicateFareRuleZoneIdFieldsValidator extends FileValidator {
-
+  private static final HashFunction HASH_FUNCTION = Hashing.farmHashFingerprint64();
   private final GtfsFareRuleTableContainer fareRuleTable;
 
   @Inject
@@ -50,7 +51,7 @@ public class DuplicateFareRuleZoneIdFieldsValidator extends FileValidator {
 
   @Override
   public void validate(NoticeContainer noticeContainer) {
-    final Map<Integer, GtfsFareRule> fareRuleByZoneIdFieldsCombination =
+    final Map<Long, GtfsFareRule> fareRuleByZoneIdFieldsCombination =
         new HashMap<>(fareRuleTable.entityCount());
     fareRuleTable
         .getEntities()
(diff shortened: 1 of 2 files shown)


main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateFareRuleZoneIdFieldsValidator.java
```diff
@@ -77,9 +78,18 @@ public void validate(NoticeContainer noticeContainer) {
    * @return the hashcode associated to the combination of this {@code GtfsFareRule} "route_id",
    *     "origin_id", "contains_id" and "destination_id".
    */
-  private int getHash(GtfsFareRule fareRule) {
-    return Objects.hash(
-        fareRule.routeId(), fareRule.originId(), fareRule.containsId(), fareRule.destinationId());
+  private static long getHash(GtfsFareRule fareRule) {
+    return HASH_FUNCTION
+        .newHasher()
+        .putUnencodedChars(fareRule.routeId())
```
lionel-nj, line 84: Why not using `putString` here?

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateFareRuleZoneIdFieldsValidator.java
```diff
@@ -77,9 +78,18 @@ public void validate(NoticeContainer noticeContainer) {
...
+        .putChar('\0')
```
lionel-nj, line 85: Why using this here?

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateRouteNameValidator.java
```diff
@@ -50,9 +52,17 @@ public void validate(NoticeContainer noticeContainer) {
...
-        route.routeLongName(), route.routeShortName(), route.routeType(), route.agencyId());
+  private static long getRouteKey(GtfsRoute route) {
+    return HASH_FUNCTION
+        .newHasher()
+        .putUnencodedChars(route.routeLongName())
+        .putChar('\0')
+        .putUnencodedChars(route.routeShortName())
+        .putChar('\0')
+        .putInt(route.routeType().getNumber())
+        .putUnencodedChars(route.agencyId())
```
lionel-nj, line 63: Should `.putChar('\0')` be used here?

```suggestion
        .putChar('\0')
        .putUnencodedChars(route.agencyId())
```

lionel-nj (commented):
Thanks @aababilov for working on that - I did not know this library. It comes very handy! 

A few questions in the comments.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateFareRuleZoneIdFieldsValidator.java
aababilov: Yep, exactly! We care for speed, so we don't bother with putString that will reencode the string.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateFareRuleZoneIdFieldsValidator.java
aababilov: You see, putUnencodedChars (and putString as well) do not put any delimiter, so

```
newHasher().putUnencodedChars("ab").putUnencodedChars("c").hash()
newHasher().putUnencodedChars("a").putUnencodedChars("bc").hash()
```

produce the same result.

Valid GTFS strings never include '\0', so I am using it as a delimiter.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateRouteNameValidator.java
aababilov: No, since this is the last string. A delimiter needs to be placed between a string and the next item.

aababilov:
Thanks for review! Let me know if my answers make sense.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateFareRuleZoneIdFieldsValidator.java
lionel-nj: Oh! Makes sense! Thanks.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DuplicateRouteNameValidator.java
lionel-nj: I get it :)

lionel-nj (approved):
Thanks @aababilov, that LGTM 🚀  If @barbeau agrees I'd say let's merge this PR!

barbeau (approved):
LGTM, thanks @aababilov!

PR #865, opened 2021-04-22
fix: Return ImmutableSet from GtfsInput.getFilenames()

aababilov:
The set of files is never modified.


lionel-nj (approved):
thanks @aababilov

PR #868, opened 2021-04-23
fix: Java style fixes

aababilov:
* it is not recommended to import classes with a generic name like Builder
* add static final modifiers
* avoid raw types
* remove redundant continue
* add missing @Override


lionel-nj (approved):
Thanks 👍🏾

PR #869, opened 2021-04-23
feat: Propagate InterruptedException during feed loading

aababilov:
InterruptedException is a sign to immediately stop validation. We may not just add a notice and keep on validating.

IBM wrote an article on the subject:
http://web.archive.org/web/20210201071452/https://www.ibm.com/developerworks/java/library/j-jtp05236/index.html

If throwing InterruptedException means that a method is a blocking method, then calling a blocking method means that your method is a blocking method too, and you should have a strategy for dealing with InterruptedException. Often the easiest strategy is to throw InterruptedException yourself, as shown in the putTask() and getTask() methods in Listing 1. Doing so makes your method responsive to interruption as well and often requires nothing more than adding InterruptedException to your throws clause.

### DIFF
--- a/core/src/main/java/org/mobilitydata/gtfsvalidator/table/GtfsFeedLoader.java
+++ b/core/src/main/java/org/mobilitydata/gtfsvalidator/table/GtfsFeedLoader.java
@@ -28,12 +28,12 @@ import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
 import org.mobilitydata.gtfsvalidator.annotation.GtfsLoader;
 import org.mobilitydata.gtfsvalidator.input.GtfsInput;
 import org.mobilitydata.gtfsvalidator.notice.NoticeContainer;
 import org.mobilitydata.gtfsvalidator.notice.RuntimeExceptionInLoaderError;
 import org.mobilitydata.gtfsvalidator.notice.ThreadExecutionError;
-import org.mobilitydata.gtfsvalidator.notice.ThreadInterruptedError;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/mobilitydata/gtfsvalidator/table/GtfsFeedLoader.java
```diff
@@ -86,74 +86,61 @@ public void setNumThreads(int numThreads) {
...
-                // the context since we know the filename here.
-                logger.atSevere().withCause(e).log("Runtime exception when loading %s", filename);
-                loaderNotices.addSystemError(
-                    new RuntimeExceptionInLoaderError(
-                        filename, e.getClass().getCanonicalName(), e.getMessage()));
-                // Since the file was not loaded successfully, we treat
-                // it as missing for continuing validation.
-                tableContainer = loader.loadMissingFile(validatorProvider, loaderNotices);
-              } finally {
-                inputStream.close();
```
lionel-nj, line 121: This line has been removed because wou wrote 
`try (InputStream inputStream = gtfsInput.getFile(filename)) {
...
}` earlier, right? Which will close our `InputStream`?

lionel-nj (approved):
Thanks @aababilov for working on this. Will keep the documentation you attached handy so that we can make sure to replicate this mechanism if needed in the future.

core/src/main/java/org/mobilitydata/gtfsvalidator/table/GtfsFeedLoader.java
aababilov: Sure, that is called try-with-resources. https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

aababilov:
Thanks for review!

Could you merge it? I still don't have permissions to do that on my own.

PR #870, opened 2021-04-23
feat: Simplify validator functions

aababilov:
If an error is stored to the NoticeContainer, then it is already clear
that the field is invalid. Returning an additional boolean is redundant
and may lead to bugs when the function returns false but no error
notices are emitted.

### DIFF
--- a/core/src/main/java/org/mobilitydata/gtfsvalidator/validator/DefaultFieldValidator.java
+++ b/core/src/main/java/org/mobilitydata/gtfsvalidator/validator/DefaultFieldValidator.java
@@ -104,7 +104,7 @@ public class DefaultFieldValidator implements GtfsFieldValidator {
     }
   }
 
-  public static boolean hasOnlyPrintableAscii(String s) {
+  static boolean hasOnlyPrintableAscii(String s) {
     for (int i = 0, n = s.length(); i < n; ++i) {
       if (!(s.charAt(i) >= 32 && s.charAt(i) < 127)) {
         return false;
--- a/core/src/main/java/org/mobilitydata/gtfsvalidator/validator/DefaultFieldValidator.java
+++ b/core/src/main/java/org/mobilitydata/gtfsvalidator/validator/DefaultFieldValidator.java
@@ -62,43 +62,36 @@ public class DefaultFieldValidator implements GtfsFieldValidator {
   }
 
   @Override
-  public boolean validateId(
-      String id, GtfsCellContext cellContext, NoticeContainer noticeContainer) {
+  public void validateId(String id, GtfsCellContext cellContext, NoticeContainer noticeContainer) {
     if (!hasOnlyPrintableAscii(id)) {
       noticeContainer.addValidationNotice(
           new NonAsciiOrNonPrintableCharNotice(
               cellContext.filename(), cellContext.csvRowNumber(), cellContext.fieldName(), id));
-      return false;
     }
-    return true;
   }
 
   @Override
-  public boolean validateUrl(
+  public void validateUrl(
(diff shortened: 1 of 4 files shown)


aababilov:
Example. Consider we want to lower InvalidPhoneNumberNotice from ERROR to WARNING in a custom implementation of GtfsFieldValidator. A valid code with the old interface must be:

```
public boolean validatePhoneNumber(
      String phoneNumber, GtfsCellContext cellContext, NoticeContainer noticeContainer) {
    if (!PhoneNumberUtil.getInstance()
        .isPossibleNumber(phoneNumber, RegionCode.valueOf(countryCode.getCountryCode()))) {
          new InvalidPhoneNumberNotice(
              ...
              SeverityLevel.WARNING));
    }
   return true;
  }
```

An invalid code may return still false but emit a WARNING.

```
public boolean validatePhoneNumber(
      String phoneNumber, GtfsCellContext cellContext, NoticeContainer noticeContainer) {
    if (!PhoneNumberUtil.getInstance()
        .isPossibleNumber(phoneNumber, RegionCode.valueOf(countryCode.getCountryCode()))) {
          new InvalidPhoneNumberNotice(
              ...
              SeverityLevel.WARNING));
      return false;   // THIS IS AN INCORRECT COPY-PASTE FROM THE UPSTREAM!
    }
   return true;
  }
```

If validatePhoneNumber returns void, then it is impossible to introduce a bug.

```
public void validatePhoneNumber(
      String phoneNumber, GtfsCellContext cellContext, NoticeContainer noticeContainer) {
    if (!PhoneNumberUtil.getInstance()
        .isPossibleNumber(phoneNumber, RegionCode.valueOf(countryCode.getCountryCode()))) {
          new InvalidPhoneNumberNotice(
              ...
              SeverityLevel.WARNING));
    }
    // Nothing to return here.
  }
```

core/src/main/java/org/mobilitydata/gtfsvalidator/validator/DefaultFieldValidator.java
```diff
@@ -108,12 +101,10 @@ public boolean validatePhoneNumber(
               cellContext.csvRowNumber(),
               cellContext.fieldName(),
               phoneNumber));
-      return false;
     }
-    return true;
   }
 
-  static boolean hasOnlyPrintableAscii(String s) {
+  public static boolean hasOnlyPrintableAscii(String s) {
```
barbeau, line 107: Why does this need to have it's visibility increased?

barbeau (commented):
Good catch @aababilov! One question in-line.

aababilov:
I will reuse that function in Google's own implementation of
Field.Validator.

Den tis 27 apr. 2021 08:01Sean Barbeau ***@***.***> skrev:

barbeau:
>I will reuse that function in Google's own implementation of Field.Validator.

Do you think it would make sense to move these types of static methods that may be reused in downstream implementations to their own utility class? Or would you prefer to leave it in the DefaultFieldValidator to make it clearer that you're using the default implementation?

aababilov:
I think it is acceptable for now. We can revisit that later.

Den tis 27 apr. 2021 08:52Sean Barbeau ***@***.***> skrev:

core/src/main/java/org/mobilitydata/gtfsvalidator/validator/DefaultFieldValidator.java
aababilov: Let's keep it package-private. That was unrelated.

barbeau (approved):
Thanks @aababilov!

PR #871, opened 2021-04-23
fix: Prevent NullPointerException when Throwable.getMessage returns null

aababilov:
Throwable.getMessage() can return null but ImmutableMap may not keep
nulls, so we store "" instead.

We pass an exception instance to the constructors for convenience.

Note that ThreadInterruptedError was not updated since we plan to remove
it.


aababilov:
@lionel-nj 

Updated to resolve merge conflicts.

barbeau (approved):
Thanks @aababilov!

PR #872, opened 2021-04-27
fix: Swap arguments of isAssignableFrom

aababilov:
Consider we want to inject a ChildCurrentDateTime class. This assignment
is invalid:

  ChildCurrentDateTime param = new CurrentDateTime();

It may be only the other way:

  CurrentDateTime param = new ChildCurrentDateTime();

See also the valid implementation:

```
  private static boolean isInjectableFromContext(Class<?> parameterType) {
    return parameterType.isAssignableFrom(CurrentDateTime.class)
        || parameterType.isAssignableFrom(CountryCode.class);
  }
```


barbeau (approved):
👍

PR #873, opened 2021-04-27
fix: Remove redundant SuppressWarnings

aababilov:
There are no unchecked type conversions in the annotated functions.


barbeau (approved):
👍

PR #874, opened 2021-04-28
feat: Catch CSV parse errors and add validation notices

aababilov:
If a CSV file failed to parse, there will be a clear user-visible
notice, e.g.:

```
        {
            "code": "csv_parsing_failed",
            "severity": "ERROR",
            "totalNotices": 1,
            "notices": [
                {
                    "filename": "stops.txt",
                    "charIndex": 278201,
                    "columnIndex": 2,
                    "lineIndex": 2148,
                    "message": "Length of parsed input (4097) exceeds the maximum number of characters defined in your parser settings (4096).",
                    "content": "..."
                }
            ]
         }
```

### DIFF
--- a/RULES.md
+++ b/RULES.md
@@ -68,6 +68,7 @@ Notices are split into three categories: `INFO`, `WARNING`, `ERROR`.
 | Name                                                                                                            	| Description                                                                                                                                                 	|
 |-----------------------------------------------------------------------------------------------------------------	|-------------------------------------------------------------------------------------------------------------------------------------------------------------	|
 | [`BlockTripsWithOverlappingStopTimesNotice`](#BlockTripsWithOverlappingStopTimesNotice)                         	| Block trips with overlapping stop times.                                                                                                                    	|
+| [`CsvParsingFailedNotice`](#CsvParsingFailedNotice)                                                             	| Parsing of a CSV file failed.                                                                                                                             	|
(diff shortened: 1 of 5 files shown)


aababilov:
Note that Univocity provides ExpandingCharAppender that can expand up to (Integer.MAX_VALUE - 8). We may switch to it in a separate PR if there is a use case for that (which I doubt because a CSV value longer than 4096 characters looks invalid to me).

lionel-nj (approved):
thanks for working on that @aababilov! Could you update `RULES.md` with this new notice description please? 

> Note that Univocity provides ExpandingCharAppender that can expand up to (Integer.MAX_VALUE - 8). We may switch to it in a separate PR if there is a use case for that (which I doubt because a CSV value longer than 4096 characters looks invalid to me).

@MobilityData/transit-specs could the GTFS specification add a requirement? "each cell value must be shorter than 4096 characters"? Unless there is a use case for as @aababilov suggested.

lionel-nj:
Also @maximearmstrong you should take a look, see how his new notice code impacts the MobilityDatabase.

aababilov:
> thanks for working on that @aababilov! Could you update `RULES.md` with this new notice description please?

Done. 

> @MobilityData/transit-specs could the GTFS specification add a requirement? "each cell value must be shorter than 4096 characters"? Unless there is a use case for as @aababilov suggested.

Good question. This requires an extra PR.

lionel-nj (approved):
Thanks!

timMillet:
> @MobilityData/transit-specs could the GTFS specification add a requirement? "each cell value must be shorter than 4096 characters"?

Issue created to keep track https://github.com/MobilityData/gtfs-tasks/issues/118

PR #875, opened 2021-04-28
chore: Remove unused methods and document GtfsTableContainer.TableStatus

aababilov:
Only isMissingFile() method is actually used in
MissingCalendarAndCalendarDateValidator.

If needed, users may call getTableStatus() method. In general,
validators do not need to handle most of the statuses. We may want to
merge INVALID_HEADERS and UNPARSABLE_ROWS to a single status
PARSING_FAILED in a separate PR.


lionel-nj (dismissed):
Thanks @aababilov! Opening an issue to merge  INVALID_HEADERS and UNPARSABLE_ROWS to a single status PARSING_FAILED.

lionel-nj (approved):
LGTM! thanks -- wanted to completely erase the review comment, but I cannot so here we are.

PR #876, opened 2021-04-28
feat: Treat --storage_directory as a directory

aababilov:
The old implementation treated --storage_directory as a target file name
instead of a directory. We bring validator behaviour to its
documentation.

createFromUrl now receives a Path instead of String for maximum
flexibility.

Validator does not remove storage directory if it already exists since it is 
a dangerous behaviour. Instead, it will throw and catch an IOException
and eventually store an IOError in `system_errors.json`.

ThreadInterruptedError is not used anymore and it was removed.

We use try-with-resources where possible.

We also rewrote unit tests for GtfsInput:

* tests for redirects are removed since we do not implement handling
  of redirects in Validator code;
* added a test for invalid URL to createFromUrlInMemory;
* made tests granular and avoided having multiple assertThat.


lionel-nj (changes requested):
Thanks @aababilov, a few comments below.

> ThreadInterruptedError is not used anymore and it was removed.

Could you update the documentation in `RULES.md` to reflect that please? 

> tests for redirects are removed since we do not implement handling
> of redirects in Validator code;

@barbeau is this advisable in the sense of 'should we effectively handle redirection'?

lionel-nj:
> Validator does not remove storage directory if it already exists since it is
> a dangerous behaviour. Instead, it will throw and catch an IOException
> and eventually store an IOError in system_errors.json.

Good catch! Thanks. 

> ThreadInterruptedError is not used anymore and it was removed.

@maximearmstrong PTAL for the MobilityDatabase

aababilov:
> Could you update the documentation in `RULES.md` to reflect that please?

Done.

> @barbeau is this advisable in the sense of 'should we effectively handle redirection'?

Lionel, as far as I know, your tests did not test redirects from the beginning. The https://octa.net/current/google_transit.zip URL is not a redirect, as far as I can tell.

lionel-nj:
> > Could you update the documentation in `RULES.md` to reflect that please?
> 
> Done.

Thanks :)

> 
> > @barbeau is this advisable in the sense of 'should we effectively handle redirection'?
> 
> Lionel, as far as I know, your tests did not test redirects from the beginning. The https://octa.net/current/google_transit.zip URL is not a redirect, as far as I can tell.
> 
> I also do not see a point in testing behaviour of an external library that we use to handle URLs.

Thanks for flagging that! I realize that is may be old code from `v.1.x` where we had to handle redirects https://github.com/MobilityData/gtfs-validator/pull/433. As you mentioned this is no longer relevant since the library handles that.

PR #889, opened 2021-05-05
feat: Export infinity and NaN values in JSON

aababilov:
Some notices may have special double values (NaN, Infinity, -Infinity)
in the context. If we do not allow their serialization, then an attempt
to store validation results ends up with a runtime exception.

Documentation for GSON library says the following.

Section 2.4 of <a href="http://www.ietf.org/rfc/rfc4627.txt">JSON specification</a> disallows
special double values (NaN, Infinity, -Infinity). However,
<a href="http://www.ecma-international.org/publications/files/ECMA-ST/Ecma-262.pdf">Javascript
specification</a> (see section 4.3.20, 4.3.22, 4.3.23) allows these values as valid Javascript
values. Moreover, most JavaScript engines will accept these special values in JSON without
problem. So, at a practical level, it makes sense to accept these values as valid JSON even
though JSON specification disallows them.

### DIFF
--- a/core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NoticeContainer.java
+++ b/core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NoticeContainer.java
@@ -34,7 +34,8 @@ import java.util.List;
  */
 public class NoticeContainer {
   private static final int MAX_EXPORTS_PER_NOTICE_TYPE = 100000;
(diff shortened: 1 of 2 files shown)


lionel-nj (approved):
Thanks @aababilov that will certainly have a positive impact on https://github.com/MobilityData/gtfs-validator/pull/848 where we deserialize the validation reports.

PR #894, opened 2021-05-05
feat: Return non-null for invalid strings and doubles

aababilov:
This is a performance improvement: this way we do not create a
temporary NoticeContainer object for each string value.

This affects the following functions:
* asLatitude
* asLongitude
* asUrl
* asEmail
* asPhoneNumber

These functions were returning null for invalid values just for
consistency with such functions as asLanguageCode or asTimezone that
are not able to return a valid Locale or ZoneId object for invalid field
values.

There is no practical need to return null for invalid values. A
TableLoader creates a separate NoticeContainer instance per each row and
if there are any errors for any cell, the whole row is considered
invalid.


aababilov:
Thanks for review!

PR #895, opened 2021-05-05
feat: Prevent OOM in NoticeContainer and speed up hasValidationErrors

aababilov:
Store up to 10 M validation notices. This is a measure to prevent OOM
in the rare case when each row in a large file (such as stop_times.txt
or shapes.txt) produces a notice. Since this case is rare, we just
introduce a total limit on the amount of notices instead of counting
amount of notices of each type.

Note that system errors are not limited since we don't expect to have
a lot of them.

Now we have a boolean field to flag if there are validation errors. This
way hasValidationErrors() now takes O(1) instead of O(n).

We also documented the methods and return unmodifiableList from getters.
We cannot use ImmutableList since the notice lists are always growing.


lionel-nj (approved):
Thanks for working on that! LGTM

aababilov:
Thanks for review!

PR #897, opened 2021-05-06
fix: Protect against start_date and end_date in wrong order

aababilov:
ServicePeriod constructor throws an exception for start and end date in
wrong order. There is a dedicated GtfsCalendarEndRangeValidator, so an
error is reported to the end user. This change avoids throwing an
exception in other validators that rely on CalendarUtil.


lionel-nj (approved):
Thanks Alexei! LGTM.

aababilov:
Thanks for review!

PR #898, opened 2021-05-06
feat: Treat empty files as parsed unsuccessfully

aababilov:
EmptyFileNotice is already an ERROR.

We have verified on 3K feeds at Google that none provides empty CSV file
(i.e. a file that contains no rows and no headers). We do not see a use
case for empty files in GTFS.

Since an empty file is an error, we do not invoke single-file validators
for it. Cross-file validations will not be invoked if any file is
completely empty.

In future we may merge EMPTY_FILE, INVALID_HEADERS and UNPARSABLE_ROWS
into a single constant.


tripd22 (approved):
LGTM.

aababilov:
Thanks, Darcy!

PR #901, opened 2021-05-11
feat: introduce -p CLI parameter to output beautified reports

### LINKED ISSUE #838: Add command-line parameter to output "pretty" JSON
My workflow when running the validator is typically:
1. Run the validator
2. Copy/paste the `report.json` file into a JSON formatter like https://jsonformatter.curiousconcept.com/ so it's readable (by default the JSON file is only one line without any whitespace). If the report file is big I'll open it in an application like NotePad++ and use the JSON formatter plugin.

**Proposed solution**

I'd like to add a command-line parameter (e.g., `-p` or `--pretty`) that would allow users to specify that they want readable JSON that can directly be viewed in any text editor without having to be formatted. Note that we want to keep the default to the more efficient single-line output for efficiency when using it primarily as a machine readable format between applications.

We're using GSON to output the JSON data, so looks like pretty print can be enabled with a single line of code when instantiating the `Gson` object:

```java
Gson gson = new GsonBuilder().setPrettyPrinting().create();
```

So this change would mainly entail implementing the command-line parameter and piping it through to the notice container.

Do nothing - keep copy/pasting the output to read it

lionel-nj:
closes #838 

**Summary:**

This PR provides support for an additional CLI parameter : `-p`/ `--pretty` which takes no value. This is used to bautfy the generated reports.

**Expected behavior:** 

Sample usage 
```
-u http://webapps.thebus.org/transitdata/Production/google_transit.zip -t 4 -p -o output
```
### `report.json`

Capture d’écran, le 2021-05-11 à 15 06 33

### `system_errors.json`

Capture d’écran, le 2021-05-11 à 15 06 41

### DIFF
--- a/core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NoticeContainer.java
+++ b/core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NoticeContainer.java
@@ -36,6 +36,8 @@ import java.util.List;
 public class NoticeContainer {
   private static final Gson DEFAULT_GSON =
       new GsonBuilder().serializeNulls().serializeSpecialFloatingPointValues().create();
+  private static final Gson PRETTY_GSON =
+      new GsonBuilder().setPrettyPrinting().serializeNulls().serializeSpecialFloatingPointValues().create();
 
   /** Limit on the amount of exported notices of the same type and severity. */
   private static final int MAX_EXPORTS_PER_NOTICE_TYPE = 100000;
@@ -100,13 +102,13 @@ public class NoticeContainer {
   }
 
   /** Exports all validation notices as JSON. */
-  public String exportValidationNotices() {
-    return exportJson(validationNotices);
+  public String exportValidationNotices(boolean isPretty) {
+    return exportJson(validationNotices, isPretty);
   }
 
   /** Exports all system errors as JSON. */
-  public String exportSystemErrors() {
-    return exportJson(systemErrors);
+  public String exportSystemErrors(boolean isPretty) {
+    return exportJson(systemErrors, isPretty);
   }
 
   /**
@@ -114,10 +116,11 @@ public class NoticeContainer {
    *
    * <p>Up to {@link #MAX_EXPORTS_PER_NOTICE_TYPE} is exported per each type+severity.
    */
-  public static <T extends Notice> String exportJson(List<T> notices) {
+  public static <T extends Notice> String exportJson(List<T> notices, boolean isPretty) {
     JsonObject root = new JsonObject();
     JsonArray jsonNotices = new JsonArray();
     root.add("notices", jsonNotices);
+    Gson GSON = isPretty ? PRETTY_GSON : DEFAULT_GSON;
 
     for (Collection<T> noticesOfType : groupNoticesByTypeAndSeverity(notices).asMap().values()) {
       JsonObject noticesOfTypeJson = new JsonObject();
@@ -135,11 +138,11 @@ public class NoticeContainer {
           // Do not export too many notices for this type.
           break;
         }
-        noticesArrayJson.add(DEFAULT_GSON.toJsonTree(notice.getContext()));
+        noticesArrayJson.add(GSON.toJsonTree(notice.getContext()));
       }
     }
 
-    return DEFAULT_GSON.toJson(root);
+    return GSON.toJson(root);
   }
 
   private static <T extends Notice> ListMultimap<String, T> groupNoticesByTypeAndSeverity(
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NoticeContainer.java
```diff
@@ -36,6 +36,8 @@
 public class NoticeContainer {
   private static final Gson DEFAULT_GSON =
       new GsonBuilder().serializeNulls().serializeSpecialFloatingPointValues().create();
+  private static final Gson PRETTY_GSON =
```
aababilov, line 39: We risk to have those Gson objects out of sync. Let's try this: 

  private static final Gson PRETTY_GSON = new GsonBuilder(DEFAULT_GSON).setPrettyPrinting().create();

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NoticeContainer.java
```diff
@@ -100,24 +102,25 @@ public boolean hasValidationErrors() {
...
    * Exports notices as JSON.
    *
    * <p>Up to {@link #MAX_EXPORTS_PER_NOTICE_TYPE} is exported per each type+severity.
    */
-  public static <T extends Notice> String exportJson(List<T> notices) {
+  public static <T extends Notice> String exportJson(List<T> notices, boolean isPretty) {
     JsonObject root = new JsonObject();
     JsonArray jsonNotices = new JsonArray();
     root.add("notices", jsonNotices);
+    Gson GSON = isPretty ? PRETTY_GSON : DEFAULT_GSON;
```
aababilov, line 123: This is not a static final, so it must be `gson` instead of `GSON`.

main/src/main/java/org/mobilitydata/gtfsvalidator/cli/Arguments.java
```diff
@@ -77,6 +77,11 @@
       help = true)
   private boolean help = false;
 
+  @Parameter(
+      names = {"-p", "--pretty"},
+      description = "Pretty json output")
+  private Boolean pretty = false;
```
aababilov, line 83: boolean - starting with lowercase.

I see that you systematically write Boolean with capital. Do you know why?

main/src/test/java/org/mobilitydata/gtfsvalidator/cli/ArgumentsTest.java
```diff
@@ -122,4 +124,41 @@ public void numThreadsShouldHaveDefaultValueIfNotProvided() {
     assertThat(underTest.getCountryCode()).matches("ca");
     assertThat(underTest.getNumThreads()).isEqualTo(1);
   }
+
```
aababilov, line 127: I am not sure that those tests add value because a single option without params is trivial. So I suggest to revert this file.

aababilov (changes requested):
Thanks for working on this!

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NoticeContainer.java
lionel-nj: True! Thanks

main/src/main/java/org/mobilitydata/gtfsvalidator/cli/Arguments.java
lionel-nj: A very bad habit of mine - I'll keep an eye open to not reproduce this in the future. Thanks Alexei!

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NoticeContainer.java
lionel-nj: Unfortunately `GsonBuilder(Gson gson)` is package-private and cannot be accessed from `NoticeContainer` class. What do you think of the following?
```java
  private static final Gson PRETTY_GSON = DEFAULT_GSON.newBuilder().setPrettyPrinting().create();
```

https://www.javadoc.io/doc/com.google.code.gson/gson/2.8.5/com/google/gson/Gson.html#newBuilder--

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NoticeContainer.java
```diff
@@ -55,6 +56,49 @@
   private final List<SystemError> systemErrors = new ArrayList<>();
   private boolean hasValidationErrors = false;
 
+  /**
```
aababilov, line 59: Please do not move large blocks of code in CLs that change logic. That hides your changes because it looks like you removed one function completely and added a new one.

aababilov (changes requested):
Good job! But please move the static functions back to their previous location.

aababilov (approved):
Very good, thank you!

PR #905, opened 2021-06-15
docs: update docs to clarify path to documentation for previous versions of the validator

lionel-nj:
**Summary:**

This PR clarifies the paths to documentation for previous versions of the validator
**Expected behavior:** 

No code change. 


barbeau (commented):
IMHO this is still a bit confusing. For example, if I were new to the repo I'd think "There is a v2.0.0 in the list, but the below README also says v2.0.0 in the documentation, so they're probably the same".

I think there needs to be some explanation of what *this* README is, in context of the previous releases.

Maybe something like:

>This README contains information for master branch of this project, which is under active development. If you'd like to view documentation for past releases of the project, see:
>* [v2.0.0](https://github.com/MobilityData/gtfs-validator/blob/v2.0.0/README.md)
>* [v1.4.0](https://github.com/MobilityData/gtfs-validator/blob/v1.4.0/README.md)

lionel-nj:
Thanks for the clarification @barbeau.

README.md
```diff
@@ -2,6 +2,10 @@
 
 A GTFS Schedule (static) [General Transit Feed Specification (GTFS)](https://gtfs.mobilitydata.org/spec/gtfs-schedule) feed validator
 
+This README contains information for master branch of this project, which is under active development. If you'd like to view documentation for past releases of the project, see:
+* [v1.4.0](https://github.com/MobilityData/gtfs-validator/blob/v1.4.0/README.md)
+* [v2.0.0](https://github.com/MobilityData/gtfs-validator/blob/v2.0.0/README.md)
+
 # Introduction
 This is a command-line tool written in Java that performs the following steps:
 1. Loads input GTFS zip file from a URL or disk
```
barbeau, line 11: In the "Run the app via command line" section of the README, the given command-line parameters won't work with the v2.0.0 release, right? Shouldn't we be referencing a snapshot JAR in the main README that's built from the master branch?

barbeau: Similar comment on the rest of the README that's still referencing v2.0.0

README.md
```diff
@@ -2,6 +2,10 @@
 
 A GTFS Schedule (static) [General Transit Feed Specification (GTFS)](https://gtfs.mobilitydata.org/spec/gtfs-schedule) feed validator
 
+This README contains information for master branch of this project, which is under active development. If you'd like to view documentation for past releases of the project, see:
+* [v1.4.0](https://github.com/MobilityData/gtfs-validator/blob/v1.4.0/README.md)
```
barbeau, line 6: Looking more closely at these two READMEs, neither actually have the completely correct information for the given release:
1. v1.4.0 docs say to "Download the latest gtfs-validator JAR (cli or web) file from our Releases page" and use `java -jar gtfs-validator-v1.3.0_cli.jar...`, which won't work
2. v2.0.0 docs say "If you're running a v1.x release JAR file you'll need Java 11, and can follow these instructions. The below instructions are for the master branch, which will be v2.0." And it doesn't actually point to where you can find the v2.0 release JAR. Also, the full command for `java -jar...` isn't mentioned.

Suggestions:
1. Update the RELEASE.md steps to include updating the README to reflect using the release JAR before doing the release so we don't make the same mistake in the future
2. Check out both the above release tags into new branches and edit the docs so they are correct (have the correct JAR link and command-line instructions) and commit and tag those (maybe `v1.4.0-docs` and `v2.0.0-docs` respectively), and point to those tags from here so the docs for how to use those past releases are correct.

README.md
```diff
@@ -11,16 +15,16 @@ This is a command-line tool written in Java that performs the following steps:
 # Run the app via command line
 ### Setup
 1. Install [Java 8 or higher](https://www.oracle.com/java/technologies/javase-downloads.html)
-1. Download [gtfs-validator-v2.0.0_cli.jar](https://github.com/MobilityData/gtfs-validator/releases/download/v2.0.0/gtfs-validator-v2.0.0_cli.jar)
+1. Download the latest snapshot jar of the validator from the artifacts generated by the [`Test Package Document`](https://github.com/MobilityData/gtfs-validator/actions/workflows/test_pack_doc.yml) workflow.
```
barbeau, line 18: A few issues here:
1. This link includes PR builds as well - we only want users downloading snapshots built from the master branch. So https://github.com/MobilityData/gtfs-validator/actions/workflows/test_pack_doc.yml?query=branch%3Amaster seems like a better link. But...
2. It's not clear to me where I get the JAR file from on this page. If you can link more directly to that location (preferrably the JAR itself) that would be very helpful - alternately, you'll need to provide more specific instructions here.

README.md
```diff
@@ -34,15 +38,15 @@ More detailed instructions are on our ["Usage"](/docs/USAGE.md) page.
 # Run the app using Docker
 ### Setup
 1. Download and install [Docker](https://docs.docker.com/get-started/)
-1. Pull the [latest Docker image for this project](https://github.com/orgs/MobilityData/packages/container/package/gtfs-validator). For example, `docker pull ghcr.io/mobilitydata/gtfs-validator:v2.0.0`.
+1. Pull the [latest Docker image for this project](https://github.com/orgs/MobilityData/packages/container/package/gtfs-validator). For example, `docker pull ghcr.io/mobilitydata/gtfs-validator:fix-debian-version-used-in-docker-image`.
```
barbeau, line 41: After this text changes I'm not sure it will be obvious what the user should run to get the latest image. I'd suggest providing more specific instructions that include copying the value show in the text box at the top of the https://github.com/orgs/MobilityData/packages/container/package/gtfs-validator page.

README.md
```diff
@@ -34,15 +38,15 @@ More detailed instructions are on our ["Usage"](/docs/USAGE.md) page.
...
 
 ### Run it
 
 #### For Mac and Linux
 
 To run the Docker image in a new container:
 
-`docker run -v /myDirectory:/theContainerDirectory -it ghcr.io/mobilitydata/gtfs-validator:v2.0.0`
+`docker run -v /myDirectory:/theContainerDirectory -it ghcr.io/mobilitydata/gtfs-validator:snapshot`
```
barbeau, line 49: Will `snapshot` always point to the latest docker image? If so, can we use that above in "Setup" section too?

README.md
barbeau: Actually, if we can just use `snapshot` instead here that would be greatly preferred (see below).

barbeau (commented):
Thanks @lionel-nj. More comments below.

I'd strongly suggest you have someone that's not familiar with the steps walk through the docs for each version (1.4.0, 2.0.0, and current master branch) after the next set of edits and try to complete the steps to make sure everything is accurate and understandable.

README.md
lionel-nj: At present, Docker images are only published at reslease or manually when needed (as we did for the [latest one](https://github.com/orgs/MobilityData/packages/container/gtfs-validator/2374496?tag=fix-debian-version-used-in-docker-image)) - Do you think that we should publish a new docker image for each snapshot version of the validator?

barbeau: >Do you think that we should publish a new docker image for each snapshot version of the validator?

Yes, I think that would be helpful if someone wants to pull in a bug fix before the release of the next version (i.e., in the case of fixing the debian version). We have the gtfs-realtime-validator set up to publish a new docker image on each commit to main branch.

lionel-nj: > Actually, if we can just use snapshot instead here that would be greatly preferred (see below).

Would following the format: snapshot-sha float our boat in this case?

README.md
lionel-nj: > This link includes PR builds as well - we only want users downloading snapshots built from the master branch. So https://github.com/MobilityData/gtfs-validator/actions/workflows/test_pack_doc.yml?query=branch%3Amaster seems like a better link. But...

True - I wanted to let the door open for someone who would like to download the jar from a different branch. But since this documentation is dedicated to the `master` branch, it makes sense to use the link you provided. Thanks!

lionel-nj: > It's not clear to me where I get the JAR file from on this page. If you can link more directly to that location (preferrably the JAR itself) that would be very helpful - alternately, you'll need to provide more specific instructions here.

I'll provide more detailed instructions since it does not appear to have a place to download these jars from a list of jars for example.

README.md
barbeau: Ideally you want something that's not going to change frequently so you don't need to keep updating the docs on every commit to the main branch.

See https://github.com/CUTR-at-USF/gtfs-realtime-validator#docker for example.

We publish the RT validator to Docker Hub and every commit gets the `latest` TAG, and Docker Hub always returns that when you do `docker pull cutrusf/gtfs-realtime-validator`.

README.md
lionel-nj: https://github.com/MobilityData/gtfs-validator/pull/913 provides support to publish a new docker image using the `latest` tag which will allow us to automatically point to it in the documentation. @barbeau  PTAL

> Will snapshot always point to the latest docker image? If so, can we use that above in "Setup" section too?

After the above is reviewed and merge to master - yes.

lionel-nj: Also, it will be "latest" instead of "snapshot".

README.md
lionel-nj: Please tell me if https://github.com/MobilityData/gtfs-validator/pull/905/commits/e776d1b751488a1048eae3c65f13da83fe42db4f clarifies this :)

lionel-nj:
@isabelle-dr https://github.com/MobilityData/gtfs-validator/pull/905/commits/e776d1b751488a1048eae3c65f13da83fe42db4f while working on documentation update, I included suggestions to fix https://github.com/MobilityData/gtfs-validator/issues/912, PTAL.

README.md
barbeau: Much better, thanks!

barbeau:
>@isabelle-dr while working on documentation update, I included suggestions to fix #912, PTAL at e776d1b .

I'd suggest nesting the 3 categories so it's clear they are all possible future rules:

* Possible future rules for:
    * [GTFS Reference](https://github.com/MobilityData/gtfs-validator/labels/Rules%20-%20GTFS%20Reference)
    * [GTFS Best Practices](https://github.com/MobilityData/gtfs-validator/labels/Rules%20-%20GTFS%20Best%20Practices)
    * [Common cases of bad data](https://github.com/MobilityData/gtfs-validator/labels/Rules%20-%20Common%20cases%20of%20bad%20data)

README.md
lionel-nj: > Update the RELEASE.md steps to include updating the README to reflect using the release JAR before doing the release so we don't make the same mistake in the future

Done in https://github.com/MobilityData/gtfs-validator/pull/905/commits/33de638eeb809cdd91ea359626350f48d5f45a43

lionel-nj: > Check out both the above release tags into new branches and edit the docs so they are correct (have the correct JAR link and command-line instructions) and commit and tag those (maybe v1.4.0-docs and v2.0.0-docs respectively), and point to those tags from here so the docs for how to use those past releases are correct.

👍🏾 two new PRs upcoming :)

lionel-nj: *v2.0.0: https://github.com/MobilityData/gtfs-validator/pull/915
*v1.4.0: https://github.com/MobilityData/gtfs-validator/pull/914

cc @kaiomi-i for more context :)

isabelle-dr:
Thank you, I added a comment proposing to switch the term "Common cases of bad data" with "Community rules". 
I think it represents better what it is and it aligns with the vision for this tool.

isabelle-dr:
@lionel-nj just one thing to change here so the title "Common cases of bad data" matches the label we chose "Community rule".
After this, I think we're good to merge this PR.

lionel-nj:
> @lionel-nj just one thing to change here so the title "Common cases of bad data" matches the label we chose "Community rule".
> After this, I think we're good to merge this PR.

Fixed in https://github.com/MobilityData/gtfs-validator/pull/905/commits/da7a817904554ebc1acd02d2a5b1039bc2d23069. Merging to master.

PR #917, opened 2021-06-24
feat: Reduce severity of SameNameAndDescriptionForRouteNotice to WARNING

aababilov:
A route that has the same route_desc and route_short_name or
route_long_name may be still served to the end user, so this is not an
error. In fact, many consumers may just ignore route_desc completely.

Numerous feeds that are consumed by Google do have this notice and they
are considered valid.

We also update the docs for the notice.


isabelle-dr:
The specification is currently being reviewed to become RFC 2119 compliant (see changes [here](https://github.com/google/transit/pull/277/files)).
The ambiguity in `route.route_desc` is planned on being solved by having:

> Must not be a duplicate of route_short_name or route_long_name.

If this does not make a dataset invalid (and triggers a WARNING in the validator), we should:
1) Modify the change planned to the spec in [PR227](https://github.com/google/transit/pull/277) to treat as a warning
2) Modify the validator to a warning to match the spec

@MobilityData/transit-specs thoughts?

aababilov:
I have posted a comment regarding route_desc on https://github.com/google/transit/pull/277

I suggest to merge this pull request #917 first because the current behaviour of Validator treats plenty of feeds as invalid. This makes more difficult adoption of the Validator by members of GTFS community.

isabelle-dr:
Hello, 

Thanks for your comment on google/transit!
I'm okay with merging this PR first @lionel-nj

lionel-nj (changes requested):
Thanks for working on that @aababilov! One small documentation change should be provided here: `SameNameAndDescriptionForRouteNotice` should be in the table of warnings instead of the table or errors in `RULES.md` (same thing for the descriptions) and `/docs/ NOTICES.md` :)

aababilov:
> Thanks for working on that @aababilov! One small documentation change should be provided here: `SameNameAndDescriptionForRouteNotice` should be in the table of warnings instead of the table or errors in `RULES.md` (same thing for the descriptions) and `/docs/ NOTICES.md` :)

Done. I think that would be great to develop a way for generating that doc from Java code.

lionel-nj (approved):
Thanks @aababilov!

Issue #899, opened 2021-05-06
Ability to dump validator notice schemas to json or yaml


machow:
## **Feature request**

**Is your feature request related to a problem? Please describe.**

Currently, the validator uses very clear schemas for structuring results internally. However, because it dumps results to a semi-structured JSON file, they're easy to read, but difficult to load into structured data sources (like SQL databases).

**Proposed solution**

Add an option to the CLI to dump a JSON file of notice schemas. Maybe something like this?:

```
{
    "invalidPhoneNumber": [
        {
            "name": "csvRowNumber",
            "type": "integer"
        },
        ...
    ],
    "suggestedExpirationDate": [
        {
            "name":  "localDate",
            "type": "record",
            "fields": [
                {"name": "year", "type": "integer"},
                ...
            ]
        }
    ],
    ...
}
```

Where the pattern here is...

```
{
    "CODE_NAME": [
        SCHEMA_ENTRY
    ],
}
```

Where SCHEMA_ENTRY is just copying [BigQuery's schema format](https://cloud.google.com/bigquery/docs/nested-repeated#example_schema)...

```
# note fields is optional, to describe nested records
{"name": string, "type": string, fields: [SCHEMA_ENTRY...]}
```

**Describe alternatives you've considered**

Currently, I'm just inferring the schema from a bunch of report data!

machow:
Hey--sorry for the wait. I was imagining it being generated once, and covering every possible notice. This way people could dump the schema once when setting up a database / dump it as a reference.

lionel-nj:
Hi @machow! We thought we could build such a schema using SQL (that is more generic) instead of BigQuery; would that meet your needs?

cc @barbeau

machow:
@lionel-nj I am not sure what that means--can you say more about what building a schema using SQL would look like?

For example, this is what a schema using SQL looks like...

```sql
CREATE TABLE table_name (
    column1 datatype,
    column2 datatype,
    column3 datatype,
   ....
);
```

These statements tend to be database / dialect dependent (esp. the types). You could dump it to json using type names that tools bridging across dialects (e.g. sqlalchemy) use. That would allow people to construct these schemas across databases.

barbeau:
@machow Yes, our internal discussions were about trying to generalize a solution to be usable beyond just BigQuery. We're open to suggestions for how exactly that is implemented to be most useful to users across many different database systems.

machow:
[...]

PR #922, opened 2021-07-07
chore: additional annotation for notice export

lionel-nj:
**Summary:**

This PR provides support to define a new annotation `NoticeExport`.  This annotation is to be used on notice constructor. This specifies the constructor to be considered while exporting notice information.

This relates to https://github.com/MobilityData/gtfs-validator/issues/899

**Expected behavior:** 

No code change for now. Same tests should pass.

### DIFF
--- a/core/src/main/java/org/mobilitydata/gtfsvalidator/annotation/NoticeExport.java
+++ b/core/src/main/java/org/mobilitydata/gtfsvalidator/annotation/NoticeExport.java
@@ -1,0 +1,7 @@
+package org.mobilitydata.gtfsvalidator.annotation;
+
+/**
+ * Annotation to be used on notice constructor. This specifies the constructor to be considered
+ * while exporting notice information.
+ */
+public @interface NoticeExport {}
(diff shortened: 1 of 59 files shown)


maximearmstrong (approved):
LGTM

core/src/main/java/org/mobilitydata/gtfsvalidator/annotation/NoticeExport.java
```diff
@@ -0,0 +1,7 @@
+package org.mobilitydata.gtfsvalidator.annotation;
+
+/**
+ * Annotation to be used on notice constructor. This specifies the constructor to be considered
+ * while exporting notice information.
+ */
+public @interface NoticeExport {}
```
barbeau, line 7: @lionel-nj One comment here - I would suggest renaming this annotation and providing a more detailed description regarding the exact use case for which it exists.

For example, I think `SchemaExport` may be a better name here, or `NoticeSchema`, or something with `schema` in it, to distinguish it from the export of the notices themselves. It's not clear by the name or description of the annotation what exactly the annotation is used for - I'd link to the class that processes it too when that exists.

PR #925, opened 2021-07-09
feat: export notice schema as json file

### LINKED ISSUE #899: Ability to dump validator notice schemas to json or yaml

lionel-nj:
closes #899 

**Summary:**

This PR provides support to export notice schema as a json file.

**Expected behavior:** 

When providing the additional CLI parameter `-n` `--export_notices_schema`: a json file named `notice_schema.json` should be created and stored in the folder specified by the `output_base` CLI parameter).

See gist linked https://gist.github.com/lionel-nj/be94d4cb18af68e743408c5c6c5129ab.


lionel-nj:
@barbeau following our discussion of Thursday: this could probably be written in a more modular way (to consider `.proto` files) - I'd be glad to have your feedback on this first version and iterate to improve it.

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NoticeContainer.java
```diff
@@ -152,4 +174,195 @@ public String exportSystemErrors(boolean isPretty) {
...
+  private static JsonArray extractNoticeProperties(Class<?> validationNoticeSubClass) {
+    Constructor<?> constructor = getAnnotatedConstructor(validationNoticeSubClass);
+    JsonArray parametersAsJsonArray = new JsonArray();
+    Arrays.stream(constructor.getParameters()).forEach(parameter -> {
+      JsonObject parameterDetails = new JsonObject();
+      parameterDetails.addProperty(NAME, parameter.getName());
+        parameterDetails.addProperty(TYPE, mapDataType(parameter));
+      parametersAsJsonArray.add(parameterDetails);
+    });
+    if (!noticeConstructorHasSeverityLevelParameter(constructor)) {
```
maximearmstrong, line 296: I'm not quite sure what is going on here. From my understanding, if a constructor has the severity level parameter, then you don't enter here, but in `NoticeContainerTest.java`, it looks like `SomeTestValidationNotice.java` has the severity level parameter in its constructor, but the property `{"SeverityLevel": "BLOB"}` is added to the notices schema (so it would have passed the condition at line 296). Am I missing something?

maximearmstrong (changes requested):

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NoticeContainer.java
lionel-nj: By default, each `Notice` has a `SeverityLevel` whether it is explicitly passed in the constructor or not. Since we base the schema elaboration on the `Notice` constructor, we have to take this into account and it is the rationale behind this `if` statement.

lionel-nj: Trying to be more clear - let's say that we elaborate the schema based on `SomeTestValidationNotice.java` and `EmptyFileNotice.java`: 

The resulting schema should specify that the two notices contain a `SeverityLevel` object - however, as you mentioned `SomeTestValidationNotice.java` already has `SeverityLevel` in its constructor parameter -> nothing to do here. For `EmptyFileNotice.java` we have to add it manually; hence the extra step provided in the`if` statement.

maximearmstrong: All good with this, but then why is the property `{"SeverityLevel": "BLOB"}` added to the notices schema for `SomeTestValidationNotice.java`?  From my understanding, it should not enter the `if` statement.

lionel-nj: Since `SeverityLevel` is one of the parameters of `SomeTestValidationNotice`'s constructor, it is added to the schema when evaluating lines 290-295
 https://github.com/MobilityData/gtfs-validator/blob/0b44560551b09033498d22ee35054b31a2edb114/core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NoticeContainer.java#L290. 

> From my understanding, it should not enter the if statement.
Absolutely, in this case the `if` statement is not entered.

lionel-nj:
@maximearmstrong PTAL, I modified the implementation regarding the suggestions you made during our last discussion.

barbeau (commented):
Thanks @lionel-nj - some preliminary feedback is below. I ran into some other issues while reviewing that I opened other issues for, so I ran out of time before I could do a full review. I'd like to look at this more, but here's my thoughts so far.

main/src/main/java/org/mobilitydata/gtfsvalidator/cli/Arguments.java
```diff
@@ -82,6 +84,11 @@
       description = "Pretty json output")
   private boolean pretty = false;
 
+  @Parameter(
```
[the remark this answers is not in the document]
lionel-nj: As of now yes, but modifying this to align with your suggestion. Thanks :)

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NoticeContainer.java
```diff
@@ -152,4 +174,196 @@ public String exportSystemErrors(boolean isPretty) {
...
+   * Exports notices information as a json file.
+   *
+   * @param isPretty will beautify the output if set to true
+   * @param noticePackageName the name of the package that contains the notices
+   * @param validatorPackageName the name of the package that contains the {@code GtfsValidator}s
+   * @return the stringified json file that contains information about all {@code ValidationNotice}s
+   * @throws IOException if the attempt to read class path resources (jar files or directories)
+   * failed.
+   * */
+  public static String exportNoticesSchema(boolean isPretty, String noticePackageName,
```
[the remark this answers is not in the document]
lionel-nj: Good catch! See https://github.com/MobilityData/gtfs-validator/pull/925/commits/1eab42ba9ae4c8034dbddf2a6b26ff5e2d4d4e48 for the modification I provided to this extent.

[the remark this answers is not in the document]
maximearmstrong: Should the classes be passed here instead of strings, then extract the name in the method? That was my understanding

core/src/test/java/org/mobilitydata/gtfsvalidator/notice/sample/AnotherTestValidationNotice.java
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * {@code ValidationNotice} defined for tests purpose.
+ */
+public class AnotherTestValidationNotice extends ValidationNotice {
+
+  @SchemaExport
+  public AnotherTestValidationNotice(String filename, long csvRowNumber, String fieldName,
+      Object fieldValue, double otherFieldValue, GtfsDate sampleDate, GtfsTime sampleTime,
+      GtfsColor sampleColor) {
+    super(new ImmutableMap.Builder<String, Object>()
```
maximearmstrong, line 36: Is there a reason why you are calling a `new ImmutableMap.Builder<>()` here instead of a `ImmutableMap.of()` like in `SomeTestValidationNotice.java` and `SampleTestValidator.java`? I would keep it this way if there is a reason behind this choice, but if it's arbitrary I'd change it to keep the consistency between the sample files.

main/src/main/java/org/mobilitydata/gtfsvalidator/cli/Main.java
```diff
@@ -56,6 +59,10 @@ public static void main(String[] argv) {
           "⚠️ Note that parameters marked with an asterisk (*) in the help menu are mandatory.");
       return;
     }
+    if (args.getExportNoticeSchema()) {
```
maximearmstrong, line 62: I agree with [this comment](https://github.com/MobilityData/gtfs-validator/pull/925#discussion_r675050396), but this modification removes the ability to extract the schema when a dataset is being validated. Is this what we want? Could we do
```
if (args.getExportNoticeSchema()) {
    exportNoticeSchema(args);
    if(hasNoOtherArgs(args)) { 
        return; 
    }
}
```
or something like that?

maximearmstrong (commented):
Some questions and comments inline before approval :)

main/src/main/java/org/mobilitydata/gtfsvalidator/cli/Main.java
lionel-nj: Makes sense to me! What do you think about adding a new CLI parameter `--abort_after_schema_extraction, -a` to represent this behavior? The default value would be set to `true`.

core/src/test/java/org/mobilitydata/gtfsvalidator/notice/sample/AnotherTestValidationNotice.java
lionel-nj: As discussed, for some reason the method `ImmutableMap.of()` does not allow to create a map with `GtfsDate`, or `GtfsTime` values, hence the recourse to `new ImmutableMap.Builder<>()`. For the sakes of consistency, I'll change the constructors (for `SomeTestValidationNotice `, `AnotherTestValidationNotice` and `NestedTestValidatorNotice`) to use `new ImmutableMap.Builder<>()`.

main/src/main/java/org/mobilitydata/gtfsvalidator/cli/Main.java
lionel-nj: edit: default value is set to `false` (see https://github.com/MobilityData/gtfs-validator/pull/925/commits/6e97a97e038720fb700bcb17337746bcbacd02c5)

docs/USAGE.md
```diff
@@ -9,6 +9,7 @@
 
 | Short name 	| Long name                     	| required? 	| Description                                                                                                               	|
 |------------	|-------------------------------	|-----------	|---------------------------------------------------------------------------------------------------------------------------	|
+| `-a`       	| `--abort_after_schema_export`    	| Optional  	| Aborts the application process.                                                                   	|
```
maximearmstrong, line 12: I'd remove the "s" to "Aborts" here to be consistent with the other descriptions.
```suggestion
| `-a`       	| `--abort_after_schema_export`    	| Optional  	| Abort the application process.                                                                   	|
```

maximearmstrong (approved):
One small change, then LGTM :)

core/src/test/java/org/mobilitydata/gtfsvalidator/notice/NoticeContainerTest.java
```diff
@@ -165,4 +168,87 @@ public void addAll() {
     assertThat(c1.getValidationNotices()).containsExactly(n1, n2);
     assertThat(c1.getSystemErrors()).containsExactly(e1, e2);
   }
+
+  @Test
+  public void exportNoticesSchema() throws IOException {
```
aababilov, line 173: This test is unstable, it fails on my machine. Did you consider to run it several times?


```
diff:
    @@ -1,5 +1,5 @@
     {
    -  "AnotherTestValidationNotice": [
    +  "SomeTestValidationNotice": [
         {
           "name": "filename",
           "type": "VARCHAR"
    @@ -17,27 +17,11 @@
           "type": "BLOB"
         },
         {
    -      "name": "otherFieldValue",
    -      "type": "DOUBLE"
    -    },
    -    {
    -      "name": "sampleDate",
    +      "name": "severityLevel",
           "type": "BLOB"
    -    },
    -    {
    -      "name": "sampleTime",
    -      "type": "BLOB"
    -    },
    -    {
    -      "name": "sampleColor",
    -      "type": "VARCHAR"
    -    },
    -    {
    -      "name": "SeverityLevel",
    -      "type": "BLOB"
         }
       ],
    -  "SomeTestValidationNotice": [
    +  "AnotherTestValidationNotice": [
         {
           "name": "filename",
           "type": "VARCHAR"
    @@ -55,8 +39,24 @@
           "type": "BLOB"
         },
         {
    -      "name": "severityLevel",
    +      "name": "otherFieldValue",
    +      "type": "DOUBLE"
    +    },
    +    {
    +      "name": "sampleDate",
           "type": "BLOB"
    +    },
    +    {
    +      "name": "sampleTime",
    +      "type": "BLOB"
    +    },
    +    {
    +      "name": "sampleColor",
    +      "type": "VARCHAR"
    +    },
    +    {
    +      "name": "SeverityLevel",
    +      "type": "BLOB"
         }
       ],
       "NestedTestValidatorNotice": [
	at org.mobilitydata.gtfsvalidator.notice.NoticeContainerTest.exportNoticesSchema(NoticeContainerTest.java:177)
```

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NoticeContainer.java
```diff
@@ -152,4 +173,195 @@ public String exportSystemErrors(boolean isPretty) {
     }
     return noticesByType;
   }
+
```
aababilov, line 176: The added code does not seem to be closely related to NoticeContainer - which is a container for notices, not an RTTI machine for schema generation. Please create a new class for the new code.

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NoticeContainer.java
```diff
@@ -152,4 +173,195 @@ public String exportSystemErrors(boolean isPretty) {
...
+   * {@code  NoticeExport}.
+   *
+   * @param validationNoticeSubClass the {@code ValidationNotice} sub class to extract information
+   *                                from
+   * @return a {@code JsonArray} that contains information about the type of each parameter of
+   * said {@code ValidationNotice} using the constructor of the class that is annotated by
+   * {@code NoticeExport}.
+   */
+  private static JsonArray extractNoticeProperties(Class<?> validationNoticeSubClass) {
+    Constructor<?> constructor = getAnnotatedConstructor(validationNoticeSubClass);
```
aababilov, line 286: This variable is used only once and it can be inlined.

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NoticeContainer.java
```diff
@@ -152,4 +173,195 @@ public String exportSystemErrors(boolean isPretty) {
...
+        ClassPath.from(ClassLoader.getSystemClassLoader()).getTopLevelClasses(noticePackageName)) {
+      clazz = noticeClass.load();
+      if (ValidationNotice.class.isAssignableFrom(clazz)) {
+        // do not iterate over ValidationNotice.java which is the base class for all validation
+        // notices
+        if (clazz.equals(ValidationNotice.class)){
+          continue;
+        }
+        toReturn.add(
+            noticeClass.getSimpleName(),
```
aababilov, line 241: This differs from the error code (e.g., NumberOutOfRangeNotice vs number_out_of_range). Would you like to be consistent? Also below (however, I recommend to avoid code duplication and to add a function that processes a notice class).

aababilov (changes requested):
Thanks! Some first comments.

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NoticeContainer.java
```diff
@@ -152,4 +173,195 @@ public String exportSystemErrors(boolean isPretty) {
...
+    JsonArray parametersAsJsonArray = new JsonArray();
+    Arrays.stream(constructor.getParameters()).forEach(parameter -> {
+      JsonObject parameterDetails = new JsonObject();
+      parameterDetails.addProperty(JSON_KEY_NAME, parameter.getName());
+        parameterDetails.addProperty(JSON_KEY_TYPE, mapDataType(parameter));
+      parametersAsJsonArray.add(parameterDetails);
+    });
+    if (!hasSeverityLevelParameter(constructor)) {
+      JsonObject severityDetails = new JsonObject();
+      severityDetails.addProperty(JSON_KEY_NAME, CLASS_SIMPLE_NAME_SEVERITY_LEVEL);
```
aababilov, line 296: CLASS_SIMPLE_NAME_SEVERITY_LEVEL is `"SeverityLevel"` while the param is `"severityLevel"` (starts from lowercase).

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NoticeContainer.java
```diff
@@ -152,4 +173,195 @@ public String exportSystemErrors(boolean isPretty) {
...
+   * constructor for schema export i.e. no constructor for the given class ses {@code NoticeExport}
+   * annotation.
+   *
+   * @param clazz the class to extract the annotated constructor from
+   * @return the annotated constructor of a {@code ValidationNotice} subclass.
+   * Throws {@code AnnotationFormatError} if the {@code ValidationNotice} subclass does not define a
+   * constructor for schema export i.e. no constructor for the given class uses {@code NoticeExport}
+   * annotation.
+   */
+  private static Constructor<?> getAnnotatedConstructor(Class<?> clazz) {
```
aababilov, line 357: Please support more than one constructor with `@SchemaExport` annotation.

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NoticeContainer.java
```diff
@@ -152,4 +173,195 @@ public String exportSystemErrors(boolean isPretty) {
...
+    Arrays.stream(constructor.getParameters()).forEach(parameter -> {
+      JsonObject parameterDetails = new JsonObject();
+      parameterDetails.addProperty(JSON_KEY_NAME, parameter.getName());
+        parameterDetails.addProperty(JSON_KEY_TYPE, mapDataType(parameter));
+      parametersAsJsonArray.add(parameterDetails);
+    });
+    if (!hasSeverityLevelParameter(constructor)) {
+      JsonObject severityDetails = new JsonObject();
+      severityDetails.addProperty(JSON_KEY_NAME, CLASS_SIMPLE_NAME_SEVERITY_LEVEL);
+      severityDetails.addProperty(JSON_KEY_TYPE, BLOB_DATA_TYPE);
```
aababilov, line 297: A severityLevel is a string from a fixed list (WARNING, ERROR, INFO). Why are you calling it BLOB?

core/src/main/java/org/mobilitydata/gtfsvalidator/annotation/SchemaExport.java
```diff
@@ -15,8 +15,22 @@
...
-public @interface SchemaExport {}
+@Retention(RetentionPolicy.RUNTIME)
+public @interface SchemaExport {
+  String JSON_KEY_NAME = "name";
+  String JSON_KEY_TYPE = "type";
+  String BLOB_DATA_TYPE = "BLOB";
+  String VARCHAR_DATA_TYPE = "VARCHAR";
+  String BIGINT_DATA_TYPE = "BIGINT";
+  String INTEGER_DATA_TYPE = "INTEGER";
+  String CLASS_SIMPLE_NAME_SEVERITY_LEVEL = "SeverityLevel";
```
aababilov, line 34: What are those constants doing in this class? Can they be moved to a place where they are actually used?

aababilov (changes requested):

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NoticeContainer.java
lionel-nj: It's true, I was considering the type of the parameter passed to the Notice constructor instead of the type of the element that ends up in the final notice report. Modifying this.

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NoticeContainer.java
lionel-nj: It is actually used one more time in the subsequent lines.

core/src/main/java/org/mobilitydata/gtfsvalidator/annotation/SchemaExport.java
lionel-nj: I was addressing [this review comment](https://github.com/MobilityData/gtfs-validator/pull/925#discussion_r674420430). Will be moved to `NoticeSchemaGenerator.java` after creation.

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NoticeContainer.java
lionel-nj: In this case, how would you make the difference between the two constructors for the same notice in the JSON schema?

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NoticeContainer.java
```diff
@@ -152,4 +173,195 @@ public String exportSystemErrors(boolean isPretty) {
...
+   * @param validationNoticeSubClass the {@code ValidationNotice} sub class to extract information
+   *                                from
+   * @return a {@code JsonArray} that contains information about the type of each parameter of
+   * said {@code ValidationNotice} using the constructor of the class that is annotated by
+   * {@code NoticeExport}.
+   */
+  private static JsonArray extractNoticeProperties(Class<?> validationNoticeSubClass) {
+    Constructor<?> constructor = getAnnotatedConstructor(validationNoticeSubClass);
+    JsonArray parametersAsJsonArray = new JsonArray();
+    Arrays.stream(constructor.getParameters()).forEach(parameter -> {
```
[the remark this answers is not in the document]
lionel-nj: It seemed more straightforward while writing the code - however it might not be for someone not so familiar with the implementation. Will remove as suggested.

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NoticeContainer.java
lionel-nj: Will push the changes on this branch after clarifying this last point :)

aababilov: All params from all constructors annotated with `@SchemaExport` should be aggregated. Consider:

```
@SchemaExport
MyCoolNotice(long csvRowNumber, String routeId) {}
@SchemaExport
MyCoolNotice(long csvRowNumber, String tripId) {}
```

aababilov:
Now I am thinking if using Map for the notice context was the best solution. An alternative is to introduce usual member fields and then rely on RTTI to discover them for JSON export. Something like that:

```
public class InvalidIntegerNotice extends ValidationNotice {
  private String filename;
  private long csvRowNumber;
  private String fieldName;
  private String fieldValue;

  @SchemaExport
  public InvalidIntegerNotice(
      String filename, long csvRowNumber, String fieldName, String fieldValue) {
    super(SeverityLevel.ERROR);
    this.filename = filename;
    this.csvRowNumber = csvRowNumber;
    this.fieldName = fieldName;
    this.fieldValue = fieldValue;
  }
}
```

Logically optional fields can be initialized with `null`.

core/src/test/java/org/mobilitydata/gtfsvalidator/notice/NoticeContainerTest.java
lionel-nj: I modified the code so that the keys of the json file are sorted alphabetically. I ran it several times on my machine and it seems to work. Could you confirm the new behavior on your side please @aababilov?

lionel-nj:
> Now I am thinking if using Map for the notice context was the best solution. An alternative is to introduce usual member fields and then rely on RTTI to discover them for JSON export.

Do you see that constructor used in addition to the constructors that use the map for `context`, or as a replacement? For the validation report exports, the contexts are very handy and I think we could keep them...but I am open to rethink the process if needed though :)

main/src/main/java/org/mobilitydata/gtfsvalidator/cli/Main.java
barbeau: IMHO adding yet another command-line parameter for this behavior is confusing. As a user, I would expect that if I provide the parameter to dump the schema it dumps the schema, if I provide the validation parameters (in particular `input`) then it validates the dataset, and if I provide both it does both.

aababilov:
I am considering to give up the `Map context` and switch to usual fields. This gives us the following advantages.

1. Each field gets usual Java documentation. This is more convenient and reliable than put all docs in a single `NOTICES.md` file (see https://github.com/MobilityData/gtfs-validator/issues/943).
2. JSON schema can be generated based on class fields, so there is no need to provide additional constructors with `@SchemaExport`.

We need to estimate what are the costs of the proposed alternative.

1. Obviously, refactoring. Not a big deal.
2. Which alternative consumes more memory: Map or fields?
3. There will be some changes in the speed of exporting JSON via RTTI but I would say they are negligible, especially in compare to the total time of feed validation.

main/src/main/java/org/mobilitydata/gtfsvalidator/cli/Main.java
lionel-nj: Changed in https://github.com/MobilityData/gtfs-validator/pull/925/commits/3f591e2725b64f412907c4e2df7a6c05b9439036.

lionel-nj:
> 1. Each field gets usual Java documentation. This is more convenient and reliable than put all docs in a single `NOTICES.md` file (see https://github.com/MobilityData/gtfs-validator/issues/943).
> 2. JSON schema can be generated based on class fields, so there is no need to provide additional constructors with `@SchemaExport`.

Indeed, that would overall simplify the documentation tasks and the work within this PR. 

> Which alternative consumes more memory: Map or fields?

Do you have documentation that you suggest reading, or would this be an empirical measurement?

aababilov:
> > Which alternative consumes more memory: Map or fields?
> 
> Do you have documentation that you suggest reading, or would this be an empirical measurement?

I tried https://github.com/stokito/IdeaJol and it nicely shows the actual consumption.

tl;dr Fields consume *less* memory than Maps.

The current "Map" approach uses two fields in the `Notice` class:
* `Map<String, Object> context` - 4 bytes for the pointer
* `SeverityLevel severityLevel` - 4 bytes for the enum

The map is an instance of `RegularImmutableMap` that immediately consumes 40 bytes for its object header (12 bytes) and 7 fields (4 bytes per each: entrySet, keySet, values, multimapView, mask, entries, table).
* `table` is an array of `ImmutableMapEntry` that consume 24 bytes per item
* `entries` is an array of `SimpleImmutableEntry` that consume 24 bytes per item

So, we have at least 48 bytes per each field and also a fixed 44 bytes overhead for the `RegularImmutableMap`.

Let's compare memory consumption for small and large notice classes. We do not take `SeverityLevel` bytes into account.

| Notice class  | # fields | Map bytes | Fields bytes |
|----|----|---|
| BlockTripsWithOverlappingStopTimesNotice | 8 |  424 | 40 |
| InvalidTimeNotice | 4 | 232 | 20 |
| MissingRequiredFileNotice | 1 | 88 | 4 |

Note that map also keeps key names as a separate string objects that also consume space. If we switch to fields, we won't consume them.

aababilov:
Have I convinced you that we need fields? :)

aababilov:
@barbeau @lionel-nj @maximearmstrong 

If you agree that we should use fields instead of map, then I am proposing the following plan.

Pull request 1. Convert all notice classes to fields instead of maps. I will do that.
Pull request 2. Add export of JSON schema based on RTTI. I will do that.
Pull request 3. Document all notice fields based on NOTICES.md. @lionel-nj Can I ask you to do that?

lionel-nj:
Thanks for the explanation and the reference to the IntelliJ plugin : will definitely reuse that in the future👀

> Have I convinced you that we need fields? :)

That looks like a good idea to me!

> Pull request 3. Document all notice fields based on NOTICES.md. @lionel-nj Can I ask you to do that?

👍🏾

aababilov:
It looks like this PR is still using constructor parameters. Unfortunately, this won't work with all notice classes. Please let me prepare my code that uses reflections.

aababilov:
Please see my https://github.com/MobilityData/gtfs-validator/pull/951

core/src/test/java/org/mobilitydata/gtfsvalidator/notice/NoticeContainerTest.java
aababilov: Thanks for fixing!

lionel-nj:
Thanks for #951 @aababilov - integrated the changes provided to this PR, does that seem alright to you? @maximearmstrong @barbeau

main/build.gradle
```diff
@@ -77,7 +77,6 @@ dependencies {
     implementation 'org.locationtech.spatial4j:spatial4j:0.7'
     testImplementation group: 'junit', name: 'junit', version: '4.13'
     testImplementation "com.google.truth:truth:1.0.1"
-    testImplementation 'com.fasterxml.jackson.core:jackson-databind:2.10.0'
```
aababilov, line 80: Nice catch! Let's fully remove it: https://github.com/MobilityData/gtfs-validator/pull/955

main/src/main/java/org/mobilitydata/gtfsvalidator/cli/Main.java
```diff
@@ -171,4 +180,19 @@ private static void exportReport(final NoticeContainer noticeContainer, final Ar
       logger.atSevere().withCause(e).log("Cannot store report files");
     }
   }
+
+  private static void exportNoticeSchema(final Arguments args) {
+    new File(args.getOutputBase()).mkdirs();
+          Gson gson = createGson(args.getPretty());
```
aababilov, line 186: This is an experiment that the code will be formatted before merging this PR, right?

aababilov (approved):
Thanks a lot! Very nice.

main/src/main/java/org/mobilitydata/gtfsvalidator/cli/Main.java
lionel-nj: Yes!

PR #933, opened 2021-07-22
chore: add springboot module for deployment on k8s

lionel-nj:
relates to #734 

**Summary:**

This PR provides support to deploy the `gtfs-validator` to Kubernetes. This comes as a required step for the acceptance tests to be up and running in #848 

**Expected behavior:** 
- [x] Gradle should be updated to version 6.8
- [x]  CLI Application should be packaged and build as usual.
- [x]  Spring Boot Application should be packaged and build successfully.
- [x] CLI app jar should be persisted
- [x] jar dedicated to the Spring Boot layer should be persisted
- [ ] All tests reports should be collected in `build/reports/allTests`
- [x] A Docker image should be generated for the new jar. 
- [x] Said Docker image should be published on Google Artifacts Registry and be usable by GKE.


lionel-nj:
# Bug description

## CLI entry point

- 🆗 Console output 
Capture d’écran, le 2021-07-22 à 13 10 32

- 🆗 [Validation report](https://gist.github.com/lionel-nj/7265f2cf40347afde626428f21ee02f3)


## Spring Boot entry point
- ❌ Console output when requesting `http://localhost:8080/url=http://webapps.thebus.org/transitdata/Production/google_transit.zip`: 
Capture d’écran, le 2021-07-22 à 12 51 55
- ❌ output: no output, the process is not killed - no validation report is generated.

# Expected behavior
- When requesting `http://localhost:8080/url=http://webapps.thebus.org/transitdata/Production/google_transit.zip`, the validation report should be generated as it is usually done by the CLI application.
- The console output should be similar to the one observed when running the CLI application

lionel-nj:
Instructions to run the springboot entry point jar: `java -jar gtfs-validator*_springboot.jar` from `main/springboot/build/libs`

opyh:
For somebody without the resources to reproduce the issue locally, it might be helpful to see the first log entries of the failing case, not something in the middle of the logs.

Could you pipe the logs into a file and upload them somewhere? The first few bytes might be hint at encoding issues, for example. And everything that happens before the http debug output is probably most interesting.

Another idea (just a wild guess): The logs might be correct, but the console output is very slow, so you didn't get to the end result yet. What happens if you switch off the http output debug logging?

lionel-nj:
Hi @opyh thanks for helping the gtfs-validator moving forward. 

> Could you pipe the logs into a file and upload them somewhere? The first few bytes might hint at encoding issues, for example. And everything that happens before the http debug output is probably most interesting.

Absolutely, you will find the Iogs attached [here](https://gist.github.com/lionel-nj/9b87b3f8cf5d57f4fb282c58973d8219).

Here is the command used to export these logs: 

```
 java -jar main/springboot/build/libs/gtfs-validator-null_springboot.jar
```
... and the HTTP request: 
```
http://localhost:8080/?url=http://webapps.thebus.org/transitdata/Production/google_transit.zip&country_code=us
```

> Another idea (just a wild guess): The logs might be correct, but the console output is very slow, so you didn't get to the end result yet. What happens if you switch off the http output debug logging?

1. Adding the following line to `main/springboot/resources/application.properties` did not succeed in turning off the debug output. 

```
logging.level.org.springframework=INFO
```
_adapted from https://stackoverflow.com/questions/35353869/how-turn-off-debug-log-messages-in-spring-boot_


2. However the [logs referred to earlier ](https://gist.github.com/lionel-nj/9b87b3f8cf5d57f4fb282c58973d8219) indicate that the console output is very slow as you suggested

3. The whole validation is actually executed. The validation report can also be consulted in the `./output` folder. 

`(1)+(2)+(3)` makes me believe that this might be related to the logging libraries that are used throughout the project (Flogger, and log4j used by Spring boot ).

lionel-nj:
As of today we could:
- create a kubernetes cluster
- deploy the application to the said cluster

Next steps:
- store and retrieve data generated while running the gtfs-validator in the different clusters
- make sure that the machines are correctly automatically scaled (horizontally and vertically)
- integrate this PR to #848 

FYI @isabelle-dr, @maximearmstrong, @barbeau.

.github/workflows/test_pack_doc.yml
```diff
@@ -103,7 +104,12 @@ jobs:
         uses: actions/upload-artifact@v2
         with:
           name: Application - cli executable .jar -- ${{ steps.prep.outputs.versionTag }}
-          path: main/build/libs/gtfs-validator-${{ steps.prep.outputs.versionTag }}_cli.jar
+          path: main/cli/build/libs/gtfs-validator-${{ steps.prep.outputs.versionTag }}_cli.jar
+      - name: Persist spirng boot app jar
```
f8full: Typo

lionel-nj: Good catch! Thanks :)

lionel-nj: Updated in one of the earlier commits.

lionel-nj:
@barbeau PTAL

lionel-nj:
This PR depends on https://github.com/MobilityData/gtfs-validator/pull/962

lionel-nj:
This PR depends also on https://github.com/MobilityData/gtfs-validator/pull/963

main/springboot/build.gradle
```diff
@@ -0,0 +1,94 @@
...
+
+sourceCompatibility = JavaVersion.VERSION_1_8
+mainClassName = 'org.mobilitydata.gtfsvalidator.springboot.Main'
+
+repositories {
+    mavenCentral()
+}
+
+dependencies {
+    compile 'com.google.cloud:google-cloud-storage'
```
barbeau, line 60: Should use `implementation` instead of `compile`, and also needs a version number.

main/springboot/build.gradle
```diff
@@ -0,0 +1,94 @@
...
+    implementation project(':main:cli')
+    implementation platform('com.google.cloud:libraries-bom:20.9.0')
+    implementation 'com.beust:jcommander:1.48'
+    implementation 'com.google.flogger:flogger:0.5.1'
+    implementation 'com.google.flogger:flogger-system-backend:0.5.1'
+    implementation 'org.locationtech.spatial4j:spatial4j:0.7'
+    implementation 'org.springframework.boot:spring-boot-starter'
```
barbeau, line 67: All dependencies should include version numbers so builds are reproducible

barbeau (commented):
@lionel-nj Some initial feedback as I work through this. 133 files changed in a single PR is a lot :).

I think an overall architecture diagram and explanation of how Spring Boot fits into a Kubernetes deployment would help me wrap my head around this. The PR does a pretty good job of explaining what Spring Boot does, but not why it does it. Could you add this here or somewhere else?

lionel-nj:
> @lionel-nj Some initial feedback as I work through this. 133 files changed in a single PR is a lot :).

100% agree - which is why I split the refactor and gradle update in #962 and #963 

> I think an overall architecture diagram and explanation of how Spring Boot fits into a Kubernetes deployment would help me wrap my head around this. The PR does a pretty good job of explaining what Spring Boot does, but not why it does it. Could you add this here or somewhere else?

Sure, will do!

lionel-nj:
> I think an overall architecture diagram and explanation of how Spring Boot fits into a Kubernetes deployment would help me wrap my head around this. The PR does a pretty good job of explaining what Spring Boot does, but not why it does it. Could you add this here or somewhere else?

@barbeau PTAL at the following. I tries to describe how the springboot layer stands in the kubernetes. It is basically needed to expose the gtfs-validator service via a container.
Blank Diagram (6)

maximearmstrong (changes requested):
Some comments in-line :)

barbeau (commented):
Thanks @lionel-nj, some comments in-line.

docs/USAGE.md
```diff
@@ -85,3 +85,41 @@ java -jar gtfs-validator-SNAPSHOT.jar --export_notice_schema --url https://url/t
  1. Download the GTFS feed at the URL `https://url/to/dataset.zip` and name it `input.zip`  
  1. Validate the GTFS data and output the results to the directory located at `relative/output/path`. Validation results are exported to JSON by default.
 Please note that since downloading will take time, we recommend validating repeatedly on a local file.
+
+## via spring boot app
+
+### Launch the service
+1. Install [Java 8 or higher](https://www.oracle.com/java/technologies/javase-downloads.html).
+1. Download the latest gtfs-validator_springboot JAR file from our [Releases page](https://github.com/MobilityData/gtfs-validator/releases) or snapshot artifact from [GitHub Actions](https://github.com/MobilityData/gtfs-validator/actions?query=branch%3Amaster).
+1. Authenticate to Google Storage
```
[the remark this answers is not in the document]
lionel-nj: Thanks.

main/springboot/src/main/java/org/mobilitydata/gtfsvalidator/springboot/GtfsValidatorController.java
```diff
@@ -0,0 +1,129 @@
...
+            String.format("%s/%s/%s", commitSha, datasetId, args.getValidationReportName()));
+    BlobInfo blobInfo = BlobInfo.newBuilder(blobId).build();
+    try {
+      storage.create(
+          blobInfo,
+          Files.readAllBytes(
+              Paths.get(
+                  String.format("%s/%s", args.getOutputBase(), args.getValidationReportName()))));
+      builder.append("Execution fo the validator was successful.");
+    } catch (IOException ioException) {
```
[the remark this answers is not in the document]
lionel-nj: > Related to status codes - does the Google Cloud SDK provide more precise exceptions for unauthorized, etc.?

Yes, that is the case. 

> If so I'd suggest catching these separately (or using SDK methods to check explicitly first, if that's supported) so that you can return status codes specific to unauthorized, network problem, etc.

Modified.

main/springboot/src/main/java/org/mobilitydata/gtfsvalidator/springboot/GtfsValidatorController.java
```diff
@@ -0,0 +1,129 @@
...
+      "-t", threads,
+      "-c", country_code,
+      "-u", url,
+      "-v", validation_report_name,
+      "-e", system_error_report_name
+    };
+    Arguments args = new Arguments();
+    JCommander jCommander = new JCommander(args);
+    jCommander.parse(argv);
+    Main.main(argv);
```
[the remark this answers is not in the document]
lionel-nj: With the latest modifications, if `Main.main` ever throws an exception, it will be propagated by `SystemLambda.catchSystemExit`: HttpStatus is set to `500` and a message refers to the execution logs to debug.

docs/USAGE.md
```diff
@@ -85,3 +85,41 @@ java -jar gtfs-validator-SNAPSHOT.jar --export_notice_schema --url https://url/t
...
+Sample query:
+
+```
+http://localhost:8080/?url=http://webapps.thebus.org/transitdata/Production/google_transit.zip&dataset_id=id_value&commit_sha=sha_value
+```
+
+which will:
+1. execute the validator on the dataset available at: http://webapps.thebus.org/transitdata/Production/google_transit.zip
+1. store the validation report as report.json in the bucket specified by environment variable `VALIDATION_REPORT_BUCKET` sha_value/id_value.
+
```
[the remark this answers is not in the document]
lionel-nj: > The documentation above is great to launch the spring boot app locally. Is the documentation about deploying it on Kubernetes is planned in this PR? I think it would be a great addition to document how
> 
> 1. the app can be deployed on Kubernetes;
> 2. the pipeline from having a list of URLs to storing the reports in Google Cloud Storage, passing by the spring boot app, docker and Kubernetes;
> 3. a quick explanation on why spring boot was necessary.
> 
> I think your [architecture](https://github.com/MobilityData/gtfs-validator/pull/933#issuecomment-902075651) is a really good start for points 2 and 3.

I was not planning on including the documentation to deploy to Kubernetes within this PR but in https://github.com/MobilityData/gtfs-validator/pull/848 - However, given your feedback, I will include documentation that explains why this springboot layer is needed to run the acceptance tests on Kubernetes.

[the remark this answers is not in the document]
lionel-nj: > Thanks @lionel-nj for putting [the architecture diagram](https://github.com/MobilityData/gtfs-validator/pull/933#issuecomment-902075651) together! I agree with @maximearmstrong, I'd put it in the documentation in this PR directly so it's included in the project.
> 
> However, I had a hard time deciphering exactly what was happening looking at the current architecture diagram. I think you're mixing architecture with process diagrams here which I found confusing. The GitHub Action shading encompasses the entire data flow, which makes it seem like everything is running on GitHub's infrastructure (it's not, right? The actual validation runs as the Spring Boot application on Google Cloud infrastructure?).

You're right, the Github action is the starting point of the pipeline: but the operations do not run in the Github Actions infrastructure.

> My understanding is that the overall architecture logical data flow looks like the following:
> 
> * GitHub Action (Validator is compiled) -> Kubernetes on Google Cloud (Validators are executed) -> GitHub Action (Output files from executed validators are compared)
> 
> Is that right?

Absolutely!

> It would also be helpful to label the parts of the architecture with what is completed (i.e., in this PR or already exists) and what is not yet implemented).

Will do.

main/springboot/build.gradle
```diff
@@ -0,0 +1,94 @@
+/*
```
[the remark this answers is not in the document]
lionel-nj: > Have you thought of the tradeoffs between putting this in it's own top-level module of springboot vs a sub-module under main?

> To me it seems like it would be more natural in it's own top-level module (i.e., it's not really a part of the "main" application), but you may have thought of reasons why that shouldn't exist. Also, it seems a little strange to have a "main/cli" module that actually includes a lot more than the CLI.


During my experiment for some reason having springboot as a top-level module was causing issue - however not it seems to work fine. So, per your suggestion I'll refactor.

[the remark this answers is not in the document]
lionel-nj: > Also, no matter what, the BUILD.md file needs to be updated to explain how many JARs are generated when you execute ./gradlew clean build and where they are located and what they do, and how to build the sub-modules if you only want to build the CLI (e.g., if building springboot takes longer).

Will do.

main/springboot/src/main/java/org/mobilitydata/gtfsvalidator/springboot/GtfsValidatorController.java
```diff
@@ -0,0 +1,129 @@
...
+
+  @GetMapping("/")
+  @ResponseBody
+  /**
+   * Runs {@code org.mobilitydata.gtfsvalidator.cli.Main.main} with the arguments extracted via the
+   * query parameters. Returns the status of the validation and validation report storage. Requires
+   * authentification to be set prior execution i.e. environment variable
+   * GOOGLE_APPLICATION_CREDENTIALS has to be defined.
+   */
+  public String run(
```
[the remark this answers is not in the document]
lionel-nj: > Is it possible to force Spring Boot to only use HTTPS without a lot of manual configuration? If not, you should also call out that HTTP communication is used and is inherently insecure. It would be good to default to HTTPS throughout, including in the documentation.

It is possible to require HTTPS, but from my early research it seems that this required to expose `.p12`. and `.crt` files that might contain credentials, on Github. I will look further, but I'd rather not publish certificates on GitHub - there might be a trick to avoid this.

lionel-nj:
Thanks for reviewing! @barbeau and @maximearmstrong! Next thing for this PR consists in integrating documentation (a diagram) to explain why this is needed.  Meanwhile could you take a look at the logic that was reworked please?

springboot/src/main/java/org/mobilitydata/gtfsvalidator/springboot/GtfsValidatorController.java
```diff
@@ -0,0 +1,219 @@
...
+      if (exitCode != 0) {
+        messageBuilder.append(
+            "Internal error. Please check execution logs for more information.\n");
+        root.getAsJsonObject(PROPERTIES_JSON_KEY)
+            .addProperty(MESSAGE_JSON_KEY, messageBuilder.toString());
+        status = HttpStatus.INTERNAL_SERVER_ERROR;
+        return new ResponseEntity<>(DEFAULT_GSON.toJson(root), status);
+      }
+    } catch (AssertionError assertionError) {
+      messageBuilder.append("Execution of the validator was successful.\n");
```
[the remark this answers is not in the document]
lionel-nj: This was addressed in https://github.com/MobilityData/gtfs-validator/pull/993, PTAL. I also include the changes that result from #993 to this PR.

springboot/src/main/java/org/mobilitydata/gtfsvalidator/springboot/GtfsValidatorController.java
```diff
@@ -0,0 +1,329 @@
...
+      messageBuilder.append(ioException.getMessage());
+      status = HttpStatus.BAD_REQUEST;
+      root.getAsJsonObject(PROPERTIES_JSON_KEY)
+          .addProperty(
+              MESSAGE_JSON_KEY, messageBuilder.append(ioException.getMessage()).toString());
+      exportReport(noticeContainer, args);
+      return new ResponseEntity<>(GSON.toJson(root), status);
+    } catch (URISyntaxException uriSyntaxException) {
+      logger.atSevere().withCause(uriSyntaxException);
+      status = HttpStatus.NOT_FOUND;
```
barbeau, line 178: Shouldn't this be `HttpStatus.BAD_REQUEST` as well? It would indicate "if URL is malformed" according to GtfsInput doc and https://docs.oracle.com/javase/8/docs/api/java/net/URISyntaxException.html

springboot/src/main/java/org/mobilitydata/gtfsvalidator/springboot/GtfsValidatorController.java
```diff
@@ -0,0 +1,329 @@
...
+      @RequestParam(required = false, defaultValue = DEFAULT_COUNTRY_CODE) String country_code,
+      @RequestParam() String url,
+      @RequestParam(required = false, defaultValue = Arguments.VALIDATION_REPORT_NAME_JSON)
+          String validation_report_name,
+      @RequestParam(required = false, defaultValue = Arguments.SYSTEM_ERRORS_REPORT_NAME_JSON)
+          String system_error_report_name,
+      @RequestParam(required = false, defaultValue = "dataset id value") String dataset_id,
+      @RequestParam(required = false, defaultValue = "commit sha value") String commit_sha) {
+
+    Arguments args =
```
barbeau, line 127: Do you still need this step to convert to JCommander Arguments? The only potential benefit I could see is that maybe it's doing some kind of validation, but looking at the docs quickly it doesn't look like it does by default:
https://jcommander.org/#_parameter_validation

So I think it can be removed and remove the dependency too from build.gradle.

springboot/src/main/java/org/mobilitydata/gtfsvalidator/springboot/GtfsValidatorController.java
lionel-nj: Indeed! Thanks.

springboot/src/main/java/org/mobilitydata/gtfsvalidator/springboot/GtfsValidatorController.java
lionel-nj: I think that we should keep it since we use `Arguments` and its methods in the code below.

springboot/src/main/java/org/mobilitydata/gtfsvalidator/springboot/GtfsValidatorController.java
lionel-nj: `BAD_REQUEST` seems to be appropriate for `IOEception`s. Will modify accordingly.

springboot/src/main/java/org/mobilitydata/gtfsvalidator/springboot/GtfsValidatorController.java
barbeau: Ah, ok - I thought that Arguments was a JCommander library class, but I see it's not now. It's still a little strange IMHO that the web implementation has an indirect dependency on the CLI arguments library (JCommander), but I see why you're doing it (to avoid duplicating code to hold parameter data).

springboot/src/main/java/org/mobilitydata/gtfsvalidator/springboot/GtfsValidatorController.java
```diff
@@ -0,0 +1,356 @@
...
+   * @param dataset_id the id of the dataset validated
+   * @param commit_sha the commit SHA
+   * @return the {@code ResponseEntity} that contains information about the response's {@code
+   *     HttpStatus} and a message (as a {@code String}) that gives more information about success
+   *     or failure. <a href=https://cloud.google.com/storage/docs/json_api/v1/status-codes>Possible
+   *     HTTP status codes</a>
+   */
+  @GetMapping(value = "/", produces = "application/json; charset=UTF-8")
+  @ResponseBody
+  public ResponseEntity<HashMap<String, String>> run(
```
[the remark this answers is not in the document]
lionel-nj: Indeed I was returning a `Json` here before, I switched to `HashMap` because `ResponseEntity` automatically converts the response body as a json string, so no need to build one from scratch. I used a map to keep the following structure (which seems more clear than the former one):
- current response body structure:

```json
{ 
  "message": "message body",
  "status": "status code"
}
```

- former response body structure:

```json
{ 
  "properties": 
  {
    "message": "message body",
  }
}
```

springboot/src/main/java/org/mobilitydata/gtfsvalidator/springboot/GtfsValidatorController.java
```diff
@@ -0,0 +1,356 @@
...
+        ValidationContext.builder()
+            .setCountryCode(
+                CountryCode.forStringOrUnknown(
+                    args.getCountryCode() == null ? CountryCode.ZZ : args.getCountryCode()))
+            .setCurrentDateTime(new CurrentDateTime(ZonedDateTime.now(ZoneId.systemDefault())))
+            .build();
+
+    try {
+      feedContainer =
+          Main.loadAndValidate(
```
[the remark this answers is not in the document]
lionel-nj: Thanks for flagging that, I implemented websocket via springboot, will take this into consideration when deploying to Kubernetes (will likely follow the steps described [here](https://medium.com/k8scaleio/running-websocket-app-on-kubernetes-2e13eabb4c4f)).

springboot/src/main/java/org/mobilitydata/gtfsvalidator/springboot/Main.java
```diff
@@ -0,0 +1,40 @@
...
+@SpringBootApplication
+public class Main {
+
+  public static void main(String[] args) {
+    final FluentLogger logger = FluentLogger.forEnclosingClass();
+
+    logger.atWarning().log(
+        "Before running this, the user should understand that: There isn't any authentication "
+            + "(i.e. any incoming HTTP request is processed) and this web application relies"
+            + " entirely on JCommander to parse query parameters. To be fully secure, this web"
```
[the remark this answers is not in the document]
lionel-nj: Addressed in https://github.com/MobilityData/gtfs-validator/pull/933/commits/afa807901be2a08af0162f58a0d7aa296414c771.

lionel-nj:
@barbeau PTAL.

springboot/src/main/java/org/mobilitydata/gtfsvalidator/springboot/GtfsValidatorController.java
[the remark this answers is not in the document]
barbeau: >Thanks for flagging that, I implemented WebSocket via springboot (addressed in https://github.com/MobilityData/gtfs-validator/pull/933/commits/57f3e71953a771a131a460caa7c4f29e86ff7886) and will take this into consideration when deploying to Kubernetes (will likely follow the steps described [here](https://medium.com/k8scaleio/running-websocket-app-on-kubernetes-2e13eabb4c4f)).

I've never used Spring Boot myself, but I'm not sure this is enough to "implement" web sockets? For example see https://stackoverflow.com/questions/50984497/how-to-write-websockets-for-rest-calls-in-spring-boot.

This code in this PR looks like it's using annotations to handle incoming HTTP GET requests in the REST model and then provide a response. I imagine you'd need to re-write this using the classes in the Spring Boot `web.socket` package, as websockets don't fit in the request/response REST model.

As I said above, I don't know if websockets are necessary, but if you use the REST request/response model you'll have clients waiting for a long time for responses because it's synchronous code, so the validation needs to complete before the server can return the response. I'd test first and see if this is a problem if it involves a lot of effort to re-write this for websockets.

springboot/src/main/java/org/mobilitydata/gtfsvalidator/springboot/GtfsValidatorController.java
```diff
@@ -0,0 +1,420 @@
...
+  private static final String PRIVATE_KEY = "PRIVATE_KEY";
+  private static final String CLIENT_EMAIL = "CLIENT_EMAIL";
+  private static final String CLIENT_ID = "CLIENT_ID";
+  private static final String AUTH_URI = "AUTH_URI";
+  private static final String TOKEN_URI = "TOKEN_URI";
+  private static final String AUTH_PROVIDER_X509_CERT_URL = "AUTH_PROVIDER_X509_CERT_URL";
+  private static final String CLIENT_X509_CERT_URL = "CLIENT_X509_CERT_URL";
+
+  /**
+   * Runs {@code org.mobilitydata.gtfsvalidator.cli.Main.main} with the arguments extracted via the
```
barbeau, line 91: Javadocs need to be updated - this no longer runs `Main.main()`

springboot/src/main/java/org/mobilitydata/gtfsvalidator/springboot/GtfsValidatorController.java
```diff
@@ -0,0 +1,420 @@
...
+   * codes</a>
+   *
+   * @param output_base Base directory to store the outputs
+   * @param threads Number of threads to use
+   * @param country_code Country code of the feed, e.g., `nl`. It must be a two-letter country code
+   *     (ISO 3166-1 alpha-2)")
+   * @param url The fully qualified URL to download GTFS archive
+   * @param validation_report_name The name of the validation report including .json extension.
+   * @param dataset_id the id of the dataset validated
+   * @param commit_sha the commit SHA
```
barbeau, line 131: `commit_sha` could use additional explanation here - what `commit_sha` are we talking about and how it is used?

springboot/src/main/java/org/mobilitydata/gtfsvalidator/springboot/GtfsValidatorController.java
```diff
@@ -0,0 +1,420 @@
...
+   * <a href=https://cloud.google.com/storage/docs/json_api/v1/status-codes>Possible HTTP status
+   * codes</a>
+   *
+   * @param output_base Base directory to store the outputs
+   * @param threads Number of threads to use
+   * @param country_code Country code of the feed, e.g., `nl`. It must be a two-letter country code
+   *     (ISO 3166-1 alpha-2)")
+   * @param url The fully qualified URL to download GTFS archive
+   * @param validation_report_name The name of the validation report including .json extension.
+   * @param dataset_id the id of the dataset validated
```
barbeau, line 130: What is this `dataset_id` and where do I find it?

springboot/src/main/java/org/mobilitydata/gtfsvalidator/springboot/GtfsValidatorController.java
```diff
@@ -0,0 +1,420 @@
...
+   *       href=https://cloud.google.com/docs/authentication/getting-started#setting_the_environment_variable>this
+   *       documentation for authentication procedure</a>
+   * </ul>
+   *
+   * @param commitSha the commit SHA
+   * @param datasetId the id of the dataset
+   * @param args the {@code Argument} generated from the query parameters
+   * @return the {@code HttpStatus} of the validation report storage process
+   */
+  private HttpStatus pushValidationReportToCloudStorage(
```
barbeau, line 296: This method throws IOException, which would then also be thrown by `run()`, which I assume would crash the thread handling the incoming HTTP request. Is this what you want? Seems like IOExceptions here should be handled and response with status code returned to client.

springboot/src/main/java/org/mobilitydata/gtfsvalidator/springboot/GtfsValidatorController.java
```diff
@@ -0,0 +1,420 @@
...
+   * @param output_base Base directory to store the outputs
+   * @param threads Number of threads to use
+   * @param country_code Country code of the feed, e.g., `nl`. It must be a two-letter country code
+   *     (ISO 3166-1 alpha-2)").
+   * @param url The fully qualified URL to download GTFS archive
+   * @param validation_report_name The name of the validation report including .json extension.
+   * @param system_error_report_name The name of the system error report including .json extension.
+   * @return the {@code Argument} instance generated from the query parameters passed.
+   */
+  private Arguments queryParametersToArguments(
```
barbeau, line 231: In this method signature and in `run()` pipe variables are using snake case (with `_` - `output_base`) instead of the normal Java convention camel case (`outputBase`). Seems like at a minimum this method should be changed to use camel case, and `run()` as well unless Spring Boot requires snake case for some reason.

build.gradle
```diff
@@ -35,11 +35,27 @@ repositories {
 
 apply plugin: 'nebula-aggregate-javadocs'
 
-// a task to run all tests and combine them in one report
-task testReport(type: TestReport) {
-    print("System.getenv(\"CI\"): " + System.getenv("CI"))
-    print(" - System.getenv(\"versionTag\"): " + System.getenv("versionTag"))
+// A resolvable configuration to collect test reports data
```
f8full, line 38: 👍

springboot/src/main/java/org/mobilitydata/gtfsvalidator/springboot/GtfsValidatorController.java
lionel-nj: At first I thought query parameters had to be in snake case by convention. However, checking it seems that camelCase is also accepted. https://github.com/MobilityData/gtfs-validator/pull/933/commits/3bfbdda1e34f02b5979c7884920ae6ba3ad85aa5 changes snake case to camel case.

https://api.gov.au/standards/national_api_standards/naming-conventions.html

springboot/src/main/java/org/mobilitydata/gtfsvalidator/springboot/GtfsValidatorController.java
lionel-nj: Additional documentation has been provided in https://github.com/MobilityData/gtfs-validator/pull/933/commits/3bfbdda1e34f02b5979c7884920ae6ba3ad85aa5.

springboot/src/main/java/org/mobilitydata/gtfsvalidator/springboot/GtfsValidatorController.java
lionel-nj: Additional documentation has been provided in https://github.com/MobilityData/gtfs-validator/pull/933/commits/3bfbdda1e34f02b5979c7884920ae6ba3ad85aa5.

springboot/src/main/java/org/mobilitydata/gtfsvalidator/springboot/GtfsValidatorController.java
lionel-nj: My bad - I forgot to push code I had on my laptop. https://github.com/MobilityData/gtfs-validator/pull/933/commits/3bfbdda1e34f02b5979c7884920ae6ba3ad85aa5 does not throw IOException.

lionel-nj:
Closing since it is no longer relevant: we use github actions instead of kubernetes technologies.

Issue #763, opened 2021-02-19
Check allowed location_types for stops referenced by stop_times.stop_id 


barbeau:
## **Feature request**
Discussion started here as part of review of StopTooFarFromTripShapeValidator changes:
https://github.com/MobilityData/gtfs-validator/pull/750#discussion_r577892892

Looking at [the spec](https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md#stop_timestxt) (it was recently changed in https://github.com/google/transit/pull/258), it now says:

>Referenced locations must be stops/platforms, i.e. their stops.location_type value must be 0 or empty.

**Is your feature request related to a problem? Please describe.**
Currently the validator does not check to ensure the location_types of stops that are referenced in stop_times.txt

**Proposed solution**
The validator should throw errors if stops with the wrong location_type are referenced from stop_times.stop_id.

PR #938, opened 2021-07-26
chore: new rule - additional check on stops.location_type

lionel-nj:
closes https://github.com/MobilityData/gtfs-validator/issues/763

**Summary:**

This PR provides support to implement a new rule.

GTFS specification reference: 

> Referenced locations must be stops/platforms, i.e. their `stops.location_type` value must be 0 or empty.

from _https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md#stoptimestxt_

This is `WARNING` to be upgraded to `ERROR` in the future.

**Expected behavior:** 

New notice: `WrongStopTimeStopLocationTypeNotice`

A notice should be emitted if: 

- [x] the stop referred to in `stop_times.stop_id` is not a "STOP" or a "PLATFORM", i.e its value for the field `stops.location_type` is different from 0 or empty.

### DIFF
--- a/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTimeStopLocationTypeValidator.java
+++ b/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTimeStopLocationTypeValidator.java
@@ -1,0 +1,99 @@
+/*
+ * Copyright 2021 MobilityData IO
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.mobilitydata.gtfsvalidator.validator;
+
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Multimaps;
+import java.util.List;
+import java.util.Optional;
+import javax.inject.Inject;
+import org.mobilitydata.gtfsvalidator.annotation.GtfsValidator;
+import org.mobilitydata.gtfsvalidator.annotation.SchemaExport;
+import org.mobilitydata.gtfsvalidator.notice.NoticeContainer;
+import org.mobilitydata.gtfsvalidator.notice.SeverityLevel;
+import org.mobilitydata.gtfsvalidator.notice.ValidationNotice;
+import org.mobilitydata.gtfsvalidator.table.GtfsLocationType;
+import org.mobilitydata.gtfsvalidator.table.GtfsStop;
+import org.mobilitydata.gtfsvalidator.table.GtfsStopTableContainer;
+import org.mobilitydata.gtfsvalidator.table.GtfsStopTime;
+import org.mobilitydata.gtfsvalidator.table.GtfsStopTimeTableContainer;
+
+/**
+ * Validates {@code stop_times.stop_id} refers to a stop/platform  i.e. their stops.location_type
+ * value must be 0 or empty.
+ *
+ * <p>Generated notices:
+ *
+ * <ul>
+ *   <li>{@link WrongStopTimeStopLocationTypeNotice} - a single departure_time or
+ *       arrival_time is defined for a row (both or none are expected)
+ * </ul>
+ */
+@GtfsValidator
+public class StopTimeStopLocationTypeValidator extends FileValidator {
+
+  private final GtfsStopTimeTableContainer stopTimeTable;
+  private final GtfsStopTableContainer stopTable;
+
+  @Inject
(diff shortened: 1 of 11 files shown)


main/src/test/java/org/mobilitydata/gtfsvalidator/validator/StopTimeStopLocationTypeValidatorTest.java
```diff
@@ -0,0 +1,123 @@
...
+        .validate(noticeContainer);
+    return noticeContainer.getValidationNotices();
+  }
+
+  @Test
+  public void stopTimeRefersToStop_noNotice() {
+    assertThat(generateNotices(
+        ImmutableList.of(
+            createStopTime(1, "trip id value", "stop id value", 5),
+            createStopTime(5, "other  id value", "other id value", 8)
```
maximearmstrong, line 71:
```suggestion
            createStopTime(5, "other trip  id value", "other stop id value", 8)
```

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/StopTimeStopLocationTypeValidatorTest.java
```diff
@@ -0,0 +1,123 @@
...
+        )
+    )).isEmpty();
+  }
+
+  @Test
+  public void wrongForeignKey_noNotice() {
+    assertThat(generateNotices(
+        ImmutableList.of(
+            createStopTime(1, "trip id value", "stop id value", 5),
+            createStopTime(5, "other  id value", "wrong foreign key", 8)
```
maximearmstrong, line 85:
```suggestion
            createStopTime(5, "other trip id value", "wrong foreign key", 8)
```

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/StopTimeStopLocationTypeValidatorTest.java
```diff
@@ -0,0 +1,123 @@
...
+      long csvRowNumber, String tripId, String stopId, int stopSequence) {
+    return new GtfsStopTime.Builder()
+        .setCsvRowNumber(csvRowNumber)
+        .setTripId(tripId)
+        .setStopId(stopId)
+        .setStopSequence(stopSequence)
+        .build();
+  }
+
+  private static GtfsStop createStop(long csvRowNumber, String stopId, Integer locationTypeValue) {
```
maximearmstrong, line 48: Should we use GtfsLocationtype in the signature instead since it is an Enum with limited Integer possibilities? No strong feeling about this, I just think it would be closer to the real world behaviour where receiving a value outside the Enum possibilities would be a problem.

maximearmstrong (changes requested):

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/StopTimeStopLocationTypeValidatorTest.java
lionel-nj: Good catch - yes!

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTimeStopLocationTypeValidator.java
```diff
@@ -0,0 +1,99 @@
...
+import org.mobilitydata.gtfsvalidator.notice.ValidationNotice;
+import org.mobilitydata.gtfsvalidator.table.GtfsLocationType;
+import org.mobilitydata.gtfsvalidator.table.GtfsStop;
+import org.mobilitydata.gtfsvalidator.table.GtfsStopTableContainer;
+import org.mobilitydata.gtfsvalidator.table.GtfsStopTime;
+import org.mobilitydata.gtfsvalidator.table.GtfsStopTimeTableContainer;
+
+/**
+ * Validates that each {@code stop_times.stop_id} refers to a stop/platform, i.e. its stops.location_type value must be 0 or empty.
+ * value must be 0 or empty.
```
maximearmstrong, line 37: I think the end of the previous line is duplicated here
```suggestion
 *
```

maximearmstrong (approved):
One small change, then LGTM :)

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTimeStopLocationTypeValidator.java
```diff
@@ -0,0 +1,99 @@
...
+      GtfsStopTableContainer stopTable) {
+    this.stopTimeTable = stopTimeTable;
+    this.stopTable = stopTable;
+  }
+
+  @Override
+  public void validate(NoticeContainer noticeContainer) {
+    for (List<GtfsStopTime> stopTimeList : Multimaps.asMap(stopTimeTable.byTripIdMap()).values()) {
+      for (GtfsStopTime stopTime : stopTimeList) {
+        Optional<GtfsStop> stop = Optional.ofNullable(stopTable.byStopId(stopTime.stopId()));
```
aababilov, line 63: I would suggest to simplify this code. Optional and lambda are not necessary to achieve your goal here. Instead, you can just compare the stop with null and write a usual `if`.

We already have an implementation of that rule at Google and it also warns about unused stops. I would suggest to properly sync with you guys to make sure that you are not rewriting the rules that we already have and wish to open-source.

Issue #943, opened 2021-08-02
Fix typos and other errors in NOTICES.md or remove it


aababilov:
Fix typos and other errors in NOTICES.md or remove it. This file is hard to maintain manually because it aggregates all the notices produced by validator.

Some errors that I have found so far:

```
| `routeId`        	| The id of the faulty record.            	| Long   	|
```
routeId must be string, not long.

```
| `csvRowNumber`   	| The row number of the faulty record.    	| String 	|
```
csvRowNumber must be long, not string.

```
| `block_trips_with_overlapping_stopTimes`               	| [`BlockTripsWithOverlappingStopTimesNotice`](#BlockTripsWithOverlappingStopTimesNotice)                         	|
| `decreasing_or_equal_stopTime_distance`                	| [`DecreasingOrEqualStopTimeDistanceNotice`](#DecreasingOrEqualStopTimeDistanceNotice)                           	|
```

Wrong capitalization.

isabelle-dr:
Thank you for opening this issue, the typos you mentioned + some other similar ones have been fixed in #940 

What was the reason for having two files (NOTICES and RULES)? I feel like both files are needed to either: 
1. interpret the validation report 
2. understanding the rules

RULES.md has:
- a quick description (in table), a longer description (below) and the provenance of the rule (python validator, spec...)

NOTICES.md has:
- the notice code as seen in the validation report
- the affected files
- the schema of the results

@lionel-nj do you see any potential issues or confusion to having only one file?

lionel-nj:
> What was the reason for having two files (NOTICES and RULES)?

We initially saw `RULES.md` as a file that would describe the validation rules and `NOTICES.md` as a way to give more information about a specific notice.

> @lionel-nj do you see any potential issues or confusion to having only one file?

Indeed, having information in both files is confusing - I would advocate to only have one: `RULES.md` because since `NOTICES.md` is manually edited (as @aababilov mentioned: this leads to merging conflicts, typos, and possible inaccuracy); the file is hard to maintain.  Also in https://github.com/MobilityData/gtfs-validator/pull/925 there is a discussion to get the information from `NOTICES.md` into the javadocs.

PR #944, opened 2021-08-02
feat: Check that vehicles do not travel too fast

aababilov:
The new StopTimeTravelSpeedValidator was implemented at used at Google
for years. It replaces TooFastTravelValidator.

The new validator has the following differences from the old one.
* Use Google S2 Geometry library instead of Spatial4J, so it can be built at Google
* Check both the speed between consequent stops as well as the speed between far stops
* Each line variant is validated only once. This is a major speed up
  because one line variant typically runs multiple trips.
* Different route types have different speed thresholds, so a train may
  run much faster than a bus.


aababilov:
Here I am publishing the code that we use at Google. It was reviewed by Google engineers and its logic was used by us for years.

I am not using the SchemaExport annotation because I would have to create constructors with 15 parameters. This is error-prone since it is easy to pass params in wrong order. I think we need another way to export the schema.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTimeTravelSpeedValidator.java
```diff
@@ -0,0 +1,360 @@
...
+  }
+
+  /**
+   * Returns coordinates of the stop. If they are missing, coordinates of the parent are used.
+   *
+   * <p>Returns (0, 0) if no coordinates are found in parent chain. That case is reported in another
+   * validator.
+   */
+  static S2LatLng getStopLatLng(GtfsStopTableContainer stopTable, String stopId) {
+    for (; ; ) {
```
lionel-nj, line 259: Why using an infinite loop here? @aababilov

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTimeTravelSpeedValidator.java
```diff
@@ -0,0 +1,360 @@
...
+                stop1,
+                stopsOnTrip.get(i + 1),
+                stop2,
+                speedKph,
+                distanceKm));
+      }
+    }
+  }
+
+  static double getSpeedKphBetweenStops(
```
lionel-nj, line 202: Should this be private?

lionel-nj (approved):
Thanks @aabilov for working on that! LGTM @barbeau, thoughts?

core/src/main/java/org/mobilitydata/gtfsvalidator/util/S2Earth.java
```diff
@@ -0,0 +1,70 @@
...
+     6371.01 +/- 0.02 km.  The equatorial radius is 6378.136 km, and the polar
+     radius is 6356.752 km.  They differ by one part in 298.257.
+
+     Reference: http://ssd.jpl.nasa.gov/phys_props_earth.html, which quotes
+     Yoder, C.F. 1995. "Astrometric and Geodetic Properties of Earth and the
+     Solar System" in Global Earth Physics, A Handbook of Physical ants,
+     AGU Reference Shelf 1, American Geophysical Union, Table 2.
+  */
+  public static double getRadiusMeters() {
+    return 6371010.0;
```
lionel-nj, line 68: Should this be a class constant?

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTimeTravelSpeedValidator.java
aababilov: StopTimeTravelSpeedValidatorTest uses getSpeedKphBetweenStops. Let me put `@VisibleForTesting` here.

core/src/main/java/org/mobilitydata/gtfsvalidator/util/S2Earth.java
aababilov: I would have made it a class constant if I had developed it from scratch. However, S2Earth is widely used in Google. I wish to open-source it some day, but for now I suggest to copy its tiny code to GTFS Validator. If S2Earth gets published, then the transition will be very easy.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTimeTravelSpeedValidator.java
```diff
@@ -0,0 +1,362 @@
...
+   * <p>If there is a fast travel detected, then a separate notice is issued for each trip.
+   */
+  private void validateConsecutiveStops(
+      List<List<GtfsStopTime>> trips,
+      double[] distancesKm,
+      double maxSpeedInKph,
+      NoticeContainer noticeContainer) {
+    final List<GtfsStopTime> stopTimes = trips.get(0);
+
+    for (int i = 0; i < distancesKm.length; ++i) {
```
barbeau, line 175: If I understand this logic correctly, `validateConsecutiveStops()` won't validate GTFS without the `timepoint` column. In other words, if there are rows in stop_times.txt without times included (i.e., the rows with times are implicitly timepoints):

|  stop_sequence | arrival_time | departure_time |
|  ------ | ------ | ------ | 
1 | 00:00:00 | 00:00:00  |  
2 |  |  |  
3 |  |  |  
4 | 00:10:00 | 00:10:00  |  

...because it would simply skip over the empty rows and look for the next row with a departure time.

But, these cases would still be evaluated by `validateFarStops()`.

Is that right @aababilov?

If so it might be useful to include some explanation of this in the comments to explain why the two different rule implementations exist.

Related - https://github.com/google/transit/issues/61.

RULES.md
```diff
@@ -584,6 +585,24 @@ A row in the input file has only spaces.
...
+
+A transit vehicle moves too fast between two consecutive stops. The speed threshold depends on route type.
+
+##### References:
+* [Original Python validator implementation](https://github.com/google/transitfeed)
+
+#### FastTravelBetweenFarStopsNotice
+
+A transit vehicle moves too fast between far consecutive stops (more than in 10 km apart). 
+This normally indicates a more serious problem than too fast travel between consecutive stops.
```
barbeau, line 600: If this is a "more serious problem", should it be elevated to an error instead of warning?

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/StopTimeTravelSpeedValidatorTest.java
```diff
@@ -0,0 +1,315 @@
...
+import org.mobilitydata.gtfsvalidator.table.GtfsStopTimeTableContainer;
+import org.mobilitydata.gtfsvalidator.table.GtfsTrip;
+import org.mobilitydata.gtfsvalidator.table.GtfsTripTableContainer;
+import org.mobilitydata.gtfsvalidator.type.GtfsTime;
+import org.mobilitydata.gtfsvalidator.util.S2Earth;
+import org.mobilitydata.gtfsvalidator.validator.StopTimeTravelSpeedValidator.FastTravelBetweenConsecutiveStopsNotice;
+import org.mobilitydata.gtfsvalidator.validator.StopTimeTravelSpeedValidator.FastTravelBetweenFarStopsNotice;
+
+@RunWith(JUnit4.class)
+public final class StopTimeTravelSpeedValidatorTest {
```
barbeau, line 47: Suggest adding test case for legacy GTFS without timepoint column that may have missing arrival and departure times for non-timepoint records:

|  stop_sequence | arrival_time | departure_time |
|  ------ | ------ | ------ | 
1 | 00:00:00 | 00:00:00  |  
2 |  |  |  
3 |  |  |  
4 | 00:10:00 | 00:10:00  |

core/build.gradle
```diff
@@ -38,6 +38,7 @@ dependencies {
     implementation 'commons-io:commons-io:2.8.0'
     implementation 'com.univocity:univocity-parsers:2.9.0'
     implementation 'com.google.code.gson:gson:2.8.6'
+    implementation 'com.google.geometry:s2-geometry:2.0.0'
```
barbeau, line 41: 🎉

barbeau (commented):
Thanks @aababilov for this contribution - great work! 💯 

Some comments in-line.

@lionel-nj I suggest highlighting the change of this rule in the release notes for the next release to give people a heads up that the thresholds are changing and that the old notice now maps to two new notices.

isabelle-dr:
Thank you for this contribution @aababilov! 🙏

RULES.md
```diff
@@ -584,6 +585,24 @@ A row in the input file has only spaces.
 ##### References:
 * [GTFS file requirements](http://gtfs.org/reference/static/#file-requirements)
 
+<a name="FastTravelBetweenConsecutiveStopsNotice"/>
+
+#### FastTravelBetweenConsecutiveStopsNotice
+
+A transit vehicle moves too fast between two consecutive stops. The speed threshold depends on route type.
```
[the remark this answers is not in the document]
aababilov: Done.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTimeTravelSpeedValidator.java
```diff
@@ -0,0 +1,362 @@
...
+
+  /**
+   * The maximum distance between any two stops for which it is allowed to go over the speed limit.
+   */
+  private static final double MAX_DISTANCE_OVER_MAX_SPEED_IN_KMS = 10.0;
+
+  private static final int NUM_SECONDS_PER_MINUTE = 60;
+  private static final int NUM_SECONDS_PER_HOUR = 3600;
+
+  private static long tripFprint(GtfsTrip trip, List<GtfsStopTime> stopTimes) {
```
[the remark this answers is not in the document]
aababilov: You see, a line variant does not have a single definition. A line variant is roughly defined as a trip without time information (that is, a line variant is defined by route_id and a sequence of stops).

I am removing the words "line variant" and instead say "all trips that belong to the same route and visit the same stops at the same times".

Speaking about StopTooFarFromTripShape: it needs a different logic to find tripFprint (see below). This proves that "line variant" is a vague concept.

```
 private static long tripFprint(List<GtfsStopTime> stopTimes) {
    Hasher hasher = HASH_FUNCTION.newHasher().putInt(stopTimes.size());
    for (GtfsStopTime stopTime : stopTimes) {
      hasher
          .putInt(stopTime.stopId().length())
          .putUnencodedChars(stopTime.stopId())
          .putDouble(stopTime.shapeDistTraveled());
    }
    return hasher.hash().asLong();
  }
```

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTimeTravelSpeedValidator.java
```diff
@@ -0,0 +1,362 @@
...
+    return timeBetweenStops;
+  }
+
+  /**
+   * The maximum distance between any two stops for which it is allowed to go over the speed limit.
+   */
+  private static final double MAX_DISTANCE_OVER_MAX_SPEED_IN_KMS = 10.0;
+
+  private static final int NUM_SECONDS_PER_MINUTE = 60;
+  private static final int NUM_SECONDS_PER_HOUR = 3600;
```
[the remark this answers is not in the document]
aababilov: Which manual conversion errors?

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTimeTravelSpeedValidator.java
```diff
@@ -0,0 +1,362 @@
...
+      }
+      if (stop.hasParentStation()) {
+        stopId = stop.parentStation();
+      } else {
+        return S2LatLng.CENTER;
+      }
+    }
+  }
+
+  private static double getMaxVehicleSpeed(GtfsRouteType routeType) {
```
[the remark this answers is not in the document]
aababilov: Done.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTimeTravelSpeedValidator.java
aababilov: `timepoint` column does not affect this function, but it indeed depends on presence of `arrival_time` and `departure_time`.

Two rules exist to catch different situations:

* too high speed between consequent stops
* too high speed between stops that are far away (more than 10 km apart)

This is already explained twice: as a comment for the notice classes and in `RULES.md`.

RULES.md
aababilov: I do not want to reject 225 feeds that have this problem, so I would like to keep it as warning.

By the way, can you run this validator on several thousands feeds that you have in MobilityData?

core/build.gradle
aababilov: Fresh, tiny and neat library just for our favorite GTFS validator!

aababilov:
I got a merge conflict for `RULES.md` and `docs/NOTICES.md`. I think that other developers would face the same problem.

One of the goals of my validator architecture was to avoid modifying the same file when you add a validation since this leads to merge conflicts. Now `RULES.md` and `docs/NOTICES.md` are the reliable source of merge conflicts and also a risk for outdated and incorrect documentation (see https://github.com/MobilityData/gtfs-validator/issues/943).

Let's find a mechanism to generate those files instead of maintaining them manually.

aababilov:
Thanks for review!

maximearmstrong (approved):
LGTM

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTimeTravelSpeedValidator.java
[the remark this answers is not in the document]
barbeau: Ah, yes, good point.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTimeTravelSpeedValidator.java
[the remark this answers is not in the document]
barbeau: This is effectively `60 * 60`. In projects I maintain I've tried to get away from these calculated constants as much as possible and use the built-in Java conversions instead because they protect against wasting time on silly mistakes. This specific example is obviously a super-simple case, so not a big deal to leave as is if you have a strong preference :).

RULES.md
barbeau: Makes sense. 

@lionel-nj would know more about the state of running against the large feed database. Related, I noticed that even the end-to-end-100 job is failing to complete on recent commits - https://github.com/MobilityData/gtfs-validator/runs/3246259762.

docs/NOTICES.md
```diff
@@ -928,17 +930,75 @@
...
+| `stopId1`             | `stop_id` of the first stop.            | String 	|
+| `stopName1`           | `stop_name` of the first stop.          | String 	|
+| `departureTime1`      | `departure_time` of the first stop.     | Time 	|
+| `csvRowNumber2`       | The row number of the second stop time. | Long 	|
+| `stopSequence2`       | `stop_sequence` of the second stop.     | Integer	|
+| `stopId2`             | `stop_id` of the second stop.           | String 	|
+| `stopName2`           | `stop_name` of the second stop.         | String 	|
+| `arrivalTime2`        | `arrival_time` of the second stop.      | Time 	|
+
+##### Speed thresholds
```
barbeau, line 954: @lionel-nj Is this table better off in RULES.md?

My impression from other comments is that NOTICES.md will go away when you can export the notice schemas.

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/StopTimeTravelSpeedValidatorTest.java
barbeau: >By the way, timepoint column is optional. If arrival_time and departure_time are provided, there still may be no timepoint column.

Yes, that's correct.

>What is legacy GTFS that you mentioned?

Basically GTFS without the timepoint column, which has two flavors:
1. The above table, where non-timepoint records omitted arrival and departure times. This format strictly conformed to the original GTFS spec before the timepoint field was added.
2. Arrival and departure times were provided for all records in stop_times.txt. This format did not strictly conform to the original GTFS spec (which said that times should only be provided for timepoints), but it became a common practice because agencies wanted different consumers to show the same scheduled times for stops (vs. interpolating a scheduled time themselves, which might differ between consumer implementations). The timepoint field was added to allow producers to provide times for all stops, but also explicitly mark certain stops as timepoints. So ideally these types of datasets should eventually be phased out of existence by the agency adding the timepoint column - I don't know what the current state of the world is for this now.

See this thread with more examples and details - https://github.com/google/transit/issues/61#issuecomment-702329132 - which also discusses potential changes to the spec language to make the expected use of the field clearer.

docs/NOTICES.md
lionel-nj: Indeed, this would be better in `RULES.md` in order to document the rule's logic.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTimeTravelSpeedValidator.java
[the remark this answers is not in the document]
aababilov: That is nice! You see, my memory keeps 3600 as a constant and not as a product `60 * 60`. I remember it from childhood this way.

I find `TimeUnit.HOURS.toSeconds(N)` convenient for `N > 1`. However, when it comes to a single hour, then I find easier to say 3600 than to rely on a library.

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/StopTimeTravelSpeedValidatorTest.java
aababilov: Thanks a lot for explanation, this is very nice to know!

RULES.md
aababilov: I see, the Validator was out of memory. May we give it 3 GiB? I think it is usually enough.

docs/NOTICES.md
aababilov: Sure, moved.

aababilov:
> Thank you for this contribution @aababilov!

My pleasure!

RULES.md
aababilov: Validator was out of memory because se-trafiklab reports trains using extended vehicle type (100, 101, 102 etc). The upstream validator does not support extended vehicle types - it shows a warning for unknown enum value. StopTimeTravelSpeedValidator uses 50 km/h threshold for "unknown" vehicle types which is too low for trains, so GTFS Validator produces too many warnings for "fast travel" and gets out of memory.

Mitigation. Use 200 km/h threshold for unknown vehicle types: https://github.com/MobilityData/gtfs-validator/pull/946

Solution. Add support for extended route types. At Google. we convert them to basic route types.

```java
/**
   * Converts GTFS route type to a narrow range of basic types.
   *
   * <p>Google supports extended route types (Hierarchical Vehicle Type, HVT) that are not a part of
   * GTFS standard but they are widely used. This function converts HVT to basic route types.
   *
   * <p>Google also supports non-standard route types (intercity bus and commuter train) that are
   * also converted to basic ones.
   */
  public static BasicType toBasicType(GtfsRouteType routeType) {
    switch (routeType) {
      case LIGHT_RAIL:
        return BasicType.LIGHT_RAIL;
      case SUBWAY:
        return BasicType.SUBWAY;
      case RAIL:
      case COMMUTER_TRAIN:
        return BasicType.RAIL;
      case INTERCITY_BUS:
      case BUS:
      case HVT_COMMUNAL_TAXI:
        return BasicType.BUS;
      case FERRY:
        return BasicType.FERRY;
      case CABLE_TRAM:
        return BasicType.CABLE_TRAM;
      case AERIAL_LIFT:
        return BasicType.AERIAL_LIFT;
      case FUNICULAR:
        return BasicType.FUNICULAR;
      case TROLLEYBUS:
        return BasicType.TROLLEYBUS;
      case MONORAIL:
      case HVT_MONORAIL:
        return BasicType.MONORAIL;
      case HORSE_CARRIAGE:
      case HVT_HORSE_DRAWN_CARRIAGE:
        return BasicType.HORSE_CARRIAGE;
      default:
        break;
    }

    // Support for Hierarchical Vehicle Types.
    switch (routeType.getNumber() / 100) {
      case 1:
        return BasicType.RAIL;
      case 2:
      case 7:
        return BasicType.BUS;
      case 4:
        return BasicType.SUBWAY;
      case 8:
        return BasicType.TROLLEYBUS;
      case 9:
        return BasicType.LIGHT_RAIL;
      case 10:
      case 12:
        return BasicType.FERRY;
      case 13:
        return BasicType.AERIAL_LIFT;
      case 14:
        return BasicType.FUNICULAR;
      default:
        return BasicType.MISCELLANEOUS;
    }
  }
```

lionel-nj: > By the way, can you run this validator on several thousands feeds that you have in MobilityData?

The process is under construction: https://github.com/MobilityData/gtfs-validator/pull/848 & https://github.com/MobilityData/gtfs-validator/pull/933. We are considering using Kubernetes to run the validator on all datasets available on the mobility archives (1300+). 

> I see, the Validator was out of memory. May we give it 3 GiB? I think it is usually enough.

Yes, see #947.

PR #945, opened 2021-08-09
feat: Use member fields instead of a map to keep notice context

aababilov:
Each field of child `Notice` classes is exported as a part of the context.
Note that fields that are `null` are skipped from the context. This allows us to have optional fields, e.g., `DuplicateKeyNotice` has optional `fieldName2` and `fieldValue2`. Prior to this change, notice context was an ImmutableMap that does not allow `null` values, so explicit `null` was never exported anyway.

Rationale
--------------
1. Regular fields will allow us to generate JSON schema for each notice and also document each notice field.
2. Regular fields consume much less space than maps.
  * One map has a fixed 44 bytes overhead plus more than 52 bytes per each key-value pair.
  * A regular field normally consumes 4 or 8 bytes.


core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NoticeContainer.java
```diff
@@ -34,9 +35,7 @@
  * own NoticeContainer, and after execution is complete the results are merged.
  */
 public class NoticeContainer {
-  private static final Gson DEFAULT_GSON =
-      new GsonBuilder().serializeNulls().serializeSpecialFloatingPointValues().create();
-  private static final Gson PRETTY_GSON = DEFAULT_GSON.newBuilder().setPrettyPrinting().create();
+  private static final Gson PRETTY_GSON = GSON.newBuilder().setPrettyPrinting().create();
```
lionel-nj, line 38: Why only using the `PRETTY_GSON` here?

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/Notice.java
```diff
@@ -45,10 +62,9 @@ public SeverityLevel getSeverityLevel() {
   /**
    * Returns a descriptive type-specific name for this notice based on the class simple name.
    *
-   * @return notice code, e.g., "foreign_key_error".
+   * @return notice code, e.g., "foreign_key_violation".
```
lionel-nj, line 65: Thanks

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/Notice.java
```diff
@@ -93,4 +105,36 @@ public int hashCode() {
...
     return getSeverityLevel().ordinal() >= SeverityLevel.ERROR.ordinal();
   }
+
+  /** JSON exclusion strategy for notice context. It skips {@link Notice#severityLevel}. */
+  private static class NoticeExclusionStrategy implements ExclusionStrategy {
+    public boolean shouldSkipClass(Class<?> clazz) {
+      return false;
+    }
+
+    public boolean shouldSkipField(FieldAttributes f) {
```
lionel-nj, line 115: Why should we skip severityLevel?

lionel-nj (commented):
Thanks @aabilov for this contribution! A few questions below - I am currently update the documentation in #948.

lionel-nj (approved):

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/Notice.java
aababilov: You are welcome!

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/Notice.java
aababilov: It was not a part of the context originally. When we export notices, we group them by (code, severityLevel) and then export contexts for each group.


```
    "notices": [
        {
            "code": "decreasing_or_equal_shape_distance",
            "severity": "ERROR",
            "totalNotices": 2517999,
            "notices": [
                {
                    "shapeId": "112321",
                    "csvRowNumber": 232516,
                    "shapeDistTraveled": 955.3,
                    "shapePtSequence": 10002,
                    "prevCsvRowNumber": 232515,
                    "prevShapeDistTraveled": 955.3,
                    "prevShapePtSequence": 10001
                },
                {
                    "shapeId": "112321",
                    "csvRowNumber": 232517,
                    "shapeDistTraveled": 955.3,
                    "shapePtSequence": 10003,
                    "prevCsvRowNumber": 232516,
                    "prevShapeDistTraveled": 955.3,
                    "prevShapePtSequence": 10002
                },
```

aababilov:
Thanks for review!

PR #949, opened 2021-08-10
fix: Remove unused annotation NoticeExport

aababilov:
It was replaced by SchemaExport that was removed.


lionel-nj (approved):
LGTM, thanks!

PR #951, opened 2021-08-11
feat: Add NoticeSchemaGenerator

aababilov:
It is currently not plugged into cli/Main.java.

The schema generator uses Java reflection to find all notice 
classes and describe their fields.

The schema generator currently exports JSON schema for the notices. It
may also export other schema formats in future (e.g., SQL-friendly
schema).

Sample output.

```
{
  "attribution_without_role": {
    "type": "object",
    "properties": {
      "attributionId": {
        "type": "string"
      },
      "csvRowNumber": {
        "type": "integer"
      }
    }
  },
  "block_trips_with_overlapping_stop_times": {
    "type": "object",
    "properties": {
      "blockId": {
        "type": "string"
      },
      "csvRowNumberA": {
        "type": "integer"
      },
      "csvRowNumberB": {
        "type": "integer"
      },
      "intersection": {
        "type": "string"
      },
      "serviceIdA": {
        "type": "string"
      },
      "serviceIdB": {
        "type": "string"
      },
      "tripIdA": {
        "type": "string"
      },
      "tripIdB": {
        "type": "string"
      }
    }
  },
```

### DIFF
--- a/core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NoticeSchemaGenerator.java
+++ b/core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NoticeSchemaGenerator.java
@@ -1,0 +1,263 @@
+/*
+ * Copyright 2021 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.mobilitydata.gtfsvalidator.notice;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableList;
+import com.google.common.reflect.ClassPath;
+import com.google.gson.JsonArray;
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonPrimitive;
+import java.io.IOException;
+import java.lang.reflect.Field;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.TreeMap;
+import org.mobilitydata.gtfsvalidator.type.GtfsColor;
+import org.mobilitydata.gtfsvalidator.type.GtfsDate;
+import org.mobilitydata.gtfsvalidator.type.GtfsTime;
+import org.mobilitydata.gtfsvalidator.validator.FileValidator;
+import org.mobilitydata.gtfsvalidator.validator.SingleEntityValidator;
+
+/** Exports schema describing all possible notices and their contexts. */
+public class NoticeSchemaGenerator {
+  /** Default packages to find notices in open-source validator. */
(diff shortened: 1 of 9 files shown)


core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NoticeSchemaGenerator.java
```diff
@@ -0,0 +1,263 @@
...
+  static List<Class<Notice>> findNoticeSubclasses(List<String> packages) throws IOException {
+    List<Class<Notice>> notices = new ArrayList<>();
+    for (ClassPath.ClassInfo classInfo :
+        ClassPath.from(ClassLoader.getSystemClassLoader()).getTopLevelClasses()) {
+      if (!belongsToAnyPackage(classInfo, packages)) {
+        continue;
+      }
+      Class<?> clazz = classInfo.load();
+      maybeAddNoticeClass(clazz, notices);
+      if (isValidatorClass(clazz)) {
```
lionel-nj, line 182: Alright, I see how you avoided code duplication here (in comparison to #925).

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NoticeSchemaGenerator.java
```diff
@@ -0,0 +1,263 @@
...
+    }
+    return fields;
+  }
+
+  private static boolean isSubclassOf(Class<?> parent, Class<?> child) {
+    return !child.equals(parent) && parent.isAssignableFrom(child);
+  }
+
+  @SuppressWarnings("unchecked")
+  private static void maybeAddNoticeClass(Class<?> clazz, List<Class<Notice>> notices) {
```
lionel-nj, line 147: For clarity:

```suggestion
  private static void addNestedNoticeClass(Class<?> clazz, List<Class<Notice>> notices) {
```

core/src/test/java/org/mobilitydata/gtfsvalidator/notice/NoticeSchemaGeneratorTest.java
```diff
@@ -0,0 +1,124 @@
...
+            "fieldValue",
+            String.class,
+            "parentFieldName",
+            String.class,
+            "parentFilename",
+            String.class);
+  }
+
+  @Test
+  public void jsonSchemaForNotice_foreignKeyViolationNotice() {
```
lionel-nj, line 94:
```suggestion
  public void jsonSchemaForNotice_duplicateKeyNotice() {
```

lionel-nj (changes requested):
Thanks @aababilov, a few comments and suggestions in-line.

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NoticeSchemaGenerator.java
aababilov: No, it is not necessarily nested. It is called twice: first time for top-level classes, second for nested ones.

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NoticeSchemaGenerator.java
aababilov: Thanks!

core/src/test/java/org/mobilitydata/gtfsvalidator/notice/NoticeSchemaGeneratorTest.java
aababilov: Good catch! That was bad copypaste :)

aababilov:
Thanks for review!

PR #952, opened 2021-08-11
feat: Return JsonObject from NoticeContainer methods

aababilov:
The caller can decide how to format JSON string:
* pretty printing
* HTML safe (escape <>&=)
* lenient
* etc.


main/src/main/java/org/mobilitydata/gtfsvalidator/cli/Main.java
```diff
@@ -146,18 +148,25 @@ public static void main(String[] argv) {
     System.out.println(feedContainer.tableTotals());
   }
 
+  private static Gson createGson(boolean pretty) {
```
lionel-nj, line 151: Thanks!

lionel-nj (approved):
Thanks @aababilov, LGTM.

Issue #615, opened 2021-01-21
Automatically apply formatting to code


barbeau:
**Is your feature request related to a problem? Please describe.**
As more people contribute to the project, we need a common way of automatically formatting code that doesn't conflict with various contributors.

**Describe the solution you'd like**
We're started using this format:
https://github.com/google/google-java-format

...but by default this needs to be applied manually in your IDE or via the command line.

I looked at integrating this into the Gradle build process via Spotless:
https://github.com/diffplug/spotless/tree/main/plugin-gradle#google-java-format

...but it doesn't integrate into IDEs like IntelliJ unfortunately (an open issue for this is at https://youtrack.jetbrains.com/issue/IDEA-180965).

So it seems like our best bet at this point is to integrate this via GitHub Actions, so the Action would automatically apply the format to PRs:
https://github.com/axel-op/googlejavaformat-action

**Describe alternatives you've considered**
See above for Gradle plugin

barbeau:
Re-opening, as the current implementation fails to run against PRs from forks of the project due to lack of permissions. For example, see https://github.com/MobilityData/gtfs-validator/pull/690.

barbeau:
`pull_request_target` doesn't seem to work - it runs but the formatter doesn't seem any changes - https://github.com/MobilityData/gtfs-validator/pull/698/checks?check_run_id=1778650169. So re-opening. Let's just run on push to master.

barbeau:
Implemented in https://github.com/MobilityData/gtfs-validator/commit/58a361a4d2e9f2e3907e0d947976d9f5f2244f8e.

barbeau:
Re-opening - it seems that because master is a protected branch that requires both the license check as well as the semantic formatting check, we can't push directly to it via the formatting bot. 

Here's our current setup:

![image](https://user-images.githubusercontent.com/928045/106198694-5daa0b80-6182-11eb-9cc8-84559bc42b07.png)

If either is checked the formatting bot fails. If both are unchecked, it succeeds.

barbeau:
I'm trying authenticating the bot with the GitHub token in https://github.com/MobilityData/gtfs-validator/commit/ce71b93b25c14e60b1b91c5cad17e19ee86a5c78, we'll see if this helps.

barbeau:
Closing, as this seems to be working after the last change (see https://github.com/MobilityData/gtfs-validator/commit/2273573156167881fc6b3acbde8725d4e2cf4092).

PR #953, opened 2021-08-12
ci: check code formatting on pull requests

lionel-nj:
relates to https://github.com/MobilityData/gtfs-validator/issues/615

**Summary:**

This PR modifies the `formatting` workflow so that it is executed in pull request rather than on push on `master` branch.

The goal here is to avoid cluttering the commit history of the master branch and make sure that code is formatted before it is merged. 

**Expected behavior:** 

- [x] workflow should be executed on pull requests
- [x] workflow should fail on not properly formatted PRs
- [x] github-action bot should comment instructions to fix the formatting errors 

Capture d’écran, le 2021-08-12 à 17 07 19

Please check the workflow logs to have details about the list of misformatted files (e.g. https://github.com/MobilityData/gtfs-validator/runs/3313714565?check_suite_focus=true)

### DIFF
--- a/.github/workflows/formatting.yml
+++ b/.github/workflows/formatting.yml
@@ -10,8 +10,20 @@ jobs:
   formatting:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2 # v2 minimum required
-      - uses: axel-op/googlejavaformat-action@v3
+      - uses: actions/checkout@v2
+      - name: Set up JDK 1.8
+        uses: actions/setup-java@v1
         with:
-          commitMessage: "chore: Google Java Format"
-          githubToken: ${{ secrets.GITHUB_TOKEN }}
+          java-version: 1.8
+      - name: Cache Gradle packages
+        uses: actions/cache@v2
+        with:
+          path: ~/.gradle/caches
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle') }}
+          restore-keys: ${{ runner.os }}-gradle
+      - name: Check code compliance to google java format standards
+        uses: eskatos/gradle-command-action@v1
+        env:
+          versionTag: ${{ steps.prep.outputs.versionTag }}
+        with:
+          arguments: verGJF
--- a/.github/workflows/formatting.yml
+++ b/.github/workflows/formatting.yml
@@ -37,5 +37,5 @@ jobs:
               issue_number: context.issue.number,
               owner: context.repo.owner,
               repo: context.repo.repo,
-              body: "This contribution does not follow the conventions set by the [Google Java style guide](https://google.github.io/styleguide/javaguide.html). Please [download the formatter](https://github.com/google/google-java-format/releases) and run it with ```java -jar /path/to/google-java-format-1.11.0-all-deps.jar -i $(find . -name '*.java')``` to fix formatting errors."
+              body: "This contribution does not follow the conventions set by the [Google Java style guide](https://google.github.io/styleguide/javaguide.html). Please  run the following command line at the root of the project to fix formatting errors: ```./gradlew verGJF```."
             })
--- a/.github/workflows/formatting.yml
+++ b/.github/workflows/formatting.yml
@@ -28,14 +28,18 @@ jobs:
         with:
           arguments: verGJF
   comment-on-failure:
-    build:
-      name: Comment a pull_request
-      runs-on: ubuntu-latest
-#      if: ${{ (needs.formatting.result=='failure') }}
-#      needs: [ formatting ]
-      steps:
-        - name: github-action-comment-pull-request
-          uses: mb2dev/github-action-comment-pull-request@1.0.0
-          with:
-          message: "This PR code does not follow the standard defined by the Google Java style guide (see https://google.github.io/styleguide/javaguide.html). Please download the formatter at https://github.com/google/google-java-format/releases and run it with ``` -i $(find . -name '*.java')```."
+    runs-on: ubuntu-latest
+    steps:
+      - name: Add comment to PR
+        if: (needs.job1.result=='failure')
+        needs: [formatting]
+        env:
+          URL: ${{ github.event.issue.comments_url }}
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          curl \
+            -X POST \
+            $URL \
+            -H "Content-Type: application/json" \
+            -H "Authorization: token $GITHUB_TOKEN" \
+            --data '{ "body": "Check format." }'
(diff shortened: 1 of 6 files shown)


barbeau (commented):
@lionel-nj See https://github.com/MobilityData/gtfs-validator/issues/615 - I believe that this will fail for PRs using branches from forks due to lack of permissions to write to that branch from the fork. IIRC we tried a variation of this config and had that problem. See the history for this file too - https://github.com/MobilityData/gtfs-validator/commits/fb4939e29c584cccb4b46286e9f8fd27194ceed2/.github/workflows/formatting.yml.

If we want this to run on PRs versus the master, I think we'll just to implement it as a check that fails and then prompts the PR submitter to do the format themselves.

lionel-nj:
> If we want this to run on PRs versus the master, I think we'll just to implement it as a check that fails and then prompts the PR submitter to do the format themselves.

👍🏾

@barbeau I think that  https://github.com/MobilityData/gtfs-validator/pull/953/commits/ab7add78bd7ae08b6e23f96c2b602c51a735d391 and https://github.com/MobilityData/gtfs-validator/pull/953/commits/9bdf080001feab38372e09282a5836b379c1562f fix this.

Capture d’écran, le 2021-08-12 à 12 14 39

The workflow log prompts the user the list of files that are not properly formatted - it no longer tries to commit on a branch:

Capture d’écran, le 2021-08-12 à 12 11 43

barbeau (commented):
@lionel-nj Thanks! This works, but I don't think it's very clear to the PR submitter what to do after it fails. First, they have to dig through the logs to find the error that caused the build to fail, and after than there are no instructions for what to do next.

Things that I think could improve this:
1. Make the output more visible - Bots like the CLAbot comment on the PR to make it clear when the CLA has been signed and hasn't been signed. Ideally a bot would do the same with this.
2. Add clear instructions for what to do next - If a bot in 1) above exists, it could comment with the exact command to run to format the files. If it does not, then adding output to the log with the exact command to run to format the files would be helpful.

barbeau:
Could you include the full `java -jar /path/to/google-java-format-1.11.0-all-deps.jar...`  in the instructions?

core/build.gradle
```diff
@@ -16,6 +16,7 @@
 
 plugins {
     id 'java'
+    id 'com.github.sherter.google-java-format' version '0.9'
```
barbeau, line 19: I believe this should just be moved to the root `build.gradle` instead?

.github/workflows/formatting.yml
```diff
@@ -10,8 +10,32 @@ jobs:
...
+        if: ${{ failure() }}
+        uses: actions/github-script@v4
+        with:
+          github-token: ${{secrets.GITHUB_TOKEN}}
+          script: |
+            github.issues.createComment({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              body: "This contribution does not follow the conventions set by the [Google Java style guide](https://google.github.io/styleguide/javaguide.html). Please [download the formatter](https://github.com/google/google-java-format/releases) and run it with ```java -jar /path/to/google-java-format-1.11.0-all-deps.jar -i $(find . -name '*.java')``` to fix formatting errors."
```
barbeau, line 40: Actually, if we're including a dependency on `com.github.sherter.google-java-format` as part of the project, can't the developer just run `./gradlew goJF` themselves to format the files without needing to download the Google JAR separately?

An alternate approach here is to always run this Gradle task ourselves for every Gradle build, to ensure that any contributions are formatted before the developer commits the files locally. This is definitely more convenient, but I don't know what the build performance hit would be. I believe you could filter the input so that only changed files are formatted on every build, to reduce the processing on each build.

barbeau (commented):
Nice work @lionel-nj! Some thoughts in-line below.

.github/workflows/formatting.yml
lionel-nj: > Actually, if we're including a dependency on com.github.sherter.google-java-format as part of the project, can't the developer just run ./gradlew goJF themselves to format the files without needing to download the Google JAR separately?

Indeed!

An alternate approach here is to always run this Gradle task ourselves for every Gradle build, to ensure that any contributions are formatted before the developer commits the files locally. This is definitely more convenient, but I don't know what the build performance hit would be. I believe you could filter the input so that only changed files are formatted on every build, to reduce the processing on each build, if that's an issue.
```suggestion
              body: "This contribution does not follow the conventions set by the [Google Java style guide](https://google.github.io/styleguide/javaguide.html). Please [download the formatter](https://github.com/google/google-java-format/releases) and run it with ```java -jar /path/to/google-java-format-1.11.0-all-deps.jar -i $(find . -name '*.java')``` to fix formatting errors."
```
Are you talking a sort of pre-commit pre hook?

core/build.gradle
lionel-nj: Addressed in https://github.com/MobilityData/gtfs-validator/pull/953/commits/13bda0f82c0d9d6dbb9290d554d615b29c26150e.

.github/workflows/formatting.yml
barbeau: No, you can tell Gradle to run a task as the final stage in the build pipeline, and the Google Java format task would be the final task.

See https://barbeau.medium.com/how-to-avoid-duplicate-code-for-similar-sdks-7e7812e67708 for an example. In this example, I needed to copy source code files to a different package and use a regex to replace some class names and packages on each build. This way the same source code could be used to support different SDKs that had *slight* differences in the API implements (e.g., Google Maps vs Amazon Maps). So when you tell Gradle to build, it executes the normal build and then these tasks as the end.

lionel-nj: Got it! As discussed we will implement the gradle task you mentioned in a separate PR so that we can handle the case where someone would submit code via the GitHub web editor interface.

.github/workflows/formatting.yml
```diff
@@ -10,8 +10,32 @@ jobs:
...
+        if: ${{ failure() }}
+        uses: actions/github-script@v4
+        with:
+          github-token: ${{secrets.GITHUB_TOKEN}}
+          script: |
+            github.issues.createComment({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              body: "This contribution does not follow the conventions set by the [Google Java style guide](https://google.github.io/styleguide/javaguide.html). Please  run the following command line at the root of the project to fix formatting errors: ```./gradlew goJF```."
```
barbeau, line 40:
```suggestion
              body: "This contribution does not follow the conventions set by the [Google Java style guide](https://google.github.io/styleguide/javaguide.html). Please run the following command line at the root of the project to fix formatting errors: ```./gradlew goJF```."
```

barbeau (approved):
Nice work @lionel-nj! 👍

lionel-nj:
Thanks for reviewing!

PR #955, opened 2021-08-12
fix: Remove unused dependency on jackson-databind

aababilov:
Gson is used instead of Jackson.

Jackson2 has a terrible security record, with new remote code execution
vulnerabilities being published once every other month:
https://medium.com/@cowtowncoder/on-jackson-cves-dont-panic-here-is-what-you-need-to-know-54cd0d6e8062


lionel-nj (approved):
Thanks @aababilov

PR #958, opened 2021-08-13
feat: Serialize S2LatLng fields in Notice subclasses

aababilov:
We will add notices with S2LatLng fields and we need to serialize them
to JSON. An S2LatLng will be serialized as array of two numbers:
latitude and longitude, in degrees:

```
"montreal": [45.505331312, -73.55249779]
```

This is similar to GeoJSON format for points:

```
"geometry": {
    "type": "Point",
    "coordinates": [102.0, 0.5]
}
```


lionel-nj:
Closing and reopening to apply the workflow defined in https://github.com/MobilityData/gtfs-validator/pull/953 that aims at checking code formatting.

lionel-nj (approved):
LGTM, thanks @aababilov.

PR #960, opened 2021-08-16
feat: Validate how locations are referenced from stop_times.stop_id

aababilov:
* every stop (or platform) should have stop times
* every non-stop location (station, entrance etc) may not have stop times

We produce one notice per location to avoid producing too many notices.


aababilov:
I am not updating NOTICES.md since we plan to dissolve/remove this file.

aababilov:
This PR is open-sourcing a validator rule used at Google.

lionel-nj:
Closing and reopening to apply the workflow defined in https://github.com/MobilityData/gtfs-validator/pull/953 that aims at checking code formatting.

RULES.md
```diff
@@ -829,7 +837,7 @@ Trips should be referred to at least once in `stop_times.txt`.
 ##### References:
 * [Original Python validator implementation](https://github.com/google/transitfeed)
 
-#### WrongStopTimeStopLocationTypeNotice
+#### LocationWithUnexpectedStopTimeNotice
```
lionel-nj, line 840: It seems that the alphabetical order is not followed here, do you mind changing the location for this part of the doc please?

lionel-nj (changes requested):
Thanks @aababilov, one question: how are boarding areas treated? From #961 it seems that a boarding area should be allowed to have stop_times only if its parent station has `location_type=0`, if it is not the case then it should not. If said boarding area qualifies to have stop_times, then it should be used. 

@barbeau thoughts?

aababilov:
This is a good question! Our current spec for stop_times.stop_id says:

Referenced locations must be stops/platforms, i.e. their stops.location_type value must be 0 or empty. 

So, boarding areas may not be referenced from stop_times.

lionel-nj:
Makes senses @aababilov! Thanks for working on that. Could you fix [this](https://github.com/MobilityData/gtfs-validator/pull/960#discussion_r691289756) please, and then we'll be good to merge.

RULES.md
aababilov: Done.

lionel-nj (approved):
Thanks @aababilov! LGTM.

PR #968, opened 2021-08-24
feat: Validate field presence and key integrity for translations.txt

aababilov:
Checks that fields and values provided in translations are correct and checks that a
translation references existing entity.

This validator gracefully handles legacy Google translations format. If translations.txt
does not follow the new format, then no validations are performed.

### DIFF
--- a/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TranslationFieldAndReferenceValidator.java
+++ b/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TranslationFieldAndReferenceValidator.java
@@ -1,0 +1,237 @@
+/*
+ * Copyright 2021 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.mobilitydata.gtfsvalidator.validator;
+
+import com.google.common.collect.ImmutableList;
+import java.util.Optional;
+import javax.inject.Inject;
+import org.mobilitydata.gtfsvalidator.annotation.GtfsValidator;
+import org.mobilitydata.gtfsvalidator.notice.MissingRequiredFieldNotice;
+import org.mobilitydata.gtfsvalidator.notice.NoticeContainer;
+import org.mobilitydata.gtfsvalidator.notice.SeverityLevel;
+import org.mobilitydata.gtfsvalidator.notice.ValidationNotice;
+import org.mobilitydata.gtfsvalidator.table.GtfsFeedContainer;
+import org.mobilitydata.gtfsvalidator.table.GtfsTableContainer;
+import org.mobilitydata.gtfsvalidator.table.GtfsTranslation;
+import org.mobilitydata.gtfsvalidator.table.GtfsTranslationTableContainer;
+import org.mobilitydata.gtfsvalidator.table.GtfsTranslationTableLoader;
+
+/**
+ * Validates that translations are provided in accordance with GTFS Specification.
+ *
+ * <p>Checks that fields and values provided in translations are correct and checks that a
+ * translation references existing entity.
+ *
+ * <p>This validator gracefully handles legacy Google translations format. If {@code
+ * translations.txt} does not follow the new format, then no validations are performed.
+ */
+@GtfsValidator
+public class TranslationFieldAndReferenceValidator extends FileValidator {
+  private final GtfsTranslationTableContainer translationTable;
+  private final GtfsFeedContainer feedContainer;
+
+  @Inject
+  TranslationFieldAndReferenceValidator(
+      GtfsTranslationTableContainer translationTable, GtfsFeedContainer feedContainer) {
+    this.translationTable = translationTable;
+    this.feedContainer = feedContainer;
+  }
+
+  @Override
+  public void validate(NoticeContainer noticeContainer) {
+    // The legacy Google translation format does not define `translations.table_name` field.
+    if (!translationTable.getHeader().hasColumn(GtfsTranslationTableLoader.TABLE_NAME_FIELD_NAME)) {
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/mobilitydata/gtfsvalidator/validator/ValidatorLoader.java
```diff
@@ -141,12 +141,18 @@ public ValidatorLoader(ImmutableList<String> validatorPackages) throws Validator
       throws ValidatorLoaderException {
     Constructor<T> constructor = chooseConstructor(validatorClass);
 
+    // Indicates that the full GtfsFeedContainer needs to be injected.
```
lionel-nj, line 144: Could you explain why this change was needed please?

processor/src/main/java/org/mobilitydata/gtfsvalidator/processor/TableContainerGenerator.java
```diff
@@ -201,10 +203,58 @@ public TypeSpec generateGtfsContainerClass() {
...
+          .endControlFlow();
+      method.addStatement(
+          "return Optional.ofNullable($L.getOrDefault(keys[0], null))",
+          byKeyMapName(fileDescriptor.primaryKey().get().name()));
+    } else if (fileDescriptor.sequenceKey().isPresent() && fileDescriptor.firstKey().isPresent()) {
+      method
+          .beginControlFlow("if (keys.length != 2)")
+          .addStatement("return Optional.empty()")
+          .endControlFlow();
+      // TODO(aababilov): Implement this.
```
lionel-nj, line 243: Is it still something to do here?

lionel-nj (changes requested):
Thanks @aababilov for this contribution! I like how this adapts to unofficial tables! You will find a few questions in-line. 
Could you also add documentation for the new classes/methods as well as unit test for `TranslationValidator` please? :)

isabelle-dr:
Thank you for this contribution @aababilov!

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TranslationsValidator.java
```diff
@@ -0,0 +1,184 @@
...
+import org.mobilitydata.gtfsvalidator.notice.ValidationNotice;
+import org.mobilitydata.gtfsvalidator.table.GtfsFeedContainer;
+import org.mobilitydata.gtfsvalidator.table.GtfsTableContainer;
+import org.mobilitydata.gtfsvalidator.table.GtfsTranslation;
+import org.mobilitydata.gtfsvalidator.table.GtfsTranslationTableContainer;
+import org.mobilitydata.gtfsvalidator.table.GtfsTranslationTableLoader;
+
+/**
+ * Validates that translations are provided in accordance with GTFS Specification. Supports two
+ * types of specifications * unofficial Google extension * official GTFS specification Checks that
```
barbeau, line 34: >unofficial Google extension

Is this extension documented somewhere that you could link to from here?

Also, it looks like the auto-formatting destroyed the bulleted list here - I'd suggest re-formatting in a way that keeps the list and complies with the Google Java Format.

barbeau: @lionel-nj As you work on profiles, perhaps pulling checks for the unofficial Google extension into a profile would be a good use case?

core/src/main/java/org/mobilitydata/gtfsvalidator/validator/ValidatorLoader.java
aababilov: Please see usage example in TranslationsValidator.java. translations.table_name may reference any table, so we locate it by name in GtfsFeedContainer.

```
  @Inject
  TranslationsValidator(
      GtfsTranslationTableContainer translationTable, GtfsFeedContainer feedContainer) {
    this.translationTable = translationTable;
    this.feedContainer = feedContainer;
  }
  ...

    Optional<GtfsTableContainer<?>> table = feedContainer.getTable(translation.tableName());
    if (!table.isPresent()) {
      noticeContainer.addValidationNotice(new TranslationUnknownTableNameNotice(translation));
      return;
    }

```

processor/src/main/java/org/mobilitydata/gtfsvalidator/processor/TableContainerGenerator.java
aababilov: Yes, we need to implement this. One way to do it is to introduce a hash map<composite_key, entity>. The existing containers for stop_times and shapes do not support quick lookup by composite key - they just allow to retrieve a list of all entities that have the same first key (e.g., trip_id).

aababilov:
> Thank you for this contribution @aababilov!

My pleasure!

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TranslationsValidator.java
aababilov: Sorry for bad formatting!

Google translation.txt extension is hard to support in open-source since the "Google" translations.txt has different columns:

* `trans_id`
* `lang` (not `language`!)
* `translation`

Fields `table_name`, `field_name` and `language` are required in GTFS standard but not used in Google extension. So, it is not enough to add a plugin for validation of Google translations - instead, the schema has to be patched and officially required fields become optional in order to support Google translations.

aababilov:
I am splitting this change into pieces. https://github.com/MobilityData/gtfs-validator/pull/979

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TranslationsValidator.java
aababilov: Yes, that was a bad copypaste from the internal code. Fixed.

asvechnikov2 (approved):
Thank you for exporting this, looks much better with GtfsFeedContainer! LGTM!

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TranslationFieldAndReferenceValidator.java
```diff
@@ -0,0 +1,237 @@
...
+   * @return whether the field is missing or unexpected
+   */
+  private static boolean isMissingOrUnexpectedField(
+      GtfsTranslation translation,
+      boolean actualPresence,
+      boolean expectedPresence,
+      String fieldName,
+      String fieldValue,
+      NoticeContainer noticeContainer) {
+    if (expectedPresence != actualPresence) {
```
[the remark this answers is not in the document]
aababilov: Sure, good point!

aababilov:
Thanks for review!

Issue #976, opened 2021-08-30
Validator fails to report invalid dates


asvechnikov2:
## **Bug report** 

**Describe the bug**
RowParser doesn't handle cases when date has an incorrect value for instance `20219999`. `GtfsData.fromString()` [calls](https://github.com/MobilityData/gtfs-validator/blob/52434454b747bb4085038f27563e98fee6018ca5/core/src/main/java/org/mobilitydata/gtfsvalidator/type/GtfsDate.java#L49) `LocalDate.of()` method which  may return `DateTimeException`.

**How we reproduce the bug**
Steps to reproduce the behaviour:
1. Add test to [GtfsDateTest](https://github.com/MobilityData/gtfs-validator/blob/52434454b747bb4085038f27563e98fee6018ca5/core/src/test/java/org/mobilitydata/gtfsvalidator/type/GtfsDateTest.java#L36) to parse string `20219999`
2. Run test and observe DateTimeException instead of `IllegalArgumentException`

Alternatively,

1. Modify a date in `calendar.txt` to has incorrect date `20219999`
2. Run validator for the feed
3. Observe no `InvalidDateNotice` issues in a report

**Expected behaviour**
GtfsDate must handle this type of exceptions and report them as `IllegalArgumentException`.

**Observed behaviour**
GtfsDate returns `DateTimeException` which is not known how to handle by RowParser and SystemError reported instead of `InvalidDateNotice`

lionel-nj:
Thanks for flagging @asvechnikov2! It looks like @aababilov fixed this in #977. Thanks 🙏🏾

PR #977, opened 2021-08-30
fix: Raise IllegalArgumentException for invalid date

aababilov:
Fix issue #976

Previously, DateTimeException was raised for dates like 20219999. This
exception was not caught in RowParser:

```
  private <T> T parseAsType(
  ...
    try {
      return parsingFunction.apply(s);
    } catch (IllegalArgumentException | ZoneRulesException e) {

```

We prefer to change GtfsDate instead of RowParser for avoiding
complicating the generic RowParser code.


asvechnikov2:
Thank you, Alexei! The fix looks good!

PR #979, opened 2021-08-30
feat: Return optional from getTableForFilename

aababilov:
Function name was changed to getTableForFilename to make clear that it
accepts file name including ".txt" extension.

This function will be used by translations validator that must be ready
to handle invalid table names, that is why the return type is changed to
an explicit Optional.


lionel-nj (approved):
Thanks @aababilov, LGTM!

PR #992, opened 2021-08-31
feat: Support injection of the whole GtfsFeedContainer

aababilov:
This will be used by GTFS translations validator that needs to work with
arbitrary tables.

flogger-system-backend was added to test dependencies. It is required
for testing ValidatorLoader class that initializes a static
FluentLogger.

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -43,6 +43,7 @@ dependencies {
     implementation 'commons-validator:commons-validator:1.6'
     implementation 'com.googlecode.libphonenumber:libphonenumber:8.12.13'
     implementation 'com.google.flogger:flogger:0.5.1'
+    testImplementation 'com.google.flogger:flogger-system-backend:0.5.1'
     testImplementation group: 'junit', name: 'junit', version: '4.13'
     testImplementation "com.google.truth:truth:1.0.1"
     testImplementation 'com.google.truth.extensions:truth-java8-extension:1.0.1'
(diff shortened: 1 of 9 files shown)


core/build.gradle
```diff
@@ -43,6 +43,7 @@ dependencies {
     implementation 'commons-validator:commons-validator:1.6'
     implementation 'com.googlecode.libphonenumber:libphonenumber:8.12.13'
     implementation 'com.google.flogger:flogger:0.5.1'
+    testImplementation 'com.google.flogger:flogger-system-backend:0.5.1'
```
lionel-nj, line 46: The latest version of this library is 0.6, I suggest we use it (https://mvnrepository.com/artifact/com.google.flogger/flogger-system-backend).

core/src/test/java/org/mobilitydata/gtfsvalidator/validator/ValidatorLoaderTest.java
```diff
@@ -0,0 +1,86 @@
...
+        new GtfsStopTableContainer(TableStatus.EMPTY_FILE, CsvHeader.EMPTY);
+    StopEntityValidator validator =
+        ValidatorLoader.createValidatorWithContext(StopEntityValidator.class, VALIDATION_CONTEXT);
+
+    assertThat(validator.getCountryCode()).isEqualTo(VALIDATION_CONTEXT.countryCode());
+    assertThat(validator.getCurrentDateTime()).isEqualTo(VALIDATION_CONTEXT.currentDateTime());
+  }
+
+  @Test
+  public void createSingleFileValidator_injectsTableContainerAndContext()
```
lionel-nj, line 57: Could we have a unit test that demonstrates that `createSingleFileValidator` does not inject `feedContainer` please?

lionel-nj (changes requested):
Thanks @aababilov for working on that, one comment in line.

core/build.gradle
aababilov: Let's upgrade all versions in a separate PR. We use 0.5.1 everywhere.

PR #993, opened 2021-08-31
refactor: extract methods from main cli 

lionel-nj:
Relates to #734 

**Summary:**

This PR provides support to extract intermediate validation methods. The is needed for #933.

**Expected behavior:** 

Same behavior as before method extraction.

### DIFF
--- a/main/src/main/java/org/mobilitydata/gtfsvalidator/cli/Main.java
+++ b/main/src/main/java/org/mobilitydata/gtfsvalidator/cli/Main.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020 Google LLC, MobilityData IO
+ * Copyright 2020-2021 Google LLC, MobilityData IO
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -118,6 +118,7 @@ public class Main {
 
   /**
    * Prints validation metadata.
+   *
    * @param startNanos start time as nanoseconds
    * @param feedContainer the {@code GtfsFeedContainer}
    */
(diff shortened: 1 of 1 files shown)


main/src/main/java/org/mobilitydata/gtfsvalidator/cli/Main.java
```diff
@@ -157,6 +136,79 @@ public static void main(String[] argv) {
...
+            new DefaultValidatorProvider(validationContext, validatorLoader),
+            noticeContainer);
+    return feedContainer;
+  }
+
+  /**
+   * Performs parsing and sanity checks on CLI arguments.
+   *
+   * @param argv the CLI arguments
+   * @return the {@code Argument} generated after parsing the command line
```
barbeau, line 188: Should this be `Arguments`?

main/src/main/java/org/mobilitydata/gtfsvalidator/cli/Main.java
```diff
@@ -157,6 +136,79 @@ public static void main(String[] argv) {
...
+   * @param validatorLoader the {@code ValidatorLoader} used in the process
+   * @param feedLoader the {@code GtfsFeedLoader} used in the process
+   * @param noticeContainer the {@code NoticeContainer} that will contain {@code Notice}s related to
+   *     the GTFS feed
+   * @param gtfsInput the source of data
+   * @param validationContext the {@code ValidationContext} do be used during validation
+   * @return the {@code GtfsFeedContainer} used in the validation process
+   * @throws InterruptedException if validation process was interrupted
+   */
+  public static GtfsFeedContainer validate(
```
barbeau, line 168: Should this be called `loadAndValidate()` too?

barbeau (commented):
Thanks @lionel-nj, I think this looks good. I think it definitely makes the `main()` easier to read. A few comments in-line.

So for code using gtfs-validator as a library now, the code calling the library would look a lot like the `main()` method, but with that code choosing how to handle exceptions, etc., right? So for example instead of doing `System.exit(1)` when `gtfsInput == null` it might just log the error (or respond with an error in case of a hosted web service) and return.

lionel-nj:
Thanks for reviewing @barbeau.

> So for code using gtfs-validator as a library now, the code calling the library would look a lot like the main() method, but with that code choosing how to handle exceptions, etc., right? So for example instead of doing System.exit(1) when gtfsInput == null it might just log the error (or respond with an error in case of a hosted web service) and return.

Exactly, that is the goal.

barbeau (approved):
👍

PR #994, opened 2021-08-31
feat: Validate level_id presence for elevator endpoints

### LINKED ISSUE #995: IndexOutOfBoundError is thrown when validating feeds with no pathways
## **Bug report** 

**Describe the bug**
IndexOutOfBoundError is thrown when validating feeds with no pathways. 

**How we reproduce the bug**
Steps to reproduce: 
Run the following command line on the master branch: 

`java -jar -u https://openmobilitydata.org/p/abq-ride/52/latest/download -c us -o output`
**Expected behaviour**
No error.

**Observed behaviour**

Error is thrown at runtime.

**Screenshots:** 
Capture d’écran, le 2021-08-31 à 18 29 57

**Environment versions**
- validator version: SNAPSHOT f0e7f117d1f13909c72243eaed57ec0b2054af3a
- Java version: openjdk 11.0.11 2021-04-20
- OS versions: Mac OS 11.5.1

lionel-nj:
**Summary:**

This PR provides support to check: `stops.level_id != null` when `pathways.txt` uses ELEVATOR. This is an implicit check that `levels.txt` is provided when pathways.txt uses ELEVATOR. 

This also fixes #995.

**Expected behavior:** 

- Only one notice per stop should be emitted if the said stop does not have value for `stops.level_id` when `pathways.txt` uses ELEVATOR.
- No error


aababilov:
Thanks for the fix! I would like to remind you to add me as a reviewer to all PRs that you write.

I would like to suggest you to rewrite the whole function and make the code simpler - then you may have avoided this bug from the beginning. One question to answer is: do you really need this multimap?

aababilov:
I came to conclusion that a more strict validation would be appreciated by our users. Here is what I think.

Consider that levels.txt is provided but not used by any platform that has an elevator. Is this a problem? Definitely. Can we report it? Easily!

So, I would like to ask to implement the following rule:

* both endpoints for an elevator pathway must have a level_id assigned.

We do not need to check if levels.txt is present since if stops.level_id is set, then it must reference a valid levels.level_id.

lionel-nj:
> I would like to suggest you to rewrite the whole function and make the code simpler - then you may have avoided this bug from the beginning. One question to answer is: do you really need this multimap?

I used this as a way to avoid checking if each pathway has pathway mode set to `elevator` - we could remove it, but would need to do the additional check for each pathway.

> So, I would like to ask to implement the following rule:
> 
> both endpoints for an elevator pathway must have a level_id assigned.
> We do not need to check if levels.txt is present since if stops.level_id is set, then it must reference a valid levels.level_id.

@aababilov This would come as an additional rule, not as a replacement of `MissingLevelNotice`, right?

aababilov:
> I used this as a way to avoid checking if each pathway has pathway mode set to elevator - we could remove it, but would need to do the additional check for each pathway.

Sorry, I didn't get that. How was that multimap used? 

Would this code do what we need:

```
    if (levels.entityCount() != 0) {
      return;
    }
    GtfsPathway elevatorPathway = null;
    for (GtfsPathway pathway : pathways.getEntities()) {
      if (pathway.pathwayMode().equals(GtfsPathwayMode.ELEVATOR)) {
        elevatorPathway = pathway;
        break;
      }
    }
    if (elevatorPathway != null) {
          new MissingLevelFileNotice(elevatorPathway.csvRowNumber(), elevatorPathway.pathwayId()));
    }
  }
```

aababilov:
> @aababilov This would come as an additional rule, not as a replacement of MissingLevelNotice, right?

I see it as a replacement. I am sure that we need to check if endpoints have level_id assigned, and if we have that check, then there is no need to check if levels.txt is present.

lionel-nj:
> Sorry, I didn't get that. How was that multimap used?

The multimap was used to only iterate over pathways with an elevator attached.


> ```java
>   public void validate(NoticeContainer noticeContainer) {
>     if (levels.entityCount() != 0) {
>       return;
>     }
>     GtfsPathway elevatorPathway = null;
>     for (GtfsPathway pathway : pathways.getEntities()) {
>       if (pathway.pathwayMode().equals(GtfsPathwayMode.ELEVATOR)) {
>         elevatorPathway = pathway;
>         break;
>       }
>     }
>     if (elevatorPathway != null) {
>       noticeContainer.addValidationNotice(
>           new MissingLevelFileNotice(elevatorPathway.csvRowNumber(), elevatorPathway.pathwayId()));
>     }
>   }
> ```

Yes, that works - thanks!

> I see it as a replacement. I am sure that we need to check if endpoints have level_id assigned, and if we have that check, then there is no need to check if levels.txt is present.

Understood, the rule's logic has been modified in https://github.com/MobilityData/gtfs-validator/pull/994/commits/043021ebe13529d783f7c4d4e9b3716c5d508ffc.

lionel-nj:
Thanks for reviewing @aababilov, PTAL.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/MissingLevelIdValidator.java
```diff
@@ -0,0 +1,99 @@
...
+      reportStop(noticeContainer, origin, reported);
+      reportStop(noticeContainer, destination, reported);
+    }
+  }
+
+  private static void reportStop(
+      NoticeContainer noticeContainer, GtfsStop stop, Set<String> reported) {
+    if (stop == null) {
+      return;
+    }
```
aababilov, line 70: This is a very good check! I am actually thinking of returning Optional from byXyz() functions to avoid NullPointerException. And we can also consider to rewrite the Validator into Kotlin (we can translate Java to Kotlin automatically) to avoid NullPointerException completely.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/MissingLevelIdValidator.java
```diff
@@ -0,0 +1,99 @@
...
+    for (GtfsPathway pathway : pathways.getEntities()) {
+      if (!pathway.pathwayMode().equals(GtfsPathwayMode.ELEVATOR)) {
+        continue;
+      }
+      GtfsStop origin = stops.byStopId(pathway.fromStopId());
+      GtfsStop destination = stops.byStopId(pathway.toStopId());
+      reportStop(noticeContainer, origin, reported);
+      reportStop(noticeContainer, destination, reported);
+    }
+  }
```
aababilov, line 64: This is very nice but I can suggest to avoid the temp variables and make the code shorter.

```
    Set<String> elevatorEndpoints = new HashSet<>();
    for (GtfsPathway pathway : pathways.getEntities()) {
      if (pathway.pathwayMode().equals(GtfsPathwayMode.ELEVATOR)) {
        elevatorEndpoints.add(pathway.fromStopId());
        elevatorEndpoints.add(pathway.toStopId());
      }
    }
    for (String stopId : elevatorEndpoints) {
        GtfsStop location = stops.byStopId(stopId);
        if (location != null && !location.hasLevelId()) {
              new MissingLevelIdNotice(location));
        }
    }
```

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/MissingLevelIdValidator.java
```diff
@@ -0,0 +1,99 @@
...
+
+    private final long csvRowNumber;
+    private final String stopId;
+
+    MissingLevelIdNotice(long csvRowNumber, String stopId) {
+      super(SeverityLevel.ERROR);
+      this.csvRowNumber = csvRowNumber;
+      this.stopId = stopId;
+    }
+  }
```
aababilov, line 98: Let's add stopName to the notice (as we do in many others) and pass GtfsStop to the constructor for convenience.

aababilov (changes requested):
Very good, almost there! Thanks for the updates!

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/MissingLevelIdValidator.java
```diff
@@ -0,0 +1,99 @@
...
+import org.mobilitydata.gtfsvalidator.notice.ValidationNotice;
+import org.mobilitydata.gtfsvalidator.table.GtfsPathway;
+import org.mobilitydata.gtfsvalidator.table.GtfsPathwayMode;
+import org.mobilitydata.gtfsvalidator.table.GtfsPathwayTableContainer;
+import org.mobilitydata.gtfsvalidator.table.GtfsStop;
+import org.mobilitydata.gtfsvalidator.table.GtfsStopTableContainer;
+
+/**
+ * Checks that {@code stops.level_id} is provided when pathways.txt uses {@code
+ * GtfsPathwayMode.ELEVATOR}. This is an implicit check that levels.txt is provided when pathways
```
aababilov, line 34: Here and elsewhere: you can add a nice link:

{@link GtfsPathwayMode#ELEVATOR}

Note the octothorp!

aababilov:
Could you also update title of the PR and the description?

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/MissingLevelIdValidator.java
lionel-nj: Thanks @aababilov, modified in https://github.com/MobilityData/gtfs-validator/pull/994/commits/952ad9066e41c5cadffc39d8efb9f88c44e7a3b5.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/MissingLevelIdValidator.java
lionel-nj: > I am actually thinking of returning Optional from byXyz() functions to avoid NullPointerException. 

I agree that this would be a good improvement!

> And we can also consider to rewrite the Validator into Kotlin (we can translate Java to Kotlin automatically) to avoid NullPointerException completely.

@isabelle-dr @barbeau for future reference.

isabelle-dr:
I think this additional check is great, and that there is value in being stricter than the spec here.

One question @aababilov:
If a stop is referenced in pathways.txt with an elevator and has values in level_id but the file levels.txt is missing?
Are we going to have a ForeignKeyViolationNotice emitted?

aababilov:
> refactor: LevelPresenceValidator #994

I decided to change the name of PR since refactoring is altering the code without changing its external behavior, but here we definitely have a major update in behavior.

aababilov:
> If a stop is referenced in pathways.txt with an elevator and has values in level_id but the file levels.txt is missing?
> Are we going to have a ForeignKeyViolationNotice emitted?

Yes, we will have ForeignKeyViolationNotice for the broken references.

aababilov (approved):
Thanks a lot!

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/MissingLevelIdValidator.java
barbeau: 👍 for transition to Kotlin. I've been doing this with most of my Java projects. Automatic translation isn't perfect and requires some manual changes but it's a good start. You can also transition class-by-class, so you don't need to transition the entire project at the same time. IMHO model classes are a good place to start, as they are relatively straightforward and it reduces a lot of boilerplate code.

PR #998, opened 2021-09-01
feat: Validate that a generic node is not dangling

aababilov:
A dangling node has just one incident location in the pathway graph.
Such generic node is useless because there is no benefit in visiting it.


aababilov:
I had to resolve a merge conflict in RULES.md and I also updated NOTICES.md which will lead to another merge conflict with another PR. We really need to do something with that.

isabelle-dr (approved):
Looks good, thanks for updating the documentation!

aababilov:
Thanks for review!

PR #999, opened 2021-09-01
feat: Validate that a pathway is not a loop

aababilov:
A pathway should not be a loop, i.e. should not start and end in the
same location.

### DIFF
--- a/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/PathwayLoopValidator.java
+++ b/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/PathwayLoopValidator.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright 2021 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


aababilov:
I am not modifying RULES.md since that would lead to merge conflicts with the other PRs.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/PathwayLoopValidator.java
```diff
@@ -0,0 +1,50 @@
...
+
+/** Validates that pathway is not a loop, i.e. it does not start and end at the same location. */
+@GtfsValidator
+public class PathwayLoopValidator extends SingleEntityValidator<GtfsPathway> {
+
+  @Override
+  public void validate(GtfsPathway pathway, NoticeContainer noticeContainer) {
+    if (pathway.hasFromStopId()
+        && pathway.hasToStopId()
+        && pathway.fromStopId().equals(pathway.toStopId())) {
```
lionel-nj, line 33: Should we take capitalization into account here using `equalsIgnoreCase`?

lionel-nj (commented):
Thanks for working on this @aababilov, you will find one question in line.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/PathwayLoopValidator.java
aababilov: We don't have to since these are IDs and IDs are case-sensitive (when we perform lookups based on IDs, we don't convert them to lowercase).

isabelle-dr (approved):
Looks good, thank you for this contribution!

lionel-nj:
Thanks for opening #1001 @isabelle-dr.

lionel-nj (approved):
LGTM, thanks @aababilov!

PR #1007, opened 2021-09-10
ci: execute end_to_end workflows on push and pull request events

### LINKED ISSUE #1005: ci: execute end to end workflows before merging PRs
## **Feature request**

`end_to_end` workflow are currenly run after a PR is merged: 
- https://github.com/MobilityData/gtfs-validator/blob/master/.github/workflows/end_to_end.yml
- https://github.com/MobilityData/gtfs-validator/blob/master/.github/workflows/end_to_end_100.yml
- https://github.com/MobilityData/gtfs-validator/blob/master/.github/workflows/end_to_end_big.yml

**Proposed solution**
Run workflows before merging a PR a smoke tests.

lionel-nj:
closes #1005

**Summary:**

This PR modifies the list of events that trigger the execution of end to end workflows.

**Expected behavior:** 

These workflows should be triggered on pull request and push events so that hey can be used as smoke tests before merging to master.

### DIFF
--- a/.github/workflows/end_to_end.yml
+++ b/.github/workflows/end_to_end.yml
@@ -8,6 +8,10 @@ on:
       - 'Dockerfile'
       - '.gitignore'
       - 'LICENSE'
+      - 'docker.yml'
+      - 'formatting.yml'
+      - 'test_pack_dock.yml'
+      - 'triage.yml'
   pull_request:
     branches: [ master ]
     paths-ignore:
@@ -15,6 +19,10 @@ on:
       - 'Dockerfile'
       - '.gitignore'
       - 'LICENSE'
+      - 'docker.yml'
+      - 'formatting.yml'
+      - 'test_pack_dock.yml'
+      - 'triage.yml'
 jobs:
   validate_gradle_wrapper:
     runs-on: ubuntu-latest
--- a/.github/workflows/end_to_end.yml
+++ b/.github/workflows/end_to_end.yml
@@ -2,8 +2,13 @@ name: End to end
 
 on:
   push:
-    branches: [ master, extend-end-to-end ] #<-- replace extend-end-to-end by the name of the agency/publisher
-
+    branches: [ master ]
+    paths-ignore:
+      - '**.md'
+  pull_request:
+    branches: [ master ]
+    paths-ignore:
+      - '**.md'
 jobs:
   validate_gradle_wrapper:
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


.github/workflows/end_to_end.yml
```diff
@@ -2,8 +2,13 @@ name: End to end
...
   push:
-    branches: [ master, extend-end-to-end ] #<-- replace extend-end-to-end by the name of the agency/publisher
-
+    branches: [ master ]
+    paths-ignore:
+      - '**.md'
+  pull_request:
+    branches: [ master ]
+    paths-ignore:
+      - '**.md'
```
maximearmstrong, line 11: Do we want to run these workflows if the modified files are related to Gradle, Docker, Git or the license? As it is right now, modifying `.gitignore` or `Dockerfile` would trigger the 3 end-to-end workflows, and I'm not sure this is the expected behaviour. 

If we do not want that, I'd advocate to go with the notation `paths` instead, which will trigger the workflows only if a file following the paths pattern is modified:
```
    paths:
      - 'core/**'
      - 'docs/**'
      - 'main/**'
      - 'processor/**'
      - 'scripts/**'
```

maximearmstrong (commented):
One question in-line before approval :)

.github/workflows/end_to_end.yml
barbeau: I agree with @maximearmstrong that it would be good to avoid running the tests on files that definitely won't change the build, but I'd prefer to specify excluded files instead of included files. Otherwise when we refactor we need to remember to change this file for the tests to actually run. For example, if we create a new module `newModule`, we might forget to modify this GitHub file, and then the tests don't run (and we might not realize it).

So I'd suggest specifying additional specific file types or specific files to ignore here instead.

maximearmstrong: Good point :) Thanks @barbeau

lionel-nj: Good catch! https://github.com/MobilityData/gtfs-validator/pull/1007/commits/3590a3ccfc52a348fd95002675b7f364e98cb66d disables end to end tests when the following files are modified: 
- `Dockerfile`
- `.gitignore`
- `LICENSE`

> Do we want to run these workflows if the modified files are related to Gradle [...]

> [...] it would be good to avoid running the tests on files that definitely **won't change the build,** 

⬆️ For this reason, I suggest keeping the end to end tests in case gradle build files are modified

maximearmstrong (approved):
LGTM :)

aababilov:
So, now we have extra protection with the tests on real data? That is great! Thanks for working on this!

Issue #1010, opened 2021-09-14
NullPointerException at org.mobilitydata.gtfsvalidator.validator.ParentLocationTypeValidator.validate(ParentLocationTypeValidator.java:64)


markstos:
## **Bug report** 

```
SEVERE: Runtime exception when loading stops.txt
java.lang.NullPointerException
	at org.mobilitydata.gtfsvalidator.validator.ParentLocationTypeValidator.validate(ParentLocationTypeValidator.java:64)
	at org.mobilitydata.gtfsvalidator.validator.ValidatorLoader.invokeSingleFileValidators(ValidatorLoader.java:205)
	at org.mobilitydata.gtfsvalidator.table.GtfsStopTableLoader.load(GtfsStopTableLoader.java:152)
	at org.mobilitydata.gtfsvalidator.table.GtfsFeedLoader.lambda$loadAndValidate$0(GtfsFeedLoader.java:113)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:829)
```

This happens in a private GTFS feed that I can't provide but based on the file name involved, it looks there was a mismatch between the stop ID and the parent ID. This particular feed provider in the past has had a problem using the same ID for stops and the parent station.

**Expected behaviour**

Validator should report validation problems instead of crashing.

**Observed behaviour**

Crasher.

**Environment versions**
- validator version: 2.0.0 
- Java version: openjdk 11.0.12
- OS versions: Arch Linux

lionel-nj:
Thanks @markstos for flagging that and opening the issue. Work has started in https://github.com/MobilityData/gtfs-validator/pull/1011 to fix this bug.

markstos:
I've confirmed this issue is fixed in the latest snapshot. I see foreign key validations in the error report, which is perfect.

markstos:
Thanks for helping Netflix with their transit!

isabelle-dr:
No problem!

aababilov:
My pleasure!

PR #1011, opened 2021-09-15
feat: Return Optional from byXyz methods

aababilov:
Fix issue #1010

Any foreign key reference may be potentially broken. Authors of the
validator rules must bear that in mind and implement appropriate
behaviour.

A considered alternative is to run all foreign key checks separately -
but this is not preventing all bugs. E.g., trips.service_id is
referencing either calendar.txt or calendar_dates.txt.

```
  // This may raise NPE even if all foreign key checks succeded.
  calendarContainer.byServiceId(trip.serviceId())
```

### DIFF
--- a/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTimeTravelSpeedValidator.java
+++ b/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTimeTravelSpeedValidator.java
@@ -73,27 +73,48 @@ public class StopTimeTravelSpeedValidator extends FileValidator {
 
   @Override
   public void validate(NoticeContainer noticeContainer) {
-    final ListMultimap<Long, List<GtfsStopTime>> tripsByHash = ArrayListMultimap.create();
+    final ListMultimap<Long, TripAndStopTimes> tripsByHash = ArrayListMultimap.create();
     for (List<GtfsStopTime> stopTimes : Multimaps.asMap(stopTimeTable.byTripIdMap()).values()) {
       tripTable
           .byTripId(stopTimes.get(0).tripId())
-          .ifPresent(trip -> tripsByHash.put(tripFprint(trip, stopTimes), stopTimes));
+          .ifPresent(
+              trip ->
+                  tripsByHash.put(
+                      tripFprint(trip, stopTimes), new TripAndStopTimes(trip, stopTimes)));
     }
 
-    for (List<List<GtfsStopTime>> trips : Multimaps.asMap(tripsByHash).values()) {
-      final List<GtfsStopTime> stopTimes = trips.get(0);
+    for (List<TripAndStopTimes> trips : Multimaps.asMap(tripsByHash).values()) {
+      final TripAndStopTimes tripAndStopTimes = trips.get(0);
       // All trips belong to the same route.
-      final Optional<GtfsRoute> route =
-          routeTable.byRouteId(tripTable.byTripId(stopTimes.get(0).tripId()).get().routeId());
+      final Optional<GtfsRoute> route = routeTable.byRouteId(tripAndStopTimes.getTrip().routeId());
       if (!route.isPresent()) {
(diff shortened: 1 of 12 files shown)


asvechnikov2:
Looks like this breaks API consistency. We previously implicitly considered that `by<KeyId>` method always returns non-null. I'm surprised that ParentLocationTypeValidator was called at all. The validator probably needs to provide a proper expectations for the validation rules and containers, e.g. all reference integrities are guaranteed, otherwise, `by<KeyId>` method must return Optional object.

aababilov:
What do you want me to do? I am not breaking API, as you can see. The current API may lead to NPE and I have not fixed that yet.

lionel-nj (approved):
Thanks for working on that @aababilov. 

> In future, we may consider:
> 
> run foreign key checks first and then proceed to other validators
> return Optional from byXyzKey methods

Indeed, [here](https://github.com/MobilityData/gtfs-validator/pull/994#discussion_r701498719) you mentioned the possibility to use Kotlin or have `byXyzKey()` return Optional to avoid NPE. Let's discuss these options @barbeau @aababilov @isabelle-dr. Can Kotlin be consumed at Google @asvechnikov2?

barbeau:
The Google I/O 2021 video for ["State of Kotlin on Android"](https://youtu.be/etLUpHvhNZw?t=55) says:
>100+ Google engineering teams using Kotlin on the server...Kotlin powers the servers for some of our biggest products, including Google Voice, Nest, and Google Shopping.

aababilov:
Hi all!

Sorry, I was reluctant to merge this pull request. I really feel awkward that I am fixing a single NPE here but there are many other potential NPEs that we don't track yet.

I would like to prepare another pull request that would run foreign key validators in the first round. Certainly, the new PR will be larger but it will make sure we have systematically fixed the broken link problem now and forever.

aababilov:
Artur and I came to conclusion that the most reliable way to avoid NullPointerException is to use Optional. I have updated my code accordingly - fortunately, the refactoring is not too large.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTimeTravelSpeedValidator.java
```diff
@@ -72,25 +73,51 @@
...
   @Override
   public void validate(NoticeContainer noticeContainer) {
-    final ListMultimap<Long, List<GtfsStopTime>> tripsByHash = ArrayListMultimap.create();
+    final ListMultimap<Long, TripAndStopTimes> tripsByHash = ArrayListMultimap.create();
     for (List<GtfsStopTime> stopTimes : Multimaps.asMap(stopTimeTable.byTripIdMap()).values()) {
-      tripsByHash.put(
-          tripFprint(tripTable.byTripId(stopTimes.get(0).tripId()), stopTimes), stopTimes);
+      tripTable
+          .byTripId(stopTimes.get(0).tripId())
+          .ifPresent(
```
asvechnikov2, line 80: I think this might be simplified a bit. First `tripFprint()` method should be moved into TripAndStopTimes class. Then we can have something like

tripTable
          .byTripId(stopTimes.get(0).tripId())
          .map(trip -> new TripAndStopTimes(trip, stopTimes))
          .ifPresent(tripAndStopTimes -> tripsByHash.put(tripAndStopTimes.tripFprint(), tripAndStopTimes));

asvechnikov2:
Thanks! LGTM!

@lionel-nj
Unfortunately, I don't know a lot about Kotlin. To answer whether we can consume Kotlin will require some more time for an investigation (at least there will be increased maintenance cost, since we don't have Kotlin expertise) , before that it would be really good to have pros/cons we and other companies will get by using Kotlin.

asvechnikov2 (commented):
Sorry, it looks like I don't know how to use GitHub....

aababilov:
No worries - here are my fixes: https://github.com/MobilityData/gtfs-validator/pull/1013

PR #1012, opened 2021-09-21
fix: Catch any runtime exception when parsing a GTFS field

aababilov:
ZoneId.of can throw DateTimeException for certain arguments. To be
completely safe, we are catching any RuntimeException now.


stanislavd-google-com (approved):
LGTM

aababilov:
Thanks!

PR #1013, opened 2021-09-22
fix: Address comments in pull #1011

aababilov:
* move tripFprint method into TripAndStopTimes
* use orElse
* remove Nullable annotation


asvechnikov2:
Thanks! LGTM

PR #1014, opened 2021-09-22
chore: Upgrade to Java 11

aababilov:
Google has finally upgraded to Java 11, so the open-source validator
does not have to use the old Java 8 (see pull #595 and pull #759).

Java memory is reduced from 12 to 8 GiB to make the tests pass.


aababilov:
End-to-end 100 test failed on Ruler, Norway

```
9s
0s
Run java -Xmx12G -Xms12G -jar main/build/libs/*.jar --url http://transitfeeds.com/p/ruter/240/latest/download --output_base output --country_code no --storage_directory ruter.zip
OpenJDK 64-Bit Server VM warning: INFO: os::commit_memory(0x0000000500000000, 12884901888, 0) failed; error='Not enough space' (errno=12)
#
# There is insufficient memory for the Java Runtime Environment to continue.
# Native memory allocation (mmap) failed to map 12884901888 bytes for committing reserved memory.
# An error report file with more information is saved as:
# /home/runner/work/gtfs-validator/gtfs-validator/hs_err_pid1746.log
```

aababilov:
Reducing memory from 12 to 8 GiB made a miracle - now the tests pass!

.github/workflows/docker.yml
```diff
@@ -32,10 +32,10 @@ jobs:
             VERSION_TAG=v${VERSION_TAG}-sha-${GITHUB_SHA::8}-SNAPSHOT
           fi
           echo ::set-output name=versionTag::${VERSION_TAG}
-      - name: Set up JDK 1.8
```
barbeau, line 35: While we're here, the JDK setup syntax has changed (11 instead of 1.11, includes distribution), along with a newer v2 version for the action:

```yml
      - name: Set up JDK 11
        uses: actions/setup-java@v2
        with:
          java-version: '11'
          distribution: 'adopt'
```

I'd suggest we go ahead and update to `setup-java@v2` and the new syntax.

See https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-java-with-gradle

@lionel-nj A separate issue - I just noticed that we're mixing `actions/checkout@v1` and `actions/checkout@v2` in these files too - it should be `actions/checkout@v2` thoughout.

botanize:
Should this PR also update the java version requirement in README.md?

.github/workflows/docker.yml
aababilov: Done.

aababilov:
> Should this PR also update the java version requirement in README.md?

Done.

barbeau (approved):
LGTM

aababilov:
Thank you!

PR #1038, opened 2021-10-13
ci: implement queuing mechanism for acceptance tests

lionel-nj:
**Summary:**

This PR provides support to implement a queueing mechanism dedicated to the acceptance test feature. Said mechanism has to be implemented since running the validator on 1000+ datasets sequentially takes more than 6 hours (on github) which causes the workflow to get cancelled 

**Expected behavior:** 

- [x] CI should be triggered on all commits except if the developer added the keyword "[acceptance test skip]" to its latest commit message
- [x] Older workflows (only `Rule acceptance tests`) should be cancelled when a new one is triggered to avoid queueing too many workflows for a same PR
- [x] Datasets' URL should be fetched from the MobilityDatabase 
- [x] For each dataset (i.e. URL retrieved from the database), the reference version of the validator (from the master branch) should run and produce `reference.json` and `reference_errors.json` 
- [x] For each dataset (i.e. URL retrieved from the database), the snapshot version of the validator (from the branch where this CI process is run) should run and produce `report.json` and `system_errors.json` 

⚙️ The task `get-reports` (from `acceptance_test.yml`) works as follows: 
Once 1000+ urls are retrieved from the Mobility Database as a list of 256 objects (max), each one of these objects are used to define a matrix. The latest generates jobs that run concurrently. Each one of these jobs run `queue_runner.sh` which parses the input as an array of json objects:
```
({"id:"dataset id value-1", "url":"dataset url-1"} {"id:"dataset id value-2", "url":"dataset url-2"} ... {"id:"dataset id value-n", "url":"dataset url-n"})
```

For each one of these dictionaries both snapshot and reference validator are run.

The reports are all saved in a directory called `output` - which is persisted at the end of the job, and uploaded to Google Cloud Storage to ease download.

Capture d’écran, le 2021-10-26 à 13 13 54

### DIFF
--- a/.github/workflows/acceptance_test.yml
+++ b/.github/workflows/acceptance_test.yml
@@ -201,4 +201,3 @@ jobs:
         with:
           path: ${{ github.sha }}/output
           destination: gtfs-validator-reports
-


lionel-nj:
@maximearmstrong  PTAL :)

PR #1039, opened 2021-10-18
feat: Support Hierarchical Vehicle Types in StopTimeTravelSpeedValidator

aababilov:
HVTs are non-standard but widely used. The standard GTFS validator emits
a warning for an HVT but the original value can be retrieved via
GtfsRouteType.routeTypeValue().

There are feeds that run high-speed railway services. The default 200
km/h limit is too low for them, so it is worth to apply the usual 500
km/h limit for railway.

See https://developers.google.com/transit/gtfs/reference/extended-route-types

### DIFF
--- a/main/src/main/java/org/mobilitydata/gtfsvalidator/util/HierarchicalVehicleType.java
+++ b/main/src/main/java/org/mobilitydata/gtfsvalidator/util/HierarchicalVehicleType.java
@@ -1,0 +1,145 @@
+package org.mobilitydata.gtfsvalidator.util;
+
+import org.mobilitydata.gtfsvalidator.table.GtfsRouteType;
+
+/**
+ * Hierarchical Vehicle Type (HVT) codes from European TPEG standard provide a support a more rich
+ * set of vehicle types than the standard GTFS.
+ *
+ * <p>HVT are not a part of the GTFS standard (and probably won't be) but they are widely used. The
+ * standard GTFS validator emits a warning for a hierarchical vehicle type but nevertheless keeps
+ * its constant that can be retrieved via {@code GtfsRoute.routeTypeValue()}.
+ *
+ * <p>This class provides methods for validators that want to support HVT.
+ *
+ * @see <a href="https://developers.google.com/transit/gtfs/reference/extended-route-types">Extended
+ *     GTFS Route Types</a>
+ */
+public class HierarchicalVehicleType {
+
+  /**
+   * Converts hierarchical vehicle type (HVT) to basic GTFS route types.
+   *
+   * <p>If a basic GTFS route type is passed (e.g., 0-7), then it is returned as-is.
+   *
+   * <p>Returns {@link GtfsRouteType#UNRECOGNIZED} for HVT that can't be represented by a basic GTFS
+   * route type.
+   */
+  public static GtfsRouteType toBasicGtfsRouteType(int hierarchicalVehicleType) {
+    if (hierarchicalVehicleType >= 0 && hierarchicalVehicleType <= 7
+        || hierarchicalVehicleType == 11
+        || hierarchicalVehicleType == 12) {
+      return GtfsRouteType.forNumber(hierarchicalVehicleType);
+    }
+
+    switch (hierarchicalVehicleType) {
+      case COMMUNAL_TAXI:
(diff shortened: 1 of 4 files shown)


aababilov:
> Hi @aababilov, thanks for this contribution!

My pleasure!

> Just had a look and I'm not sure I understand what exactly we're doing with the identified HVT values. thinking What exactly will be different if we merge this PR?

StopTimeTravelSpeedValidator will use the correct speed threshold for feeds that use HVT. No extra patching has to be done. UnexpectedEnumValueNotice will be produced for each route that uses HVT (this is just a warning, so validation keeps on going).

> > Just to clarify, from my POV we either can merge this and create an issue to migrate this code to the profile feature once it ready or we can implement this internally and eventually re-introduce it once the profile feature is ready.

Right now, I do not see if profiles may help with HVT. 

* Profiles are not required for parsing feeds with HVT since such feeds are still considered valid (no errors, just warnings about unexpected enum values).
* Full support of HVT requires patching GtfsRouteTypeEnum.java file which is not what profiles can help with.
 
> Since the HVT aren't in the official spec, I'd say implement this internally for now & re-introduce this with custom validation. We could use this case while building the feature and have our first custom validation set (= "profile") as the canonical validator + extended route types since we know it's a need.

All of us see that HVTs are widely used, so there is definitely a demand for supporting HVT feeds. My personal point of view is that HVTs should never become a part of GTFS standard - instead, GTFS community should adopt GTFS-Modes (it is an excellent proposal and it is sad that its official adoption is taking way too much time). However, since HVTs are already here and they are not going to go any time soon, then the canonical GTFS Validator must provide a way to deal with HVTs. This will largely simplify wider adoption of GTFS Validator in the community.

When there is a demand, there should be supply.

barbeau:
I see both sides of the argument here.

Consumers need to continue to accept GTFS data with HVTs until GTFS spec bans unofficial enum values, which could be months, years, or never (and realistically, even a long time after this, given the time it will take consumers to catch up and switch to an official alternative). Currently the validator will use the default `200` kph as the threshold for `FastTravelBetweenFarStopsNotice` for all HVTs (and other other unofficial route types), which could be misleading in many cases. To protect against these cases and avoid showing bad data to users, all apps will need to check these values another way and maintain their own fork of the project. This causes additional duplicative work for all consumers.

On the other hand, the appearance of integrating even partial support for HVTs does deviate the validator from it's "canonical" state. This gives consumers an excuse to keep using HVTs even after a legitimate solution exists. The counter argument is that the validator isn't really "supporting" HVTs, it's just providing enough information to map HVTs to their official route_type counterparts so that `FastTravelBetweenFarStopsNotices` are more accurate. You could also argue that HVTs are a special case because they date back to before the modern GTFS governance structure when Google's extensions were largely seen as "standard". But still, I think appearances matter here. Where do we draw the line in adopting extensions that appear in real-world data?

>Full support of HVT requires patching GtfsRouteTypeEnum.java file which is not what profiles can help with.

If there are certain "profile" related characteristics that result in incompatible Java class definitions, we could explore implementing these via Gradle configurations.

On Android the Gradle plugin natively supports [product flavors](https://developer.android.com/studio/build/build-variants#product-flavors) so you can produce two different builds of an app that share a lot of code but have a few class definition differences (e.g., a "free" vs "paid" app). I [wrote a blog post](https://barbeau.medium.com/how-to-avoid-duplicate-code-for-similar-sdks-7e7812e67708) on how I've used product flavors to support multiple SDKs from the same application (e.g., Google and Amazon Maps SDKs), which includes additional configuration that copies and re-writes class definitions where the only difference is import statements from different packages.

Apparently pure Gradle on desktop doesn't directly support product flavors (that's an Android thing), but you can accomplish the same thing via different `sourceSets` and configuring the JAR output, and it looks like there [are 3rd party plugins](https://github.com/uklance/gradle-java-flavours) that make this easy.

This approach would result in outputting one JAR per profile, which I believe was different from the original approach, but would allow different schema definitions, for example.

@aababilov One option could be to port Google's internal fork of the project to a `google` product flavor in this repo (assuming @isabelle-dr and @lionel-nj agree). That way you wouldn't need to manage a fork and merge conflicts, etc. and others might be able to contribute to/use that same code.

PR #1022, opened 2021-10-04
chore: upgrade StopTimeTimepointWithoutTimesNotice to ERROR

lionel-nj:
## Please review #1044 before

closes #1017 

**Summary:**

This PR provides support to upgrade `StopTimeTimepointWithoutTimeNotice` to ERROR.

**Expected behavior:** 

- `StopTimeTimepointWithoutTimeNotice`: 

> |  Field Name | Type | Presence | Description |
> |  ------ | ------ | ------ | ------ |
> |  `trip_id` | ID referencing `trips.trip_id` | **Required** | Identifies a trip.  |
> |  `arrival_time` | Time | **Conditionally Required** | Arrival time at the stop (defined by `stop_times.stop_id`) for a specific trip (defined by `stop_times.trip_id`). <br><br>If there are not separate times for arrival and departure at a stop, `arrival_time` and `departure_time` should be the same. <br><br>For times occurring after midnight on the service day, enter the time as a value greater than 24:00:00 in HH:MM:SS local time for the day on which the trip schedule begins.<br><br> If exact arrival and departure times (`timepoint=1` or empty) are not available, estimated or interpolated arrival and departure times (`timepoint=0`) should be provided.<br><br>Conditionally Required:<br>- **Required** for the first and last stop in a trip (defined by `stop_times.stop_sequence`). <br>- **Required** for `timepoint=1`.<br>- Optional otherwise.|
> |  `departure_time` | Time | **Conditionally Required** | Departure time from the stop (defined by `stop_times.stop_id`) for a specific trip (defined by `stop_times.trip_id`).<br><br>If there are not separate times for arrival and departure at a stop, `arrival_time` and `departure_time` should be the same. <br><br>For times occurring after midnight on the service day, enter the time as a value greater than 24:00:00 in HH:MM:SS local time for the day on which the trip schedule begins.<br><br> If exact arrival and departure times (`timepoint=1` or empty) are not available, estimated or interpolated arrival and departure times (`timepoint=0`) should be or
> 

_https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md#stop_timestxt_

Please make sure these boxes are checked before submitting your pull request - thanks!

- [ ] Run the unit tests with `gradle test` to make sure you didn't break anything
- [ ] Format the title like "feat: [new feature short description]". Title must follow the Conventional Commit Specification(https://www.conventionalcommits.org/en/v1.0.0/).
- [ ] Linked all relevant issues
[...]

Issue #954, opened 2021-08-12
False positives for StopTimeTimepointWithoutTimesNotice


mcplanner-zz:
## **Bug report** 

**Describe the bug**
The StopTimeTimepointWithoutTimesNotice error is displayed when the stop_times.timepoint field is blank. This is unexpected. I would expect this to be interpreted as having no timepoints.

**How we reproduce the bug**
Steps to reproduce the behaviour:
This happened with this field: http://iportal.sacrt.com/GTFS/Unitrans/google_transit.zip

**Expected behaviour**
If the timepoint field isn't being used, I expect an error about it's being missing.

**Observed behaviour**
I was alerted that every timepoint was lacking an arrival and departure time.

**Screenshots:** 

**Environment versions**
Java 11, v2.0.0 validator, https://github.com/cal-itp/gtfs-validator

lionel-nj:
Thanks @mcplanner for opening.

> The StopTimeTimepointWithoutTimesNotice error is displayed when the stop_times.timepoint field is blank. This is unexpected. I would expect this to be interpreted as having no timepoints.

The GTFS specification says that if `stop_times.timepoint = 1 or empty` then the times are considered exact. From my understanding a stop time with no value for `timepoint` and no value for `stop_times.arrival_time` or `stop_times.departure_time` should issue a `StopTimeTimepointWithoutTimesNotice`.

See [stop_times.txt reference](https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md#stop_timestxt).

> If the timepoint field isn't being used, I expect an error about it's being missing.

I am not sure if a notice should be generated here, the specification handles the case where no value is attributed to `stop_times.timepoint`: default value is 1.

barbeau:
Here's example data from the above GTFS file:

| trip_id | arrival_time | departure_time | stop_id | stop_sequence | stop_headsign | pickup_type | drop\_off\_type | shape\_dist\_traveled |
| --- | --- | --- | --- | --- | --- | --- | --- | --- |
| 784678 | 07:25:00 | 07:25:00 | 22256 | 1   |     |     |     |     |
| 784678 |     |     | 22239 | 2   |     |     |     |     |
| 784678 | 07:28:00 | 07:28:00 | 22361 | 3   |     |     |     |     |
| 784678 |     |     | 22225 | 4   |     |     |     |     |

[...]

PR #1044, opened 2021-10-26
fix: False positives for StopTimeTimepointWithoutTimesNotice 

lionel-nj:
**Summary:**

This PR provides support to no longer throw false positives in `TimepointValidator`.

**Expected behavior:** 

### **sample data 1** (no timepoint column)
stop_sequence | arrival_time | departure_time
-- | -- | --
1 | 00:00:00 | 00:00:00
2 |   |  
3 |   |  
4 | 00:10:00 | 00:10:00

`MissingTimepointColumnNotice` should be issued once.

### **sample data 2** (timepoint column + no empty values)
stop_sequence | arrival_time | departure_time | timepoint
-- | -- | -- | --
1 | 00:00:00 | 00:00:00 | 1
2 | 00:02:00 | 00:02:00 | 0
3 | 00:08:00 | 00:08:00 | 0
4 | 00:10:00 | 00:10:00 | 1

No notice

### **sample data 3** (timepoint with no times)
stop_sequence | arrival_time | departure_time | timepoint
-- | -- | -- | --
1 | 00:00:00 | 00:00:00 | 1
2 |  |  | 1
3 | 00:08:00 | 00:08:00 | 0
4 | 00:10:00 | 00:10:00 | 1

`StopTimeTimepointWithoutTimesNotice` should be issued for the 2nd row. 

### **sample data 4** (approximate, with times)
stop_sequence | arrival_time | departure_time | timepoint
-- | -- | -- | --
1 | 00:00:00 | 00:00:00 | 1
2 | 00:02:00 | 00:02:00 | 
3 | 00:08:00 | 00:08:00 | 
4 | 00:10:00 | 00:10:00 | 1

`MissingTimepointValueNotice` should be issued for the 2nd and 3rd row. 

### **sample data 5** (approximate, with no times)
stop_sequence | arrival_time | departure_time | timepoint
-- | -- | -- | --
1 | 00:00:00 | 00:00:00 | 1
2 |  |  | 
3 |  |  | 
4 | 00:10:00 | 00:10:00 | 1

`MissingTimepointValueNotice` should be issued for the 2nd and 3rd row. 


main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TimepointTimeValidator.java
```diff
@@ -32,13 +32,17 @@
...
  *       or departure_time
+ *   <li>{@link MissingTimepointValue} - value for {@code stop_times.timepoint} is missing
  * </ul>
  */
 @GtfsValidator
 public class TimepointTimeValidator extends SingleEntityValidator<GtfsStopTime> {
 
   @Override
   public void validate(GtfsStopTime stopTime, NoticeContainer noticeContainer) {
+    if (!stopTime.hasTimepoint()) {
```
maximearmstrong, line 43: Does the validator fill the column with a default value if it is not in the dataset, or are we not entering this `TimepointTimeValidator`?

maximearmstrong (changes requested):
Few questions and comments in-line before approval :)

isabelle-dr:
One question here: 
If the Sample data 1, it's written 

> MissingTimepointValueNotice should be issued for the 2nd and 3rd row.

It's not exactly what is mentioned [here](https://github.com/MobilityData/gtfs-validator/issues/954#issuecomment-952173254), is this a typo in the description of this PR? We talked about throwing the `MissingTimepointValueNotice` only if the column is present.

Also: do you need real datasets that fall in those different categories to test this PR?

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TimepointTimeValidator.java
lionel-nj: > Does the validator fill the column with a default value if it is not in the dataset,

Yes, default is `1`.

lionel-nj:
> It's not exactly what is mentioned here, is this a typo in the description of this PR? We talked about throwing the MissingTimepointValueNotice only if the column is present.

Indeed, this will be fixed in the next commit. Thanks for flagging that.

> Also: do you need real datasets that fall in those different categories to test this PR?

I'll use the one you flagged in this [document](https://docs.google.com/spreadsheets/d/1Uf4NJ8py6-IgeDgSPSDMkilE-nAi67pTLCnC5vGxuyk/edit#gid=0), thanks.

lionel-nj:
@maximearmstrong PTAL - the rule logic has been changed a bit to generate `MissingTimepointColumnNotice` when a dataset has no field `timepoint` in GTFS file `stop_times.txt. 

At present:
- 44 datasets trigger `StopTimeTimepointWithoutTimesNotice`;
- 560 datasets trigger `MissingTimepointColumnNotice`;
- 125 datasets trigger `MissingTimepointValueNotice`.

isabelle-dr:
That `StopTimeTimepointWithoutTimesNotice` number makes a lot more sense!

The number of `MissingTimepointColumnNotice` is surprisingly high. 
After looking at a few datasets that trigger it, I found a false positive 😩, [here it the dataset](http://mobilitydatabase.org/wiki/Item:Q686).

RULES.md
```diff
@@ -131,6 +131,8 @@ Additional details regarding the notices' context is provided in [`NOTICES.md`](
 | [`InconsistentAgencyLangNotice`](#InconsistentAgencyLangNotice)                   	| Inconsistent language among agencies.                                                                                                                       	|
 | [`LeadingOrTrailingWhitespacesNotice`](#LeadingOrTrailingWhitespacesNotice)         | The value in CSV file has leading or trailing whitespaces.                                                                                                  	|
 | [`MissingFeedInfoDateNotice`](#MissingFeedInfoDateNotice)                         	| `feed_end_date` should be provided if `feed_start_date` is provided. `feed_start_date` should be provided if `feed_end_date` is provided.                   	|
+| [`MissingTimepointColumn`](#MissingTimepointColumn)                         	        | `timepoint` field is missing for a dataset.                                                                                                        	        |
```
isabelle-dr, line 134: MissingTimePointColumnNotice

RULES.md
```diff
@@ -131,6 +131,8 @@ Additional details regarding the notices' context is provided in [`NOTICES.md`](
...
+| [`MissingTimepointValue`](#MissingTimepointValue)                         	        | `stop_times.timepoint` value is missing for a record.                                                                                                        	|
```
isabelle-dr, line 135: MissingTimePointValueNotice

RULES.md
```diff
@@ -735,6 +737,24 @@ Even though `feed_info.start_date` and `feed_info.end_date` are optional, if one
 
 ##### References:
 * [feed_info.txt Best practices](http://gtfs.org/best-practices/#feed_infotxt)
+ 
+<a name="MissingTimepointValue"/>
+
+#### MissingTimepointValue
```
isabelle-dr, line 743: MissingTimepointValueNotice

RULES.md
```diff
@@ -735,6 +737,24 @@ Even though `feed_info.start_date` and `feed_info.end_date` are optional, if one
...
+
+Even though the column `timepoint` is optional in `stop_times.txt` according to the specification, `stop_times.timepoint` should not be empty when provided. 
+
+##### References:
+* [stop_times.txt specification](https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md#stop_timestxt)
+
+<a name="MissingTimepointColumn"/>
+
+#### MissingTimepointColumn
```
isabelle-dr, line 752: MissingTimepointColumnNotice

RULES.md
```diff
@@ -735,6 +737,24 @@ Even though `feed_info.start_date` and `feed_info.end_date` are optional, if one
 
 ##### References:
 * [feed_info.txt Best practices](http://gtfs.org/best-practices/#feed_infotxt)
+ 
+<a name="MissingTimepointValue"/>
```
isabelle-dr, line 741: MissingTimepointValueNotice

RULES.md
```diff
@@ -735,6 +737,24 @@ Even though `feed_info.start_date` and `feed_info.end_date` are optional, if one
...
+
+#### MissingTimepointValue
+
+Even though the column `timepoint` is optional in `stop_times.txt` according to the specification, `stop_times.timepoint` should not be empty when provided. 
+
+##### References:
+* [stop_times.txt specification](https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md#stop_timestxt)
+
+<a name="MissingTimepointColumn"/>
```
isabelle-dr, line 750: MissingTimepointColumnNotice

docs/NOTICES.md
```diff
@@ -675,6 +675,8 @@
 | `inconsistent_agency_lang`                 	| [`InconsistentAgencyLangNotice`](#InconsistentAgencyLangNotice)                   	|
 | `leading_or_trailing_whitespaces`           | [`LeadingOrTrailingWhitespacesNotice`](#LeadingOrTrailingWhitespacesNotice)         |
 | `missing_feed_info_date`                   	| [`MissingFeedInfoDateNotice`](#MissingFeedInfoDateNotice)                         	|
+| `missing_timepoint_column`                   	| [`MissingTimepointColumn`](#MissingTimepointColumn)                         	        |
```
isabelle-dr, line 678: MissingTimepointColumnNotice

docs/NOTICES.md
```diff
@@ -675,6 +675,8 @@
...
+| `missing_timepoint_value`                   	| [`MissingTimepointValue`](#MissingTimepointValue)                         	        |
```
isabelle-dr, line 679: MissingTimePointValueNotice

docs/NOTICES.md
```diff
@@ -807,6 +809,28 @@
 ##### Affected files
 * [`feed_info.txt`](http://gtfs.org/reference/static#feed_infotxt)
 
+#### [MissingTimepointColumn](/RULES.md#MissingTimepointColumn)
```
isabelle-dr, line 812: MissingTimePointColumnNotice

docs/NOTICES.md
```diff
@@ -807,6 +809,28 @@
...
+##### Fields description
+
+| Field name     	| Description                                     	| Type   	|
+|----------------	|-------------------------------------------------	|--------	|
+| `filename`    	| The name of the affected file.                  	| String   	|
+
+##### Affected files
+* [`stop_times.txt`](https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md#stop_timestxt)
+
+#### [MissingTimepointValue](/RULES.md#MissingTimepointValue)
```
isabelle-dr, line 822: MissingTimepointValueNotice

isabelle-dr (changes requested):
Could you rename the notice names? :) 
- MissingTimepointColumn -> MissingTimePointColumnNotice
- MissingTimepointValue -> MissingTimepointValueNotice

lionel-nj:
> That `StopTimeTimepointWithoutTimesNotice` number makes a lot more sense!
> 
> The number of `MissingTimepointColumnNotice` is surprisingly high. After looking at a few datasets that trigger it, I found a false positive 😩, [here it the dataset](http://mobilitydatabase.org/wiki/Item:Q686).

After investigation it turns out that the url found by the script does not lead to the last dataset added to the database in all cases insofar as the datasets are not sorted in chronological order when harvesting a data source. data (openmobilitydata.org in this case).
The situation is as follows: in the `acceptance_test` workflow for the source [` charm-city-circ-gtfs-q686`](http://mobilitydatabase.org/wiki/Item:Q686) the validator is executed not on the latest version (dated 2021-06-15) but on an earlier version (dated of 2019-07-29). For reference here are the datasets concerned, the behavior of the validator as well as the associated validation report:
- [dataset of 06-15-2021](https://storage.googleapis.com/storage/v1/b/charm-city-circ-gtfs-q686_archives_2021-06-15/o/48da9d0f991702f0fc17bf431507a31c74a4ba20.zip?alt=media ): the validator does not issue any notice concerning the absence of the timepoint column, which is the expected behavior;
[download report here](https://github.com/MobilityData/gtfs-validator/files/7463776/report-2019.zip)

```json 
{
  "notices": [
    {
      "code": "invalid_phone_number",
      "severity": "ERROR",
      "totalNotices": 1,
      "sampleNotices": [
        {
          "filename": "agency.txt",
          "csvRowNumber": 2,
          "fieldName": "agency_phone",
          "fieldValue": "410.545.1956"
        }
      ]
    },
    {
      "code": "missing_feed_info_date",
      "severity": "WARNING",
      "totalNotices": 1,
      "sampleNotices": [
        {
          "csvRowNumber": 2,
          "fieldName": "feed_end_date"
        }
      ]
    },
    {
      "code": "unknown_column",
      "severity": "INFO",
      "totalNotices": 1,
      "sampleNotices": [
        {
          "filename": "trips.txt",
          "fieldName": "block_name",
          "index": 11
        }
      ]
    }
  ]
}
```

- [dataset of 07-29-2019](https://storage.googleapis.com/storage/v1/b/charm-city-circ-gtfs-q686_archives_2019-07-29/o/c2638df20c382fb5f5b79e228c6e99ba350d069e.zip?alt=media ): the validator issues a notice concerning the absence of the timepoint column, which is the expected behavior.
[download report here](https://github.com/MobilityData/gtfs-validator/files/7463781/report-2021.json.zip)

```json 
{
  "notices": [
    {
      "code": "invalid_phone_number",
      "severity": "ERROR",
      "totalNotices": 1,
      "sampleNotices": [
        {
          "filename": "agency.txt",
          "csvRowNumber": 2,
          "fieldName": "agency_phone",
          "fieldValue": "410-350-0456"
        }
      ]
    },
    {
      "code": "missing_feed_info_date",
      "severity": "WARNING",
      "totalNotices": 1,
      "sampleNotices": [
        {
          "csvRowNumber": 2,
          "fieldName": "feed_end_date"
        }
      ]
    },
    {
      "code": "missing_timepoint_column",
      "severity": "WARNING",
      "totalNotices": 1,
      "sampleNotices": [
        {
          "filename": "stop_times.txt"
        }
      ]
    }
  ]
}
```

It it probable that this behavior repeats for other data sources, @maximearmstrong and I will work on improving the script used to retrieve the latest versions of datasets stored on the database.

isabelle-dr (approved):

isabelle-dr:
Sounds good, thanks for looking into this @lionel-nj!
I think we're good to merge this PR because those three new rules work the way we want and are solving the problem for the false positives flagged in the issue linked.

@asvechnikov2 or @aababilov, can we get a review?

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TimepointTimeValidator.java
```diff
@@ -32,56 +34,99 @@
...
+      this.stopSequence = stopTime.stopSequence();
+    }
+  }
+
+  /** Column {@code stop_times.timepoint} is missing. */
+  static class MissingTimepointColumnNotice extends ValidationNotice {
+    private final String filename;
+
+    MissingTimepointColumnNotice() {
+
```
asvechnikov2, line 127: nit: extra new line

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/TimepointTimeValidatorTest.java
```diff
@@ -17,110 +17,353 @@
...
+    List<GtfsStopTime> stopTimes = new ArrayList<>();
+    stopTimes.add(
+        new GtfsStopTime.Builder()
+            .setCsvRowNumber(1)
+            .setTripId("first trip id")
+            .setArrivalTime(null)
+            .setDepartureTime(null)
+            .setStopId("stop id 0")
+            .setStopSequence(2)
+            .setTimepoint((Integer) null)
```
asvechnikov2, line 109: Why do you need to explicitly set values to null? The same for other tests as well

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/TimepointTimeValidatorTest.java
```diff
@@ -17,110 +17,353 @@
...
-                    .setCsvRowNumber(1)
-                    .setTripId("first trip id")
-                    .setArrivalTime(null)
-                    .setDepartureTime(GtfsTime.fromSecondsSinceMidnight(580))
-                    .setStopId("stop id")
-                    .setStopSequence(2)
-                    .setTimepoint(0)
-                    .build()))
-        .isEmpty();
+  public void emptyTimepoint_timesProvided_shouldNotGenerateNotice() {
```
asvechnikov2, line 344: Should or should not it generate notice?

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/TimepointTimeValidatorTest.java
```diff
@@ -17,110 +17,353 @@
...
-                new GtfsStopTime.Builder()
-                    .setCsvRowNumber(1)
-                    .setTripId("first trip id")
-                    .setArrivalTime(GtfsTime.fromSecondsSinceMidnight(450))
-                    .setDepartureTime(null)
-                    .setStopId("stop id")
-                    .setStopSequence(2)
-                    .setTimepoint(1)
-                    .build()))
+  public void timepoint_missingDepartureTimeShouldGenerateNotice() {
```
asvechnikov2, line 209: could you make or change a test to make sure that not only the first stop time may generate a notice?

asvechnikov2 (changes requested):
There seems to be a discrepancy between `sample data 4` in the [original discussion](https://github.com/MobilityData/gtfs-validator/issues/954#issuecomment-952173254) and `sample data 4` in the fix here. What is the source of truth?

isabelle-dr:
Thanks for the review @asvechnikov2!

> There seems to be a discrepancy between sample data 4 in the original discussion and sample data 4 in the fix here. What is the source of truth?

We have three cases that trigger different notices:
1. times are missing when timepoint = 1 -> `StopTimeTimepointWithoutTimesNotice`
2. the timepoint column is present and the value is missing for a record (regardless of the times) -> `MissingTimepointValueNotice`
3. the timepoint column is missing -> `MissingTimepointColumnNotice`

The tables in the original discussion are only showing the expectations for the 1rst rule, and the tables here are showing the expectation for all three. It is mentioned at the bottom of the [original discussion comment](https://github.com/MobilityData/gtfs-validator/issues/954#issuecomment-952173254)! 

Please note that after we merge this PR, we'd like to see if it's possible to [upgrade the `StopTimeTimepointWithoutTimesNotice`](https://github.com/MobilityData/gtfs-validator/pull/1022) in order to solve discrepancies with the spec. With the fix done in this PR, this notice goes from being present in 150 datasets to 42!

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/TimepointTimeValidatorTest.java
lionel-nj: Should, I renamed the test to match the expected behavior. Thanks for flagging that.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TimepointTimeValidator.java
```diff
@@ -32,56 +34,99 @@
...
+              new StopTimeTimepointWithoutTimesNotice(
+                  stopTime, GtfsStopTimeTableLoader.DEPARTURE_TIME_FIELD_NAME));
+        }
+      }
     }
   }
 
   private boolean isTimepoint(GtfsStopTime stopTime) {
-    return stopTime.timepoint().equals(GtfsStopTimeTimepoint.EXACT);
+    return stopTime.hasTimepoint() && stopTime.timepoint().equals(GtfsStopTimeTimepoint.EXACT);
```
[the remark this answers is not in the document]
lionel-nj: Actually because of the fact that `timepoint` has a default value we need to check `stopTime.hasTimepoint()`. See, if a `stop_time` has no value for timepoint or has value `1` for `timepoint`, `stopTime.timepoint()` return the same value `EXACT`. Using `stopTime.hasTimepoint()` makes it possible to distinguish these two situations.

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/TimepointTimeValidatorTest.java
lionel-nj: The goal is to test what happens when no value is provided for the field `timepoint` - the validator parses an empty csv cell as an `null` object when it comes to setting `stopTime.timepoint`, which is why `null` is used here. 

See, method `asEnum()` from `RowParser`: https://github.com/MobilityData/gtfs-validator/blob/60fc8561095734417f8d198ca73da41a3180de39/core/src/main/java/org/mobilitydata/gtfsvalidator/parsing/RowParser.java#L304

RULES.md
```diff
@@ -735,6 +737,24 @@ Even though `feed_info.start_date` and `feed_info.end_date` are optional, if one
...
+Even though the column `timepoint` is optional in `stop_times.txt` according to the specification, `stop_times.timepoint` should not be empty when provided. 
+
+##### References:
+* [stop_times.txt specification](https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md#stop_timestxt)
+
+<a name="MissingTimepointColumnNotice"/>
+
+#### MissingTimepointColumnNotice
+
+The `timepoint` field should be provided.
```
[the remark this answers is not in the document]
lionel-nj: Let's use `column`, modified in https://github.com/MobilityData/gtfs-validator/pull/1044/commits/2c62b0a211fbdfa4e2d68a776968454809dd8ab0.

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/TimepointTimeValidatorTest.java
lionel-nj: Addressed in https://github.com/MobilityData/gtfs-validator/pull/1044/commits/2c62b0a211fbdfa4e2d68a776968454809dd8ab0.

lionel-nj:
Thanks for reviewing @asvechnikov2, changes have been provided, PTAL.

lionel-nj:
> @maximearmstrong PTAL - the rule logic has been changed a bit to generate `MissingTimepointColumnNotice` when a dataset has no field `timepoint` in GTFS file `stop_times.txt`.
> 
> At present, out of ~1200 datasets on the MobilityDatabse:
> 
> * 44 datasets trigger `StopTimeTimepointWithoutTimesNotice`;
> * 560 datasets trigger `MissingTimepointColumnNotice` (legacy datasets);
> * 125 datasets trigger `MissingTimepointValueNotice`.

@isabelle-dr  After adapting the script to retrieve the latest versions of all datasets in the MobilityDatabase, the numbers are the following:

* 44 datasets trigger `StopTimeTimepointWithoutTimesNotice`;
* 472 datasets trigger `MissingTimepointColumnNotice` (legacy datasets);
* 118 datasets trigger `MissingTimepointValueNotice`.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TimepointTimeValidator.java
```diff
@@ -32,56 +34,97 @@
...
-              GtfsStopTimeTableLoader.DEPARTURE_TIME_FIELD_NAME));
+    for (GtfsStopTime stopTime : stopTimes.getEntities()) {
+      if (!stopTime.hasTimepoint()) {
+        noticeContainer.addValidationNotice(new MissingTimepointValueNotice(stopTime));
+      }
+      if (isTimepoint(stopTime)) {
+        if (!stopTime.hasArrivalTime()) {
+          noticeContainer.addValidationNotice(
+              new StopTimeTimepointWithoutTimesNotice(
+                  stopTime, String.format(GtfsStopTimeTableLoader.ARRIVAL_TIME_FIELD_NAME)));
```
asvechnikov2, line 68: nit: String.format probably is not needed?

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TimepointTimeValidator.java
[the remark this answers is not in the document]
asvechnikov2: I see, in this case should we remove default value instead?

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TimepointTimeValidator.java
```diff
@@ -32,56 +34,97 @@
...
+
+    MissingTimepointValueNotice(GtfsStopTime stopTime) {
+      super(SeverityLevel.WARNING);
+      this.csvRowNumber = stopTime.csvRowNumber();
+      this.tripId = stopTime.tripId();
+      this.stopSequence = stopTime.stopSequence();
+    }
+  }
+
+  /** Column {@code stop_times.timepoint} is missing. */
```
asvechnikov2, line 121: nit: docstring is missing Severity

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/TimepointTimeValidatorTest.java
asvechnikov2: Thanks for the explanation! 

It looks like we may need to create a feature request to have an ability to specify test entities in the csv format, otherwise, we're trying to repeat logic of auto generated code, which may change and we end up with wrong tests.

asvechnikov2 (changes requested):

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TimepointTimeValidator.java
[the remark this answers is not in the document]
lionel-nj: I could see this happening after the specification is clarified to avoid making the behavior of the validator  contradictory/not in accordance with the specs. Thoughts @barbeau @isabelle-dr?

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/TimepointTimeValidatorTest.java
barbeau: >It looks like we may need to create a feature request to have an ability to specify test entities in the csv format, otherwise, we're trying to repeat logic of auto generated code, which may change and we end up with wrong tests.

💯  I'd really like to see this. This would allow us to directly use real-world data that generates false-positives, etc. in the tests, reducing the likelihood of translation errors when trying to write code for the equivalent.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TimepointTimeValidator.java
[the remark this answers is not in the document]
isabelle-dr: Agreed to remove default value.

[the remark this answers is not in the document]
barbeau: >I see, in this case should we remove default value instead?

So if we remove `@DefaultValue("1")`, the new value of this field would be `GtfsStopTimeTimepoint.UNRECOGNIZED` if the timepoint column is missing or the field has an empty value in the CSV?

I'm still not sure that's what we want.

If we define a default, I think the actual behavior we would want is that the default is set to `EXACT` IFF the timepoint column isn't provided - this matches the original GTFS spec before the `timepoint` field was added for legacy datasets. 

Otherwise, IMHO, if the column exists but only some records have a value for `timepoint` field, then `GtfsStopTimeTimepoint.UNRECOGNIZED` seems appropriate to me for the records that are empty (because the original intent of adding timepoint column was to have producers specify `0` or `1` for every record to be unambiguous about whether a record was a timepoint).

It also might be confusing as to why this value is missing a default when other fields in the spec that have a `empty` value do have a default - so we should explain in comments why.

Related - I was curious to see how other defaults were handled, but it looks like `@DefaultValue` is only sparingly used currently - in `GtfsStopTimeSchema` for timepoint and in both `GtfsStopTimeSchema` and `GtfsRouteSchema` for `continuousPickup` and `continuousDropOff`, and in `GtfsRouteSchema` for `routeColor` and `routeTextColor`. 

It looks like the defaults for  `continuousPickup` and `continuousDropOff` aren't referenced anywhere, but the defaults for `routeColor` and `routeTextColor` are used for validating contrast.

But - it looks like defaults are handled as an enum value for other enums?

Take a look at `GtfsWheelchairBoardingEnum` and `GtfsTripSchema` for trips.txt. The spec says:

>Indicates wheelchair accessibility. Valid options are:
>
>* 0 or empty - No accessibility information for the trip.
>* 1 - Vehicle being used on this particular trip can accommodate at least one rider in a wheelchair.
>* 2 - No riders in wheelchairs can be accommodated on this trip.

If I'm understanding the code correctly it looks like if the column or field value is empty/missing, `GtfsWheelchairBoarding.UNRECOGNIZED` is returned. But from reading the spec, it seems like this should actually be `GtfsWheelchairBoarding.UNKNOWN`? So should this field be annotated with `@DefaultValue("0")`?

Same with `bikes_allowed`.

IMHO this is another reason to have test CSV data, as it makes it clearer what the expected behavior of the validator is in case of defaults/empty values/columns.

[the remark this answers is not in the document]
asvechnikov2: I can see there are a couple of potential problems. 

With timepoint, if we're using different default value in this validation rule and for everything else, this will create confusion and potential bugs. We can keep it this way and then remove `DefaultValue` from `timepoint` once specification is fixed, on other hand this rule contradicts to specification and we might want to change specification first before changing the validator. I'm fine with either solution in order to unblock the PR.

For other default values, it seems, that default variable type value is used (e.g. 0 for integers), so `@DefaultValue` is used only when a default value for a field is unusual, we might want to specify `DefaultValue` explicitly everywhere.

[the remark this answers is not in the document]
barbeau: >For other default values, it seems, that default variable type value is used (e.g. 0 for integers),

@asvechnikov2 I was trying to trace the execution for this too. Where do you see this assignment?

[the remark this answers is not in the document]
asvechnikov2: I think it's this one https://github.com/MobilityData/gtfs-validator/blob/903e9af59f4b959d5186d890e4eee1bfa4fcb89b/processor/src/main/java/org/mobilitydata/gtfsvalidator/processor/EntityImplementationGenerator.java#L134-L160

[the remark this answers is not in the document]
barbeau: @asvechnikov2 Ah, thanks! Looks like you're right. I just ran this through the debugger with two different datasets to confirm:
* [hart_timepoint_header_only.zip](https://github.com/MobilityData/gtfs-validator/files/7506789/hart_timepoint_header_only.zip)
* [hart_no_timepoint_column.zip](https://github.com/MobilityData/gtfs-validator/files/7506790/hart_no_timepoint_column.zip)

This code is what auto-generates the `GtfsStopTime.DEFAULT_TIMEPOINT` variable value. 

So the current behavior is:
* If there is no `@DefaultValue` annotation, if the column is missing or the row has an empty value for an enum then the default value of `0` is used. 
* If there is a `@DefaultValue` annotation, then then provided default value is used.

So the default enum value is the value defined for `0` - for `GtfsStopTimeTimepoint` this is `APPROXIMATE(0)`. This is similar to protocol buffer bindings behavior. I agree that explicitly defining the default enum value everywhere via `@DefaultValue` would be a good best practice to avoid confusion, especially for newcomers adding new fields where is isn't obvious that the `0` enum value is the de facto default.

Back to the original question - if we remove `@DefaultValue("1")`, the new value of `timepoint` field would be `GtfsStopTimeTimepoint.APPROXIMATE` if the timepoint column is missing or the field has an empty value in the CSV. We don't want this because this either because of legacy datasets from the original GTFS spec before the timepoint field existed (because all those records should be considered `EXACT`.

IMHO the best solution is still having the default is set to `EXACT` IFF the timepoint column isn't provided, and otherwise have the default set to `UNRECOGNIZED` for empty. I think the 2nd best solution is to leave the ``@DefaultValue("1")`, because that aligns with the legacy GTFS datasets and possible interpretations of the current GTFS spec.

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/TimepointTimeValidatorTest.java
barbeau: Just to reinforce this - we really shouldn't be setting values to `null` in unit tests because that behavior will never actually happen in real-world execution of the project. As mentioned in https://github.com/MobilityData/gtfs-validator/pull/1044#discussion_r745858272, enums will always have a value, defaulting to `0`.

So the current code is fine for this PR, but we really should look at re-examining how these tests are set up.

isabelle-dr:
@asvechnikov2 can we get a review so we can merge this PR?
See [this comment](https://github.com/MobilityData/gtfs-validator/pull/1044#discussion_r746738149) for what we are proposing to do.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TimepointTimeValidator.java
```diff
@@ -32,56 +34,101 @@
...
   private boolean isTimepoint(GtfsStopTime stopTime) {
-    return stopTime.timepoint().equals(GtfsStopTimeTimepoint.EXACT);
+    return stopTime.hasTimepoint() && stopTime.timepoint().equals(GtfsStopTimeTimepoint.EXACT);
   }
 
   /**
    * Timepoint without time
    *
-   * <p>Severity: {@code SeverityLevel.WARNING}
+   * <p>Severity: {@code SeverityLevel.WARNING} - to be upgraded as {@code SeverityLevel.ERROR}
```
barbeau, line 86: Is there an issue or other link you could reference here for more information about how/why this will be upgraded? Would be nice to have a link to some context in the code itself.

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/TimepointTimeValidatorTest.java
```diff
@@ -17,110 +17,356 @@
...
+            .build());
+    stopTimes.add(
+        new GtfsStopTime.Builder()
+            .setCsvRowNumber(1)
+            .setTripId("second trip id")
+            .setArrivalTime(GtfsTime.fromSecondsSinceMidnight(450))
+            .setDepartureTime(GtfsTime.fromSecondsSinceMidnight(580))
+            .setStopId("stop id 2")
+            .setStopSequence(2)
+            .setTimepoint(1)
```
barbeau, line 147: Is it right to set timepoint equal to `1` for this test? From the method name I expected both to be set to `null`.

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/TimepointTimeValidatorTest.java
```diff
@@ -17,110 +17,356 @@
...
-                    .setArrivalTime(GtfsTime.fromSecondsSinceMidnight(450))
-                    .setDepartureTime(null)
-                    .setStopId("stop id")
-                    .setStopSequence(2)
-                    .setTimepoint(1)
-                    .build()))
+  public void timepoint_missingDepartureTimeShouldGenerateNotice() {
+    List<GtfsStopTime> stopTimes = new ArrayList<>();
+    stopTimes.add(
+        new GtfsStopTime.Builder()
```
barbeau, line 212: Do you need both stopTime records for this test? If not for simplicity I'd remove the unneeded code. Same for other tests.

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/TimepointTimeValidatorTest.java
barbeau: Follow up question after looking at other tests - how do you differentiate between "no column" and "no value" in the unit tests? 

Is the missing `timepoint`  value in `CsvHeader` header what triggers `MissingTimepointColumnNotice` (even if you call `.setTimepoint(1)` on one of the StopTime records)?

Is setting `.setTimepoint(null)` what is used to define a missing value (even if the `timepoint` value is included in header)?

I'd suggest clearly documenting this behavior in the unit tests as right now that's not clear (to me at least), even following the discussion thread in https://github.com/MobilityData/gtfs-validator/pull/1044/files#r742464479. And I'd avoid any confusing data that won't exist in the real world, like a missing column but a record with a timepoint value (i.e., this test currently).

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/TimepointTimeValidatorTest.java
```diff
@@ -17,110 +17,356 @@
...
+            .build());
+    stopTimes.add(
+        new GtfsStopTime.Builder()
+            .setCsvRowNumber(1)
+            .setTripId("second trip id")
+            .setArrivalTime(GtfsTime.fromSecondsSinceMidnight(450))
+            .setDepartureTime(GtfsTime.fromSecondsSinceMidnight(580))
+            .setStopId("stop id 2")
+            .setStopSequence(2)
+            .setTimepoint(1)
```
barbeau, line 287: Based on method name doesn't seem like this StopTime record should exist in this test

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/TimepointTimeValidatorTest.java
```diff
@@ -17,110 +17,356 @@
...
```
barbeau, line 313: Based on method name doesn't seem like this StopTime record should exist in this test

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/TimepointTimeValidatorTest.java
```diff
@@ -17,110 +17,356 @@
...
```
barbeau, line 339: Based on method name doesn't seem like this StopTime record should exist in this test

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/TimepointTimeValidatorTest.java
```diff
@@ -17,110 +17,356 @@
...
```
barbeau, line 367: Based on method name doesn't seem like this StopTime record should exist in this test

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/TimepointTimeValidatorTest.java
lionel-nj: It is actually not needed. Removed in https://github.com/MobilityData/gtfs-validator/pull/1044/commits/e33c960a486db0db6d2af6a0b705fcfd6b44b026.

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/TimepointTimeValidatorTest.java
lionel-nj: I removed this record in https://github.com/MobilityData/gtfs-validator/pull/1044/commits/e33c960a486db0db6d2af6a0b705fcfd6b44b026.

lionel-nj: > Follow up question after looking at other tests - how do you differentiate between "no column" and "no value" in the unit tests?

- no column -> these tests use he legacy header 
- no value -> these tests use the regular header, with `null` `timepoint` value. 

> Is the missing timepoint value in CsvHeader header what triggers MissingTimepointColumnNotice (even if you call .setTimepoint(1) on one of the StopTime records)?

Yes. 

> Is setting .setTimepoint(null) what is used to define a missing value (even if the timepoint value is included in header)?


Exactly. 

> I'd suggest clearly documenting this behavior in the unit tests as right now that's not 100% clear (to me at least), even following the discussion thread in https://github.com/MobilityData/gtfs-validator/pull/1044/files#r742464479. And I'd avoid any confusing data that won't exist in the real world, like a missing column but a record with a timepoint value (i.e., this test currently).

👍🏾 I removed the non-necessary (and misleading records) in https://github.com/MobilityData/gtfs-validator/commit/e33c960a486db0db6d2af6a0b705fcfd6b44b026. And added additional documentation in https://github.com/MobilityData/gtfs-validator/pull/1044/commits/84926f4eae651e30f0ac63f43c20a926056116fb.

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/TimepointTimeValidatorTest.java
```diff
@@ -17,110 +17,294 @@
...
+          SHAPE_DIST_TRAVELED_FIELD_NAME,
+          TIMEPOINT_FIELD_NAME
+        });
+  }
+
+  @Test
+  public void noTimepointColumn_noTimeProvided_shouldGenerateNotice() {
+    // unit test for legacy data that do not use timepoint column
+    // setting .setTimepoint(null) is used to define a missing value
+    // (even if the timepoint value is included in header)
```
barbeau, line 102:
```suggestion
    // Using createLegacyHeader() that omits the timestamp column will trigger the MissingTimepointColumnNotice.
    // .setTimepoint(null) is used to indicate that no value is provided, although it has no effect in this test.
```

barbeau: I'd suggest the above for all the tests that use `createLegacyHeader()`

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/TimepointTimeValidatorTest.java
```diff
@@ -17,110 +17,294 @@
...
+        new GtfsStopTime.Builder()
+            .setCsvRowNumber(1)
+            .setTripId("first trip id")
+            .setArrivalTime(GtfsTime.fromSecondsSinceMidnight(450))
+            .setDepartureTime(null)
+            .setStopId("stop id")
+            .setStopSequence(2)
+            .setTimepoint(1)
+            .build());
+    stopTimes.add(
```
barbeau, line 196: I don't think you need this StopTime for this test?

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/TimepointTimeValidatorTest.java
```diff
@@ -17,110 +17,294 @@
...
+        new GtfsStopTime.Builder()
+            .setCsvRowNumber(1)
+            .setTripId("first trip id")
+            .setArrivalTime(null)
+            .setDepartureTime(GtfsTime.fromSecondsSinceMidnight(450))
+            .setStopId("stop id")
+            .setStopSequence(2)
+            .setTimepoint(1)
+            .build());
+    stopTimes.add(
```
barbeau, line 225: I don't think you need this StopTime for this test?

barbeau (commented):
Thanks @lionel-nj, a suggestion for docs and a few more comments in-line.

barbeau (approved):
👍  LGTM

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TimepointTimeValidator.java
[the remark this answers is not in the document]
asvechnikov2: nit: I'd add a comment to clarify the behaviour, here and probably in the class docstring as well.

asvechnikov2 (commented):
> @asvechnikov2 can we get a review so we can merge this PR? See [this comment](https://github.com/MobilityData/gtfs-validator/pull/1044#discussion_r746738149) for what we are proposing to do.

LGTM! Sorry, I was sure that I'd removed "requested changes" review and the  PR had been merged already.

Issue #1058, opened 2021-11-10
False positives for InvalidPhoneNumberNotice


isabelle-dr:
The validator behaviour seems to be more strict than the specification. With the current implementation, this notice is emitted by more than 50% of the datasets available on the [Mobility Database](http://mobilitydatabase.org/wiki/Main_Page).

The GTFS Specification in `agency.agency_phone` says:

> A voice telephone number for the specified agency. This field is a string value that presents the telephone number as typical for the agency's service area. It may contain punctuation marks to group the digits of the number. Dialable text (for example, TriMet's "503-238-RIDE") is permitted, but the field must not contain any other descriptive text.

The validator flags the following phone numbers as invalid, although they are valid with regards to the specification.
- 250-390-4531 from Nanaimo's GTFS Schedule ([download link](https://transitfeeds.com/p/bc-transit/691/latest/download))
- (07) 5480 4500 from Polleys Coaches's GTFS Schedule ([download link](https://transitfeeds.com/p/f-r7ke-qconnect~gympie/latest/download))
- 02 98 80 30 30 from Bibus's GTFS Schedule ([download link](https://applications002.brest-metropole.fr/VIPDU72/GPB/Lot_BrestMetropole_Bibus.zip))

lionel-nj:
Also, there seems to be a false positive for:
-  GoRaleigh's GTFS schedule ([download link](https://transitfeeds.com/p/go-raleigh/365/latest/download)): `(919) 485-RIDE`
- GoTriangle's GTFS schedule ([download link](https://transitfeeds.com/p/triangle-transit/385/latest/download)): `(919) 485-RIDE`

barbeau:
Are the correct country codes being used when running the validator on the above GTFS datasets? Could you provide the country codes that are being used?

lionel-nj:
> Are the correct country codes being used when running the validator on the above GTFS datasets? Could you provide the country codes that are being used?

No country code was provided to run the validator on the above GTFS datasets. 

Also here is the exhaustive list of datasets that trigger this notice, and the related phone numbers: 
```json 
{
  "10-15-transit-gtfs-q9052": [
    "641-683-0695",
    "800-227-6390"
  ],
  "9-town-transit-gtfs-q9070": [
    "860-510-0429"
  ],
  "access-allegany-gtfs-q86762": [
    "585-593-1738"
  ],
  "adelaide-metro-gtfs-q38310": [
    "1300 311 108",
    "1300 311 108",
[...]

PR #1061, opened 2021-11-10
chore: downgrade InvalidPhoneNumberNotice to WARNING

lionel-nj:
closes #1060 

**Summary:**

This PR provides support to downgrade `InvalidPhoneNumberNotice` to `WARNING and no longer introduce breaking chnages between v2.0.0 and the future v3.0.0 version of the validator.

**Expected behavior:** 

Notice should be a warning.

Please make sure these boxes are checked before submitting your pull request - thanks!

- [x] Run the unit tests with `gradle test` to make sure you didn't break anything
- [x] Format the title like "feat: [new feature short description]". Title must follow the Conventional Commit Specification(https://www.conventionalcommits.org/en/v1.0.0/).
- [x] Linked all relevant issues
- ~[ ] Include screenshot(s) showing how this pull request works and fixes the issue(s)~

PR #1062, opened 2021-11-10
fix: Skip phone number validation when country code is unknown

barbeau:
**Summary:**

This PR ensures that phone number validation is skipped when country code is unknown (invalid or unprovided). Adds a unit test for this case as well.

Closes https://github.com/MobilityData/gtfs-validator/issues/1058.

**Background:**

The intent of making the country code optional in PR https://github.com/MobilityData/gtfs-validator/pull/851 was to skip phone number validation if country code wasn't provided at the command line. Apparently in the main branch phone number validation is run even when the country code is not provided or unknown. 

I'm not sure if there was a regression (perhaps in PR https://github.com/MobilityData/gtfs-validator/pull/862?) or if we just didn't have sufficient unit test coverage initially to check all cases.

In the main branch we currently have a test:

```java
    assertValid(
        nc ->
            new DefaultFieldValidator(CountryCode.forStringOrUnknown("ZZ"))
                .validatePhoneNumber("+1 (650) 253-0000", CELL_CONTEXT, nc));
```

...but this test doesn't cover this case and it still passes on the main branch. Evidently the phone number validation library considers `+1 (650) 253-0000` a valid phone number even when no valid country code is provided (I'm assuming because it's an international format).

However, the phone number validation library does consider other formats (e.g., common US format `813-254-4278`) invalid without a country code.

I've added another assertion to check an unknown country code with the US format, and this assertation fails on the current master branch but passes on this PR.

```java
    assertValid(
            nc ->
                    new DefaultFieldValidator(CountryCode.forStringOrUnknown("ZZ"))
                            .validatePhoneNumber("813-254-4278", CELL_CONTEXT, nc));
```

**Expected behavior:** 

When country code is not provided at the command-line, phone number validation should be skipped.

---

### DIFF
--- a/core/src/main/java/org/mobilitydata/gtfsvalidator/validator/DefaultFieldValidator.java
+++ b/core/src/main/java/org/mobilitydata/gtfsvalidator/validator/DefaultFieldValidator.java
@@ -94,7 +94,8 @@ public class DefaultFieldValidator implements GtfsFieldValidator {
   public void validatePhoneNumber(
       String phoneNumber, GtfsCellContext cellContext, NoticeContainer noticeContainer) {
     if (countryCode.isUnknown()) {
-      // If the user didn't provide a country code or the provided country-code was invalid, skip validation
+      // If the user didn't provide a country code or the provided country-code was invalid, skip
+      // validation
       return;
     }
     if (!PhoneNumberUtil.getInstance()
--- a/core/src/main/java/org/mobilitydata/gtfsvalidator/validator/DefaultFieldValidator.java
+++ b/core/src/main/java/org/mobilitydata/gtfsvalidator/validator/DefaultFieldValidator.java
@@ -93,6 +93,10 @@ public class DefaultFieldValidator implements GtfsFieldValidator {
   @Override
   public void validatePhoneNumber(
       String phoneNumber, GtfsCellContext cellContext, NoticeContainer noticeContainer) {
+    if (countryCode.isUnknown()) {
+      // If the user didn't provide a country code or the provided country-code was invalid, skip validation
+      return;
+    }
     if (!PhoneNumberUtil.getInstance()
         .isPossibleNumber(phoneNumber, countryCode.getCountryCode())) {
       noticeContainer.addValidationNotice(
--- a/core/src/main/java/org/mobilitydata/gtfsvalidator/validator/DefaultFieldValidator.java
+++ b/core/src/main/java/org/mobilitydata/gtfsvalidator/validator/DefaultFieldValidator.java
@@ -95,7 +95,7 @@ public class DefaultFieldValidator implements GtfsFieldValidator {
       String phoneNumber, GtfsCellContext cellContext, NoticeContainer noticeContainer) {
     if (countryCode.isUnknown()) {
       // If the user didn't provide a country code or the provided country-code was invalid, skip
-      // validation
+      // validation.
       return;
     }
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/mobilitydata/gtfsvalidator/validator/DefaultFieldValidatorTest.java
```diff
@@ -119,6 +119,19 @@ public void validatePhoneNumber_valid() {
...
                 .validatePhoneNumber("004980038762246", CELL_CONTEXT, nc));
+
+    assertValid(
+        nc ->
+            new DefaultFieldValidator(CountryCode.forStringOrUnknown("us"))
+                .validatePhoneNumber("813-254-4278", CELL_CONTEXT, nc));
+
+    // The below format fails validation without a known country code, so check that we skip
+    // validation when country
+    // code is unknown
```
aababilov, line 130: Please merge the last two lines and put a full stop at the end.

core/src/main/java/org/mobilitydata/gtfsvalidator/validator/DefaultFieldValidator.java
```diff
@@ -93,6 +93,11 @@ public void validateEmail(
   @Override
   public void validatePhoneNumber(
       String phoneNumber, GtfsCellContext cellContext, NoticeContainer noticeContainer) {
+    if (countryCode.isUnknown()) {
+      // If the user didn't provide a country code or the provided country-code was invalid, skip
+      // validation
```
aababilov, line 98: Please put a full stop at the end of comments. They must be complete sentences.

aababilov (approved):
Thank you!

core/src/test/java/org/mobilitydata/gtfsvalidator/validator/DefaultFieldValidatorTest.java
barbeau: Thanks for catching that! Not sure why the code formatter split the lines, but it's fixed now with the period too in https://github.com/MobilityData/gtfs-validator/pull/1062/commits/5223228e99241ccf2e3395c5822fe8abdf0104cd.

core/src/main/java/org/mobilitydata/gtfsvalidator/validator/DefaultFieldValidator.java
barbeau: Done in https://github.com/MobilityData/gtfs-validator/pull/1062/commits/5223228e99241ccf2e3395c5822fe8abdf0104cd.

isabelle-dr:
Thank you @barbeau!

We can then cancel PR #1061 when this is merged, problem solved 🎉

lionel-nj:
Thanks for working on that @barbeau, it seems that some unit test failed though: https://github.com/MobilityData/gtfs-validator/runs/4170664366?check_suite_focus=true.

barbeau:
@lionel-nj Whoops, I accidentally restored that test in my original commit - it should be removed because we don't want to run validation on unknown country codes. I removed it in https://github.com/MobilityData/gtfs-validator/pull/1062/commits/381683cac8e477d72ab3cfa2879b79d9b76071c8 so it should pass now.

lionel-nj (approved):
After analyzing the validation reports produced by https://github.com/MobilityData/gtfs-validator/actions/runs/1446146385 - I confirm that no more `InvalidPhoneNumberNotice` is triggered.

PR #1077, opened 2021-12-16
chore: Remove remaining references to log4j

aababilov:
GTFS Validator uses Flogger for logging. The validator is not affected by Log4Shell.

https://en.wikipedia.org/wiki/Log4Shell


asvechnikov2 (approved):
LGTM! Thank you!

Issue #1070, opened 2021-11-18
Make warning-level notice when an equal shape_dist_traveled value is encountered and the previous row in the sequence had the same GPS coordinate


evansiroky:
**What problem in GTFS datasets does this new rule address? Please describe.**

In some feeds, I have seen records in the shapes.txt file where two rows of data have the exact same GPS coordinate and exact shape_dist_traveled value. The current behavior of the validator treats this as a severe error. Although it is a somewhat duplicative record, I wouldn't think that it would qualify as a severe error and probably has no noticeable effect to an end user when shown a shape in a trip planner UI.

**Describe the new validation rule**

The code as follows should have more nuance:

https://github.com/MobilityData/gtfs-validator/blob/cbc951327e5fd5f2c7d6c2cd4ad97084b36528f0/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/ShapeIncreasingDistanceValidator.java#L52

The code should also look to see if the GPS coordinates in the previous record were the same GPS coordinates in the current record and as long as the shape_dist_traveled value is also equal, a notice should be generated, but the severity of that notice should be only a warning.

**Error vs warning**

There should still be an error if the shape_dist_traveled is equal but the GPS coordinates change. But if the coordinates are the same and the shape_dist_traveled is the same, only a warning should be generated.

lionel-nj:
Hi @evansiroky, thanks for flagging that. We will investigate the logic of this rule in one of the oncoming sprints.

evansiroky:
Thanks, @lionel-nj! I had one more thought to add here. I recently saw another feed where the validator was flagging errors for same-distance values but where the actual difference was less than 1 meter between the coordinates. Perhaps it was that the output of the distance in the feed was rounded down to a certain amount of decimal points and lost that extra precision which practically wasn't needed anyways. So I'd like to add that it'd be nice if the eventual logic used in this rule wouldn't flag same-distance values as an error if the coordinates are acceptably physically close enough (like <1 meter or something).

lionel-nj:
Interesting! It would be amazing if you could join the url to the related dataset to help troubleshooting with existing data. 

> So I'd like to add that it'd be nice if the eventual logic used in this rule wouldn't flag same-distance values as an error if the coordinates are acceptably physically close enough (like <1 meter or something).

We'll investigate how to best cover this case in our unit tests.

barbeau:
[...]

Issue #1067, opened 2021-11-17
3.1.0 Release plan


isabelle-dr:
This is an issue for planning the 3.1.0 release of the validator. It will be updated as we plan and develop the next release.
**Please share with us below your needs and ideas!** 💡
**If you'd like to contribute to this release, please get in touch with isabelle at isabelle@mobilitydata.org**🙏  

## Goal for the release 🎯
In addition to the canonical set of rules (based on the [official GTFS specification](https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md)), we want to allow introducing additional rules and create **custom validation sets** in this tool. This aims at responding to the following user needs:
- as a producer, I can validate my feed against a GTFS extension that isn't adopted yet
- as a producer, I can validate my feed against a specific consumer's set of criteria
- as a producer, I can validate my feed against several validation sets at once
- as a consumer, I can add additional rules that aren't in the canonical set of rules
- as a consumer, I can create a custom validation set by selecting a set of rules and the severity associated

#### Considerations
- we need to allow for different schemas in this feature
- all the "non-canonical" rules need to be publicly available for everyone to use
- if we validate against several validation sets at once, we need to have a way to not run the same code twice
- we need to think of a governance process for the "non-canonical" rules to ensure the stability of the custom validation sets
- we need to think of a way to flag the "non-canonical" rules that are used widely by the community and "canonalize" them
- we need a process to make sure the custom validation sets aren't getting further from the official specification as the canonical validation set evolves

#### Materials
A first brainstorm about the design of this feature is available [in this document](https://docs.google.com/document/d/1Og3WPodluc74svgqirsl-3w8ApfRzuH1nRP-2rrBptU/edit#). Feel free to leave comments directly in it!
This feature has been discussed/could help in the following use cases:
- https://github.com/MobilityData/gtfs-validator/pull/1039
- https://github.com/MobilityData/gtfs-validator/issues/927
- https://github.com/MobilityData/gtfs-validator/issues/918
- https://github.com/MobilityData/gtfs-validator/issues/882

isabelle-dr:
[Here is a summary](https://docs.google.com/document/d/189BoeeexPa0fXcMtZcCLCfEpMBQ6IKSbV0rPu-4CT44/edit#) of what was discussed during our first community meeting to confirm the needs and look at how to change the validator architecture for the 3.1 release. 
Thank you for your participation!

bdferris-v2:
[...]

PR #1083, opened 2021-12-22
fix: ShapeIncreasingDistanceValidator

lionel-nj:
closes https://github.com/MobilityData/gtfs-validator/issues/1070

**Summary:**

This PR provides support to: make warning-level notice when an equal `shape_dist_traveled` value is encountered and the previous row in the sequence had the same GPS coordinate.

**Expected behavior:** 

- `DecreasingShapeDistanceNotice` as an `ERROR`: should be issued when `shape_dist_traveled` decreases between two consecutive shape points (based on `shape_pt_sequence`);
`EqualShapeDistanceDiffCoordinatesNotice` as an `ERROR`:  should be issued when  `shape_dist_traveled` is equal between two consecutives shape points that have different GPS coordinates;
`EqualShapeDistanceSameCoordinatesNotice` as a `WARNING`: should be issued when  `shape_dist_traveled` is equal between two consecutives shape points that have the same GPS coordinates.

@evansiroky @botanize your comments and feedback are more than welcome :)


main/src/main/java/org/mobilitydata/gtfsvalidator/validator/ShapeIncreasingDistanceValidator.java
```diff
@@ -47,23 +52,39 @@ public void validate(NoticeContainer noticeContainer) {
...
+  /**
+   * Checks if two {@code GtfsShape} have the same GPS coordinates.
+   *
+   * @param shape the first {@code GtfsShape}
+   * @param otherShape the other {@code GtfsShape}
+   * @return true if both {@code GtfsShape} have the same GPS coordinates, false otherwise.
+   */
+  private static boolean sameGpsCoordinates(GtfsShape shape, GtfsShape otherShape) {
+    return shape.shapePtLon() == otherShape.shapePtLon()
+        && shape.shapePtLat() == otherShape.shapePtLat();
```
evansiroky, line 85: One of the errors we've been seeing a lot is when there are two shape points that are very close to each other (<1 meter apart) and the shape distance was equal. It could've been that the shape distance units in a particular feed were in kilometers and that due to a rounding process the distance was equal and therefore probably an OK value. It would be great if this check here could return true if the coordinates are less than 1 meter apart.

RULES.md
```diff
@@ -262,6 +264,15 @@ Empty csv file found in the archive: file does not have any headers, or is a req
 ##### References:
 * [GTFS files requirements](https://gtfs.org/reference/static#file-requirements)
 
+#### EqualShapeDistanceNotice
+
+<a name="EqualShapeDistanceNotice"/>
+
+When sorted by `shape.shape_pt_sequence`, two consecutive shape points must not have equal values for `shape_dist_traveled` if their GPS coordinates are the same.
```
botanize, line 271: I think this should read "if their GPS coordinates are not the same". It's an error to have equal distances with different coordinates, it's a warning to have equal distances and equal coordinates.

botanize: My only other concern is that at least for our feed, which generates equal distances every stop pair, we get 10s of thousands of these notices per validation. I'd love to have the ability to output an abbreviated json validation report that reported the number of times each notices happened, but only the first X occurrences.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/ShapeIncreasingDistanceValidator.java
lionel-nj: One question is why 1 meter, why not 80cm or 1.20m?

RULES.md
lionel-nj: > I think this should read "if their GPS coordinates are not the same". 

👍🏾
> It's an error to have equal distances with different coordinates, it's a warning to have equal distances and equal coordinates.

Or if said coordinates are actually less than 1 meter away according to https://github.com/MobilityData/gtfs-validator/pull/1083#discussion_r774106569. 

> My only other concern is that at least for our feed, which generates equal distances every stop pair, we get 10s of thousands of these notices per validation. I'd love to have the ability to output an abbreviated json validation report that reported the number of times each notices happened, but only the first X occurrences.

As of now a maximum of 1000 of notices are exported by type and severity, and the validation report gives information about the total number of notices per type and severity. 

https://github.com/MobilityData/gtfs-validator/blob/a871d4bad3f7eb7b2d17d1122fd825aca6883b86/core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NoticeContainer.java#L54

https://github.com/MobilityData/gtfs-validator/blob/a871d4bad3f7eb7b2d17d1122fd825aca6883b86/core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NoticeContainer.java#L190-L193

> I'd love to have the ability to output an abbreviated json validation report that reported the number of times each notices happened, but only the first X occurrences.

I see this as separate from the bug fixed here. We could allow users to reduce this limit  in a separate PR by making the limit user configurable. cc @isabelle-dr.

lionel-nj:
As of now:

- 108 datasets trigger `EqualShapeDistanceNotice` as an `ERROR`
- 272 datasets trigger `EqualShapeDistanceNotice` as a`WARNING`
- 2 datasets trigger `DecreasingShapeDistanceNotice`
- 713 datasets are valid (no error)

Before the fix:
- 277 datasets triggered `DecreasingOrEqualShapeDistanceNotice` as an `ERROR`
- 608 datasets were valid (no error)

Also, with this fix no additional error type is introduced (i.e. no dataset go from valid to invalid).

cc @isabelle-dr

isabelle-dr:
Hello, 

A few thoughts/questions about the original issue and the modification of `DecreasingOrEqualShapeDistanceNotice` in this PR.

1. Getting those stats is amazing! Seing that among all the datasets that have equal values for `shape_dist_traveled`, there is almost a 30-70% split between the ones that have the same coordinates and the ones that have different coordinates confirms that it's worth looking into what is actually a severe problem here. Thanks @botanize and @evansiroky for highlighting this issue and @lionel-nj for getting these numbers!

2. Would `EqualShapeDistanceNotice` be it the first rule that can appear as an error or a warning, depending on the logic? How would this work with our plans with the custom validation feature? (especially, if we go towards [our second option](https://docs.google.com/document/d/1Og3WPodluc74svgqirsl-3w8ApfRzuH1nRP-2rrBptU/edit#bookmark=id.z5wep6s4dqy4) using extended base validation class.)
Would it be more clear for our users if we had two distinct rules for `EqualShapeDistanceNotice` for the two cases we have here?
Again, keeping in mind that [soon](https://github.com/MobilityData/gtfs-validator/issues/1067), users will be able to upgrade or downgrade a notice. My initial feeling is that it might be more clear to have two distinct notices for different severity levels, even if part of the code is the same. I'd love to hear your thoughts here @evansiroky and @botanize. 

Here are other options we could look at:
- option 1: 
  - `DecreasingOrEqualShapeDistanceNotice` (ERROR): when `shape_dist_traveled` values are decreasing OR (when they are equal AND coordinates are different)
  - `EqualShapeDistanceNotice` (WARNING): when `shape_dist_traveled` values are equal AND coordinates are equal

- option 2: 
  - `DecreasingShapeDistanceNotice` (ERROR)
  - `EqualShapeDistanceDiffCoordinatesNotice` (ERROR)
  - `EqualShapeDistanceSameCoordinatesNotice` (WARNING)

3. Do we have the same problem for `DecreasingOrEqualStopTimeDistanceNotice`?

4. About what we consider as "same coordinates"
> It would be great if this check here could return true if the coordinates are less than 1 meter apart.

This makes a lot of sense. Like @lionel-nj mentioned, I'd be interested to look at what would be the best threshold to use.

5. Are they datasets that contain two shape points with the same coordinates and different values for `shape_dist_traveled` in shapes.txt? Should we also flag this?

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/ShapeIncreasingDistanceValidator.java
evansiroky: 1 meter is arbitrary honestly. Maybe @barbeau can provide some more educated insights into GPS precision and measurement.

evansiroky:
>    Would it be more clear for our users if we had two distinct rules for `EqualShapeDistanceNotice` for the two cases we have here?

>    Again, keeping in mind that [soon](https://github.com/MobilityData/gtfs-validator/issues/1067), users will be able to upgrade or downgrade a notice. My initial feeling is that it might be more clear to have two distinct notices for different severity levels, even if part of the code is the same. I'd love to hear your thoughts here @evansiroky and @botanize.

I'm in favor of having two distinct rules with different error levels.

> Here are other options we could look at:
> 
> * option 1:
>   
>   * `DecreasingOrEqualShapeDistanceNotice` (ERROR): when `shape_dist_traveled` values are decreasing OR (when they are equal AND coordinates are different)
>   * `EqualShapeDistanceNotice` (WARNING): when `shape_dist_traveled` values are equal AND coordinates are equal
> * option 2:
>   
>   * `DecreasingShapeDistanceNotice` (ERROR)
>   * `EqualShapeDistanceDiffCoordinatesNotice` (ERROR)
>   * `EqualShapeDistanceSameCoordinatesNotice` (WARNING)

I like option 2.

> 3. Do we have the same problem for `DecreasingOrEqualStopTimeDistanceNotice`?

I suppose this could technically be OK in a 3-dimensional sense if there was a transit vehicle that traveled along a pattern that involved an elevator with different "stops" at different levels. However, I don't think that currently exists anywhere yet, so it may be best to keep this as an error.

> 4. About what we consider as "same coordinates"
> 
> > It would be great if this check here could return true if the coordinates are less than 1 meter apart.
> 
> This makes a lot of sense. Like @lionel-nj mentioned, I'd be interested to look at what would be the best threshold to use. Would it be possible to make a 1-D plot showing values of distances between coordinates for all pairs of shape points that have the same `shape_dist_traveled` (and same shape_id) in order to see what threshold would capture all those rounded values? We can't play too much with the datasets yet at MobilityData because we don't have an API yet 😭

I'll defer to @barbeau's expertise here.

> 5. Are they datasets that contain two shape points with the same coordinates and different values for `shape_dist_traveled` in shapes.txt? Should we also flag this?

Yes, this should be an error, but I'd be surprised if it occurred in the wild.

barbeau:
>Would it be possible to make a 1-D plot showing values of distances between coordinates for all pairs of shape points that have the same shape_dist_traveled (and same shape_id) in order to see what threshold would capture all those rounded values? 

This would be ideal (i.e., create the threshold based on a percentile based on real data), but to my knowledge not easy to calculate at the moment without a lot of manual effort.

One way to approach this is based on precision of decimal places for latitude and longitude. From http://wiki.gis.com/wiki/index.php/Decimal_degrees:

| decimal places |   degrees  | distance |   
|:--------------:|:----------:|:--------
| 0              | 1.0        | 111 km   
| 1              | 0.1        | 11.1 km  
| 2              | 0.01       | 1.11 km  
| 3              | 0.001      | 111 m    
| 4              | 0.0001     | 11.1 m   
| 5              | 0.00001    | 1.11 m   
| 6              | 0.000001   | 0.111 m  
| 7              | 0.0000001  | 1.11 cm  
| 8              | 0.00000001 | 1.11 mm 

So, if you cap the number of decimal places to 5 the smallest difference between values you can represent is 1.11 meters. Therefore, it would be possible for two values less than 1.11 meters apart to be capped/rounded to the same latitude and longitude value.

So perhaps 1.11 meters is a good value? That should capture any "same" values that result from precision/rounding issues at 5 decimal places. That's also better accuracy that you'd typically see with an off-the-shelf consumer-grade GNSS unit capturing a vehicle path in motion.

botanize:
> Here are other options we could look at:
> * option 2:
>   
>   * `DecreasingShapeDistanceNotice` (ERROR)
>   * `EqualShapeDistanceDiffCoordinatesNotice` (ERROR)
>   * `EqualShapeDistanceSameCoordinatesNotice` (WARNING)
> 
I prefer option 2 here.

> 3. Do we have the same problem for `DecreasingOrEqualStopTimeDistanceNotice`?

We have historically scheduled holds at stops as two consecutive departures from the same stop, so those would have shown up as equal distance in stop_times. I believe we now schedule those holds as actual holds with arrival and departure times or post-process them to merge the pair of stops into an arrival and departure time for one stop. The spec seems to allow our old practice, so we'd expect that equal stop time distances would be allowed, if not encouraged.

> 4. About what we consider as "same coordinates"

1.11 m is fine, but realistically, anything less that 13 m (length of our regular buses) is basically the same from an operations perspective.

> 5. Are they datasets that contain two shape points with the same coordinates and different values for `shape_dist_traveled` in shapes.txt? Should we also flag this?

If the shape contains a loop then you could definitely have two different shape points with the same coordinates and different `shape_dist_traveled` values. I don't think you'd want to flag that, unless you had some way of detecting invalid looping.

isabelle-dr:
Thanks for the comments on this PR 🙏

> If the shape contains a loop then you could definitely have two different shape points with the same coordinates and different shape_dist_traveled values. I don't think you'd want to flag that, unless you had some way of detecting invalid looping.

Good point. I was thinking of consecutive stops there, which would be an error I think. Let's keep this as a potential candidate for later if we see a need.

> We have historically scheduled holds at stops as two consecutive departures from the same stop, so those would have shown up as equal distance in stop_times. I believe we now schedule those holds as actual holds with arrival and departure times or post-process them to merge the pair of stops into an arrival and departure time for one stop. The spec seems to allow our old practice, so we'd expect that equal stop time distances would be allowed, if not encouraged.

Opened #1084 to follow up.

So in conclusion, we agree on the option 2:
* DecreasingShapeDistanceNotice as an ERROR
* EqualShapeDistanceDiffCoordinatesNotice as an ERROR
    * consider as "different coordinates" pairs of lat/long that are more than 1,11 meters apart and see how many datasets would trigger this notice after the addition of this threshold (currently 108 datasets with no threshold). Do a quick inspection if needed, decide from there if the threshold seems acceptable.
* EqualShapeDistanceSameCoordinatesNotice as a WARNING
    * @botanize @evansiroky, I'm in favour of not using the threshold so the rule flags a "truly" duplicative record (i. e. pairs of lat/long that are less than 1,11 meters apart would NOT be considered equal in this context). Are you OK with this?

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/ShapeIncreasingDistanceValidator.java
```diff
@@ -47,33 +57,117 @@ public void validate(NoticeContainer noticeContainer) {
...
+   *
+   * @param shape the first {@code GtfsShape}
+   * @param otherShape the other {@code GtfsShape}
+   * @return true the two {@code GtfsShape} have distinct coordinates and if the distance that
+   *     separates them is less or equal than {@code
+   *     ShapeIncreasingDistanceValidator#MAX_DISTANCE_SHAPEPOINTS}; false otherwise.
+   */
+  private static boolean areCloseWithDifferentCoordinates(GtfsShape shape, GtfsShape otherShape) {
+    return shape.shapePtLon() != otherShape.shapePtLon()
+        && shape.shapePtLat() != otherShape.shapePtLat()
```
botanize, line 100: You can have different coordinates while either lon or lat is identical. Shouldn't this be `!(shape.shapePtLon() == otherShape.shapePtLon() && shape.shapePtLat() == otherShape.shapePtLat())`

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/ShapeIncreasingDistanceValidator.java
```diff
@@ -47,33 +57,117 @@ public void validate(NoticeContainer noticeContainer) {
...
+        }
+        if (prev.shapeDistTraveled() > curr.shapeDistTraveled()) {
+          noticeContainer.addValidationNotice(new DecreasingShapeDistanceNotice(prev, curr));
+          continue;
+        }
+        if (prev.shapeDistTraveled() != curr.shapeDistTraveled()) {
+          continue;
+        }
+        // equal distance and different coordinates
+        if (!areCloseWithDifferentCoordinates(curr, prev)) {
```
botanize: Won't this branch execute when coordinates are identical OR where coordinates are > 1.11 m?

Which means the following `else if` will never execute because when `curr.shapePtLat() == prev.shapePtLat() && curr.shapePtLon() == prev.shapePtLon()` then `!areCloseWithDifferenteCoordinates` is true, so the previous branch has already executed.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/ShapeIncreasingDistanceValidator.java
```diff
@@ -91,11 +95,11 @@ public void validate(NoticeContainer noticeContainer) {
    *     separates them is less or equal than {@code
    *     ShapeIncreasingDistanceValidator#MAX_DISTANCE_SHAPEPOINTS}; false otherwise.
    */
-  private static boolean areCloseWithDifferentCoordinates(GtfsShape shape, GtfsShape otherShape) {
+  private static boolean areFarWithDifferentCoordinates(GtfsShape shape, GtfsShape otherShape) {
     return shape.shapePtLon() != otherShape.shapePtLon()
         && shape.shapePtLat() != otherShape.shapePtLat()
```
botanize, line 100: You can have different coordinates while either lon XOR lat is identical. Shouldn't this be `!(shape.shapePtLon() == otherShape.shapePtLon() && shape.shapePtLat() == otherShape.shapePtLat())`

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/ShapeIncreasingDistanceValidator.java
botanize: But it looks like this function isn't used anywhere

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/ShapeIncreasingDistanceValidator.java
```diff
@@ -47,33 +57,121 @@ public void validate(NoticeContainer noticeContainer) {
...
           noticeContainer.addValidationNotice(
-              new DecreasingOrEqualShapeDistanceNotice(
-                  curr.shapeId(),
-                  curr.csvRowNumber(),
-                  curr.shapeDistTraveled(),
-                  curr.shapePtSequence(),
-                  prev.csvRowNumber(),
-                  prev.shapeDistTraveled(),
-                  prev.shapePtSequence()));
+              new EqualShapeDistanceSameCoordinatesNotice(prev, curr));
```
botanize, line 82: What if coordinates are different and `getDistanceMeters(curr.shapePtLatLon(), prev.shapePtLatLon()) <= MAX_DISTANCE_SHAPEPOINTS_METERS `?

lionel-nj: In this case, no notice should be triggered, see: https://github.com/MobilityData/gtfs-validator/blob/737adabd6a2358cdd5f470e741d67149db7e3184/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/ShapeIncreasingDistanceValidatorTest.java#L103

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/ShapeIncreasingDistanceValidator.java
lionel-nj: > Shouldn't this be !(shape.shapePtLon() == otherShape.shapePtLon() && shape.shapePtLat() == otherShape.shapePtLat())

Thanks for flagging that! Modified in https://github.com/MobilityData/gtfs-validator/pull/1083/commits/574cd7b1ec7691621755ab7b34892ded7bbf6462.

botanize:
I was just looking at the spec again, and shapes.shape_distance_traveled is type `non-negative integer`. The description says "Values must increase along with shape_pt_sequence". Since type and description conflict, I'd go with type, which is more clearly defined. Though I suggest someone opens an issue/files a PR to resolve the conflict, I can do that if you want, but I would suggest keeping the type and modifying the description.

If we accept that `type` overrules `description` I think the following logic makes sense:

* shape_dist_traveled decreases: DecreasingShapeDistanceNotice as an ERROR
* shape_dist_traveled increases: OK
* shape_dist_traveled identical:
  - coordinates <= 1.11 m apart: we assume insufficient precision, OK
  - coordinates > 1.11 m apart: EqualShapeDistanceDiffCoordinatesNotice as a WARNING

isabelle-dr:
> I was just looking at the spec again, and shapes.shape_distance_traveled is type non-negative integer. 

shapes.shape_distance_traveled is type non-negative float, I think you mixed up with shapes.shape_shape_pt_sequence.

botanize:
> > I was just looking at the spec again, and shapes.shape_distance_traveled is type non-negative integer.
> 
> shapes.shape_distance_traveled is type non-negative float, I think you mixed with shapes.shape_shape_pt_sequence.

Ha! I did, but my point is the same. non-negative float includes zero, so equal shape distances are allowed by the `type`, though the `description` suggests otherwise.

isabelle-dr:
> non-negative float includes zero, so equal shape distances are allowed by the type, though the description suggests otherwise.

Maybe there is something I'm not fully getting, but I don't see the contradiction.
"Values must increase" means that we don't want equal (or decreasing) values between two consecutive shape points. 
Allowing 0 or not to be this value is a different thing.

| shape_pt_sequence | shape_dist_traveled |
|-------------------|---------------------|
|                 1 |             7513.58 | 
|                 2 |             7513.58 | 
|                 3 |             7527.21 |

There is an ERROR thrown for line 2 because the value isn't increasing.

IMHO we should keep `EqualShapeDistanceDiffCoordinatesNotice` as an ERROR, but increase the 1,11m threshold because I suspect we still have a lot of false positives, considering how little the number of datasets triggering the `equal_shape_distance_diff_coordinates` changed after Lionel added the threshold (it went from 108 to 101).

botanize:
> Maybe there is something I'm not fully getting, but I don't see the contradiction 🙈. "Values must increase" means that we don't want equal (or decreasing) values between two consecutive shape points. Allowing 0 or not to be this value is a different thing.

I apologize! I was conflating shape_distance_traveled with the difference between distances.

botanize:
> IMHO we should keep `EqualShapeDistanceDiffCoordinatesNotice` as an ERROR, but increase the 1,11m threshold because I suspect we still have a lot of false positives, considering how little the number of datasets triggering the `equal_shape_distance_diff_coordinates` changed after Lionel added the threshold (it went from 108 to 101).

I think the idea of downgrading to a warning when the coordinates are the same was that the precision of `shape_dist_traveled` and the precision of the coordinates were consistent, and that there was no real harm to consumers from the duplication of shape points.

The idea of setting a threshold came from an attempt to allow some loss of precision in `shape_dist_traveled` that could still be consistent with the coordinates. The challenge with testing for consistent precision is that it's quite difficult to know the precision of the unknown units of `shape_dist_traveled`, so it's hard to say that the precision is internally consistent if the coordinates aren't exactly the same. 

Instead of testing for internal consistency when the coordinates are not identical, the test could be based on when consumers start to have issues with the data. At what spacing of shape points would a consumer or customer have problems if `shape_dist_traveled` is identical?

For example, let's say we want a map for a trip plan that shows the shape to the destination stop. When the coordinates are identical, it doesn't matter that `shape_dist_traveled` is identical because matching to either shape point gives the same result. When the shape point coordinates are a meter or two apart, the end of the shape and the stop will be very close to each other, and this doesn't seem like a problem. If two points ten meters apart share the same `shape_dist_traveled` you would introduce confusion about where the shape should really end (and possibly which stop is actually served) depending on how you choose among multiple matching points—10 m seems obviously too far. Is there a happy medium where customer information is good enough? Are there other consumer uses that have stricter accuracy requirements?

I'm starting to think the easiest, clearest and most future proof solution is if you have different coordinates you must have different distances and leave it to the producer to make sure the precision of `shape_dist_traveled` is consistent with the precision of the shape point coordinates. 

>@botanize @evansiroky, I'm in favour of not using the threshold so the rule flags a "truly" duplicative record (i. e. pairs of lat/long that are less than 1,11 meters apart would NOT be considered equal in this context). Are you OK with this?
I think, @isabelle-dr, that was the gist of your suggestion. However, I have a problem with saying, it's a warning if coordinates are exact, it's an error if coordinates are too far apart, but it's fine if they're between 0 and 1.11 m apart. Why would it be ok for the coordinates to be just a little different, but not the same!?

If we want to use a threshold we need to distinguish between coordinates that are near enough and coordinates that are too far apart:

- DecreasingShapeDistanceNotice as an ERROR
- EqualShapeDistanceFarCoordinatesNotice as an ERROR for coordinates > THRESHOLD apart
- EqualShapeDistanceNearCoordinatesNotice as a WARNING for coordinates between 0 and THRESHOLD apart
- EqualShapeDistanceSameCoordinatesNotice as INFO (is INFO a thing?)

But I suggest we drop the threshold and use:

- DecreasingShapeDistanceNotice as an ERROR
- EqualShapeDistanceDiffCoordinatesNotice as an ERROR
- EqualShapeDistanceSameCoordinatesNotice as a WARNING

isabelle-dr:
Thanks @botanize !
Not answering directly your comment here but if this could help us make an informed decision, I had a look "manually" at a few of the datasets that contain a good portion `EqualShapeDistanceDiffCoordinatesNotice`, to evaluate if we still have false positives, here are the results.

**TL;DR**: they all look like true positives to me (i.e. actual error not resulting from a rounding problem)

- dataset [bodo-verkehrsve-gtfs-q408](https://storage.googleapis.com/storage/v1/b/bodo-verkehrsve-gtfs-q408_archives_2020-06-28/o/e3d5d845163d72644771809f071ce250ad6dcc61.zip?alt=media): 472 instances of the error

| shape\_id         | shape\_pt\_sequence | shape\_pt\_lat | shape\_pt\_lon | shape\_dist\_traveled | distance with previous (m) based on lat/lon |
| ----------------- | ------------------- | -------------- | -------------- | --------------------- | ------------------------------------------- |
| 18-222-20a-6.51.R | 346                 | 47.671166      | 9.611697       | **7513.58**               |                                             |
| 18-222-20a-6.51.R | 347                 | 47.66737       | 9.61032        | **7513.58**               | 434                                         |
| 18-222-20a-6.51.R | 348                 | 47.667256      | 9.610389       | 7527.21               | 13.69                                       |
| 4-12-20a-7.41.R   | 66                  | 47.819726      | 9.606884       | **2674.92**               |                                             |
| 4-12-20a-7.41.R   | 67                  | 47.825088      | 9.597331       | **2674.92**               | 930                                         |
| 4-12-20a-7.41.R   | 68                  | 47.825113      | 9.597314       | 2677.95               | 3.06                                        |
| 8-32-20a-5.51.R   | 22                  | 47.918247      | 9.757031       | **747.87**                |                                             |
| 8-32-20a-5.51.R   | 23                  | 47.90031       | 9.758175       | **747.87**                | 1996                                        |
| 8-32-20a-5.51.R   | 24                  | 47.899781      | 9.758298       | 807.34                | 59.5                                        |

- dataset [ding-verkehrsve-gtfs-q750](https://storage.googleapis.com/storage/v1/b/ding-verkehrsve-gtfs-q750_archives_2021-06-15/o/92efbb6be0cfe9327d407c0d32bee5c080984f05.zip?alt=media): 3871 instances of the error

| shape\_id        | shape\_pt\_sequence | shape\_pt\_lat | shape\_pt\_lon | shape\_dist\_traveled | distance with previous (m) based on lat/lon |
| ---------------- | ------------------- | -------------- | -------------- | --------------------- | ---- |
| 88-59-j20-2.72.R | 1098                | 48.411279      | 10.010274      | **41348.94**              |      |
| 88-59-j20-2.72.R | 1099                | 48.411284      | 10.010374      | **41348.94**              | 7.4  |
| 88-59-j20-2.72.R | 1100                | 48.411251      | 10.010376      | 41352.56              |      |
| 20-24-j20-3.33.R | 955                 | 48.394667      | 9.981319       | **30260.76**              |      |
| 20-24-j20-3.33.R | 956                 | 48.394613      | 9.981411       | **30260.76**              | 9.07 |
| 20-24-j20-3.33.R | 957                 | 48.394817      | 9.981739       | 30293.95              |      |
| 73-21-j20-2.73.R | 700                 | 48.389429      | 9.978581       | **31227.37**              |      |
| 73-21-j20-2.73.R | 701                 | 48.38943       | 9.978597       | **31227.37**              | 1.19 |
| 73-21-j20-2.73.R | 702                 | 48.389663      | 9.978696       | 31254.29              | 26.9 |
| 87-12-j20-1.4.H  | 25                  | 48.388256      | 9.978129       | **1302.46**               |      |
| 87-12-j20-1.4.H  | 26                  | 48.388264      | 9.978089       | **1302.46**               | 3.09 |
| 87-12-j20-1.4.H  | 27                  | 48.388011      | 9.978043       | 1330.84               | 28.3 |

- dataset [greater-sydney-gtfs-q38107](https://storage.googleapis.com/storage/v1/b/greater-sydney-gtfs-q38107_archives_2019-03-18/o/1731ae94c7b080f82a7fb862b0b24745b804b0bf.zip?alt=media): 4318 instances of the error

| shape\_id        | shape\_pt\_sequence | shape\_pt\_lat | shape\_pt\_lon | shape\_dist\_traveled | distance with previous (m) based on lat/lon |
| ---------------- | ------------------- | -------------- | -------------- | --------------------- | ------------------------------------------- |
| 46-138-sj2-1.7.R | 446                 | \-32.85883235  | 151.800061     | 39854.26715           |                                             |
| 46-138-sj2-1.7.R | 447                 | \-32.86204298  | 151.7950367    | 39854.26715           | 589.6                                       |
| 46-138-sj2-1.7.R | 448                 | \-32.8623903   | 151.7946713    | 39905.89078           | 51.5                                        |
| 46-138-sj2-1.7.R | 657                 | \-32.92498991  | 151.7632449    | 56225.13937           |                                             |
| 46-138-sj2-1.7.R | 658                 | \-32.92600089  | 151.7676676    | 56225.13937           | 427.8                                       |
| 46-138-sj2-1.7.R | 659                 | \-32.92592776  | 151.7676957    | 56233.68338           |

botanize:
Thanks @isabelle-dr, those examples give me a little more confidence in my preference for eliminating the threshold and calling EqualShapeDistanceDiffCoordinatesNotice an error.

evansiroky:
> Thanks @botanize ! Not answering directly your comment here but if this could help us make an informed decision, I had a look "manually" at a few of the datasets that contain a good portion `EqualShapeDistanceDiffCoordinatesNotice`, to evaluate if we still have false positives, here are the results.
> 
> **TL;DR**: they all look like true positives to me (i.e. actual error not resulting from a rounding problem)
> 
> Details here

Hello, please take a look at [`san-joaquin-rtd-gtfs-q38059`](https://storage.googleapis.com/storage/v1/b/san-joaquin-rtd-gtfs-q38059_archives_2021-12-02/o/4cb4d8a24db4b3a33dd7bb4a45f1c27914add6fc.zip?alt=media) as an example of where the threshold could be useful for generating EqualShapeDistanceNearCoordinatesNotice as a WARNING.

lionel-nj:
After discussion with @isabelle-dr and @LeoFrachet, to apply such a threshold, the specification would need to be changed so that data producers/consumers are fully aware of it. Hence, in a first step we will drop the threshold and keep the door open to the discussion for further improvement.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/ShapeIncreasingDistanceValidator.java
```diff
@@ -47,57 +56,119 @@ public void validate(NoticeContainer noticeContainer) {
       for (int i = 1; i < shapeList.size(); ++i) {
         GtfsShape prev = shapeList.get(i - 1);
         GtfsShape curr = shapeList.get(i);
-        if (prev.hasShapeDistTraveled()
-            && curr.hasShapeDistTraveled()
-            && prev.shapeDistTraveled() >= curr.shapeDistTraveled()) {
+        if (!(prev.hasShapeDistTraveled() || curr.hasShapeDistTraveled())) {
```
maximearmstrong, line 59: Not quite sure, but do you want to verify that both `shape_dist_traveled` are present before validating them? If yes, the condition should be `!(prev.hasShapeDistTraveled() && curr.hasShapeDistTraveled())` because you want it to be false only if both `shape_dist_traveled` are present, not one or the other. In the current state, if `prev.hasShapeDistTraveled() == true` and `curr.hasShapeDistTraveled() == false`, you would have `(prev.hasShapeDistTraveled() || curr.hasShapeDistTraveled()) == (true || false) == (true)` and `!(true) == false`. We would end up validating empty/null `shape_dist_traveled` values, so potential problems here.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/ShapeIncreasingDistanceValidator.java
```diff
@@ -47,57 +56,119 @@ public void validate(NoticeContainer noticeContainer) {
...
+          noticeContainer.addValidationNotice(
+              new EqualShapeDistanceSameCoordinatesNotice(prev, curr));
         }
       }
     }
   }
 
   /**
-   * When sorted on `shapes.shape_pt_sequence` key, shape points should have strictly increasing
+   * When sorted on {@code shapes.shape_pt_sequence} key, shape points must have strictly increasing
```
maximearmstrong, line 83: Same as described in RULES.md, I would go with "must not have decreasing values" since the condition verified is `prev.shapeDistTraveled() > curr.shapeDistTraveled()`.
```suggestion
   * When sorted on {@code shapes.shape_pt_sequence} key, shape points must not have strictly decreasing
```

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/ShapeIncreasingDistanceValidator.java
```diff
@@ -47,57 +56,119 @@ public void validate(NoticeContainer noticeContainer) {
...
+      this.shapeDistTraveled = current.shapeDistTraveled();
+      this.shapePtSequence = current.shapePtSequence();
+      this.prevCsvRowNumber = previous.csvRowNumber();
+      this.prevShapeDistTraveled = previous.shapeDistTraveled();
+      this.prevShapePtSequence = previous.shapePtSequence();
+    }
+  }
+
+  /**
+   * When sorted on {@code shapes.shape_pt_sequence} key, close shape points should not have equal
```
maximearmstrong, line 113: Since we have removed the threshold, should we keep the "close"?

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/ShapeIncreasingDistanceValidator.java
```diff
@@ -47,57 +56,119 @@ public void validate(NoticeContainer noticeContainer) {
...
+      this.shapeDistTraveled = current.shapeDistTraveled();
+      this.shapePtSequence = current.shapePtSequence();
+      this.prevCsvRowNumber = previous.csvRowNumber();
+      this.prevShapeDistTraveled = previous.shapeDistTraveled();
+      this.prevShapePtSequence = previous.shapePtSequence();
+    }
+  }
+
+  /**
+   * When sorted on {@code shapes.shape_pt_sequence} key, distant shape points must not have equal
```
maximearmstrong, line 143:
```suggestion
   * When sorted on {@code shapes.shape_pt_sequence} key, shape points with different coordinates must not have equal
```

maximearmstrong (changes requested):
Some questions and comments in-line before approval :)

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTimeIncreasingDistanceValidator.java
```diff
@@ -25,12 +25,12 @@
...
 import org.mobilitydata.gtfsvalidator.table.GtfsStopTime;
 import org.mobilitydata.gtfsvalidator.table.GtfsStopTimeTableContainer;
-import org.mobilitydata.gtfsvalidator.validator.ShapeIncreasingDistanceValidator.DecreasingOrEqualShapeDistanceNotice;
+import org.mobilitydata.gtfsvalidator.validator.ShapeIncreasingDistanceValidator.DecreasingShapeDistanceNotice;
 
 /**
  * Validates: stop times of a trip have increasing distance (stops.shape_dist_traveled)
  *
- * <p>Generated notice: {@link DecreasingOrEqualShapeDistanceNotice}.
+ * <p>Generated notice: {@link DecreasingShapeDistanceNotice}.
```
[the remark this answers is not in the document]
lionel-nj: My bad - no changes should have happened in this file, `RULES.md` and `NOTICES.md` have all the information regarding `DecreasingOrEqualStopTimeDistanceNotice`.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/ShapeIncreasingDistanceValidator.java
lionel-nj: Indeed, let me remove that.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/ShapeIncreasingDistanceValidator.java
lionel-nj: Good catch! Thanks.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTimeIncreasingDistanceValidator.java
```diff
@@ -25,12 +25,11 @@
 import org.mobilitydata.gtfsvalidator.notice.ValidationNotice;
 import org.mobilitydata.gtfsvalidator.table.GtfsStopTime;
 import org.mobilitydata.gtfsvalidator.table.GtfsStopTimeTableContainer;
-import org.mobilitydata.gtfsvalidator.validator.ShapeIncreasingDistanceValidator.DecreasingOrEqualShapeDistanceNotice;
 
 /**
  * Validates: stop times of a trip have increasing distance (stops.shape_dist_traveled)
  *
- * <p>Generated notice: {@link DecreasingOrEqualShapeDistanceNotice}.
+ * <p>Generated notice: {@link DecreasingOrEqualStopTimeDistanceNotice}.
```
lionel-nj, line 32: Documentation in the master branch was erroneous.

RULES.md
```diff
@@ -262,6 +264,15 @@ Empty csv file found in the archive: file does not have any headers, or is a req
 ##### References:
 * [GTFS files requirements](https://gtfs.org/reference/static#file-requirements)
 
+#### EqualShapeDistanceDiffCoordinatesNotice
+
+<a name="EqualShapeDistanceDiffCoordinatesNotice"/>
+
+When sorted by `shape.shape_pt_sequence`, two consecutive shape points with different coordinates must not have equal values for `shape_dist_traveled`.
```
isabelle-dr, line 271: In order to give the user a bit more context, I would rephrase this to something like:
When sorted by `shape.shape_pt_sequence`, the values for `shape_dist_traveled` must increase. 
Two consecutive points with equal values for `shape_dist_traveled` and different lat / lon coordinates indicate an error.

RULES.md
```diff
@@ -685,6 +696,15 @@ A row in the input file has only spaces.
 ##### References:
 * [GTFS file requirements](http://gtfs.org/reference/static/#file-requirements)
 
+#### EqualShapeDistanceSameCoordinatesNotice
+
+<a name="EqualShapeDistanceSameCoordinatesNotice"/>
+
+When sorted by `shape.shape_pt_sequence`, two consecutive shape points should not have equal values for `shape_dist_traveled`.
```
isabelle-dr, line 703: In order to give the user a bit more context, I would rephrase this to something like:
When sorted by `shape.shape_pt_sequence`, the values for `shape_dist_traveled` must increase. 
Two consecutive points with equal values for `shape_dist_traveled` and the same values for lat / lon coordinates indicate a duplicative shape point.

isabelle-dr (approved):

maximearmstrong (approved):
LGTM!

lionel-nj:
Thank you for your contribution ! @evansiroky @botanize @maximearmstrong @isabelle-dr

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/ShapeIncreasingDistanceValidatorTest.java
```diff
@@ -34,7 +36,7 @@ public static GtfsShape createShapePoint(
       double shapePtLat,
       double shapePtLon,
       int shapePtSequence,
-      double shapeDistTraveled) {
+      Double shapeDistTraveled) {
```
barbeau, line 39: Sorry for coming in late here, but is there a reason this was changed from `double` to `Double`?

lionel-nj: This was done in order to support testing agains null value.

Issue #495, opened 2020-11-30
[REQUEST] API for Web UI


NomeQ:
Is there any plan in place to create an API for the web UI, or any other type of API wrapper for the validator?

My primary motivation is in being able to build the service into an automated pipeline; export gtfs -> POST to a running instance of the validator -> Receive results.json

I can see an argument for building it into the existing code, since there is already a Springboot web application, but the API also would not necessarily need the UI aspect, and could also be built as a separate module, something like gtfs-validator-service

f8full:
@Cristhian-HA

derhuerst:
As a "random stranger", I'm going to provide my perspective here:

From my experience, having an "official" UI that is known to work with the backend is nice to have, but often one UI doesn't fit all needs. Having a publicly documented and stable API is a big benefit, both for the official UI as well as third-party integrations.

I'm therefore in favour of decoupling the UI enough from the backend to allow for a decent development experience with third-party integrations.

Cristhian-HA:
Hello Naomi and Jannis

Thank you for your comments. We are currently working on the new version of the GTFS Validator (v2.0) that can handle larger datasets and also the processing speed has been improved. Once we will conclude the work on the validator, we will start defining the scope of work that is needed for the UI. 
We plan to split the work in 2 parts. In the first one, the UI will present the results and the second one we will integrate the validator in the we UI.

All the feedback and comments that we receive on this project, will be taken in consideration when defining the scope of work so do not hesitate to write to us and provide your valuable feedback.

Have a great day !

f8full:
Welcome around @derhuerst ! :) Thks for your input, they read everything here.

From a _"random stranger"_ to another ;)

Issue #445, opened 2020-10-21
Implement solution to gracefully shutdown web-app and enable switch to cli-app in Docker image


lionel-nj:
**Is your feature request related to a problem? Please describe.**
We have not implemented yet a solution to gracefully shutdown web-app and enable switch to cli-app in Docker image.

**Describe the solution you'd like**
- add code in spring-server to shutdown web-app gracefully. 
- implement a button in the `react-client` to shut down the process
- produce an only Docker image that will contain the `cli-app`  and the `web-app`.

**Describe alternatives you've considered**

**Additional context**

barbeau:
Closing as outdated - future work on web UI won't use the existing implementation. More info on future web UI at #495.

PR #421, opened 2020-10-13
Local web UI

lionel-nj:
**Summary:**

This PR provides support to develop a local web ui based on the existing gtfs validator using spring boot, and react/

**Expected behavior:** 

1. Run the SpringBoot application:
`./gradlew bootRun` -> will start by default on server port 8080,

To run the web app :
1. go to react-client directory
2. `yarn start`

image

**Additional information:** 
This PR takes content from #365 and #413 
This PR comes as replacement to the two aforementioned PRs. This PR was created in an attempt to solve a packaging bug. Therefore all changes from #365 were repatriated here in order to provide working code.

Please make sure these boxes are checked before submitting your pull request - thanks!

- [x] Run the unit tests with `gradle test` to make sure you didn't break anything
- [ ] Format the title like "Fix #<issue_number> - <short description of fix and changes>" (for example - "Fix #1111 - Check for null value before using field")
- [x] Linked all relevant issues
- [x] Include screenshot(s) showing how this pull request works and fixes the issue(s)

PR #1088, opened 2022-01-26
feat: html output, web front end

ed-g:
**Summary:**

### Features

* Create an html report in the same folder as the json report when run via command line.
* Add an embedded web server which can be run via a  `--web_server` option.

### Build

* Use spotless to check java syntax instead of com.github.sherter.google-java-format which appears to be unmaintained.
* Fix various small gradle warnings.
* Added a build.gradle task to compile js libraries, and bundle web front end assets to java source code.

**Expected behavior:** 

* Command line validation writes a report.html file in the output directory, the file name can be chosen by  `--validation_html_report_name`.

* `--web_server` will run an embedded web server on port 6888, where one can submit GTFS files for validation.

* `./gradlew generateViewerAssets` will re-generate ViewerAssets.java based on source html, js, and css. There might be better ways to achieve this, I'm not very familiar with best practices for java bundling or code generation. UPDATE: I've added a dependency from `build` to the js compilation and bundling steps. I think this is how gradle is supposed to work but I'd love a review from someone who is knowledgeable. See `main/build.gradle`.

https://github.com/MobilityData/gtfs-validator/issues/495


evantrillium:
a much needed feature to expand access to the validator beyond consuming software applications! Producers have relied on the quite stale feedvalidator.py script from the transitfeeds library of scripts and not having a human-readable out for this validator has hindered adoption.

isabelle-dr:
Hello, thank you so much for this contribution @ed-g! 
This is indeed one of the top requested features for this tool :) 

One question: what is the reason you chose Spark Java?
Not sure if you were aware, we previously started to work on an Interface using Spring Boot (in [PR#421](https://github.com/MobilityData/gtfs-validator/pull/421)).

ed-g:
Hello @isabelle-dr! You're very welcome :-)

My coworker and I took a look for the previous web ui, based on this comment, https://github.com/MobilityData/gtfs-validator/issues/445#issuecomment-764779942 and https://github.com/MobilityData/gtfs-validator/issues/495#issuecomment-749597294 we got the sense that 2.0 updates may have changed some internals that it relied upon?

There is no particular reason why I chose spark java over spring boot or another web framework. I think it could be ported to pretty much anything if the team has another preference.

ed-g:
I've added some basic Jasmine tests and merged report text and css updates.

Our current project funding ends today, please don't be shy if you have any quick requests for improvements.

main/src/main/java/org/mobilitydata/gtfsvalidator/viewer/ValidatorWebServer.java
```diff
@@ -0,0 +1,238 @@
...
+      return "Server error.";
+
+    } catch (InterruptedException e) {
+      logger.atSevere().withCause(e).log("Validation was interrupted");
+      res.status(500);
+      return "Server error.";
+    }
+
+    return ViewerIndex.IndexHtmlAll(
+        "var report = " + prettyGson.toJson(noticeContainer.exportValidationNotices()) + ";");
```
themightychris, line 160: Would it be trivial to check here if an `Accept: application/json` and respond with the JSON instead of an HTML render? This would make it trivial to run the existing docker container as a web service on a cluster that separately maintained and deployed static web UIs or other services could leverage. 

I'm kind of looking for the inverse of #1119—a way to run the validator as a persistent web service one could POST GTFS to and get back the JSON without any user interface being served by the validator itself. Of course sensitivity to an `Accept` request handler seems like a trivial way this PR could do both.

themightychris:
Can anyone share a brief snapshot on where this is on getting merged? Are there blockers?

main/src/main/java/org/mobilitydata/gtfsvalidator/viewer/ValidatorWebServer.java
ed-g: I'm only superficially familiar with Spark, but it looks like the https://sparkjava.com/documentation#request object will provide the `Accept` header, but it would still be necessary to parse the list of media types along with their quality factors.

That sort of thing is not built into Spark but a template engine (they support multiple: https://sparkjava.com/documentation.html#views-and-templates) might provide that feature.

It might be easier to simply add a post handler for `/validate/report.json` which always returns JSON from the form submission.

That said: this web server is meant as a proof of concept, from what I understand per #1112 the future web UI will likely be written in Spring based on work in #421. So any JSON feature may need to be rewritten against that framework once it's ready.

isabelle-dr:
Closing this PR, since this will not be worked on further

PR #1086, opened 2022-01-10
build: use mavenCentral() instead of jcenter()

ed-g:
Gradle says jcenter() has be deprecated. I've run tests before and after this change, and in both cases the `Task :output-comparator:test FAILED`, so at least there is no difference.

**Summary:**

Fixes warnings on build.

**Expected behavior:** 

Please make sure these boxes are checked before submitting your pull request - thanks!

- [x] Run the unit tests with `gradle test` to make sure you didn't break anything
> Task :output-comparator:test FAILED

org.mobilitydata.gtfsvalidator.outputcomparator.cli.MainTest > tooManyCorruptedSource_invalidTest FAILED
    com.google.common.truth.ComparisonFailureWithFacts at MainTest.java:356

8 tests completed, 1 failed

- [x] Format the title like "feat: [new feature short description]". Title must follow the Conventional Commit Specification(https://www.conventionalcommits.org/en/v1.0.0/).
- [x] Linked all relevant issues
- [ ] Include screenshot(s) showing how this pull request works and fixes the issue(s)

PR #1091, opened 2022-01-31
test: allow either order for tooManyCorruptedSource_invalidTest

ed-g:
**Summary:**

tooManyCorruptedSource_invalidTest for https://github.com/MobilityData/gtfs-validator/pull/1086#issuecomment-1026223882


output-comparator/src/test/java/org/mobilitydata/gtfsvalidator/outputcomparator/cli/MainTest.java
```diff
@@ -347,15 +349,20 @@ public void tooManyCorruptedSource_invalidTest() throws Exception {
...
-                    ACCEPTANCE_TEST_REPORT_FOLDER_NAME,
-                    SOURCES_CORRUPTION_REPORT_JSON)))
-        .isEqualTo(
+    MatcherAssert.assertThat(
+        retrieveReportString(
+          resolve(
+            NEW_NOTICES_TYPE_FOLDER_NAME,
+            ACCEPTANCE_TEST_REPORT_FOLDER_NAME,
+            SOURCES_CORRUPTION_REPORT_JSON)),
+        Matchers.anyOf(
```
maximearmstrong, line 358: I'm okay with the code. Did you encountered a use case where the order of `source-id-5` and `source-id-4` is reversed? I tried to run the original test after reversing the order and the test is failing.

maximearmstrong (commented):
Thank you for your contribution @ed-g! I have a question in-line before approving this PR. 
If you have a use case where the order of the sources is reversed, could you please provide more information on it?

output-comparator/src/test/java/org/mobilitydata/gtfsvalidator/outputcomparator/cli/MainTest.java
maximearmstrong: Oh! I just realized your explanation was in #1086. Sorry about that. I am unable to reproduce the bug. Could you provide before and after screenshots in this PR to document the fix?

maximearmstrong:
Also, the acceptance tests on PRs from forks were fixed with #1102. Could you fix the format of the code and use the acceptance tests workflow from the master branch? Then we can see if all the checks pass :)

ed-g:
The only time I've seen an reversed order is in January when I was doing initial tests.

If you're not seeing it we may not need to add the case to the test.

maximearmstrong:
I tested this again with different Java versions and I didn't experience the issue. If it happens again, please let us know :)

Issue #1094, opened 2022-02-01
Acceptance tests fail for PR opened from forked repository


barbeau:
**Describe the bug**
The acceptance tests implemented in https://github.com/MobilityData/gtfs-validator/pull/848 are important because they detect changes to validation output on all datasets in the MobilityDatabase based on changes to validation rule logic.

However, as seen in https://github.com/MobilityData/gtfs-validator/pull/1090#pullrequestreview-869252576, these tests fail to execute for forks of the gtfs-validator repo (i.e., all external contributions to the project):

>For information Rule acceptance tests fails because the PR merges code from another github repository that does not have access to the credentials required to set up and authorize the usage GCP.

**How we reproduce the bug**
Steps to reproduce the behaviour:
1. Fork the gtfs-validator
2. Change code and open a PR

**Expected behaviour**
Acceptance tests should complete with a report on changes in validation output for MobilityData datasets as seen in https://github.com/MobilityData/gtfs-validator/pull/1083#issuecomment-1018599733.

**Observed behaviour**
It appears the reports can't be persisted to Google Cloud storage because of permissions issues on the credentials, causing the process to fail.

**Screenshots:** 
See https://github.com/MobilityData/gtfs-validator/pull/1090#pullrequestreview-869252576

**Environment versions**
- validator version: Master branch as of https://github.com/MobilityData/gtfs-validator/commit/d797a13e4610f0cacb28e5fb573d94764537857f
- Java version: GitHub Action (see master branch)
- OS versions: GitHub Action (see master branch)

Issue #755, opened 2021-02-17
GTFS Validator as library


damianVarela:
Are there any plans on making this awesome validator a Java Library we can import thru maven?

carlfredl:
👋  for us at MobilityData this hasn't come up as a priority. We have captured this idea haven't planned work on this yet on our side but we would be very happy to see others contribute something.

carlfredl:
As we discussed, @barbeau noted that in theory anyone could use the current JAR file as a dependency in their Java project (whether or not it's available on Maven Central) and have access to the Java objects that are currently implemented. Most of the classes are public IIRC, so they should be accessible as a dependency. Without looking further we don't know how practical it would be as-is, given it's design as it wasn't specifically designed as a library. It might work as-is.

carlfredl:
There are a couple use cases we could think of. Do you mind sharing yours @damianVarela ?

**Integration of objects into java application use case**
Library would allow you to integrate the code/objects directly into your Java application without having to consume the JSON output of the validator. This would be most useful for people that want to extend the validator to implement their own custom rules and notices. For example, on the GTFS-realtime validator, there was a split into a library and application component to make it easier to use as a library - https://github.com/CUTR-at-USF/gtfs-realtime-validator#building-the-project. 

MobilityData's higher priority in this direction is creating the functionality of "profiles" connected to various additional validation rules to give data producers transparency on what is being looked at by various consumers beyond the spec its self.

**Data processing use case**
Use case for use as a library would be if you have a Java application that you want to do a lot of data processing and manipulation with, and as part of that application you want to run validation. For example, you have an app, and you want to integrate the validator output into your existing data processing pipeline. If you can use the gtfs-validator as a library, you can execute it directly from the Java code and get the results in the form of a `List`. If you can't use the validator as a library, then in your data pipeline you need to figure out a way to halt your existing pipeline processing, interact with the command-line somehow via scripts to launch the validator, figure out how to wait for the results to be output from the validator (when you don't know how long it's going to take to execute), and then process the JSON output file when validation is complete, and then resume your processing pipeline. Using directly as a library simplifies all this significantly because you can just execute the code synchronously within your own app, and when validation is complete your pipeline just picks back up automatically. An example project that uses the GTFS-rt validator this way is  https://github.com/CUTR-at-USF/transit-feed-quality-calculator.

barbeau:
[...]

PR #1102, opened 2022-02-09
ci: Remove upload of validation reports to Google Cloud

lionel-nj:
closes https://github.com/MobilityData/gtfs-validator/issues/1094

**Summary:**

This PR removes the usage of GCP cloud storage. The content of this PR fixes the behavior described [here](https://github.com/MobilityData/gtfs-validator/issues/1094)

**Expected behavior:** 

On forks, acceptance tests should run and not fail because of missing credentials: upload to cloud storage is not used - on this GitHub repository, google cloud storage is used as a backup. 

### DIFF
--- a/.github/workflows/acceptance_test.yml
+++ b/.github/workflows/acceptance_test.yml
@@ -186,16 +186,6 @@ jobs:
         with:
           name: reports_all
           path: ${{ github.sha }}/output
-      - name: Set up and authorize Cloud
-        uses: google-github-actions/auth@v0
-        with:
-          credentials_json: ${{ secrets.VALIDATOR_SA_KEY }}
-      - name: Upload reports to Google Cloud Storage
-        id: upload-files
-        uses: google-github-actions/upload-cloud-storage@main
-        with:
-          path: ${{ github.sha }}/output
-          destination: gtfs-validator-reports
   compare-outputs:
     needs: [ get-reports ]
     runs-on: ubuntu-latest
--- a/.github/workflows/acceptance_test.yml
+++ b/.github/workflows/acceptance_test.yml
@@ -168,8 +168,7 @@ jobs:
       - name: Check for secrets needed to upload validation reports to GCP Cloud Storage
         id: check-secrets
         run: |
-          if [ -n "${{ secrets.VALIDATOR_SA_KEY }}" ]
-          then
+          if [ "${{ secrets.VALIDATOR_SA_KEY }}" == "" ]; then
             echo "Secrets to upload validations reports to Google Cloud are configured in the repo."
             echo "::set-output name=ok::true"
           else
--- a/.github/workflows/acceptance_test.yml
+++ b/.github/workflows/acceptance_test.yml
@@ -168,7 +168,7 @@ jobs:
       - name: Check for secrets needed to upload validation reports to GCP Cloud Storage
         id: check-secrets
         run: |
-          if [ -n "${{ secrets.VALIDATOR_SA_KEY }}" ]
+          if [ ! -z "${{ secrets.VALIDATOR_SA_KEY }}" ]
           then
             echo "Secrets to upload validations reports to Google Cloud are configured in the repo."
             echo "::set-output name=ok::true"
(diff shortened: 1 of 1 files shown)


.github/workflows/acceptance_test.yml
```diff
@@ -186,16 +186,6 @@ jobs:
         with:
           name: reports_all
           path: ${{ github.sha }}/output
-      - name: Set up and authorize Cloud
```
barbeau, line 189: Could you describe more what we lose by disabling this? Can we still download the reports as linked in the comment generated by the acceptance test (e.g., "Download the full acceptance test report for commit https://github.com/MobilityData/gtfs-validator/commit/6e94daab39c7560d996a3dc5d86b21c15aac49ce [here] (report will disappear after 90 days)"? If not does that comment need to change?

If it possible to check here if the Action has permission to use the credentials? If so, we could still run this block for branches on this repo, but just not for 3rd party forks.

barbeau (commented):
Thanks @lionel-nj! Comments in-line.

barbeau:
Also, I believe the link you have in your original post of "The process of the acceptance tests will solely rely on the internal Github storage in order to avoid the behavior described https://github.com/MobilityData/gtfs-validator/issues/755." is pointing to the wrong issue.

.github/workflows/acceptance_test.yml
lionel-nj: > Could you describe more what we lose by disabling this?
By doing so, we loose the ability to store validation reports on GCP Cloud storage. But it is still possible to download them from the internal storage of github. 

> Can we still download the reports as linked in the comment generated by the acceptance test (e.g., "Download the full acceptance test report for commit https://github.com/MobilityData/gtfs-validator/commit/6e94daab39c7560d996a3dc5d86b21c15aac49ce [here] (report will disappear after 90 days)"? If not does that comment need to change?

Yes - this remains doable. 

> If it possible to check here if the Action has permission to use the credentials? If so, we could still run this block for branches on this repo, but just not for 3rd party forks.
 
Let me see.

lionel-nj: From a [quick search](https://github.community/t/how-can-i-test-if-secrets-are-available-in-an-action/17911) it appears that checking the presence of a repo secret in a `if` block of a job's step might not be possible. However, it might be possible to include an intermediate step that consists in checking the presence of said secret and store the boolean in an environmental variable, then execute `Set up and authorize Cloud` and `Upload reports to Google Cloud Storage` conditionally. Unfortunately I will not have the time to do it 😔.

barbeau: Looks like this is the issue regarding making execution conditional on secrets - https://github.com/actions/runner/issues/520

barbeau:
@maximearmstrong @isabelle-dr I've expanded on the work that @lionel-nj started here to effectively skip all validation report uploads to Google Cloud, which should fix acceptance tests on PRs from forks and therefore close https://github.com/MobilityData/gtfs-validator/issues/1094.

Ideally we'd conditionally upload reports to Google Cloud only if the PR is for a branch on this repo (and not upload if the PR is from a fork). However, this IF statement in the GitHub Action isn't currently working as expected.

Given that the acceptance tests running on forks is a critical part of the workflow of this repo for evaluating PRs like https://github.com/MobilityData/gtfs-validator/pull/1125, and uploading reports to Google Cloud is effectively a "nice to have" feature, I propose that we merge this PR as-is to fix the acceptance tests for forks. 

We can then fix the Google Cloud uploads in future work. I've opened issue https://github.com/MobilityData/gtfs-validator/issues/1128 to track re-enabling Google Cloud uploads in the future by tweaking the existing code in this PR. I've also marked the code that's not working as intended with a `FIXME` statement pointing to that issue.

So as long as all the acceptance tests pass for my latest push to this branch, then as far as I know this should be ok to merge if you agree.

maximearmstrong:
Thank you for this work @barbeau! 

@isabelle-dr Do you see value in keeping the reports for more than 90 days? If yes, I agree with Sean that we should merge this PR as-is to fix the problem and then address the Google Cloud issue. If no, I suggest that we remove the code related to Google Cloud right away.

barbeau:
@maximearmstrong @isabelle-dr Sounds good - I've removed the Google Cloud upload code entirely in https://github.com/MobilityData/gtfs-validator/pull/1102/commits/ced54b9b200b4b7624c8a8dc8cc5a45590220cc7.

After the tests run this PR should be ready for review/merge.

maximearmstrong (approved):
Perfect! Thank you @barbeau and @lionel-nj for this contribution :)

PR #1119, opened 2022-04-14
feat: html output

ed-g:
**Summary:**

### Features

* Create an html report in the same folder as the json report when run via command line.

### Build

* Use spotless to check java syntax instead of com.github.sherter.google-java-format which appears to be unmaintained.
* Fix various small gradle warnings.
* Added a build.gradle task to compile js libraries, and bundle web front end assets to java source code.

**Expected behavior:** 

* Command line validation writes a report.html file in the output directory, the file name can be chosen by  `--validation_html_report_name`.

* `./gradlew generateViewerAssets` will re-generate ViewerAssets.java based on source html, js, and css. There might be better ways to achieve this, I'm not very familiar with best practices for java bundling or code generation. UPDATE: I've added a dependency from `build` to the js compilation and bundling steps. I think this is how gradle is supposed to work but I'd love a review from someone who is knowledgeable. See `main/build.gradle`.

https://github.com/MobilityData/gtfs-validator/issues/495


ed-g:
This is the same as https://github.com/MobilityData/gtfs-validator/pull/1088 minus the web UI.

main/build.gradle
```diff
@@ -112,3 +143,99 @@ configurations {
...
+ext {
+    viewerDir = "$projectDir/src/main/java/org/mobilitydata/gtfsvalidator/viewer/"
+}
+
+task installSnabbdom(type: Exec) {
+  inputs.file "$projectDir/js/snabbdom-umd/package.json"
+  outputs.dir "$projectDir/js/snabbdom-umd/node_modules"
+
+  workingDir "$projectDir/js/snabbdom-umd"
+  executable 'npm'
```
barbeau, line 156: @ed-g Do I need to install NPM for the project build?

I'm getting this when I run `./gradlew build` or try to run the CLI `Main.java` from within IntelliJ:

```
> Task :main:installSnabbdom FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':main:installSnabbdom'.
> A problem occurred starting process 'command 'npm''
```

I'm on Windows 10 Home on this machine so I'm wondering if this is a Windows issue.

ed-g: @barbeau 

Oh right, the snabbdom update processes uses `npm`.

The git repository ships with a previously downloaded copy of snabbdom as `main/src/main/java/org/mobilitydata/gtfsvalidator/viewer/snabbdom.browser.js`

The purpose of the `rollupSnabbdomUmd` gradle task is to refresh from upstream so project maintainers don't end up with an outdated version of snabbdom over time, and to act as some executable documentation for how that file was originally made. This was my attempt to implement dependency management for the javascript side based on the gradle reference I was using, there might be a more idiomatic way.

If you want to skip it for the initial test, the line

```
    dependsOn rollupSnabbdomUmd
```

can be temporarily commented out of `main/build.gradle`

main/build.gradle
```diff
@@ -15,12 +15,36 @@
...
 plugins {
-    id 'com.github.johnrengelman.shadow' version '5.2.0'
+    id 'com.github.johnrengelman.shadow' version '7.1.2'
     id 'application'
     id 'java'
+    id "com.diffplug.spotless" // Version provided in ../build.gradle.
     id 'maven-publish'
 }
 
+spotless {
```
barbeau, line 25: I don't believe you should need this, configuring once in the root `build.gradle` should be enough

main/build.gradle
```diff
@@ -85,11 +110,17 @@ dependencies {
     implementation 'com.google.flogger:flogger-system-backend:0.6'
     implementation 'com.univocity:univocity-parsers:2.9.0'
     implementation 'com.google.geometry:s2-geometry:2.0.0'
+    implementation 'com.sparkjava:spark-core:2.9.3' // Spark web framework.
```
barbeau, line 113: This dependency for Spark can be removed?

main/build.gradle
barbeau: Thanks! What I don't understand, though, is why GitHub Action CI is passing and successfully packaging the app if I can't do the same with `./gradlew build` or `./gradlew shadowJar` on my local machine, which made me wonder if it's a specific Windows issue. Because you didn't configure the GitHub Action to have NPM installed, right?

Another related item is that we don't want unnecessary Gradle tasks to run on each build for performance reasons. Does `generateViewerAssets` need to run on every build? If so, could we limit the scope so it only runs if files in a particular package are changed? There are some examples of how to do this here - https://barbeau.medium.com/how-to-avoid-duplicate-code-for-similar-sdks-7e7812e67708.

barbeau: @ed-g Could you merge the master branch into this PR? The CI should now run GitHub Actions on Windows as well which should help troubleshoot the above issue.

ed-g: @barbeau

I've merged master into this PR. There was a minor doc conflict, I believe my merge makes sense there.

> Another related item is that we don't want unnecessary Gradle tasks to run on each build for performance reasons. Does generateViewerAssets need to run on every build? If so, could we limit the scope so it only runs if files in a particular package are changed? There are some examples of how to do this here - https://barbeau.medium.com/how-to-avoid-duplicate-code-for-similar-sdks-7e7812e67708.

The task should be looking at $projectDir/js/snabbdom-umd/{package.json,node_modules} and not updating unless one these change (in other words, if the version is updated). On my machine it only runs once. However this is my first experience with Gradle and while I followed the docs as far as I could, there might be a better way to accomplish this.

main/build.gradle
ed-g: Yes absolutely

main/build.gradle
ed-g: Uncertain. I think it wasn't running unless it was also defined in the sub directories. If I recall it was saying everything was fine, but not actually looking at the code under main.

I'm a little out of my depth and could use help/advice from a gradle java expert here.

main/build.gradle
barbeau: Looks like the GitHub CI on Windows is failing with the same error I saw previously:
https://github.com/MobilityData/gtfs-validator/runs/6102573605?check_suite_focus=true


```
> Task :core:jar
> Task :main:installSnabbdom FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':main:installSnabbdom'.
> A problem occurred starting process 'command 'npm''
```

...but the Ubuntu build is fine:
https://github.com/MobilityData/gtfs-validator/runs/6102573447?check_suite_focus=true

```
> Task :core:jar

> Task :main:installSnabbdom

added 3 packages, and audited 4 packages in 2s

found 0 vulnerabilities

> Task :main:rollupSnabbdomUmd

> Task :main:generateViewerAssets

./node_modules/snabbdom/build/index.js → ../../src/main/java/org/mobilitydata/gtfsvalidator/viewer/snabbdom.browser.js...
created ../../src/main/java/org/mobilitydata/gtfsvalidator/viewer/snabbdom.browser.js in 183ms
```

So it does look like a Windows thing. https://stackoverflow.com/a/39070534/937715 says:

>Also there is a better way of running node scripts and npm tasks in gradle. Check out the [node-gradle-plugin](https://github.com/srs/gradle-node-plugin) for that. It is a neat wrapper over the exec task you are using.

So it would probably be better to manage npm using something like https://github.com/srs/gradle-node-plugin or https://github.com/node-gradle/gradle-node-plugin.

main/build.gradle
ed-g: Ok I've run some experiments.

You're correct in that we don't need both, but it turns out the one we need is `./main/build.grade`.

What's needed in ./build.gradle is the spotless plugin version, but spotless (for whatever reason) won't look at .java files within ./main/ unless spotless is specifically configured within the main/build.gradle. This is different from how google-java-format works which is why it was confusing and took me a while to figure out (and how I ended up with redundant configuration here).

Made these changes in https://github.com/Trillium-Solutions/gtfs-validator/commit/4feba98edef19342ed3e9c376e01a98bc38c1559

main/build.gradle
ed-g: Excellent! I've swapped in gradle-node-plugin in https://github.com/Trillium-Solutions/gtfs-validator/commit/4feba98edef19342ed3e9c376e01a98bc38c1559 works on local, fingers crossed it works on Windows just as well.

ed-g: It appears that rollup npm command expects bash to be available.

https://github.com/MobilityData/gtfs-validator/runs/6200904503?check_suite_focus=true

main/build.gradle
ed-g: @barbeau are you happy with the changes here?

maximearmstrong: @ed-g Do you think we could create another PR for the spotless task? We can discuss it there and then merge it if it's ready :)

isabelle-dr:
Closing, the work in this PR will be used when we work on a second iteration of the HTML output.

PR #1120, opened 2022-04-15
feat(ci): handle gradle build in Dockerfile

themightychris:
**Summary:**

As a new contributor wanting to try building/running PR branches locally, I was surprised that the `Dockerfile` didn't actually build the application but rather relied on it being built first by a toolchain on the host machine. The intent of `Dockerfile` is usually to capture and pin the build dependencies as well as runtime environment.

**Expected behavior:** 

`docker build .` always goes from branch source to ready-to-run container image

**Proposal:**

Ideally `Dockerfile` handles a full build from source. I also configured the default entrypoint to handle running the CLI, as the instructions indicated running the Docker container would give you a shell you could run the jar on but you'd need to override the entrypoint to `sh` to get that rather than the jShell the JDK base image comes with.

**Testing:**

```bash
docker build . --build-arg=VERSION_TAG=3.4.5 -t gtfs-validator:latest
docker run --rm gtfs-validator:latest --help
```

**Potential improvements:**

- ~~Inserting `RUN apk add npm` after the first `FROM` line is needed on branches that include the HTML UI~~
- [X] Update/extend the Docker instructions in the README if these changes are acceptable in spirit
- [X] Use `openjdk:11-slim` or `openjdk:19-alpine` as the runtime base image
   - Alpine would be ideal as it offers the smallest base image, but `-alpine` variants are only available starting with the latest major version 19, so we would need to verify that everything functions correctly under JDK 19
   - This might best be done in a separate PR since it introduces a higher risk of runtime behavior change and this PR as-is is fairly isolated from causing any runtime behavior changes (see #1185)

**Open questions:**

- Is this an acceptable change?
- Are there any other processes besides the GitHub workflow I updated in this PR that script what happens before `docker build` that would need updating too?

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+.git/
+Dockerfile
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,2 +1,4 @@
 .git/
 Dockerfile
+build/
+config/
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -61,6 +61,7 @@ jobs:
           if [[ $GITHUB_REF == refs/tags/* ]]; then
             VERSION=${GITHUB_REF#refs/tags/}
             VERSION=${VERSION#v}
+            echo ::set-output name=gradleVersionTag::${VERSION}
           elif [[ $GITHUB_REF == refs/heads/* ]]; then
             VERSION=$(echo ${GITHUB_REF#refs/heads/} | sed -r 's#/+#-#g')
           elif [[ $GITHUB_REF == refs/pull/* ]]; then
@@ -70,7 +71,6 @@ jobs:
           if [ "${{ github.event_name }}" = "push" ]; then
             TAGS="${DOCKER_IMAGE}:latest"
           fi
-          echo ::set-output name=gradleVersionTag::${VERSION}
           echo ::set-output name=tags::${TAGS}
           echo ::set-output name=created::$(date -u +'%Y-%m-%dT%H:%M:%SZ')
       - name: Set up JDK 11
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -95,8 +95,6 @@ jobs:
         with:
           context: .
           file: ./Dockerfile
-          build-args: |
-            VERSION_TAG=${{ steps.prep.outputs.gradleVersionTag }}
           push: true
           tags: ${{ steps.prep.outputs.tags }}
           labels: |
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -56,10 +56,11 @@ jobs:
       - name: Prepare
         id: prep
         run: |
-          DOCKER_IMAGE=ghcr.io/mobilitydata/gtfs-validator
+          DOCKER_IMAGE=ghcr.io/${GITHUB_REPOSITORY}
           VERSION=edge
           if [[ $GITHUB_REF == refs/tags/* ]]; then
             VERSION=${GITHUB_REF#refs/tags/}
+            VERSION=${VERSION#v}
           elif [[ $GITHUB_REF == refs/heads/* ]]; then
             VERSION=$(echo ${GITHUB_REF#refs/heads/} | sed -r 's#/+#-#g')
           elif [[ $GITHUB_REF == refs/pull/* ]]; then
@@ -69,6 +70,7 @@ jobs:
           if [ "${{ github.event_name }}" = "push" ]; then
             TAGS="${DOCKER_IMAGE}:latest"
           fi
+          echo ::set-output name=gradleVersionTag::${VERSION}
           echo ::set-output name=tags::${TAGS}
           echo ::set-output name=created::$(date -u +'%Y-%m-%dT%H:%M:%SZ')
       - name: Set up JDK 11
@@ -95,6 +97,8 @@ jobs:
         with:
           context: .
           file: ./Dockerfile
+          build-args: |
+            VERSION_TAG=${{ steps.prep.outputs.gradleVersionTag }}
           push: true
           tags: ${{ steps.prep.outputs.tags }}
           labels: |
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -56,7 +56,7 @@ jobs:
       - name: Prepare
         id: prep
         run: |
-          DOCKER_IMAGE=ghcr.io/${GITHUB_REPOSITORY}
+          DOCKER_IMAGE=ghcr.io/${GITHUB_REPOSITORY,,}
           VERSION=edge
           if [[ $GITHUB_REF == refs/tags/* ]]; then
             VERSION=${GITHUB_REF#refs/tags/}
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -82,10 +82,6 @@ jobs:
           path: ~/.gradle/caches
           key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle') }}
           restore-keys: ${{ runner.os }}-gradle
-      - name: Package cli app jar with Gradle
-        uses: eskatos/gradle-command-action@v1
-        with:
-          arguments: shadowJar
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v1
       - name: Login to GitHub Container Registry
@@ -99,6 +95,8 @@ jobs:
         with:
           context: .
           file: ./Dockerfile
+          build-args: |
+            VERSION_TAG=${{ steps.prep.outputs.gradleVersionTag }}
           push: true
           tags: ${{ steps.prep.outputs.tags }}
           labels: |
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -53,25 +53,31 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - name: Prepare
+      - name: Initialize Gradle
+        run: |
+          # The first time gradlew is invoked it downloads gradle and outputs progress about that to STDOUT.
+          # This "dummy" invokation gets that out of the way so future gradlew commands have clean output
+          ./gradlew --version
+      - name: Prepare version metadata
         id: prep
         run: |
+          # ghcr.io path should match current repo but be all lowercase
           DOCKER_IMAGE=ghcr.io/${GITHUB_REPOSITORY,,}
-          VERSION=edge
-          if [[ $GITHUB_REF == refs/tags/* ]]; then
-            VERSION=${GITHUB_REF#refs/tags/}
-            VERSION=${VERSION#v}
-            echo ::set-output name=gradleVersionTag::${VERSION}
-          elif [[ $GITHUB_REF == refs/heads/* ]]; then
-            VERSION=$(echo ${GITHUB_REF#refs/heads/} | sed -r 's#/+#-#g')
-          elif [[ $GITHUB_REF == refs/pull/* ]]; then
-            VERSION=pr-${{ github.event.number }}
+
+          # delegate to axion-release-plugin to generate version string from Git repository state
+          AXION_VERSION="$(./gradlew currentVersion -q -Prelease.quiet)"
+
+          # determine docker tags
+          if [[ "${GITHUB_EVENT_NAME}" == "release" ]]; then
+            # tag releases with version determined by axion-release-plugin
+            DOCKER_TAGS="${DOCKER_IMAGE}:${AXION_VERSION}"
+          else
+            # tag pushes to master as "latest"
+            DOCKER_TAGS="${DOCKER_IMAGE}:latest"
           fi
-          TAGS="${DOCKER_IMAGE}:${VERSION}"
-          if [ "${{ github.event_name }}" = "push" ]; then
-            TAGS="${DOCKER_IMAGE}:latest"
-          fi
-          echo ::set-output name=tags::${TAGS}
+
+          echo ::set-output name=version::${AXION_VERSION}
+          echo ::set-output name=tags::${DOCKER_TAGS}
           echo ::set-output name=created::$(date -u +'%Y-%m-%dT%H:%M:%SZ')
       - name: Set up JDK 11
         uses: actions/setup-java@v2
@@ -98,11 +104,12 @@ jobs:
           context: .
           file: ./Dockerfile
           build-args: |
-            VERSION_TAG=${{ steps.prep.outputs.gradleVersionTag }}
+            VERSION_TAG=${{ steps.prep.outputs.version }}
           push: true
           tags: ${{ steps.prep.outputs.tags }}
           labels: |
             org.opencontainers.image.source=https://github.com/mobilitydata/gtfs-validator.git
             org.opencontainers.image.created=${{ steps.prep.outputs.created }}
             org.opencontainers.image.revision=${{ github.sha }}
+            org.opencontainers.image.version=${{ steps.prep.outputs.version }}
 # should be org.opencontainers.image.source=${{ github.event.repository.clone_url }} but caps in MobilityData seem to break something
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -10,13 +10,13 @@ jobs:
   validate_gradle_wrapper:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - uses: gradle/wrapper-validation-action@v1
   test:
     needs: [ validate_gradle_wrapper ]
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v3
       - name: Set up JDK 11
         uses: actions/setup-java@v2
         with:
@@ -52,7 +52,7 @@ jobs:
     name: Build and push Docker image
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
         with:
           fetch-depth: 0 # need full clone so ./grade
       - name: Set up JDK 11
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -118,4 +118,3 @@ jobs:
             org.opencontainers.image.created=${{ steps.prep.outputs.created }}
             org.opencontainers.image.revision=${{ github.sha }}
             org.opencontainers.image.version=${{ steps.prep.outputs.version }}
-# should be org.opencontainers.image.source=${{ github.event.repository.clone_url }} but caps in MobilityData seem to break something
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -53,6 +53,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0 # need full clone so ./grade
       - name: Initialize Gradle
         run: |
           # The first time gradlew is invoked it downloads gradle and outputs progress about that to STDOUT.
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -108,7 +108,7 @@ jobs:
           push: true
           tags: ${{ steps.prep.outputs.tags }}
           labels: |
-            org.opencontainers.image.source=https://github.com/mobilitydata/gtfs-validator.git
+            org.opencontainers.image.source=https://github.com/${GITHUB_REPOSITORY,,}.git
             org.opencontainers.image.created=${{ steps.prep.outputs.created }}
             org.opencontainers.image.revision=${{ github.sha }}
             org.opencontainers.image.version=${{ steps.prep.outputs.version }}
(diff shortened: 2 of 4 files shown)


Dockerfile
```diff
@@ -1,3 +1,18 @@
+FROM gradle:7-jdk11-alpine AS build
+
+COPY --chown=gradle:gradle . /build
+WORKDIR /build
+
+ARG VERSION_TAG
+ENV versionTag=$VERSION_TAG
+RUN gradle build --no-daemon
```
barbeau, line 8: To get a fat JAR with all the dependencies bundled you'll need to use `gradle shadowJar` here instead of `gradle build`.

f8full:
Nice ! I definitely didn't know much about Docker when originally implementing so thank you for the new knowledge!

themightychris:
@f8full @barbeau I updated the build to use `shadowJar` and got my CLA signed, should be good to merge now! Let me know if there's anything else I can address

themightychris:
Rebased on top of 3.1.0

README.md
```diff
@@ -87,6 +87,20 @@ where:
 
 The validator can then be executed via bash commands. See the [preceeding instructions for command line usage](#run-the-app-via-command-line).
 
+### Build Docker container locally
```
maximearmstrong, line 90: I think we could add this in the "### Setup section". We would have the subsections "#### Build Docker container locally" and "#### Pull the latest docker image". The current step "1. Download and install Docker" should also be added to this.

maximearmstrong (changes requested):
Thank you @themightychris! Only one thing to change in `README.md`.

More importantly, I get this error when I run the command `docker build . --build-arg=VERSION_TAG=v3 -t gtfs-validator:latest`. Did you experience something similar?

Capture d’écran, le 2022-06-07 à 18 21 19

themightychris:
@maximearmstrong I suspect the error you see is related to existing build/cache assets in your working directory from doing non-dockerized builds getting pulled into the Docker environment. I did this work so I could skip ever installing/running gradle on my host machine so I would be able to test this scenario

Could you try adding `build/` and any other generated artifacts you see in your existing local working tree to the `.dockerignore` file and see if that does the trick?

maximearmstrong:
@themightychris Indeed, that was the problem. Adding the following lines solved the problem.
```
build/
config/
```

.dockerignore
```diff
@@ -0,0 +1,2 @@
+.git/
+Dockerfile
```
maximearmstrong, line 2:
```suggestion
.git/
Dockerfile
build/
config/
```

maximearmstrong (commented):
@themightychris I think we just need to fix the `.dockerignore` and the documentation comment [here](https://github.com/MobilityData/gtfs-validator/pull/1120#discussion_r891761057), then we are ready to merge :)

themightychris:
@maximearmstrong I've:

- incorporated your `.dockerignore` additions
- rebased and updated `Dockerfile` to accommodate the removal of `versionTag`
- updated the README per your recommendation and some other thoughts I had

I think we should be good to go now!

themightychris:
@maximearmstrong can you approve the workflow runs?

Dockerfile
```diff
@@ -1,3 +1,13 @@
...
+
+RUN gradle shadowJar --no-daemon
+
+
 FROM openjdk:11
-COPY main/build/libs/*.jar /
+COPY --from=build /build/main/build/libs/*.jar /
 WORKDIR /
+
+ENTRYPOINT [ "java", "-jar", "gtfs-validator-0.1.0-SNAPSHOT-cli.jar" ]
```
maximearmstrong, line 13: It looks like we don't compute the project version using [this line](https://github.com/MobilityData/gtfs-validator/blob/ef868fa2cce573f1afc22ac73b757025e17f8662/build.gradle#L33) when building via Docker.

This makes the `ENTRYPOINT` stable and works well for local testing, but creates a problem for the ones using the actual docker image. This implementation doesn't allow users to know the real validator version, which is usually displayed in the reports, which I think is very valuable. Any thoughts on this @isabelle-dr?

 If the shadow jar was created using the correct project version, we could use the file name in the `ENTRYPOINT`. @bdferris Any thoughts on this? Do you have an idea why the axion-release plugin would not work here?

maximearmstrong (commented):
Thanks for the changes @themightychris! That one thing aside, I think we are ready to merge. Also, could you sign the CLA agreement?

Dockerfile
bdferris-v2: I believe if you update the actions/checkout@v2 action in the `docker.yml` file to include tag info then the version number will be computed correctly.  See https://github.com/MobilityData/gtfs-validator/blob/master/.github/workflows/package_installers.yml#L33 for an example.

maximearmstrong: Thanks, that's helpful. Do you know if there is a way to do this locally as well? Currently, building with Docker does not calculate the version. We wouldn't need to hardcode the jar file name if it did.

maximearmstrong: That's what I thought too. Currently, when I run this command
```
docker build . -t gtfs-validator:latest
```
I end up building a shadow jar named `gtfs-validator-0.1.0-SNAPSHOT-cli.jar`, which seems also to be the case on Chris' end since the entry point was hardcoded as such. Though, by running `./gradlew shadowJar --no-daemon` locally, the jar file is built with the correct tag version. 

Ideally, we would create an entry point using the tag detected locally and on GitHub. So I'm looking to find out why the axion-release plugin is not detected in the docker build locally. So far, I have had no luck with the documentation of the axion-release plugin on what might be the issue.

themightychris: @maximearmstrong my original version of this PR replicated what the GitHub workflow for building docker did, passing in `$VERSION_TAG` and using that to generate the entrypoint.

I removed that in 6aa301a147d54ed4f367624abb8f393f728b96fe because the jar build process stopped respecting the `$versionTag` input and started building `gtfs-validator-0.1.0-SNAPSHOT-cli.jar` regardless of input.

I found that b93dc1b75c504eac14277d8c8ba97c8f62bf9291, which got merged after I started my PR, removed that functionality so I updated this branch accordingly. We can just revert 6aa301a147d54ed4f367624abb8f393f728b96fe if the preference is to restore `$versionTag`

bdferris-v2: Apologies, I'm new to Docker, took me a second to figure out what's going on.

1) Why wasn't the Docker build resolving the version number correctly in local builds?  Recall that the axion-release plugin uses Git tags in the cloned repo to resolve the current version.  Since `.git` was included in the `.dockerignore` file, none of the Git repro metadata was being copied over with the Docker build context (aka the `COPY --chown=gradle:gradle . /build` directive).  By removing `.git` from `.dockerignore`, I was able to build local Docker images that correctly resolved the release version number.  @themightychris did you add `.git` to `.dockerignore` for a specific reason?  I couldn't think of any reasons, but it doesn't mean they don't exist :)

2) The upshot of (1) is that gradle is now building a jar like `gtfs-validator-3.1.1-SNAPSHOT-cli.jar`.  How to specify the entry point now that the version number is no longer fixed?  I'd use something like the following:

```
COPY --from=build /build/main/build/libs/gtfs-validator-*-cli.jar /gtfs-validator-cli.jar
...
ENTRYPOINT [ "java", "-jar", "gtfs-validator-cli.jar" ]
```

Here, we are copying the gtfs-validator-VERSION-cli.jar (@themightychris note how I didn't copy *.jar, since that picks up `main-VERSION.jar` as well) and rename it to a file without a version number.  That way we can get a consistent reference in the ENTRYPOINT specifier.

I've verified the combo of (1) and (2) works for me on local Docker builds.  I think it should also work for the `docker.yml` release builds, assuming we make the `action/checkout` tweak I mentioned.

I'll admit that I'm not entirely sure what would happen if you attempted `docker build https://github.com/MobilityData/gtfs-validator.git#some_tag`.  Namely, I'm not sure if Docker will preserve enough git metadata to let version resolution work.  Worse-case scenario, it'll fail to resolve the version number, it will default back to 0.1.0-SNAPSHOT, but the ENTRYPOINT should still work correctly.

maximearmstrong: Thanks a lot @bdferris-v2! I tested it locally and it works for me too. The correct version of the validator is displayed in the report. @themightychris If you can do these changes, we will be ready to merge.

themightychris: @bdferris-v2 It's a bit of a bad practice to pull `.git/` into the Docker build context—it can lead to extra or even sensitive content accidentally getting built into the container, and it's a criss-crossing of concerns to have your Dockerfile build process using a git client and knowledge of your release process to figure things out about the release itself.

A preferred approach is to let the system invoking the build determine the designated version and pass that in via a `--build-arg`. We can see that workflow happening previously in 6aa301a147d54ed4f367624abb8f393f728b96fe when `gradle shadowJar` could accept `versionTag` as an environment variable.

If I understand correctly, a plugin being used within the `gradle` workflow is handling extracting the current version from a local `.git/` repo? If that's the case, it would be ideal if that could be made to be overridden environmentally, such that a CI system can pass one in and deactivate the build process trying to inspect `.git/`—Would that be difficult? That's a less quick fix, but this sort of separation of build and release concerns is important for anyone who needs to run their own builds and keep their forks minimal. And it's fine if the `gradle` build layer has some smarts when its run directly as long as they can be overridden in CI via environment

bdferris-v2: High-level: I hear what you are saying.  That said, I do wonder if we'll just be introducing complexity elsewhere for the sake of purity in the Docker file.  But more on that later.

For context, the versioning changes were relatively recent in b93dc1b75c504eac14277d8c8ba97c8f62bf9291 to better support X.Y.Z version numbers, as used in app packaging, version checks, etc.  This moved us away from the previous scheme of extracting version info from $GITHUB_REF in GitHub workflows.  One nice by-product of this change is that version info was now correctly computed for local builds as well.

So how would we make this work in a Docker build without access to .git?

First, it is possible to override the version from the Gradle command-line.  Per the [axion-release plugin docs](https://axion-release-plugin.readthedocs.io/en/latest/configuration/force_version/), you can force the version on the command-line in the following way:

```
./gradlew -Prelease.forceVersion=3.0.0
```

Second, it's possible to output the current version as deduced from .git using the Gradle [currentVersion command](https://axion-release-plugin.readthedocs.io/en/latest/configuration/tasks/#currentversion) as well:

```
./gradlew cV -q -Prelease.quiet
```

So put that all together:

1) In the docker workflow, we'd compute the current version as described above.
2) We'd pass that as a `--build-arg` to the Docker build context.
3) We'd pass that further into the actual gradle build to override the version.

Kind of roundabout, but I think it technically works?  My only question is if it's possible to gracefully fallback for local docker build such that the user doesn't have to remember to manually pass in the version argument.

.github/workflows/docker.yml
```diff
@@ -69,6 +70,7 @@ jobs:
           if [ "${{ github.event_name }}" = "push" ]; then
             TAGS="${DOCKER_IMAGE}:latest"
           fi
+          echo ::set-output name=gradleVersionTag::${VERSION}
```
bdferris-v2, line 73: The release version in Gradle is now expected to be the form X.Y.Z (with an optional -SNAPSHOT suffix).  I think it would be best if you got rid of version resolution via $GITHUB_REF and use the `./gradle currentVersion` mechanism I described in my previous comment.

Dockerfile
```diff
@@ -1,3 +1,20 @@
...
+
+ARG VERSION_TAG
+RUN ./gradlew shadowJar \
+    --no-daemon \
+    -Prelease.forceVersion=$VERSION_TAG
+
+
 FROM openjdk:11
-COPY main/build/libs/*.jar /
+COPY --from=build /build/main/build/libs/*.jar /
```
bdferris-v2, line 13: Per my previous comment, you are still pulling in multiple jars with this, only one of which is needed.  I think you can simplify this a lot by going with the following pattern:

```
COPY --from=build /build/main/build/libs/gtfs-validator-*-cli.jar /gtfs-validator-cli.jar
...
ENTRYPOINT [ "java", "-jar", "gtfs-validator-cli.jar" ]
```

Specifically, you won't need to include the version tag in the output jar with this approach.

Dockerfile
```diff
@@ -1,3 +1,20 @@
...
 WORKDIR /
+
+ARG VERSION_TAG
+ENV VERSION_TAG=$VERSION_TAG
+RUN echo "#!/bin/bash\nexec java -jar /gtfs-validator-${VERSION_TAG}-SNAPSHOT-cli.jar \"\$@\"" > /entrypoint.sh \
```
bdferris-v2, line 18: I'm not sure there will always be a -SNAPSHOT suffix for the jar, depending on the version number passed in.  Again, another reason to avoid encoding the version number in the jar here.

Dockerfile
```diff
@@ -1,3 +1,20 @@
+FROM gradle:7-jdk11-alpine AS build
+
+COPY --chown=gradle:gradle . /build
+WORKDIR /build
+
+ARG VERSION_TAG
```
bdferris-v2, line 6: What happens when you run a local Docker build without specifying a VERSION_TAG from the command line?

Dockerfile
themightychris: I dig it!

Dockerfile
themightychris: I did some digging on this, and it seems like the underlying lib offers a `forceSnapshot` option to force the `-SNAPSHOT` suffix, but none to disable it. It seems like as long as we're not feeding it a git repo + clean working tree + a tag that matches HEAD, it's always going to append this

Dockerfile
themightychris: Good question, looks like the lower level defaults to 0.1.0 upon receiving a blank `-Prelease.forceVersion=` option—same as it does if none were provided. The wrapper script would then be broken though. I've implemented your above suggestion to work around the problem and verified that local builds work with or without VERSION_TAG provided as a build arg now

.github/workflows/docker.yml
themightychris: I've seen approaches like this run into problems in the past: developers are free to push tags without any restrictions GitHub can implement, and it's nondeterministic which of multiple tags for the current HEAD such client-side tooling will select. When it comes to GitHub actions being triggered by the `release` events though, the maintainer creating the GitHub release selected a specific tag to use/create and GitHub is providing that to the workflow as a singular input via `GITHUB_REF`. The real world case here is developers pushing pre-release tags for testing, and then the GitHub release flow arbitrarily picking one of those to tag a release with rather than what the maintainer creating the official release tag selected.

Besides, this was already how the current workflow was sourcing the Docker image tag to apply and these changes are just also routing that same designation into the container build. So now they're both chained off the same input and it would be a broader scope of change to source that version designation differently. @bdferris-v2 I'd be happy to take that change on within this PR though if that's still preferred given the gotcha I described above.

The one substantial change I DID make here was dropping the `v` prefix when reading the Git tag. While it **is** a git convention to prefix version number tags with `v`, the convention with Docker image tags is to drop it. The `-Prelease.forceVersion=` option was also expecting it without the `v` prefix, so I did it upstream here to drive both the Docker tag and the Gradle version. So as of right now if this got merged, future docker container image publishings would drop the`v`—which seems like a good move to me this early in publishing one to bring in line with convention.

themightychris:
@bdferris-v2 I applied most of your suggested, except for one which I pushed back on a bit with additional context—happy to make that change to though if you think it makes this PR more mergeable after reviewing my comment.

I also made c96b228 after testing and finding the gradle build would crash if given any of the non-version tag values (e.g. `pr-123` that block of code might come up with for the the Docker container image tag in the cases of being triggered by pushes to master instead of by creating releases

.github/workflows/docker.yml
bdferris-v2: I wouldn't describe it as non-deterministic.  The `axios-release-plugin` has a [defined strategy](https://axion-release-plugin.readthedocs.io/en/latest/configuration/version/#first-tag-encountered) for resolving the current version from the Git commit history.  It's smart enough to ignore arbitrary tags like `i-love-tacos` but it is true that it will complain if you create a version tag like `v1.tacos`.

So I'm agreed that this requires us to use consistent vX.Y.Z version tags when tagging releases, but that doesn't seem too terrible.  More importantly, this wasn't introduced arbitrarily.  The `jpackage` app packaging phase specifically fails if passed a non-X.Y.Z version number, which is what led to this upstream change in the first place.  We are going to need some consistency in our version tags one way or another.  If there are other approaches that play well with the free-wheeling nature of Git versioning, I'm interested in learning more about them.

But if we stick with what we've got, then I'd argue that you should drop the code to resolve the version from `GITHUB_REF` and instead use the `currentVersion` approach I discussed earlier so that we get consistent version numbers whether we are building from a release tag, a PR, etc.  I'll admit it's a hack (thus my inclination to just include the .git metadata in the Docker build itself, but I digress).

Unfortunately, while playing around with `forceVersion` behavior, I discovered that if you pass in `1.2.3-SNAPSHOT` as the forced version, the plugin will happily append an additional -SNAPSHOT to make the version `1.2.3-SNAPSHOT-SNAPSHOT`.  So there's probably an argument for making the version resolution command:

```
$(./gradlew cV -q -Prelease.quiet | sed 's/-SNAPSHOT//')
```

themightychris:
@bdferris-v2 regarding https://github.com/MobilityData/gtfs-validator/pull/1120#discussion_r919435662

Someone testing release processes will often push semver-formatted prerelease tags like `vX.Y.Z-alpha.1`. Sounds like `axios-release-plugin` _tries_ to pick the highest version and commits to semver rules, so someone properly picking pre-release versions shouldn't cause any trouble...

> If there are other approaches that play well with the free-wheeling nature of Git versioning, I'm interested in learning more about them.

IME what works best is treating GitHub Releases as authoritative and the driver of published releases. That leaves behind a tag but the release drives publishing, not the tag. Axion's job seems to be to assign version numbers, but is the person creating releases in GitHub using that or handling picking version numbers manually in parallel anyway?

I'm going ahead and implement your suggestion though. In doing so I've had trouble reconciling all the use cases:

- The GitHub workflow has code to handle generating version strings for PRs and arbitrary branches, but the workflow is only hooked up to releases and pushes to master—so that code is unreachable and therefor my changes aren't testable
- You mention stripping `-SNAPSHOT`, but this workflow is set to run either on releases or any push to master, so every push to master between releases would generate a `-SNAPSHOT` release and by stripping it we'd be publishing a falsely-labeled next minor release to `ghcr.io` on each push to master (if I'm thinking this through correctly)
- The general convention with Docker images is to push trunk builds to `:latest`

So I took a pass at trying to reconcile this all and narrowed the code to specifically deal with the two cases we have GitHub actions triggers for right now:

- Releases publish container builds with a `:X.Y.Z` version tag generated by `gradlew cV` (see [test run](https://github.com/JarvusInnovations/gtfs-validator/actions/runs/2666651673) and [published image](https://github.com/JarvusInnovations/gtfs-validator/pkgs/container/gtfs-validator/28426348?tag=4.5.11))
- Pushes to `master` publish container builds with a `:latest` tag (see [test run](https://github.com/JarvusInnovations/gtfs-validator/actions/runs/2666655409) and [published image](https://github.com/JarvusInnovations/gtfs-validator/pkgs/container/gtfs-validator/28426429?tag=latest))
- All published containers have the `gradlew cV`-generated version passed into their build via `forceVersion` and applied to the container labels (which may include `X.Y.{Z+1}-SNAPSHOT` versions)

Take another look and see if this looks good to merge to you now

themightychris:
The `./gradlew currentVersion` functionality of digging up past tags in order to generate a `X.Y.{Z+1}-SNAPSHOT` version for non-tagged HEADs wasn't working correctly in my last push because the GitHub `actions/checkout@v2` action was only shallow cloning by default. I fixed that in d2551c214f058c31870834c55bddbab5808fb393

I did some other misc cleanup of the action in d2551c214f058c31870834c55bddbab5808fb393...f6a6a89645a36942bb44225a82d5eae2a639859f

.github/workflows/docker.yml
```diff
@@ -52,25 +52,9 @@ jobs:
...
-          fi
-          TAGS="${DOCKER_IMAGE}:${VERSION}"
-          if [ "${{ github.event_name }}" = "push" ]; then
-            TAGS="${DOCKER_IMAGE}:latest"
-          fi
-          echo ::set-output name=tags::${TAGS}
-          echo ::set-output name=created::$(date -u +'%Y-%m-%dT%H:%M:%SZ')
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0 # need full clone so ./grade
```
bdferris-v2, line 57: This comment looks like it got cut off?

Dockerfile
```diff
@@ -1,3 +1,18 @@
...
+    --no-daemon \
+    -Prelease.forceVersion="${VERSION_TAG%-SNAPSHOT}"
+
+
+FROM openjdk:11-slim
+COPY --from=build /build/main/build/libs/gtfs-validator-*-cli.jar /gtfs-validator-cli.jar
 WORKDIR /
+
+ARG VERSION_TAG
+ENV VERSION_TAG=$VERSION_TAG
```
bdferris-v2, line 17: Mostly out of curiosity, why do you need the VERSION_TAG here?

bdferris-v2:
LGTM.  I only had a comment nit + a question, but I don't think they are overall blocking.

.github/workflows/docker.yml
themightychris: heck... fixed

Dockerfile
bdferris-v2: Fair enough.  Thanks!

bdferris-v2 (approved):

PR #1127, opened 2022-04-28
feat: Implements the HTML output MVP

### LINKED ISSUE #1126: Implement HTML output for validation report (MVP)
## **Feature request**
Currently, the validation report is only available as a JSON file. Adding an HTML output would improve the user experience.

Browsing through the JSON validation report can be complicated and tedious, especially for user with low to mid technical knowledge. This is an obstacle to the use of the product.

**Proposed solution**
Inspired by Trillium work in PR #1119, we suggest implement a lean HTML + CSS + JavaScript output. The file will be self-contained so it can live on its own. We suggest using jQuery instead of a front-end framework like React or Snabbdom as we are no longer going for a local/hosted web UI app.

Steps:
- [X] Output prototype
- [X] Add the jQuery dependency
- [X] Add the components to generate the HTML output
- [X] Modify the validator main to add the generation of the HTML output during the export of the validation report.
- [X] Add tests.

We considered using React or Snabbdom at first. The initial idea was to create a local web UI, then to host it. This is not a feature we are considering doing anymore.

**Additional context**
Trillium did great work in PR #1119 to implement a front-end framework and a UI.
Issue #1112 explains the previous feature we were considering, the local web UI.
Epic #1124 explains the current feature we are considering, a packaged executable and an HTML output.

maximearmstrong:
**Summary:**

Fixes #1126: Implement HTML output for validation report (MVP)

This PR implements `HtmlOutputUtil.java`, `script.js` and `style.css` to generate a HTML version of the validation report.

**Expected behavior:** 

The HTML report output, defaulted to `report.html` is located in the output folder. It is a standalone HTML file (all dependencies are included in it). It can be viewed in a local browser by opening the file.

Example:

Capture d’écran, le 2022-04-28 à 20 02 38


main/src/main/java/org/mobilitydata/gtfsvalidator/cli/Main.java
```diff
@@ -225,6 +226,14 @@ public static void exportReport(final NoticeContainer noticeContainer, final Arg
       Files.write(
           Paths.get(args.getOutputBase(), args.getValidationReportName()),
           gson.toJson(noticeContainer.exportValidationNotices()).getBytes(StandardCharsets.UTF_8));
+      Files.write(
+          Paths.get(args.getOutputBase(), args.getHtmlReportName()),
+          HtmlOutputUtil.outputBuilder(
+                  noticeContainer.exportValidationNotices(),
```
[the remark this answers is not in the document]
barbeau: On a similar note, though, I think it would be safer to build the HTML report off of `ValidationReport` instead of the JSON output.

Right now the code is doing this process:
validate->`ValidationReport`->JSON->HTML

The problem with this is that if the JSON representation changes it will break HTML output. And the current `HtmlOutputUtilTest.outputBuilderTest()` would still pass because it's just using a hard-coded representation of the JSON output.

If you take a look at `ValidationReportDeserializer.serialize()`, at the end of the method it builds a `new ValidationReport(noticeReports)` and then converts that to a `JsonObject`.

You should be able to refactor this to effectively build the `ValidationReport` once and then use that as the input to the JSON and HTML output. So the new workflow would look like:
validate->`ValidationReport`->JSON
                                            \->HTML

This would also enable you to build better tests where you could build `Notice`s programmatically and export them to HTML and check the HTML output against that.

main/src/main/java/org/mobilitydata/gtfsvalidator/util/HtmlOutputUtil.java
```diff
@@ -0,0 +1,199 @@
...
+      Scanner scanner = new Scanner(stream).useDelimiter("\\A");
+      content = scanner.hasNext() ? scanner.next() : "";
+    } catch (IOException e) {
+      System.out.println("The resource was not found at \"" + resourcePath + "\".");
+    }
+    return content;
+  }
+
+  protected static String noticeColumnsBuilder(ArrayList<String> noticeFields) {
+    StringBuilder columns = new StringBuilder("                ");
```
barbeau, line 78: The current HTML representation is definitely an improvement over digging through raw JSON, but I think the workflow for the HTML report could be improved further. Currently it's still very difficult to track down what an error actually means.

Right now, we show the user the list of errors/warnings:

image

The natural question is "what does this error mean?"

So I would click on NOTICES.md and RULES.md and search for the code like `equal_shape_distance_diff_coordinates` to find more info.

However, if I search RULES.md, I don't get any matches because the rules are listed by notice class name and not the error code.

If I search NOTICES.md, I do get a hit in a list of notice codes mapped to notice class names:

image

...but that's not very helpful by itself. If I click on the notice class name I'm taken to an anchor lower on the same NOTICES.md page that describes what the fields within that notice mean:

https://github.com/MobilityData/gtfs-validator/blob/master/docs/NOTICES.md#EqualShapeDistanceDiffCoordinatesNotice

image

But that still doesn't describe what the actual error means. If I click on the notice class name linked in this anchor THEN I get taken over to RULES.md for the description of the error that I was originally looking for:

https://github.com/MobilityData/gtfs-validator/blob/master/RULES.md#EqualShapeDistanceDiffCoordinatesNotice

image

IMHO in the HTML report itself we should link directly to these error descriptions so the user has a single click to find out what the error means. We did this in the GTFS Realtime Validator HTML so the user is directly linked to, for example:
https://github.com/MobilityData/gtfs-realtime-validator/blob/master/RULES.md#e050---timestamp-is-in-the-future

...when clicking on that error in the report.

Maybe a simple approach is just to add the error code to RULES.md for now, maybe as section anchors in the markdown if that's possible so you can link directly to them?

That's what we did with RT validator - see https://raw.githubusercontent.com/MobilityData/gtfs-realtime-validator/master/RULES.md and anchors like `<a name="E003"/>`.

Likely beyond the scope of this issue, but I think we should also move the field descriptions in NOTICES.md into RULES.md, possibly as [collapsed text](https://gist.github.com/pierrejoubert73/902cc94d79424356a8d20be2b382e1ab) that you can expand if you want the field details.

main/src/main/java/org/mobilitydata/gtfsvalidator/util/HtmlOutputUtil.java
```diff
@@ -0,0 +1,199 @@
...
+    for (String field : noticeFields) {
+      columns.append("<th>").append(field).append("</th>");
+    }
+    columns.append("\n");
+    return columns.toString();
+  }
+
+  protected static String noticeRowsBuilder(
```
barbeau, line 86: Maybe it's just me, but I find it hard to read some of the data that's dumped into tables now, like this:

image

Creating human-readable sentences would obviously be an improvement - is that outside of the scope of this PR?

I was thinking similar to what we have on the GTFS Realtime Validator:

image

main/src/main/java/org/mobilitydata/gtfsvalidator/util/HtmlOutputUtil.java
```diff
@@ -0,0 +1,199 @@
...
+  }
+
+  public static String resourceLoader(String resourcePath) {
+    String content = "";
+    try (InputStream stream =
+        HtmlOutputUtil.class.getClassLoader().getResourceAsStream(resourcePath)) {
+      Scanner scanner = new Scanner(stream).useDelimiter("\\A");
+      content = scanner.hasNext() ? scanner.next() : "";
+    } catch (IOException e) {
+      System.out.println("The resource was not found at \"" + resourcePath + "\".");
```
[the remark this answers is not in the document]
barbeau: Also, I think the main result of this would be the HTML file would be malformed/broken? If so you should probably add that to the output message, otherwise the user might not tie these two things together.

barbeau (commented):
Thanks for putting this together @maximearmstrong! HTML file is definitely an improvement over just the JSON. Some thoughts below.

main/src/main/java/org/mobilitydata/gtfsvalidator/util/HtmlOutputUtil.java
```diff
@@ -0,0 +1,199 @@
...
+        outputHeader(
+            (noticesQty.get(ERROR) + noticesQty.get(WARNING) + noticesQty.get(INFO)),
+            noticesQty.get(ERROR),
+            noticesQty.get(WARNING),
+            noticesQty.get(INFO));
+    return outputFrame(
+        jqueryDependency, styleCss, scriptJs, outputHeader, notices.toString(), outputFooter());
+  }
+
+  public static String resourceLoader(String resourcePath) {
```
barbeau, line 65: Have you tested this using the shaded JAR? If not I'd suggest doing this, because resource loading can work differently that way vs. testing in IntelliJ (which is where I tested it).

bdferris-v2:
I wanted to link to the [comment I made on the feature request issue](https://github.com/MobilityData/gtfs-validator/issues/1126#issuecomment-1113679498) about the technical approach here, since I accidentally posted it there instead of here on the PR.  Curious if @maximearmstrong or @barbeau have opinions?

barbeau:
@bdferris-v2 ThymeLeaf (https://www.thymeleaf.org/) looks really nice - looks like there is an [IntelliJ plugin too](https://www.jetbrains.com/help/idea/thymeleaf.html). I agree that having a framework that allows us to manage the HTML as HTML content (as opposed to strings within a Java class) would greatly reduce the amount of effort to maintain and enhance the HTML output. The internationalization feature also looks like it would be very useful.

maximearmstrong:
@bdferris-v2 I also agree. I will implement it.

main/src/main/java/org/mobilitydata/gtfsvalidator/util/HtmlOutputUtil.java
isabelle-dr: @barbeau this is a great piece of feed-back. 
We were thinking of having this PR merged as a first step and then implementing (at least) the two additional parts you flagged here: 
1. A description of the notice (what is this problem?)
2. Where exactly is the issue and how to fix it (How to fix my dataset?)

Before implementing these parts, we would like to (1) gather user feedback & do some user testing on different options for the next version of the HTML output, (2) discuss the best strategy to integrate the info from RULES.md and NOTICES.md into the results (knowing that we will want to translate).

> Maybe a simple approach is just to add the error code to RULES.md for now, maybe as section anchors in the markdown if that's possible so you can link directly to them?
I really like this idea and this would be actionable for the next release, and I'm willing to PR this one. What about merging RULES.md and NOTICES.md in one document, using the collapsed text to make it more digestible?

barbeau: >What about merging RULES.md and NOTICES.md in one document, using the collapsed text to make it more digestible?

👍

main/src/main/java/org/mobilitydata/gtfsvalidator/util/HtmlOutputUtil.java
```diff
@@ -0,0 +1,199 @@
...
+        + outputNotices
+        + outputFooter
+        + "<script>\n"
+        + scriptJs
+        + "</script>\n"
+        + "</body>\n"
+        + "</html>\n";
+  }
+
+  protected static String noticeFrame(
```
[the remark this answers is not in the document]
maximearmstrong: Good catch!

main/src/test/java/org/mobilitydata/gtfsvalidator/util/HtmlOutputUtilTest.java
```diff
@@ -0,0 +1,266 @@
...
+                + "<p>This validation report was generated using the <a href=\"https://github.com/MobilityData/gtfs-validator\">GTFS Schedule validator</a>.</p>\n"
+                + "<script>\n"
+                + "scriptJsTest"
+                + "</script>\n"
+                + "</body>\n"
+                + "</html>\n");
+  }
+
+  @Test
+  public void resourceLoaderTest() {
```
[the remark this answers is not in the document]
maximearmstrong: Removed in [71a6271](https://github.com/MobilityData/gtfs-validator/pull/1127/commits/71a62714ec8421dc6b98174fe8de908ab6540725)

main/src/main/java/org/mobilitydata/gtfsvalidator/cli/Main.java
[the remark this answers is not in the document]
maximearmstrong: This have been fixed with the Thymeleaf approach.

main/src/main/java/org/mobilitydata/gtfsvalidator/util/HtmlOutputUtil.java
```diff
@@ -0,0 +1,199 @@
...
+        + "      <th>Code</th><th>Severity</th><th>Total</th>\n"
+        + "    </tr>\n"
+        + "  </thead>\n"
+        + "  <tbody>\n";
+  }
+
+  protected static String outputFooter() {
+    return "  </tbody>\n"
+        + "</table>\n"
+        + "<p>This validation report was generated using the <a href=\"https://github.com/MobilityData/gtfs-validator\">GTFS Schedule validator</a>.</p>\n";
```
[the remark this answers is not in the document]
maximearmstrong: I added
- the arguments passed to the validator, which includes the dataset input name
- the date and time
- the validator version

I did not add the GTFS dataset version for now since `feed_info.txt` is optional. @isabelle-dr, thoughts?

maximearmstrong:
@barbeau My implementation using Thymeleaf is ready for review. I have fixed most of the comments you made in your initial review! It was very thorough, thank you! We kept the "table" display for now.

cc @bdferris-v2 @isabelle-dr

main/src/main/java/org/mobilitydata/gtfsvalidator/util/HtmlOutputUtil.java
isabelle-dr: ✅ Done in PR #1132, and the anchor links have been added in the HTML output

isabelle-dr (changes requested):
Thank you for working on this @maximearmstrong!
I added a few comments for copywriting. 
It would be good to update the README.md with how the user can see the results after this PR is merged (and then we will update the instructions again when the PR with the Packaged EXE gets merged). Here is a proposal
1. In the **Introduction** section, add a forth step:
```
4. Provides an easy-to-use validation report in an HTML format that can be opened in the browser and shared with other parties. The report is also available in a JSON format that can be used for parsing and running additional analytics.
```
2. In each of the sections **Runs the app via command line** and  **Run the app using Docker**, can we add a sub section containing something like:
```
### Visualize the results
In the output directory, the following files will be created:
- `report.html`: the validation report in an html format. It can be opened in a browser
- `report.json`: the validation report in a JSON format
- `system_errors.json`: this file will be created every-time the validator is run. If no system errors were encountered, this fille will be empty.
```

main/src/main/resources/report.html
```diff
@@ -0,0 +1,229 @@
...
+    <tr>
+    <th>Notice Code</th><th>Notice Name</th><th>Severity</th><th>Total</th>
+    </tr>
+    </thead>
+    <tbody>
+    <span th:each="noticesBySeverityLevel: ${summary.noticesMap}">
+        <span th:each="noticeByCode: ${noticesBySeverityLevel.value}">
+            <tr class="notice">
+                <td th:text="${noticeByCode.key}" />
+                <td th:text="${noticeByCode.value.get(0).getName()}" />
```
bdferris-v2, line 114: Both here and throughout the HTML file, I think it is worth being consistent with the use of method calls vs OGNL property/getter syntax.  I think in OGNL style, you'd just have "noticesByCode.value[0].name" I believe?

Obviously, if you are actually calling a method (ala toLowerCase() or something similar), you'd leave it as a method call.

main/src/main/resources/report.html
```diff
@@ -0,0 +1,229 @@
...
+                                    <th:block th:each="field: ${noticeByCode.value.get(0).getFields()}" >
+                                        <th th:text="${field.getName()}" />
+                                    </th:block>
+                                </tr>
+                            </thead>
+                            <span th:each="notice: ${noticeByCode.value}">
+                                <tbody>
+                                <tr>
+                                <th:block th:each="field: ${notice.getFields()}">
+                                        <td th:text="${notice.getContext().getAsJsonObject().get(field.getName())}" />
```
bdferris-v2, line 135: I have to admit I'm not a hug fan of the approach here of exposing a reflected Field[] list on each Notice and then looking up values via the serialized JSON object.  Java field reflection may not always be equivalent to JSON serialization (e.g. maybe someone marks that a field shouldn't be serialized in JSON?).  It also repeatedly serializes the same object as json in this loop.

I'd propose an alternative approach where you introduce a presentation model object, which is a wrapper around a Notice, that provides the specific methods needed for display.  This is the spirit behind ReportSummary, which is a presentation-level model object for NoticeContainer.

Such a Notice wrapper could call JSON serialization once on construction and use that representation to provide method for display.

Thoughts?

bdferris-v2 (commented):
Left a couple of high-level comments.  I didn't review the tests yet.

main/src/main/resources/report.html
maximearmstrong: Thanks for this. I fixed it

main/src/main/resources/report.html
maximearmstrong: I initially used `notice.context.getAsJsonObject().entrySet()` and looped over the entries to get the keys and values, which was working well, but I wasn't sure about the approach.

> I'd propose an alternative approach where you introduce a presentation model object, which is a wrapper around a Notice, that provides the specific methods needed for display. This is the spirit behind ReportSummary, which is a presentation-level model object for NoticeContainer.

My concern is that the fields aren't the same for each Notice. Could we do this at the Notice abstract class level?

bdferris-v2: I think you'd have something like:

```java
public class NoticeView {
  private final Notice notice;
  private final JsonObject json; 
  public NoticeView(Notice notice) {
    this.notice = notice;
    this.json = notice.getContext().getAsJsonObject();
  }
  String getNameAsMarkdownId() { return notice.getClass().getSimpleName().toLowerCase(); }
  List<String> getFields() { ... }
  Object getValueForField(String field) { ... }  
}
```

Because Notice is an abstract base class that defines the getContext() method, then you can pass in any time of concrete Notice sub-class and it will still work the same way, regardless of the fields present.  Does that make sense?

maximearmstrong: Oh! Yes, it does. Thank you for the clarification.

maximearmstrong: @bdferris-v2 I added NoticeView.java and NoticeViewTest.java to fix this. Please let me know if it makes more sense.

main/src/main/resources/report.html
```diff
@@ -0,0 +1,234 @@
...
+<p>Use this report alongside the <a href="https://github.com/MobilityData/gtfs-validator/blob/master/RULES.md">RULES.md</a> file to get more details about the validation issues.</p>
+<table class="accordion">
+    <thead>
+    <tr>
+    <th>Notice Code</th><th>Notice Name</th><th>Severity</th><th>Total</th>
+    </tr>
+    </thead>
+    <tbody>
+    <span th:each="noticesBySeverityLevel: ${summary.noticesMap}">
+        <span th:each="noticeByCode: ${noticesBySeverityLevel.value}">
```
bdferris-v2, line 112: nit: noticeByCode => noticesByCode since there are multiple notices bundled here?  Also to match plurality of noticesBySeverityLevel?

main/src/test/java/org/mobilitydata/gtfsvalidator/report/model/NoticeViewTest.java
```diff
@@ -0,0 +1,42 @@
...
+import org.mobilitydata.gtfsvalidator.notice.MissingRequiredFieldNotice;
+import org.mobilitydata.gtfsvalidator.notice.Notice;
+
+public class NoticeViewTest {
+  private static final Notice notice = new MissingRequiredFieldNotice("test.txt", 1, "field");
+  private static final NoticeView noticeView = new NoticeView(notice);
+
+  @Test
+  public void getNameTest() {
+    assertEquals(noticeView.getName(), notice.getClass().getSimpleName());
```
bdferris-v2, line 16: I think both here and below, I'd avoid just re-using the internal impl code as the expected test value.  Aka, instead of testing against "notice.getClass().getSimpleName()", I'd just hard-code "MissingRequiredFieldNotice" instead.  The general idea being that if getClass().getSimpleName() were to change in some unexpected way (unlikely, I know), then the test would keep passing, but we might start spitting our weird values in our validation report.

main/src/test/java/org/mobilitydata/gtfsvalidator/report/model/NoticeViewTest.java
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  @Test
+  public void getFieldsTest() {
+    assertEquals(
+        noticeView.getFields(), new ArrayList<>(notice.getContext().getAsJsonObject().keySet()));
```
bdferris-v2, line 22: Ditto here.  Replace the ArrayList with hard-coded field values.

main/src/test/java/org/mobilitydata/gtfsvalidator/report/model/NoticeViewTest.java
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  @Test
+  public void getValueForFieldTest() {
+    for (String field : noticeView.getFields()) {
+      assertEquals(
+          noticeView.getValueForField(field), notice.getContext().getAsJsonObject().get(field));
+    }
```
bdferris-v2, line 30: Ditto here.

README.md
```diff
@@ -55,6 +62,12 @@ where:
 
 The validator can then be executed via bash commands. See the [preceeding instructions for command line usage](#run-the-app-via-command-line).
 
+### Visualize the results
+In the output directory, the following files will be created:
+- `report.html`: the validation report in HTML format. It can be opened in a browser.
+- `report.json`: the validation report in JSON format.
+- `system_errors.json`: this file will be created every-time the validator is run. If no system errors were encountered, this file will be empty.
+
```
bdferris-v2, line 70: I wonder if it's worth just linking up to the previous section where you document the output files, instead of repeating it verbatim here?  It seems like this will just get out of sync over time.

main/src/main/resources/report.html
maximearmstrong: I agree, it makes more sense.

main/src/test/java/org/mobilitydata/gtfsvalidator/report/model/NoticeViewTest.java
maximearmstrong: It makes sense, I will change it.

maximearmstrong:
I think we may be close to the finish line here. @bdferris-v2 I made the changes you suggested. Let me know if you think there is anything else that needs to be changed.

cc @barbeau

maximearmstrong:
Thank you @bdferris-v2, I fixed it. @barbeau, if you agree with the changes, I think we are ready to merge.

barbeau:
When I try to run this PR in IntelliJ with `-i tma-tn-us.zip -o output -c us` and [tma-tn-us.zip](https://github.com/MobilityData/gtfs-validator/files/8670555/tma-tn-us.zip) (after merging master into this PR branch, to see what it will look like after merge), I'm getting this error at runtime and there is no output from the validator:

```
INFO: Cache for trips.txt block_id: size = 7, lookup count = 142, hits = 95.07%, misses = 4.93%
May 11, 2022 9:55:42 AM org.mobilitydata.gtfsvalidator.table.GtfsTripTableLoader load
INFO: Cache for trips.txt shape_id: size = 6, lookup count = 142, hits = 95.77%, misses = 4.23%
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
Exception in thread "main" java.lang.IllegalArgumentException: Attribute 'Validator-Version' not found in MANIFEST.MF file(s) among 42 other attribute(s): ["Archiver-Version", "Automatic-Module-Name", "Bnd-LastModified", "Build-Jdk", "Build-Jdk-Spec", "Built-By", "Bundle-ContactAddress", "Bundle-Description", "Bundle-DocURL", "Bundle-License", "Bundle-ManifestVersion", "Bundle-Name", "Bundle-RequiredExecutionEnvironment", "Bundle-SymbolicName", "Bundle-Vendor", "Bundle-Version", "Created-By", "Export-Package", "Extension-Name", "Ignore-Package", "Implementation-Build", "Implementation-Title", "Implementation-URL", "Implementation-Vendor", "Implementation-Vendor-Id", "Implementation-Version", "Import-Package", "Include-Resource", "JCabi-Build", "JCabi-Date", "JCabi-Version", "Main-Class", "Manifest-Version", "Require-Capability", "Specification-Title", "Specification-Vendor", "Specification-Version", "Target-Label", "Tool", "X-Compile-Source-JDK", "X-Compile-Target-JDK", "url"]
	at com.jcabi.manifests.Manifests.read(Manifests.java:274)
	at org.mobilitydata.gtfsvalidator.report.HtmlReportGenerator.generateReport(HtmlReportGenerator.java:45)
	at org.mobilitydata.gtfsvalidator.cli.Main.exportReport(Main.java:231)
	at org.mobilitydata.gtfsvalidator.cli.Main.main(Main.java:118)

Execution failed for task ':main:Main.main()'.
> Process 'command 'C:/Program Files/Java/jdk-11.0.12/bin/java.exe'' finished with non-zero exit value 1
```

barbeau:
>(after merging master into this PR branch, to see what it will look like after merge)

An aside - I believe in the GitHub branch protection settings you can configure the repo to always require merge from master prior to merging PR to catch these types of issues, although it does get annoying as those merges are always required even when they have no impact on the PR (e.g., documentation changes).

maximearmstrong:
> > (after merging master into this PR branch, to see what it will look like after merge)
> 
> An aside - I believe in the GitHub branch protection settings you can configure the repo to always require merge from master prior to merging PR to catch these types of issues on CI, although it does get annoying as those merges are always required even when they have no impact on the PR (e.g., documentation changes).

Good catch, we did that for the Mobility Database Catalogs. I configured it for this repo as well. You are right, it does get annoying sometimes, but I think it's a reasonable tradeoff.

> When I try to run this PR in IntelliJ with `-i tma-tn-us.zip -o output -c us` and [tma-tn-us.zip](https://github.com/MobilityData/gtfs-validator/files/8670555/tma-tn-us.zip) (after merging master into this PR branch, to see what it will look like after merge), I'm getting this error at runtime and there is no output from the validator:

Oh, the problem is that we set the `Validator-Version` attribute in `main/build.gradle` to be in the manifest stored in the JAR file. We could use the `Implementation-Version` attribute instead, which solves the problem in IntelliJ, but the version displayed in the report would be `v3.0-cli` instead of `v3.0`. What do you think would be better?

maximearmstrong:
> > When I try to run this PR in IntelliJ with `-i tma-tn-us.zip -o output -c us` and [tma-tn-us.zip](https://github.com/MobilityData/gtfs-validator/files/8670555/tma-tn-us.zip) (after merging master into this PR branch, to see what it will look like after merge), I'm getting this error at runtime and there is no output from the validator:
> 
> Oh, the problem is that we set the `Validator-Version` attribute in `main/build.gradle` to be in the manifest stored in the JAR file. We could use the `Implementation-Version` attribute instead, which solves the problem in IntelliJ, but the version displayed in the report would be `v3.0-cli` instead of `v3.0`. What do you think would be better?

After discussion with @bdferris-v2, we decided to use the `Implementation-Version` attribute for now, as it might be refactored and fixed along with issue #1139. Running the validator in IntelliJ now works, however the HTML output obtained this way does not display the `Implementation-Version` that we would get using the JAR file. Since the HTML output obtained with the JAR file is correct, I think we are ready to merge.

README.md
```diff
@@ -33,6 +34,12 @@ You can run this validator using a GTFS dataset on your computer, or from a URL.
 
 More detailed instructions with all the parameters that exists are available on our ["Usage"](/docs/USAGE.md) page.
 
+### <a name="visualize-results"></a>Visualize the results
```
barbeau, line 37: Nit - You don't need the anchor here, you can just link to `#visualize-the-results` below

main/src/main/java/org/mobilitydata/gtfsvalidator/report/model/ReportSummary.java
```diff
@@ -0,0 +1,98 @@
...
+
+  public ReportSummary(NoticeContainer container) {
+    this.container = container;
+    this.severityCounts =
+        container.getValidationNotices().stream()
+            .collect(Collectors.groupingBy(Notice::getSeverityLevel, Collectors.counting()));
+    this.noticesMap =
+        container.getValidationNotices().stream()
+            .map(NoticeView::new)
+            .collect(Collectors.toList())
```
barbeau, line 42: I don't think you need the intermediate `.collect(Collectors.toList()).stream()` here? The tests pass without it.

barbeau (commented):
Thanks @maximearmstrong, looks good! Two comments in-line. Note there are conflicts now that need to be resolved too.

barbeau:
@maximearmstrong Also, do you want to open a separate issue for human-readable sentences in the HTML report? That was discussed in this PR but my understanding is it's getting tabled for future work.

bdferris-v2:
(Apologies in advance for the merge conflicts, @maximearmstrong!)

README.md
maximearmstrong: Interesting! I thought it wouldn't work because we have two "Visualize the results" titles, but the first one is properly referenced. We don't need it in this case, but do you know a way to reference the second one using the same method?

main/src/main/java/org/mobilitydata/gtfsvalidator/report/model/ReportSummary.java
maximearmstrong: Good catch, thank you!

maximearmstrong:
No worries @bdferris-v2! I merged the master branch. I had to modify the behaviour around the arguments table in `report.html` because we don't have access to the raw arguments anymore - a ValidatorRunnerConfig object is passed to the ValidatorRunner, not an Arguments one.

I removed the arguments `exportNoticeSchema` and `help` because we don't keep them in the the config object - I don't think it's a problem since they were not really relevant. The arguments `input` and `URL` are now both related to `config.gtfsSource`, so I had to check wether the URI scheme was a file or a URL one to display the correct one. The `countryCode` is not displayed when unknown because we don't have access to the raw data.

I think our initial idea of displaying the raw arguments was probably better, but it seems that we would need to break the ValidatorRunner design to make it possible. Thoughts? @barbeau @isabelle-dr Any concerns on your side?

maximearmstrong:
No worries @bdferris-v2! I merged the master branch. I had to change the behavior around the argument table in `report.html` because we no longer have access to raw arguments - a ValidatorRunnerConfig object is passed to the ValidatorRunner, not an Arguments.

I removed the `exportNoticeSchema` and `help` arguments because we don't keep them in the config object - I don't think this is a problem since they weren't really relevant. The `input` and `URL` arguments are now both represented by `config.gtfsSource`, so I had to check whether the URI scheme was related to a file or a URL to display the correct one. The `countryCode` is not displayed when it is unknown because we don't have access to the raw data.

I think our original idea of displaying the raw arguments was probably better, but it looks like we would have to break the ValidatorRunner design to make this possible. Any thoughts? @barbeau @isabelle-dr Any concerns on your end?

bdferris-v2:
I understand the motivation to display the raw arguments, but I think given the ValidationRunner refactor, there is value in just displaying the contents of the config object, as you have done, as that ultimately controls the execution.  As such, I wouldn't be opposed if you got rid of the idea of command-line arguments in the html report entirely and just described them as the validation configuration.  That'll probably scale better to other execution environments for the validator in the future.

If you really wanted to record the raw cli arguments for debugging purposes, I would add an Optional<List<String>> rawArguments() field to the config where you can optionally pass those in if it makes sense for the execution environment.

maximearmstrong:
@bdferris-v2 Thank you, that's helpful. I thought about changing the argument table to a configuration table and changing the "Argument Description" and "Argument Short Name" columns to "Related Argument" and "Related Argument Short Name". @isabelle-dr Do you think it would be relevant for the end user?

isabelle-dr:
For the arguments, my only concern is consistency between (1) the table in [USAGE.md](https://github.com/MobilityData/gtfs-validator/blob/master/docs/USAGE.md#via-cli-app) and (2) the way the params are written in the "Advanced" section in PR #1146. 
If a user ticks a box that says "number of threads for execution" on the GUI and then sees "threads" and "-t" in the HTML report, this is _slightly_ confusing (although pretty straight forward to guess).

I think from a user perspective, the best would be to have the same short descriptions than in the GUI, and remove the argument short name. There is a link to the USAGE.md file if users want to map this with the exact params used in the cli. The short descriptions could be:
- Input directory
- Output directory
- Number of threads used to run the validator
- Country code (for phone validation)
- Validation report name
- System errors report name
- HTML report name

maximearmstrong:
@isabelle-dr I agree. I made the changes. I used "GTFS input (ZIP file, directory or URL)" instead of "Input directory". This would be the final version:
Capture d’écran, le 2022-05-16 à 18 26 15

isabelle-dr:
This looks great @maximearmstrong. Thanks for adding the sentence in the value for country code.
One last tiny change 😃: can you replace _Argument description_ by _Parameter description_? We use the term _Parameter_ above and in the documentation.
After this, ready to merge from my perspective, I'm approving now. great work!

isabelle-dr (approved):

Issue #1129, opened 2022-04-28
End to End workflow failure of PR


bdferris-v2:
## **Bug report** 

On PR #1125, I had a couple of acceptance tests fail.  While some of those might be due to issue #1094, I think there might be other bugs as well.  For example, consider the "End to end" test:

https://github.com/MobilityData/gtfs-validator/actions/runs/2234494855

I see the following error in the `run-on-data` step:

```
Run java -jar main/build/libs/*.jar --url https://openmobilitydata.org/p/transport-for-nsw/237/latest/download --output_base output --country_code au --storage_directory gs.zip
Error: Exception in thread "main" com.beust.jcommander.ParameterException: Was passed main parameter 'main/build/libs/main-v-pr-1125-sha-0b91b063-SNAPSHOT_cli.jar' but no main parameter was defined
```

I think this is an issue with the following line in [end_to_end.yml](https://github.com/MobilityData/gtfs-validator/blob/master/.github/workflows/end_to_end.yml#L74):

```
java -jar application/cli-app/build/libs/*.jar
```

I think the issue is that the :main sub-project can actually generate two jars.  One regular jar (`main-$version.jar`) and one shadowJar (`gtfs-validator-$version_cli.jar`) containing all the dependencies for the CLI.  The shadow jar is what you want here and is indeed what's built by the workflow, but I'm wondering if something is causing the build to be left in a weird state (possibly due to a previous workflow failure?) where the other jar is still around, such that the `*.jar` wild-card in the command actually expands to two jars, causing the command to fail?  That's my interpretation of the error message, at least.

Could probably be fixed by adding a `gtfs-validator-` prefix to the `*.jar` wildcard.  Or, alternatively, doing a full clean before running the workflow step?

I'll admit I'm not familiar enough with the workflow setup to know the right course of action here.

barbeau:
>Could probably be fixed by adding a gtfs-validator- prefix to the *.jar wildcard.

As discussed in today's meeting I think this would be a good first fix and we can see what happens from here. 👍

bdferris-v2:
My end-to-end tests are now successfully running for PR #1125 with the new workflow changes.  While there may be other issues affecting workflows and acceptance test, I believe this specific instance appears to be fixed.

PR #1131, opened 2022-05-02
fix: Modify end_to_end.yml to be more specific about which gtfs-validator jar is used

bdferris-v2:
Per discussion in issue #1129, I think I ran into an edge-case where there are multiple jars in the main/build/libs directory (maybe from a previous build?) that caused the script to fail.  Attempting to fix by being more specific about which of the :main project jars to use (hint: we want the shadow jar).

### DIFF
--- a/.github/workflows/end_to_end.yml
+++ b/.github/workflows/end_to_end.yml
@@ -71,30 +71,30 @@ jobs:
         with:
           arguments: shadowJar
       #- name: Validate dataset from -- ACRONYM #<-- uncomment this line, replace ACRONYM by the name of the agency/publisher acronym
-      #  run: java -jar application/cli-app/build/libs/*.jar --url DATASET_PUBLIC_URL --input [[[ACRONYM]]].zip --extract pathToExtractedZipContent --output validationResultDirectory #<-- uncomment this line,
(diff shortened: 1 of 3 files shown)


barbeau (commented):
These changes LGTM, although the same pattern exists in end_to_end_big.yml and end_to_end_100.yml as well. I'd suggest we change all these occurrences to be consistent.

bdferris-v2:
@barbeau good call.  Done.

barbeau (approved):
Thanks @bdferris-v2, LGTM!

The acceptance tests are still blocked by https://github.com/MobilityData/gtfs-validator/issues/1094, so if we want them to run before merge this PR would need to be merged after PR https://github.com/MobilityData/gtfs-validator/pull/1102.

bdferris-v2:
All the end_to_end*.yml tests look to be passing, which I think is the most important thing to check here.  So I propose to submit without waiting for #1102.  I will do so this evening unless I hear objections.

bdferris-v2:
But maybe you were instead pointing out that I won't be able to submit this PR until #1102 is in, regardless?

barbeau:
No objections here. I think the repo permissions allow merge even with failing/canceled tests.

maximearmstrong:
Thank you for this contribution @bdferris-v2 and @barbeau ! We didn’t get a chance to add this to the contribution guidelines yet, but can you make sure there is a review from one MobilityData staff before merging please? :slightly_smiling_face:

cc @isabelle-dr

Issue #1139, opened 2022-05-10
Refactor handling of project version numbers and tags


bdferris-v2:
Per discussion in https://github.com/MobilityData/gtfs-validator/pull/1125#discussion_r869435195, we'd like to consider a few potential changes to how `gtfs-validator` handles version numbers.  Two potential issues:

*  The use of a leading `v` in the version tag (ala v3.1.0) causes problems for some tooling (Java Modules, Mac OS app packaging) that expects numeric major.minor.patch version numbers.
* Our use of a `versionTag` environment variable for passing in version information is maybe not idiomatic Gradle?  My limited research has suggested that overriding `project.version` via a command-line flag ala `-Pversion=1.2.3` is a more consistent way of setting the project version for the entire build.

maximearmstrong:
As discussed [here](https://github.com/MobilityData/gtfs-validator/pull/1127#issuecomment-1125486282) in #1127, the work done for this issue may affect the version of the validator we display in the HTML output.

isabelle-dr:
> The use of a leading v in the version tag

Do we want to do this one right now? (i. e. rename previous versions & only use major.minor.patch from now on)

bdferris-v2:
@isabelle-dr the axion-release plugin in PR #1147 seems to be able to handle the leading 'v' on the version tag, so I don't think any changes are needed there.

PR #1147, opened 2022-05-14
feat: Refactor handling of project version numbers and tags

bdferris-v2:
Per issue #1139, refactor how versions are handled in project.

Currently, project version info is passed in via a `versionTag` env variable, which is only specified in the Github workflow environment.  Consequently, `project.version` ends up being `null` when run locally.  This causes weirdness for tools that expect a valid version number.

This change switches the project to use the axion-release plugin, which determines the project version by looking at the git repo's tags.

Along with that switch, I've moved the project group and version specification into the root build.gradle allprojects {} such that these values do not need to be specified in each sub-poject.

I also did some related clean-up for our jitpack configuration (and Maven repository modules, should we every decide to publish them).  I believe our current jitpack config doesn't actually work correctly because we only publish artifacts from the main package, but not core.

You can see this in practice when looking at:
https://jitpack.io/com/github/MobilityData/gtfs-validator/3.0.1/gtfs-validator-3.0.1.pom

Notice how it has a dependency on the `core` module but we don't actually publish it anywhere.  If you actually attempt to use our project as a jitpack dependency, it will fail.

The change was to also publish the core module and to specify more natural artifact names.


bdferris-v2:
(This isn't ready for review yet.  I just wanted to see what would happen with all the workflow tasks.)

.github/workflows/docker.yml
```diff
@@ -45,8 +45,6 @@ jobs:
           restore-keys: ${{ runner.os }}-gradle
       - name: Run Java tests
         uses: eskatos/gradle-command-action@v1
-        env:
-          versionTag: ${{ steps.prep.outputs.versionTag }}
```
maximearmstrong, line 49: I agree with removing this. I'm not sure if we need the `Prepare version name` job at all. I see that it used below in two jobs name, but I wonder if we could just get rid of that as well. Thoughts? cc @barbeau

.github/workflows/test_pack_doc.yml
```diff
@@ -102,15 +102,13 @@ jobs:
           restore-keys: ${{ runner.os }}-gradle
       - name: Package cli app jar with Gradle
         uses: eskatos/gradle-command-action@v1
-        env:
-          versionTag: ${{ needs.prepare-version-name.outputs.versionTag }}
```
maximearmstrong, line 106: Same as mentioned in `docker.yml`. We are still using the versionTag a lot, but mostly for job names. I'm not sure if this is relevant anymore if the validator is responsible for its own versioning now? cc @barbeau

maximearmstrong (commented):
Thanks @bdferris-v2, this looks good! Some thoughts in-line.

.github/workflows/test_pack_doc.yml
barbeau: Yes, I think we should remove all version tag references from the test_pack_doc GitHub Action? I believe it should be printed to the log as part of the commands shown in the JAR file name?

.github/workflows/docker.yml
barbeau: I think so - from looking at the script it seems like the commands that set up the Docker image version tag are separate from the script setting the validator version tag.

bdferris-v2: I went ahead and dropped all the versioning code.

.github/workflows/test_pack_doc.yml
bdferris-v2: I removed the versionTag references here as well.

.github/workflows/docker.yml
```diff
@@ -73,34 +56,7 @@ jobs:
...
-            VERSION=$(echo ${GITHUB_REF#refs/heads/} | sed -r 's#/+#-#g')
-          elif [[ $GITHUB_REF == refs/pull/* ]]; then
-            VERSION=pr-${{ github.event.number }}
-          fi
-          TAGS="${DOCKER_IMAGE}:${VERSION}"
-          if [ "${{ github.event_name }}" = "push" ]; then
-            TAGS="${DOCKER_IMAGE}:latest"
-          fi
-          echo ::set-output name=version::${VERSION}
-          echo ::set-output name=tags::${TAGS}
```
barbeau, line 90: I think we need this line that sets up the output for `tags`, as that seems to be used later to set the Docker image tag in line 89:

```
tags: ${{ steps.prep.outputs.tags }}
```

bdferris-v2: Ah, I totally missed that.  I restored the TAGS code.

barbeau (approved):
LGTM

barbeau:
@bdferris-v2 The branch rules require that the main branch be merged into the PR, if you could do that.

bdferris-v2:
Hopefully should be merged at this point.

isabelle-dr:
I forgot to add my review before merging: I approve this PR, thanks!

Issue #885, opened 2021-04-29
Best Practice: feed_info should include start_date, end_date, version, email, url (WARNING)


e-lo:
`feed_info.txt` should Includes valid fields for:

- feed_start_date, feed_end_date 
- Feed_version.
- Feed_email
- feed_url
-

lionel-nj:
Agreed, as suggested per best practice:

> `feed_info.txt` should be included, with all fields below.
> 
> | Field Name | Recommendation |
> | --- | --- |
> | feed_start_date & feed_end_date | Should be included |
> | feed_version | Should be included |
> | feed_contact_email & feed_contact_url | Provide at least one |

https://github.com/MobilityData/GTFS_Schedule_Best-Practices/blob/master/en/feed_info.md

KClough:
@isabelle-dr can we mark this issue as closed? 

This has been implemented in #1157 , I probably should have put "resolves" instead of "related" in the PR description.

isabelle-dr:
I think what closes automatically an issue is writing "closes #issue number" in a PR :)

Closing this issue, thanks for adding this to this project!

PR #1123, opened 2022-04-25
feat: recommended file annotation #877

KClough:
**Summary:**

Resolves #877 

Adds a new annotation for recommended files and applies this annotation to `feed_info.txt` and `shapes.txt`

**Expected behavior:** 

A new validator warning is generated when either `feed_info.txt` or `shapes.txt` is missing

Please make sure these boxes are checked before submitting your pull request - thanks!

- [x] Run the unit tests with `gradle test` to make sure you didn't break anything
- [x] Format the title like "feat: [new feature short description]". Title must follow the Conventional Commit Specification(https://www.conventionalcommits.org/en/v1.0.0/).
- [x] Linked all relevant issues
- [x] Include screenshot(s) showing how this pull request works and fixes the issue(s)

In liue of a screenshot, here is the sample output from the validator for a data set with missing `feed_info.txt` and `shapes.txt`:

```json
{
  "notices": [
    {
      "code": "missing_recommended_file",
      "severity": "WARNING",
      "totalNotices": 2,
      "sampleNotices": [
        {
          "filename": "shapes.txt"
        },
        {
          "filename": "feed_info.txt"
        }
      ]
    }
]}
```

PR #1149, opened 2022-05-16
feat: recommended annotation

### LINKED ISSUE #877: Best Practice:  feed_info.txt should be provided (WARNING)
Lack of  `feed_info.txt` should all trigger a WARNING because the file is required in the GTFS Best Practices. 

Should probably mirror or extend [MissingRequiredFileNotice](https://github.com/MobilityData/gtfs-validator/blob/master/RULES.md#MissingRequiredFileNotice) to allow for a parameter to indicate `SeverityLevel.WARNING` 

https://github.com/MobilityData/gtfs-validator/blob/52434454b747bb4085038f27563e98fee6018ca5/core/src/main/java/org/mobilitydata/gtfsvalidator/notice/MissingRequiredFileNotice.java#L27-L29

KClough:
**Summary:**

Related to issue #885

Adds a new annotation for recommended fields

**Expected behavior:** 

Provides additional functionality around the `recommended` annotation. It can now be applied to fields. 

~- [ ] Include screenshot(s) showing how this pull request works and fixes the issue(s)~

@isabelle-dr per [your request](https://github.com/MobilityData/gtfs-validator/pull/1123#pullrequestreview-958240912), this pull request contains just the annotation. I don't think a screenshot makes sense in this scenario.

### DIFF
--- a/core/src/main/java/org/mobilitydata/gtfsvalidator/annotation/Recommended.java
+++ b/core/src/main/java/org/mobilitydata/gtfsvalidator/annotation/Recommended.java
@@ -1,0 +1,38 @@
+/*
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.mobilitydata.gtfsvalidator.annotation;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Adds a validation that the field or a file is recommended.
+ *
+ * <p>Example.
+ *
+ * <pre>
+ *   {@literal @}Recommended
+ *   {@literal @}GtfsTable(value = "feed_info.txt", singleRow = true) {
+ *     ...
+ *   }
+ * </pre>
+ */
+@Target({ElementType.METHOD, ElementType.TYPE})
+@Retention(RetentionPolicy.SOURCE)
+public @interface Recommended {}
--- a/core/src/main/java/org/mobilitydata/gtfsvalidator/parsing/RowParser.java
+++ b/core/src/main/java/org/mobilitydata/gtfsvalidator/parsing/RowParser.java
@@ -22,6 +22,7 @@ import java.util.Currency;
 import java.util.Locale;
 import java.util.function.Function;
 import javax.annotation.Nullable;
+import org.mobilitydata.gtfsvalidator.annotation.FieldLevelEnum;
 import org.mobilitydata.gtfsvalidator.notice.EmptyRowNotice;
 import org.mobilitydata.gtfsvalidator.notice.InvalidColorNotice;
 import org.mobilitydata.gtfsvalidator.notice.InvalidCurrencyNotice;
@@ -52,8 +53,6 @@ import org.mobilitydata.gtfsvalidator.validator.GtfsFieldValidator;
  */
 public class RowParser {
 
-  public static final boolean REQUIRED = true;
-  public static final boolean OPTIONAL = false;
   private final String fileName;
   private final CsvHeader header;
   private final GtfsFieldValidator fieldValidator;
@@ -107,9 +106,9 @@ public class RowParser {
   }
 
   @Nullable
-  public String asString(int columnIndex, boolean required) {
+  public String asString(int columnIndex, FieldLevelEnum level) {
     String s = row.asString(columnIndex);
-    if (required && s == null) {
+    if (level == FieldLevelEnum.REQUIRED && s == null) {
       noticeContainer.addValidationNotice(
           new MissingRequiredFieldNotice(
               fileName, row.getRowNumber(), header.getColumnName(columnIndex)));
@@ -126,23 +125,23 @@ public class RowParser {
   }
 
   @Nullable
-  public String asText(int columnIndex, boolean required) {
-    return asString(columnIndex, required);
+  public String asText(int columnIndex, FieldLevelEnum level) {
+    return asString(columnIndex, level);
   }
 
   @Nullable
-  public String asId(int columnIndex, boolean required) {
-    return asValidatedString(columnIndex, required, fieldValidator::validateId);
+  public String asId(int columnIndex, FieldLevelEnum level) {
+    return asValidatedString(columnIndex, level, fieldValidator::validateId);
   }
 
   @Nullable
-  public String asUrl(int columnIndex, boolean required) {
-    return asValidatedString(columnIndex, required, fieldValidator::validateUrl);
+  public String asUrl(int columnIndex, FieldLevelEnum level) {
+    return asValidatedString(columnIndex, level, fieldValidator::validateUrl);
   }
 
   @Nullable
-  public String asEmail(int columnIndex, boolean required) {
-    return asValidatedString(columnIndex, required, fieldValidator::validateEmail);
+  public String asEmail(int columnIndex, FieldLevelEnum level) {
+    return asValidatedString(columnIndex, level, fieldValidator::validateEmail);
   }
 
   /**
@@ -151,45 +150,44 @@ public class RowParser {
    * unknown, only phone number starting by "+" are validated.
    *
    * @param columnIndex the column index
-   * @param required true is the value is required, false otherwise
+   * @param level whether the value is required, recommended or optional according to GTFS
    * @return the string value of the phone number to be validated if a valid number according to the
    *     {@code CountryCode}, returns {@code null} otherwise. Note that if {@code CountryCode} is
    *     unknown, only phone number starting by "+" are validated.
    */
   @Nullable
-  public String asPhoneNumber(int columnIndex, boolean required) {
-    return asValidatedString(columnIndex, required, fieldValidator::validatePhoneNumber);
+  public String asPhoneNumber(int columnIndex, FieldLevelEnum level) {
+    return asValidatedString(columnIndex, level, fieldValidator::validatePhoneNumber);
   }
 
   @Nullable
-  public Locale asLanguageCode(int columnIndex, boolean required) {
-    return parseAsType(
-        columnIndex, required, Locale::forLanguageTag, InvalidLanguageCodeNotice::new);
+  public Locale asLanguageCode(int columnIndex, FieldLevelEnum level) {
+    return parseAsType(columnIndex, level, Locale::forLanguageTag, InvalidLanguageCodeNotice::new);
   }
 
   @Nullable
-  public ZoneId asTimezone(int columnIndex, boolean required) {
-    return parseAsType(columnIndex, required, ZoneId::of, InvalidTimezoneNotice::new);
+  public ZoneId asTimezone(int columnIndex, FieldLevelEnum level) {
+    return parseAsType(columnIndex, level, ZoneId::of, InvalidTimezoneNotice::new);
   }
 
   @Nullable
-  public Currency asCurrencyCode(int columnIndex, boolean required) {
-    return parseAsType(columnIndex, required, Currency::getInstance, InvalidCurrencyNotice::new);
+  public Currency asCurrencyCode(int columnIndex, FieldLevelEnum level) {
+    return parseAsType(columnIndex, level, Currency::getInstance, InvalidCurrencyNotice::new);
   }
 
   @Nullable
-  public Double asFloat(int columnIndex, boolean required) {
-    return parseAsType(columnIndex, required, Double::parseDouble, InvalidFloatNotice::new);
+  public Double asFloat(int columnIndex, FieldLevelEnum level) {
+    return parseAsType(columnIndex, level, Double::parseDouble, InvalidFloatNotice::new);
   }
 
   @Nullable
-  public Double asFloat(int columnIndex, boolean required, NumberBounds bounds) {
-    return checkBounds(asFloat(columnIndex, required), 0.0, columnIndex, "float", bounds);
+  public Double asFloat(int columnIndex, FieldLevelEnum level, NumberBounds bounds) {
+    return checkBounds(asFloat(columnIndex, level), 0.0, columnIndex, "float", bounds);
   }
 
   @Nullable
-  public Double asLatitude(int columnIndex, boolean required) {
-    Double value = asFloat(columnIndex, required);
+  public Double asLatitude(int columnIndex, FieldLevelEnum level) {
+    Double value = asFloat(columnIndex, level);
     if (value != null && !(-90 <= value && value <= 90)) {
       noticeContainer.addValidationNotice(
           new NumberOutOfRangeNotice(
@@ -204,8 +202,8 @@ public class RowParser {
   }
 
   @Nullable
-  public Double asLongitude(int columnIndex, boolean required) {
-    Double value = asFloat(columnIndex, required);
+  public Double asLongitude(int columnIndex, FieldLevelEnum level) {
+    Double value = asFloat(columnIndex, level);
     if (value != null && !(-180 <= value && value <= 180)) {
       noticeContainer.addValidationNotice(
           new NumberOutOfRangeNotice(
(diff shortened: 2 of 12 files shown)


core/src/main/java/org/mobilitydata/gtfsvalidator/parsing/RowParser.java
```diff
@@ -107,12 +110,16 @@ public boolean checkRowLength() {
...
   @Nullable
-  public String asString(int columnIndex, boolean required) {
+  public String asString(int columnIndex, FieldLevelEnum level) {
     String s = row.asString(columnIndex);
-    if (required && s == null) {
+    if (level == FieldLevelEnum.REQUIRED && s == null) {
       noticeContainer.addValidationNotice(
           new MissingRequiredFieldNotice(
               fileName, row.getRowNumber(), header.getColumnName(columnIndex)));
+    } else if (level == FieldLevelEnum.RECOMMENDED && s == null) {
```
maximearmstrong, line 119: If possible, this could also go in another PR specifically for the notices.

core/src/main/java/org/mobilitydata/gtfsvalidator/parsing/RowParser.java
```diff
@@ -341,7 +347,7 @@ public enum NumberBounds {
    * parsing failed.
    *
    * @param columnIndex index of the column to parse
-   * @param required whether the value is required according to GTFS
+   * @param level whether the value is required or recommended according to GTFS
```
maximearmstrong, line 350:
```suggestion
   * @param level whether the value is required, recommended or optional according to GTFS
```

core/src/main/java/org/mobilitydata/gtfsvalidator/parsing/RowParser.java
```diff
@@ -376,14 +382,14 @@ private <T> T parseAsType(
    * error is emitted, the value is considered invalid.
    *
    * @param columnIndex index of the column to parse
-   * @param required whether the value is required according to GTFS
+   * @param level whether the value is required or recommended according to GTFS
```
maximearmstrong, line 385:
```suggestion
   * @param level whether the value is required, recommended or optional according to GTFS
```

core/src/test/java/org/mobilitydata/gtfsvalidator/notice/NoticeContainerTest.java
```diff
@@ -51,6 +51,18 @@ public void exportNotices() {
                 + "\"message\":\"Index 0 out of bounds\"}]}]}");
   }
 
+  @Test
+  public void exportRecommendedFileNotices() {
```
maximearmstrong, line 55: This could also be in the other PR for the notices.

core/src/test/java/org/mobilitydata/gtfsvalidator/testgtfs/GtfsStopTableContainer.java
```diff
@@ -58,4 +58,9 @@ public Optional<GtfsStop> byPrimaryKey(String id, String subId) {
   public boolean isRequired() {
     return true;
   }
+
+  @Override
+  public boolean isRecommended() {
```
maximearmstrong, line 63: I'm not sure if I'm missing something here, but is there any reason why we would need this for a test related to `GtfsStop`? Also, there seems to be `no usages found` according to IntelliJ.

main/src/main/java/org/mobilitydata/gtfsvalidator/table/GtfsFeedInfoSchema.java
```diff
@@ -38,16 +40,21 @@ public interface GtfsFeedInfoSchema extends GtfsEntity {
...
+  @Recommended
   String feedVersion();
 
+  @Recommended
   @FieldType(FieldTypeEnum.EMAIL)
   String feedContactEmail();
 
+  @Recommended
   @FieldType(FieldTypeEnum.URL)
   String feedContactUrl();
```
maximearmstrong, line 59: According to the [spec](https://gtfs.org/schedule/reference/#feed_infotxt), this fields are optional, not recommended. I understand though that the file is recommended. @isabelle-dr What is your input on this? Should fields included in a recommended file be recommended although the spec says they are optional? I think they should stay optional, but I would be happy to hear other thoughts on this.

processor/src/main/java/org/mobilitydata/gtfsvalidator/processor/TableLoaderGenerator.java
```diff
@@ -436,6 +447,11 @@ private MethodSpec generateLoadMissingFileMethod() {
                 classNames.tableContainerTypeName(),
                 classNames.tableContainerTypeName(),
                 TableStatus.class)
+            .beginControlFlow("if (isRecommended())")
+            .addStatement(
+                "noticeContainer.addValidationNotice(new $T(gtfsFilename()))",
+                MissingRecommendedFileNotice.class)
+            .endControlFlow()
```
maximearmstrong, line 454: I think this could be added in the other PR, but everything else in this class could stay.

maximearmstrong (changes requested):
Thank you for your contribution @KClough ! This is great work. Overall, if possible, it may be better to split this PR in two so that the notices are added in a second one. Otherwise, only a few comments before approval.

core/src/test/java/org/mobilitydata/gtfsvalidator/testgtfs/GtfsStopTableContainer.java
KClough: I believe I hit some errors where the existing tests were failing without this logic, I will double check to confirm.

KClough:
Sorry, it looks like I made a mistake when separating the PRs, I meant for this one to just have the annotation logic. I'll take another pass at it.

maximearmstrong:
No worries @KClough! I'll take a look at it right after.

main/src/main/java/org/mobilitydata/gtfsvalidator/table/GtfsFeedInfoSchema.java
maximearmstrong: This is related to #885.

isabelle-dr:
Hi @KClough, thanks for splitting the PRs, this makes it simpler :) 

Here is what I understand from the opened PRs:
- PR #1149 (this one) to add a @Recommended annotation in the validator
- PR #1151 which closes issue #877 to add the `missing_required_file` notice, which will apply (for now) only to the`feed_info.txt` file, but that could be used for other files when the Best Practices evolves.
- PR #1152 which closes issue #885 to add the `missing_required_field` notice, which will apply (for now) only to the `feed_info.txt` fields mentioned in the issue. Again, we can then use this notice for other fields recommended by the best practices or spec.

If this is accurate, here are some changes to do, in order to make it easier when we look back at these PRs in the future:
1. Can you please update the description of this PR to remove the link to issue #887, and update the "Expected behaviour section"?
2. Remove the @recommended in the `GTFSFeedInfoSchema` file in this PR, to have it only in PR #1152

main/src/main/java/org/mobilitydata/gtfsvalidator/table/GtfsFeedInfoSchema.java
maximearmstrong: Perfect, thank you!

core/src/test/java/org/mobilitydata/gtfsvalidator/testgtfs/GtfsStopTableContainer.java
KClough: Confirmed, I get this error when running `gradlew test` without this change. 

```
> Task :core:compileTestJava FAILED
/Users/kclough/projects/jarvus/gtfs-validator/core/src/test/java/org/mobilitydata/gtfsvalidator/testgtfs/GtfsStopTableContainer.java:26: error: GtfsStopTableContainer is not abstract and does not override abstract method isRecommended() in GtfsTableContainer
public class GtfsStopTableContainer extends GtfsTableContainer<GtfsStop> {
       ^
1 error

FAILURE: Build failed with an exception.
```

core/src/main/java/org/mobilitydata/gtfsvalidator/parsing/RowParser.java
maximearmstrong: I believe we will need to remove the `else if` statement here as the `MissingRecommendedFieldNotice` is not yet implemented in this PR. It should be in #1157.

core/src/main/java/org/mobilitydata/gtfsvalidator/annotation/Recommended.java
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * Adds a validation that the field or a file is recommended.
+ *
+ * <p>Example.
+ *
+ * <pre>
+ *   @Recommended
+ *   {@literal @}GtfsTable(value = "feed_info.txt", singleRow = true) {
+ *     ...
+ *   }
```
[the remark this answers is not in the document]
maximearmstrong: We would need this before we merge the PR.

core/src/main/java/org/mobilitydata/gtfsvalidator/parsing/RowParser.java
maximearmstrong: Actually, most of the checks failing in this PR are because of this - we are importing a class that does not exist in the import section.

core/src/test/java/org/mobilitydata/gtfsvalidator/testgtfs/GtfsStopTableContainer.java
maximearmstrong: Ah, right, it has to be implemented because the class inherits the abstract method. All good.

maximearmstrong:
Than you for the changes @KClough! 

I believe we are missing a few things before we can merge:
- there are two unresolved items above - I believe one of these will fix most of the checks failing in this PR. 
- the code format workflow fails. Did you format the code using `gradle googleJavaFormat` or `gradle goJF` before pushing your code?
- The branch is out-of-date with the `master` branch. We will need it to be up-to-date to be able to merge.

maximearmstrong:
@bdferris-v2, I've added you as a reviewer for this PR. If you get a chance to look at it, your feedback is welcome :)

core/src/main/java/org/mobilitydata/gtfsvalidator/parsing/RowParser.java
```diff
@@ -157,39 +159,38 @@ public String asEmail(int columnIndex, boolean required) {
    *     unknown, only phone number starting by "+" are validated.
    */
   @Nullable
-  public String asPhoneNumber(int columnIndex, boolean required) {
-    return asValidatedString(columnIndex, required, fieldValidator::validatePhoneNumber);
+  public String asPhoneNumber(int columnIndex, FieldLevelEnum level) {
```
maximearmstrong, line 162: We will need to change the doc string here. I think it makes the Task :aggregateJavadoc to fail [here](https://github.com/MobilityData/gtfs-validator/runs/6619597119?check_suite_focus=true) and [here](https://github.com/MobilityData/gtfs-validator/runs/6619597193?check_suite_focus=true).

core/src/main/java/org/mobilitydata/gtfsvalidator/annotation/Recommended.java
```diff
@@ -0,0 +1,39 @@
...
+import java.lang.annotation.Target;
+
+/**
+ * Adds a validation that the field or a file is recommended.
+ *
+ * <p>Example.
+ *
+ * <pre>
+ *   {@literal @}GtfsTable(value = "feed_info.txt", singleRow = true) {
+ *   @Recommended
```
maximearmstrong, line 31: My bad, there was an error in my previous suggestion. It makes the Task :aggregateJavadoc to fail [here](https://github.com/MobilityData/gtfs-validator/runs/6619597119?check_suite_focus=true) and [here](https://github.com/MobilityData/gtfs-validator/runs/6619597193?check_suite_focus=true).
```suggestion
 *   {@literal @}Recommended
```

maximearmstrong (commented):
2 things that should fix the checks that are failing right now.

core/src/main/java/org/mobilitydata/gtfsvalidator/parsing/RowParser.java
```diff
@@ -52,8 +53,9 @@
  */
 public class RowParser {
 
-  public static final boolean REQUIRED = true;
-  public static final boolean OPTIONAL = false;
+  public static final FieldLevelEnum OPTIONAL = FieldLevelEnum.OPTIONAL;
+  public static final FieldLevelEnum RECOMMENDED = FieldLevelEnum.RECOMMENDED;
+  public static final FieldLevelEnum REQUIRED = FieldLevelEnum.REQUIRED;
```
bdferris-v2, line 58: Are these used anywhere?

bdferris-v2: Never mind.  Ignore this.

bdferris-v2 (commented):
I've got a high-level macro comment that may just highlight my ignorance of how this code works, but let me give it a shot.

I feel like there is a model mismatch in how you are treating Required vs Recommended vs Optional in this PR.  On one hand, you've got the file and field descriptors, which have separate booleans for isRequired() and isRecommended().  On the other hand, you've introduced a new FieldLevelEnum for passing into RowParser specifying the acceptance level for a given field.  This PR, as best I can tell, doesn't actually update the logic in TableLoaderGenerator for setting that FieldLevelEnum when calling RowParser methods, instead relying on the existing isRequired() method and switching between REQUIRED or OPTIONAL, but never actually passing in RECOMMENDED.  Maybe that's a function of the way you split up the PR into parts?

So at a tactical level, I think it would be good to update that logic in this PR.

But more generally, it made me wonder how this should actually be modeled overall.  So I have questions like:

1) Are all Required fields also Recommended by implication?
2) Should the boolean isRequired() and isRecommended() methods in the descriptors be refactored to use FieldLevelEnum directly instead as a single state?

maximearmstrong:
@bdferris-v2 Very good point, I hadn't noticed that this logic implementation was missing in `TableLoaderGenerator`. I also think it should be updated in this PR.

For your questions

1. This is a good question. A field is Required because the specification says it is required. A field is Recommended because the best practices say that the field "should" be there - even if "must" is used in the best practices, we consider it is recommended. Then yes, most probably we can say that Required fields are Recommended by implication. We would not want notices to be created for both though - the highest requirement level reached for a field is the one considered. @isabelle-dr Do you have thoughts on this?

2. Now that you mention it, I think it could be the best. It seems that the `required()` method in `GtfsFieldDescriptor` is only used in `TableLoaderGenerator` to generate the `load()` and `getRequiredColumnNames()` methods of the table loaders. Most likely using Enum values would be better. Though, would we keep the `isRecommended()` and `isRequired()` methods for the file descriptors as is?

KClough:
@bdferris-v2 it looks like I messed up this PR when splitting out notices from annotations. I think we're in good spot now.

@maximearmstrong I believe I've resolved all javaDoc issues. The build step passes locally for me.

Once this PR is merged, I'll rebase my other PRs and resolve any other outstanding issues.

bdferris-v2:
Apologies, but I think my original concern still exists.  Namely, you've updated all the methods in `RowParser` to accept the new `FieldLevelEnum`, but you haven't updated the call site generated in `TableLoaderGenerator` to ever pass in a `RECOMMENDED` value where appropriate.

Specifically, [TableLoaderGenerator L331](https://github.com/MobilityData/gtfs-validator/blob/master/processor/src/main/java/org/mobilitydata/gtfsvalidator/processor/TableLoaderGenerator.java#L331):

```
"rowParser.$L($L, $T.$L$L)",
    gtfsTypeToParserMethod(field.type()),
    fieldColumnIndex(field.name()),
    RowParser.class,
    field.required() ? "REQUIRED" : "OPTIONAL",  // <= This line right here
    ...
```

That's the piece I'm still looking for in this PR.

processor/src/main/java/org/mobilitydata/gtfsvalidator/processor/TableLoaderGenerator.java
```diff
@@ -328,7 +329,7 @@ private MethodSpec generateLoadMethod() {
               gtfsTypeToParserMethod(field.type()),
               fieldColumnIndex(field.name()),
               RowParser.class,
-              field.required() ? "REQUIRED" : "OPTIONAL",
+              field.required() ? "REQUIRED" : (field.recommended() ? "RECOMMENDED" : "OPTIONAL"),
```
bdferris-v2, line 332: Apologies for the nitpick, but I think this would be more straightforward if you referenced `FieldLevelEnum.class` directly here, instead of indirectly referencing the value via RowParser.  Aka replace the value on line 331.

core/src/main/java/org/mobilitydata/gtfsvalidator/parsing/RowParser.java
bdferris-v2, line 58: You could drop these three lines entirely if you made the suggested change below.

KClough: I didn't see a suggested change below, but I believe I understand what you're proposing. This does make it a little easier to follow. It doesn't make sense for `RowParser` to have to expose `FieldLevelEnum` values this way. 

Please see my updated commit. 

Thanks!

bdferris-v2 (approved):

isabelle-dr:
Thanks for following-up on this @bdferris-v2! 
@maximearmstrong, can you have another look?
I have been following from far, do we need to open an issue to re-iterate the logic for adding the @recommended annotation in a later stage?

maximearmstrong (approved):
Thank you @KClough and @bdferris-v2! The build passes for me as well. We are ready to merge 🎉 

@isabelle-dr I don't think we need to open an issue to re-iterate the logic for adding the `@recommended` annotation. I think this implementation works well.

PR #1146, opened 2022-05-13
feat: Initial support for validation app GUI

bdferris-v2:
Provides initial implementation for the simple GUI to be used with the packaged validation app.  Details of the UI product requirements and proposed implementation are documented in https://bit.ly/gtfs-validator-packaged-exe

Screen Shot 2022-05-13 at 1 42 35 PM
.

PR #1155, opened 2022-05-18
test: Add path parsing tests

barbeau:
**Summary:**

In PR https://github.com/MobilityData/gtfs-validator/pull/1146 we observed that changes were made to the input GTFS file path handling that broke execution (CLI and the GUI) on Windows. However, CI never caught this because currently all end-to-end tests are run only on `ubuntu-latest`.

~~This PR changes the CI configuration to run the smallest end-to-end test on `windows-latest` as well.~~ 

Given the complexity that's emerged of running end-to-end Windows CI  (see below comments), I've changed this PR to add simple unit tests to catch the same issue (https://github.com/MobilityData/gtfs-validator/issues/1158).

This should catch future instances of path issues on Windows prior to PR merges.

**Expected behavior:** 

~~CI should catch changes that break end-to-end functionality on Windows as well as Linux.~~

Unit test should catch path parsing errors


barbeau:
The Windows end-to-end script currently fails because of the version tag script:
https://github.com/MobilityData/gtfs-validator/runs/6492006893

```
VERSION_TAG=edge
  if [[ $GITHUB_REF == refs/tags/* ]]; then
    VERSION_TAG=${GITHUB_REF#refs/tags/}
  elif [[ $GITHUB_REF == refs/heads/* ]]; then
    VERSION_TAG=-$(echo ${GITHUB_REF#refs/heads/} | sed -r 's#/+#-#g')
  elif [[ $GITHUB_REF == refs/pull/* ]]; then
    VERSION_TAG=-pr-1155
  fi
  if [ ${VERSION_TAG} != ${GITHUB_REF#refs/tags/} ]; then
    VERSION_TAG=v${VERSION_TAG}-sha-${GITHUB_SHA::8}-SNAPSHOT
  fi
  echo ::set-output name=versionTag::${VERSION_TAG}
  shell: C:\Program Files\PowerShell\7\pwsh.EXE -command ". '{0}'"
ParserError: D:\a\_temp\92c1c[3](https://github.com/MobilityData/gtfs-validator/runs/6492006893?check_suite_focus=true#step:3:3)78-c3da-4a31-9920-9071a4c0ff72.ps1:3
Line |
   3 |  if [[ $GITHUB_REF == refs/tags/* ]]; then
     |    ~
     | Missing '(' after 'if' in if statement.

Error: Process completed with exit code 1.
```

This script is removed in PR https://github.com/MobilityData/gtfs-validator/pull/1147 as part of the refactor of version number handling, so let's revisit this PR after PR https://github.com/MobilityData/gtfs-validator/pull/1147 is merged, as that should resolve the issue.

barbeau:
So apparently running the validator JAR file with wildcards like:

```
java -jar main/build/libs/gtfs-validator-*.jar ...
```

...doesn't work on Windows PowerShell.

Here's the failed CI run with PowerShell:
https://github.com/MobilityData/gtfs-validator/runs/6562700554

...with the error:

```
Run java -jar main/build/libs/gtfs-validator-*.jar --url https://openmobilitydata.org/p/transport-for-nsw/237/latest/download --output_base output --country_code au --storage_directory gs.zip
  java -jar main/build/libs/gtfs-validator-*.jar --url https://openmobilitydata.org/p/transport-for-nsw/[2](https://github.com/MobilityData/gtfs-validator/runs/6562700554?check_suite_focus=true#step:6:2)[3](https://github.com/MobilityData/gtfs-validator/runs/6562700554?check_suite_focus=true#step:6:3)7/latest/download --output_base output --country_code au --storage_directory gs.zip
  shell: C:\Program Files\PowerShell\7\pwsh.EXE -command ". '{0}'"
  env:
    JAVA_HOME: C:\hostedtoolcache\windows\Java_Temurin-Hotspot_jdk\11.0.15-10\x6[4](https://github.com/MobilityData/gtfs-validator/runs/6562700554?check_suite_focus=true#step:6:4)
Error: Unable to access jarfile main/build/libs/gtfs-validator-*.jar
Error: Process completed with exit code 1.
```

I can reproduce this locally on Windows 10 Home w/ PowerShell too.

I tried switching to using `shell: bash` on Windows, and this fixes the wildcard issue. 

However, using bash hides the path issue that we saw in https://github.com/MobilityData/gtfs-validator/issues/1158 (i.e., the path issue only happens when using PowerShell) and defeats the purpose of having a separately Windows CI task. Here's a successful CI run when it should have failed due to the path handling bug (which I added to this PR to test):
https://github.com/MobilityData/gtfs-validator/runs/6579581437

@bdferris-v2 Based on your work in PR https://github.com/MobilityData/gtfs-validator/pull/1147 do you have ideas for the most elegant way to handle this lack of wildcard expansion on Windows PowerShell?

Some options I can think of:
1. Change the GitHub Action script to run different commands for Linux and Windows to handle wildcard expansion on Windows - Pros is no changes to the app, con is that it's a bit messy for CI.
2. Pass the version number from Gradle build to CI (e.g., via environmental variable) so the exact JAR file name can be executed on CI - Pros is CI script remains same across platforms, con is changes required to the app.

bdferris-v2:
First, a meta question:

Is there really value in running the entire suite of end-to-end feed tests on both Windows and Linux?  I think the general classes of errors we are catching here ("does this feed cause the validator to crash unexpectedly?") will be the same on both platforms.  Instead, there's an alternate approach where we have a single specific Windows test with a single feed that's just designed to capture platform quirks of Windows.

Unfortunately, even if you buy that argument, I'm not sure how much it actually buys us because we'd still need to duplicate major portions of the job definition for Windows.

So on to some alternatives.  A few that I can think of:

1) We use the axion-release plugin option to [force override the version](https://axion-release-plugin.readthedocs.io/en/latest/configuration/force_version/) aka -Prelease.forceVersion=1.0.0 such that the produced jar has a consistent version number that can be hard-coded in the test.  I think this would mostly work, except the plugin still appends -SNAPSHOT for non-release runs.  Thus, I think this would break under an actual release build (where axion would helpfully drop the -SNAPSHOT), thus we are back needing a wildcard...
2) We try to extract the version using the `./gradlew currentVersion` command.  More specifically, we'd use the concise output version: `./gradlew cV -q -Prelease.quiet`.  That could be a pre-run step and then used to specify the command-line.  I think this should work?
3) We use a run step to construct the command-line for the jar with some os-specific logic as a pre-step before running all the actual validation runs.
4) We switch the shell to `bash` for all environments.  I know it's not ideal, but I do believe it would have still caught #1158 even in that configuration.

barbeau:
>Is there really value in running the entire suite of end-to-end feed tests on both Windows and Linux?

Thanks @bdferris-v2, that's a good point. That got me thinking - given these issues, I think a simple unit test for path parsing is probably the best/simplest approach here that gives us coverage for #1158.

I added such a test in https://github.com/MobilityData/gtfs-validator/pull/1155/commits/e64e23435080c5e171b2d6d126b223b6355e27d2 and removed the end-to-end Windows CI from this PR (we still run tests on Windows CI, though). I intentionally reintroduced the bug from #1158 here in this PR to see how the CI handles it. Hopefully it fails as it should. If that works, I'll remove the bug and hopefully we'll have green lights.

Thoughts and feedback are welcome.

barbeau:
Ok, looks good. With the bug from #1158 still in this PR, the new unit test is failing on Windows as expected:
https://github.com/MobilityData/gtfs-validator/runs/6595978098

...but passing on Linux:
https://github.com/MobilityData/gtfs-validator/runs/6595977965

I just removed the #1158 bug from this PR in https://github.com/MobilityData/gtfs-validator/pull/1155/commits/62d434ddc5594205d97d500de2dc151b58754d57, so hopefully everything should be green (I tested it locally on Windows and it works).  

This PR should be ready for review.

barbeau:
Looks like the acceptance tests are failing:
https://github.com/MobilityData/gtfs-validator/runs/6596556690

...with:

```
Run actions/download-artifact@v2
Starting download for gtfs-validator-master
Error: Unable to find an artifact with the name: gtfs-validator-master
```

@bdferris-v2 Is this related to the below TODO statement in acceptance_test.yml?

```yml
    steps:
      - uses: actions/checkout@v1
        with:
          ref: master
      # We still need to compute the versionTag for the master branch until
      # changes in PR #1147 have gone in, at which point we can clean up this
      # codepath to match the snapshot.
      - name: Prepare version name
```

bdferris-v2:
@barbeau tracking the fix in PR #1168.

bdferris-v2 (approved):
LGTM

Issue #1163, opened 2022-05-20
Add the GUI app package to the test_pack_doc.yml workflow


bdferris-v2:
So that the GUI application package can be automatically built as part of the release process (along with other CI runs), we'd like to add a phase to the existing `test_pack_doc.yml` workflow to package the application.  Ideally, we'd run it on Windows and Mac OS (and maybe Linux?), though I don't know if GitHub supports all those flavors.

barbeau:
Windows may be a little tricky as it requires (summarized from the [build docs](https://github.com/MobilityData/gtfs-validator/blob/master/docs/BUILD.md#packaging-as-installable-application)):
* JDK >= 15
* https://wixtoolset.org/ on the path
* .NET Framework 3.5.1

I imagine WiX can be fetched dynamically within a GitHub Action.

.NET Framework might be bundled with the Windows GitHub Action runner (?). From a quick look at the Windows 2022 runner docs, I don't see .NET Framework 3.5.1 specifically:
https://github.com/actions/virtual-environments/blob/main/images/win/Windows2022-Readme.md

...but the Windows 2019 runner lists `Microsoft.Net.Component.3.5.DeveloperTools`:
https://github.com/actions/virtual-environments/blob/main/images/win/Windows2019-Readme.md

So if `windows-latest` doesn't work (which is currently the same as `windows-2022`, we might want to try `windows-2019`.

bdferris-v2:
Are there any concerns about how often we trigger full packaging runs?  I'm not totally clear on what tier MobilityData is in, but I assume they are paying for GitHub actions, both in minutes run + storage.  Packaging will certainly use minutes + storage, so I wondered if this is something that should only be run on release (similar to Docker?) versus every PR?

barbeau:
@isabelle-dr could answer the GitHub tier question definitively, but I'm not aware of any concerns (i.e., maybe on free tier?).

I think mirroring Docker is fine in terms of making the artifacts available - if someone wants bleeding edge they will probably be running command line. However, knowing if something is breaks in the packaging execution would be valuable prior to release, and I believe the only way to do that as of now is to run on every PR?

bdferris-v2:
[...]

PR #1166, opened 2022-05-25
feat: Initial entry for a workflow to package the gui application.

bdferris-v2:
Implementation for a GitHub workflow to build the GUI application installers and packages on all key environments.

### DIFF
--- a/.github/workflows/package_app.yml
+++ b/.github/workflows/package_app.yml
@@ -7,7 +7,7 @@ on:
     branches: [ master ]
   release:
     types: [ prereleased, released ]
-  workflow_dispatch
+  workflow_dispatch:
 
 jobs:
   validate_gradle_wrapper:
--- a/.github/workflows/package_app.yml
+++ b/.github/workflows/package_app.yml
@@ -3,8 +3,6 @@ name: Package Application
 on:
   push:
     branches: [ master ]
-  pull_request:
-    branches: [ master ]
   release:
     types: [ prereleased, released ]
   workflow_dispatch:
@@ -22,15 +20,19 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ macos-latest, windows-latest ]
+        os: [ macos-latest, windows-latest, ubuntu-latest ]
     steps:
       - uses: actions/checkout@v2
         with:
+          # We need to download all tags so that the axion-release-plugin
+          # can resolve the most recent version tag.
           fetch-depth: 0
       - name: Set up JDK 17
         uses: actions/setup-java@v2
         with:
+          # We need a recent version of Java with jpackage included.
           java-version: '17'
+          # We use the zulu distribution, which is an OpenJDK distro.
           distribution: 'zulu'
       - name: Package GUI app installer with Gradle
         uses: gradle/gradle-build-action@v2
--- a/.github/workflows/package_app.yml
+++ b/.github/workflows/package_app.yml
@@ -49,3 +49,4 @@ jobs:
           path: |
             app/pkg/build/jpackage/*.msi
             app/pkg/build/jpackage/*.dmg
+            app/pkg/build/jpackage/*.deb
--- a/.github/workflows/package_app.yml
+++ b/.github/workflows/package_app.yml
@@ -7,7 +7,7 @@ on:
     branches: [ master ]
   release:
     types: [ prereleased, released ]
-  workflow_dispatch:
+  workflow_dispatch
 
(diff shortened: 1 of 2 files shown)


bdferris-v2:
Per https://github.com/MobilityData/gtfs-validator/actions/runs/2382685953, this workflow is successfully packaging for Windows, Mac, and Linux.  Ready for review.

.github/workflows/package_app.yml
```diff
@@ -0,0 +1,52 @@
...
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+    paths:
+      - '**/package_app.yml'
+  release:
+    types: [ prereleased, released ]
+  workflow_dispatch:
```
maximearmstrong, line 12: Do we need the `:` here since we are not providing anything after?

.github/workflows/package_app.yml
```diff
@@ -0,0 +1,52 @@
+name: Package Application
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+    paths:
+      - '**/package_app.yml'
```
maximearmstrong, line 9: @isabelle-dr mentioned it would be useful to run this workflow on every commit in a PR so that we have access to the artifact to test the validator using the package. Thoughts?

maximearmstrong (commented):
Two small comments before approval. Thanks @bdferris-v2

maximearmstrong:
Also, now that #1168 is merged, the acceptance tests should pass.

.github/workflows/package_app.yml
bdferris-v2: I [asked about this](https://github.com/MobilityData/gtfs-validator/issues/1163#issuecomment-1136251359) in issue #1163 and hadn't heard anything from @isabelle-dr there.  However, if you are hearing that it'd be helpful to run this on every PR, I'm happy to remove the path restriction.

.github/workflows/package_app.yml
bdferris-v2: I think we do need it.  I tried removing it and got a workflow error:

https://github.com/bdferris-v2/gtfs-validator/actions/runs/2387663309

More importantly, it seems pretty common to include the `:` in other projects:

https://sourcegraph.com/search?q=context:global+f:.github/workflows+workflow_dispatch&patternType=regexp

barbeau (commented):
Wow, I'm amazed that this was so straightforward 💯.

I tested the Windows installer that was uploaded and it works correctly 👍.

One small item in-line.

.github/workflows/package_app.yml
maximearmstrong: Perfect, thank you!

.github/workflows/package_app.yml
```diff
@@ -0,0 +1,50 @@
+name: Package Application
```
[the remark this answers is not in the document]
bdferris-v2: Seems reasonable. Done.

maximearmstrong (approved):
One small thing aside, we are ready to merge :)

isabelle-dr (changes requested):
I believe we could update README.md in this PR to add the instructions? 
We could add a section "Run the validator app" describing (1) how to download the app and (2) how to use it.
If we want to mirror what we have with the JAR file, we will also need to include a DOWNLOAD_SNAPSHOT_APP.md similar to [DOWNLOAD_SNAPSHOT_JAR.md](https://github.com/MobilityData/gtfs-validator/blob/master/docs/DOWNLOAD_SNAPSHOT_JAR.md).

.github/workflows/package_installers.yml
```diff
@@ -0,0 +1,50 @@
+name: Package Installers
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
```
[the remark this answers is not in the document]
bdferris-v2: Done.

bdferris-v2:
@isabelle-dr I'm happy to write that documentation, but I think it's getting beyond the original scope of this PR?  Would it be ok if I updated documentation in a follow-up PR?

isabelle-dr:
@bdferris-v2 No problem updating the documentation in another PR, I thought it was in scope.

isabelle-dr (approved):

isabelle-dr:
We just need to update this branch with `master` and we can merge.

bdferris-v2:
Done.

maximearmstrong:
Thank you @bdferris-v2 !

PR #1169, opened 2022-05-25
feat: Persist GUI app settings between executions

### LINKED ISSUE #1165: Persist GUI app settings between executions
Currently, the GUI app does not remember settings (input and output paths, advanced settings) between different executions of the app.  It would be a nice-to-have to persist these between runs so that a user doesn't have to update them each time they use the app.

bdferris-v2:
Currently, the GUI app does not remember settings (input and output paths, advanced settings) between different executions of the app.   This PR adds support for persisting settings between runs of the validator, using the Java Preferences API for storage.

Closes #1165.

- [X] Run the unit tests with `gradle test` to make sure you didn't break anything
- [X] Format the title like "feat: [new feature short description]". Title must follow the Conventional Commit Specification(https://www.conventionalcommits.org/en/v1.0.0/).
- [X] Linked all relevant issues

### DIFF
--- a/app/gui/src/main/java/org/mobilitydata/gtfsvalidator/app/gui/GtfsValidatorApp.java
+++ b/app/gui/src/main/java/org/mobilitydata/gtfsvalidator/app/gui/GtfsValidatorApp.java
@@ -25,6 +25,8 @@ import java.io.File;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.ResourceBundle;
 import javax.swing.BorderFactory;
 import javax.swing.Box;
@@ -68,6 +70,8 @@ public class GtfsValidatorApp extends JFrame {
   private final ValidationDisplay validationDisplay;
   private final ResourceBundle bundle;
 
+  private final List<Runnable> preValidationCallbacks = new ArrayList<>();
+
   public GtfsValidatorApp(
       MonitoredValidationRunner validationRunner, ValidationDisplay validationDisplay) {
     super("GTFS Schedule Validator");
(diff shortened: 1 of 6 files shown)


maximearmstrong (commented):
Everything is fine with the code, but I encounter a strange behavior on macOS. After building the package, I use the .dmg. The application opens then closes automatically, without even displaying the application window. Here are the logs: 

```May 27, 2022 12:59:38 PM org.mobilitydata.gtfsvalidator.app.gui.Main main
INFOS: gtfs-validator: start
May 27, 2022 12:59:39 PM org.mobilitydata.gtfsvalidator.app.gui.Main main
INFOS: gtfs-validator: exit```

Any ideas on what the problem might be?

bdferris-v2:
The issue was that I hadn't added the "java.prefs" module to the list of required modules in the apps module spec.  I fixed that in 09d79fb348d993f93a027bf4290f88fb796bf324.

But more broadly, the exception that was thrown was getting added to the run.log because it was an uncaught exception in the UI thread.  In 101b2da1e0265ae7e0242457445b44926714c29f, I added a custom uncaught exception handler that logs these exceptions (and anything else we missed) into the application log.  I've verified that the logging properly logs the missing error above.  Doesn't stop the crash but does make it easier to debug.

This all makes me wonder if we can actually run the full application in the CI environment somehow to catch issues like this.  Not sure if a GUI application would be supported though...

maximearmstrong:
> This all makes me wonder if we can actually run the full application in the CI environment somehow to catch issues like this. Not sure if a GUI application would be supported though...

If we can do that, that would be great. We could open an issue to investigate it. I don't think it's required for this PR though.

maximearmstrong (approved):
LGTM

PR #1185, opened 2022-06-07
feat(ci): use slim base for runtime container

themightychris:
Produces a smaller runtime Docker container image, does not appear to cause any issue in runtime


themightychris:
Folding this into #1185

Issue #1183, opened 2022-06-07
Validator didn't open on MacOS


dget:
## **Bug report** 

**Describe the bug**
I got the error below after trying to open the validator.

 
<img width="417" alt="Screen Shot 2022-06-07 at 9 47 47 AM" src="https://user-images.githubusercontent.com/307164/172396754-d8175932-3799-4e0c-9dc6-d4b8113af392.png">

**How we reproduce the bug**
1. Downloaded the MacOS release of 3.1.0
2. Try to open it

**Screenshots:** 
Visual examples can help us better understand the issue.

**Environment versions**
- validator version: 3.1.0
- Java version: 
- OS versions: MacOS 10.15.7

bdferris-v2:
I've been able to repro this on my own laptop with the release image, but not with one built locally.  I'm not sure what's going on, but I'll do some investigation on the Github workflow to start.

bdferris-v2:
I've found a similar error described in https://bugs.openjdk.org/browse/JDK-8276150.  Indeed, when I'm able to repro this myself: when I upload my locally built dmg (which I've verified runs successfully) to a website and download it back, the same file is now reported as damaged.  I think it all boils down to a code signing issue.

Interestingly, this was considered to be an actual bug in the JDK and it seems that while they attempted a fix in 2021, it didn't take and the fix was reapplied to the JDK just 30 days ago: https://bugs.openjdk.org/browse/JDK-8277493

I'm still investigating if that fix has made it into a production JDK yet.  Also evaluating if we could/should sign the app ourselves to get around some of these errors.

bdferris-v2:
Ok, did a little more digging and here is what I've found.

Per the [original JDK bug](https://bugs.openjdk.org/browse/JDK-8276150), this issue occurs when a "signed" (in the Apple code-signing sense) Java SDK is used to build a local jpackage application.  The resulting application still has the original signature, but a slightly different binary, so Mac OS thinks the binary is damaged (paraphrasing a bit here).

I've been able to see this by building the app (`./gradlew jpackage`) and looking at the code signing details of the app:

```
codesign -v -d app/pkg/build/jpackage/GTFS\ Validator.app
```

On my Mac with the default Java 15 runtime, I see output like the following:

```
Executable=/Users/bdferris/IdeaProjects/gtfs-validator/app/pkg/build/jpackage/GTFS Validator.app/Contents/MacOS/GTFS Validator
Identifier=jpackageapplauncher
[...]

PR #1186, opened 2022-06-09
fix: Temporary fix for MacOS validator app corruption

bdferris-v2:
Per #1183, I'm investigating the corruption of `dmg` package on MacOS for the packaged app.  This PR introduces a temporary fix to unsign the app via the GitHub `package_installers.yml` workflow.  A more permanent fix will be to actually sign the app.


maximearmstrong (approved):
LGTM, thank you @bdferris-v2!

PR #1204, opened 2022-07-06
feat: Updates acceptance tests to use the Mobility Database Catalogs

### LINKED ISSUE #1143: Update the Acceptance Tests by using the Mobility Database Catalogs
## **Feature request**
At the moment, the Acceptance tests fetch the urls from the Wikibase instance of the Mobility Database. This process takes time because the Wikibase instance API is relatively slow, mainly because we are using the Wikibase docker image for our instance.

Instead, we could use the CSV artifact created by the Mobility database catalogs. This would be a great improvement because:
- the latest dataset for each GTFS Schedule source can be downloaded using the latest URLs listed in the CSV.
- this solution is much faster
- the data is more accurate and recent
- 
The Acceptance Tests take a long time to run, partly because of the Wikibase instance of the Mobility Database.

**Proposed solution**
Using the [CSV artifact](https://storage.googleapis.com/storage/v1/b/mdb-csv/o/sources.csv?alt=media) as described in the [Mobility Database Catalogs](https://github.com/MobilityData/mobility-database-catalogs/pull/153#event-6593548165).

maximearmstrong:
**Summary:**

Closes #1143 

This PR update the acceptance tests so that they use the Mobility Database Catalogs instead of the outdated Wikibase instance.

**Expected behavior:** 

Acceptance tests should be run on the sources included in the Mobility Database Catalogs. Since the catalogs can be downloaded as a CSV file, acceptance tests should be much faster when fetching the URLs (no more waiting on the Wikibase API!). Also, the data is in greater quantity and more accurate.


scripts/mobility-database-harvester/harvest_latest_versions.py
```diff
@@ -11,79 +11,28 @@
...
 
-# Catalog constants
-GTFS_CATALOG_ID = "Q6"
-
-# Archives ids file constants
-ARCHIVES_IDS = "archives_ids"
-HARVESTING_DATE = "harvesting_date"
-EPOCH_DATE = "1970-01-01T00:00:00Z"
+# Mobility Database Catalogs constants
+CATALOGS_CSV = "https://bit.ly/catalogs-csv"
```
bdferris-v2, line 26: I'll admit I was a little nervous seeing a bit.ly link here, since I wasn't totally sure what it pointed to.  Given that you are using the expanded url below for URL_PREFIX, would it be reasonable to use the expanded url here as well?

scripts/mobility-database-harvester/harvest_latest_versions.py
```diff
@@ -11,79 +11,28 @@
...
+LATEST = "urls.latest"
```
bdferris-v2, line 27: nit: I think this would be clearer as "LATEST_URL"

.github/workflows/acceptance_test.yml
```diff
@@ -116,8 +116,8 @@ jobs:
       - name: Set URL matrix
         id: set-matrix
         run: |
-          python3 scripts/mobility-database-harvester/harvest_latest_versions.py -d scripts/mobility-database-harvester/datasets_metadata -l gtfs_latest_versions.json -o gtfs_older_versions.json
-          DATASETS=$(jq . scripts/mobility-database-harvester/datasets_metadata/latest_versions.json -c)
+          python3 scripts/mobility-database-harvester/harvest_latest_versions.py -d scripts/mobility-database-harvester
+          DATASETS=$(jq . scripts/mobility-database-harvester/latest_versions.json -c)
```
bdferris-v2, line 120: I'm curious why the jq line is need here.  Isn't the data already compact json when output from the python script?

scripts/mobility-database-harvester/harvest_latest_versions.py
```diff
@@ -11,79 +11,28 @@
...
 import os
 from os import path
-import json
-from tqdm import tqdm
 
-###############################################################################
-# This script harvests the latest dataset versions on the Mobility Database.
+#####################################################################################
+# This script harvests the latest dataset versions on the Mobility Database Catalogs.
 # Made for Python 3.9. Requires the modules listed in requirements.txt.
```
bdferris-v2, line 22: I think in general this script could use more documentation to help subsequent developers (and code reviewers like me) understand the behavior, the inputs, and the outputs of this script without having to dig through so much code.

scripts/mobility-database-harvester/harvest_latest_versions.py
```diff
@@ -145,23 +51,29 @@ def save_content_to_file(content, data_path, filename):
         json.dump(content, f)
 
 
-def load_content(data_path, filename):
-    file_path = path.join(data_path, filename)
-    with open(file_path, "r") as f:
-        content = json.load(f)
-    return content
+def harvest_latest_versions():
```
bdferris-v2, line 54: A docstring describing the format of the returned data would be helpful here.

scripts/mobility-database-harvester/harvest_latest_versions.py
```diff
@@ -145,23 +51,29 @@ def save_content_to_file(content, data_path, filename):
...
+    latest_urls = catalogs[LATEST].loc[catalogs[DATA_TYPE] == GTFS]
+    for index, value in latest_urls.items():
+        latest_url = value
+        source_file_name = latest_url.replace(URL_PREFIX, "").replace(URL_SUFFIX, "")
+        latest_versions[source_file_name] = latest_url
+
+    return latest_versions
 
 
 def apply_github_matrix_formatting(json_data):
```
bdferris-v2, line 67: I'm not sure json_data is an accurate or useful variable name here.  It's just a dictionary of name => url, yeah?

scripts/mobility-database-harvester/harvest_latest_versions.py
```diff
@@ -145,23 +51,29 @@ def save_content_to_file(content, data_path, filename):
         json.dump(content, f)
```
bdferris-v2, line 51: Instead of writing content to a file with a hard-coded name, I'd consider just writing to stdout directly and capturing that output directly in the workflow script, especially if you don't actually need the jq step.  If that doesn't work for some reason, I'd still inline the function and consider having the script accept a full path for the output file instead of a directory.  It wasn't quite clear in the workflow script where the `latest_versions.json` file was coming from since it hadn't been specified as an argument to the script.

bdferris-v2:
Meta comment: I don't think I'd ever looked too closely at the `acceptance_test.yml` script before.  This thing is a beast!   How many urls are in the old database?  I count 1300+ in the new database.  Is that an increase?  I guess the project doesn't get pull requests /too/ often, but I hadn't realized just how many feeds were being downloaded with each update to a PR.

scripts/mobility-database-harvester/harvest_latest_versions.py
maximearmstrong: Sure.

scripts/mobility-database-harvester/harvest_latest_versions.py
maximearmstrong: Right. I changed it to `latest_urls`.

.github/workflows/acceptance_test.yml
maximearmstrong: I finally removed that line after reading your other comments. Now using stdout only.

scripts/mobility-database-harvester/harvest_latest_versions.py
maximearmstrong: I agree, it makes more sense.

maximearmstrong:
> Meta comment: I don't think I'd ever looked too closely at the `acceptance_test.yml` script before. This thing is a beast! How many urls are in the old database? I count 1300+ in the new database. Is that an increase? I guess the project doesn't get pull requests /too/ often, but I hadn't realized just how many feeds were being downloaded with each update to a PR.

We have 1332 GTFS Schedule sources in the Mobility Database catalogs. Previously, 1117 on the Wikibase instance (more actually, but we noticed that many were duplicates). So we will run acceptance tests on 215 additional sources.

A lot of downloads every time a PR is updated, indeed! It is possible to skip the acceptance tests by using the `[acceptance test skip]` string in a commit, so that the tests are only run when necessary.

bdferris-v2 (approved):
LGTM - assuming the acceptance test runs to completion, this looks ready to go.

PR #1230, opened 2022-08-08
Docs: Update CONTRIBUTING.md

isabelle-dr:
Why does this always happen 🥲
Right after merging, I noticed there is a typo in the first sentence of the doc... always!


emmambd:
@isabelle-dr No need for a typo fix here! Your original version was correct, since "look", the singular verb, is describing "processes", which is plural. I hope that makes sense. Let me know if you need any clarification.

maximearmstrong:
Oh no worries @isabelle-dr ! I agree with @emmambd, I think the original sentence was correct. I'm closing this PR with this comment. Feel free to re-open it if we are missing something :)

isabelle-dr:
Thank you to you both 🙏😄

Issue #1232, opened 2022-08-09
Improve acceptance rules


maximearmstrong:
## **Feature request**
It would be beneficial to add new rules to the acceptance tests. Currently, two rules are tested: 
- Does the validator that uses the new code contribution generate more new errors than allowed?
- Were there too many corrupted sources/datasets during validation?
Although these rules are fundamental to the acceptance tests, other aspects are not considered: disappearing errors or warnings, unusual info notices, etc.

**Is your feature request related to a problem? Please describe.**
Specifically in #1188, the acceptance tests passed, even though the initial implementation prevented the validator from checking most of the validation rules. The reports generated did not contain any errors or warnings. Also, the reports contained the `unknown_file` info notice for required files such as `stop_times.txt`. Although the acceptance tests were not originally designed to cover this use case, it would be nice if they did.

**Proposed solution**
First, I think adding two new rules would be helpful:
- Are too many errors disappearing?
- Are too many warnings disappearing?

Another potential additional rule would be
- Do the reports contain the `unknown_file` notice for the required files?
since this is an indicator of a problem, but it is very specific.

isabelle-dr:
Thanks for opening this! 
There is another issue about a next iteration on the acceptance tests: #1076. This one isn't a duplicate, both flag different improvements to be made.

bdferris-v2:
I was looking at the acceptance test implementation and I noticed that there is apparently [code in the acceptance_test.yml](https://github.com/MobilityData/gtfs-validator/blob/master/.github/workflows/acceptance_test.yml#L185) to post a comment to the pull request with acceptance test status.  I've never actually seen this in action before.  Does it actually work?

Asking because I was taking an initial stab at improving the logic here and was thinking of refactoring some things, but I wanted to make sure I understood the existing behavior first.

maximearmstrong:
Oh, it looks like it's working when PRs are made from this repo (see [here](https://github.com/MobilityData/gtfs-validator/pull/1204#issuecomment-1178040173) and [here](https://github.com/MobilityData/gtfs-validator/pull/1212#issuecomment-1191414051)), but not from forks.

Looking more closely to the run logs, it looks like the problem is related to the GitHub token. [Here](https://github.com/MobilityData/gtfs-validator/runs/7448936134?check_suite_focus=true) (from our repo) vs [here](https://github.com/MobilityData/gtfs-validator/runs/7738417161?check_suite_focus=true#logs) (from a fork).

[...]

Issue #1076, opened 2021-12-10
Modify the logic for the acceptance tests


isabelle-dr:
As discussed [in this comment](https://github.com/MobilityData/gtfs-validator/pull/848#issuecomment-988812929):

Modify the logic of the acceptance tests so a PR is flagged if there is at least one additional error in **one** dataset from the Mobility Database (currently it's flagged if there is at least one additional error in at least 1% of datasets).

barbeau:
>Modify the logic of the acceptance tests so a PR is flagged if there is at least one additional error in one dataset from the Mobility Database

Is this one additional error type, or one additional occurrence of an error type (or both)?

If you're treating this more like a unit test and expecting the same output from both validators, then IMHO count occurrences of errors is important. In other words, you'd want to know if a code change made the output jump from 1 occurrence of `block_trips_with_overlapping_stop_times` to 1000 occurrences.

isabelle-dr:
👋 Hello, 

> Is this one additional error type, or one additional occurrence of an error type (or both)?

We are thinking one additional occurrence of an error type. Thanks for clarifying this!

PR #1238, opened 2022-08-15
feat: Refactor the output comparator to consider dropped errors

bdferris-v2:
Also includes a general refactoring of the report generation logic:
* pull more logic out of Main
* adding specific model classes for JSON serialization

As discussed in #1232.


bdferris-v2:
@maximearmstrong this is mostly ready for review and I'd love your feedback if you have cycles.

It isn't fully ready to review/commit because I haven't updated the `comment_generator.py` script yet.  That said, I wanted to propose a secondary refactor there as well.  The outputcomparator is already producing a block of text to be logged by the output comparator that seems pretty similar to what's being generated by the `comment_generator.py` script.  Maybe we could have outputcomparator just save its log lines to a file and use that in the PR comment, scraping the script entirely?

maximearmstrong:
> Maybe we could have output comparator just save its log lines to a file and use that in the PR comment, scraping the script entirely?

I agree. I think it would also be better from a design standpoint if the output comparator took on that responsibility.

output-comparator/src/main/java/org/mobilitydata/gtfsvalidator/outputcomparator/cli/Main.java
```diff
@@ -115,92 +106,45 @@ public static void main(String[] argv) {
...
-        noticeComparisonReport.update(
-            sourceId,
-            latestReport.getErrorNoticeReportByNoticeCode(noticeCode).getTotalNotices(),
-            sourceUrlContainer);
-      }
-      if (referenceReport.getNewErrorsListing(latestReport).size() >= args.getNewErrorThreshold()) {
-        ++badDatasetCount;
-      }
+      newErrors.compareValidationReports(sourceId, sourceUrl, referenceReport, latestReport);
+      droppedErrors.compareValidationReports(sourceId, sourceUrl, latestReport, referenceReport);
```
maximearmstrong, line 115: This is great. This also means that we apply the same threshold for both new errors and dropped errors. 

@isabelle-dr Do you see a case where it would be important to have a different threshold for each case? Eg. 10% above for new errors, 20% under for dropped errors. I'm fine with having the same for both cases, just wondering what are your thoughts on this.

output-comparator/src/main/java/org/mobilitydata/gtfsvalidator/outputcomparator/model/report/AffectedSource.java
```diff
@@ -0,0 +1,26 @@
...
+  public abstract String sourceId();
+
+  public abstract String sourceUrl();
+
+  /**
+   * The number of notices of a particular type.
+   *
+   * @return
+   */
+  public abstract int count();
```
maximearmstrong, line 21: When you say "the number of notices of a particular type", this should return the number of errors or warning, for instance? 

I'm not the most familiar with Autovalue, but won't the method `create` below assign `totalNotices` to `count`? Or will the library will create a new variable `self.totalNotices`? I'm just wondering what is the expected behaviour between `count` and `totalNotices`.

output-comparator/src/main/java/org/mobilitydata/gtfsvalidator/outputcomparator/io/ChangedNoticesCollector.java
```diff
@@ -0,0 +1,121 @@
...
+ * collects the set of newly-generated error notices that changed between the two validator
+ * versions.
+ */
+public class ChangedNoticesCollector {
+
+  /**
+   * If the number of new error notices in a report meets or exceeds this threshold, the
+   * corresponding dataset will be considered "invalid".
+   */
+  private final int newErrorThreshold;
```
maximearmstrong, line 24: I'm wondering if this shouldn't be a percentage threshold as well, instead of being an absolute value. 

In practice, 1 new error in a dataset that already has over 100 errors is not much, but 10 new errors in a dataset that previously had 2 errors seems like a lot. @isabelle-dr do you have thoughts on this?

maximearmstrong (commented):
I've added a few comments, but overall it's great! The new design brings a lot of clarity to the code. Thanks a lot @bdferris-v2!

output-comparator/src/main/java/org/mobilitydata/gtfsvalidator/outputcomparator/model/report/AffectedSource.java
bdferris-v2: I think it would always be the number of errors in practice based on the calling methods.  We could clarify that here.

As for @AutoValue behavior, it will assigned `totalNotices` to `count`.  Internally, @AutoValue is generating a constructor that has a param for each defined field, in field-declaration order.  I'm not sure `count` is the most descriptive name, but I kept it to match the schema of the original JSON output.  We could probably change that without breaking too much if you thought it was appropriate?

maximearmstrong: I mistook "type" for "severity" here. Since we count the errors by source, wouldn't it be "The number of notices of a particular severity" instead?

If by `type` we mean the `count` per notice, eg. the `count` for `missing_required_field`, I think it makes sense.

`totalNotices` is used in the report as the number of notices per type of notice, so if this is what we want to do here, it makes sense too. I think in both cases, we could just make the docstring more descriptive and that would do the trick.

> I'm not sure count is the most descriptive name, but I kept it to match the schema of the original JSON output.

Do you mean the original output of the acceptance tests? If yes, we can change `count` for `totalNotices` or another more descriptive name without any problem. It wouldn't be a breaking change, only the validation report must stay the same.

bdferris-v2: It's definitely the notice type, not the severity.  I renamed the field + tried to reword the comment for clarity.

bdferris-v2:
I've updated the comment generation code, as discussed.  This is ready for final review.

output-comparator/src/main/java/org/mobilitydata/gtfsvalidator/outputcomparator/cli/Main.java
isabelle-dr: Interesting question, I believe the current threshold is 1%. 

In the short term: 
I'd like to have a threshold as: every new occurrence of a notice (error or warning) is captured.
I am saying this because I'd like to use this test more for analytics and impact management (and depending on the results: contact every agency concerned), more than a green light to merge a PR or not (this was the initial idea, and this is why it's called "acceptance test").
So I'd like the threshold to not be a % anymore: just a count, and if it is it >1, it appears in the report + message.
This was the idea behind issue #1076.

Longer term if we see a need: a custom threshold that the developer can define depending on the granulometry needed.

If we go this route we might want to:
1.  rename this test (maybe "impact report")
2. run it ad-hoc instead of on every commit, knowing that it would definitely be run before merging as part of the review process.

@bdferris-v2 and @maximearmstrong what do you think of this?

isabelle-dr: And we would use `end_to_end` on every commit just to make sure nothing is broken.

output-comparator/src/main/java/org/mobilitydata/gtfsvalidator/outputcomparator/io/ChangedNoticesCollector.java
isabelle-dr: See my comment above 😊. I'd like this test to capture as much information as possible.

isabelle-dr:
Thanks for working on this @bdferris-v2 !
I have added a comment above. We might need to update [ACCEPTANCE_TESTS.md](https://github.com/MobilityData/gtfs-validator/blob/master/docs/ACCEPTANCE_TESTS.md) to reflect the changes in this PR.

What is going to be the schema of the new report with these changes?

output-comparator/src/main/java/org/mobilitydata/gtfsvalidator/outputcomparator/cli/Main.java
bdferris-v2: I still like the idea of keeping acceptance test with a relatively strict threshold.

The end-to-end tests are really only catching issues where a new change causes the validator to crash, which is not a particularly frequent failure case.  On the other hand, the diff-style acceptance test is capturing any regression in added/removed errors (we should probably add warnings in a follow-up).  I think we should be running this for every change, both for those that are changing notices and those that are not.  It's a good sanity check.

@isabelle-dr to your point about analytics and impact management, what are you hoping to measure here?

maximearmstrong: > I think we should be running this for every change, both for those that are changing notices and those that are not. It's a good sanity check.

Here do you mean for every commit or for every PR? Isabelle and I discussed the idea of requesting the acceptance tests to succeed on the last commit, before a PR is merged, but we are wondering if it is relevant to run them on each commit?

Our plan was to allow the acceptance test to be run "on-demand", by adding `[acceptance test run]` at the end of the commit message for instance, instead of adding `[acceptance test skip]` to skip them. One could run them when they think their code is ready to test. We would then use the end-to-end workflow as a "safety net", but I agree with you that the use case where the validator would crash is uncommon. I've refactored the end-to-end workflow in #1243 so it better fits our needs. Your feedback is welcome.

> I still like the idea of keeping acceptance test with a relatively strict threshold.

I'm fine with this.

output-comparator/src/main/java/org/mobilitydata/gtfsvalidator/outputcomparator/model/report/AffectedSource.java
maximearmstrong: Perfect, thanks!

output-comparator/src/main/java/org/mobilitydata/gtfsvalidator/outputcomparator/cli/Main.java
isabelle-dr: @bdferris-v2 

> to your point about analytics and impact management, what are you hoping to measure here?

I am hoping to get:
1. Getting general stats on what are the common error and warnings in order to improve the spec and the validator documentation. 
2. Get a list of producers and consumers that are affected by a change in data validity (additional or disappearing errors) and make sure they are aware (for example: by publishing a the list of datasets affected in the release note, and later down the road: set-up a notification system)

I have gathered my thoughts very quickly [here](https://docs.google.com/document/d/1Im8ZVDbGpZu20OzFa9fBY0ubi9y8502xtMxM0aFEsQs/edit#), there are still some questions to explore (how to best let the producers know that a change affects their dataset? How to let consumers know that there is a validity change in a dataset they are consuming?). I hope to include the PR process graph in the contribution guideline so code contributors know what to expect.

I think what we see with the addition of Fares v2 is a good example to put this into practice.

bdferris-v2: I think output comparator, as implemented, will support task #2.  For task #1, I would think you'd just periodically run the full validator on all feeds in the Mobility Database, without needing to compare it against some other version of the validator.

In general, I'm supportive of all of this.  However I do wonder if we are getting beyond the scope of this PR?

isabelle-dr: That makes sense.
You're right: this discussion is getting beyond the scope of this PR 😊, I'll let @maximearmstrong give the final review, thanks for this addition @bdferris-v2!

maximearmstrong (approved):

maximearmstrong:
I've approved the PR. Thanks @bdferris-v2!

> We might need to update [ACCEPTANCE_TESTS.md](https://github.com/MobilityData/gtfs-validator/blob/master/docs/ACCEPTANCE_TESTS.md) to reflect the changes in this PR.
> 
> What is going to be the schema of the new report with these changes?

@isabelle-dr do we need clarity on these items before merging?

isabelle-dr:
Thanks Maxime, I think it is in the scope of this PR to update [ACCEPTANCE_TESTS.md](https://github.com/MobilityData/gtfs-validator/blob/master/docs/ACCEPTANCE_TESTS.md) with the changes (addition of dropped errors, report schema change).

bdferris-v2:
@isabelle-dr I updated the `ACCEPTANCE_TESTS.md` to reflect the changes in this PR.

maximearmstrong (approved):
Perfect! Thanks @bdferris-v2

PR #1240, opened 2022-08-16
fix(docs): remove duplicate prefix in example commits

themightychris:
Minor typos


isabelle-dr (approved):
Thanks @themightychris 😊

Issue #1206, opened 2022-07-07
End to end workflow fails because of a SSL certificate


maximearmstrong:
**Describe the bug**
The end to end workflow fails when the SSL certificate of an URL we download a dataset from is currently invalid. See [here](https://github.com/MobilityData/gtfs-validator/runs/7241493846?check_suite_focus=true#step:10:99).

**How we reproduce the bug**
Run the end to end workflow.

**Expected behaviour**
The workflow should fail if the validation fails, not because of a SSL certificate problem.

**How to fix**
We should be using the datasets from the Mobility Database (in our archives), so we should not encounter that problem anymore.

**Environment versions**
- validator version: 3.1.2-SNAPSHOT
- Java version: JDK 11 temurin
- OS versions: [ubuntu-20.04](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-Readme.md)

PR #1243, opened 2022-08-18
feat: Lighten end_to_end.yml workflow

### LINKED ISSUE #1233: Lighten end-to-end workflows or remove them
## **Feature request**
Currently, the end-to-end workflows test over 100 datasets at every commit. All datasets from the Mobility Database catalogs are also being tested for each commit excluding `[acceptance test skip]` in the commit message.

Since end-to-end workflows were created before the acceptance tests (partially after that [conversation](https://github.com/MobilityData/gtfs-validator/pull/712#issuecomment-776110813) but even before [here](https://github.com/MobilityData/gtfs-validator/pull/414)), we are wondering if they are still relevant now that the acceptance tests are in place. Originally, they were created to check that new code contributions do not break the validation process.

**Proposed solution**
3 potential solutions:
1. we remove them completely and use only the acceptance instead
2. Lighten the end-to-end workflows, but keep hardcoded URLs as it is right now.
3. Lighten them, but by sampling (say 5%) of the sources in the catalogs.

By lightening end-to-end workflows, but keeping them, users could push commits to PRs and have an overview of how the validation behave, while skipping acceptance tests.

maximearmstrong:
**Summary:**

Closes #1233 and #1206 

This PR lightens the end-to-end workflow. It concentrates the end-to-end workflow efforts in a single file and samples GTFS Schedule sources from the Mobility Database catalogs. This way, each time the workflow is run, the snapshot validator is tested using different sources.

Using the latest URLs from the Mobility Database fixes issue #1206, since the SSL error was raised when using agency or transitfeeds URLs.

Changes:

- The `end_to_end_big.yml` and `end_to_end_100.yml` workflows are removed so that everything about the end-to-end workflow happens in `end_to_end.yml`.
- `end_to_end.yml` is refactored to mimic the behavior of the `acceptance_test.yml` workflow. The artifacts are named similarly to those in the acceptance test workflow.
- `harvest_latest_versions.py` is updated to allow sampling. 5% of the GTFS Schedule sources in the database are used for the end-to-end workflow (excluding sources requiring authentication for simplicity).
- `queue_runner.sh` is udpated to add a flag. It the flag is set to `true`, the queue runner will validate the datasets using both the snapshot and master validators, otherwise only the snapshot one.

**Expected behavior:** 

The end-to-end workflow is run on each commit and samples different GTFS Schedule sources from the Mobility Database each time.


bdferris-v2:
One high-level comment: am I reading the code correctly that a different sub-set of feeds will be selected with each run?  I feel like that could be problematic.  For example, if a particular feed causes a PR to fail, you wouldn't necessarily get the same feed on the next run, making it hard to repro.  I think I'd be more in favor of a consistent set of feeds for each run.

maximearmstrong:
@bdferris-v2 Our original idea was that we would cover more feeds this way, with different feeds being tested at each commit, but you bring a very good point. I changed the behaviour and added a consistent set of 50 feeds. 

@emmambd Do you think the [selection](https://github.com/MobilityData/gtfs-validator/pull/1243/files#diff-8beeab471004ef094ac241e660d83defe9b443c79f024105dacbbf792634d8a7) makes sense? I've selected feeds from various locations, some with features, some aggregated. Let me know if you think more or other feeds should be added.

isabelle-dr:
@maximearmstrong do we need to update (END_TO_END.md)[https://github.com/MobilityData/gtfs-validator/blob/master/docs/END_TO_END.md]?

emmambd:
> Do you think the [selection](https://github.com/MobilityData/gtfs-validator/pull/1243/files#diff-8beeab471004ef094ac241e660d83defe9b443c79f024105dacbbf792634d8a7) makes sense? I've selected feeds from various locations, some with features, some aggregated. Let me know if you think more or other feeds should be added.

@maximearmstrong I don't see any feed types missing so this list looks good to me! My one thought/consideration might be to look at end-to-end test failures in the past and see which feeds it failed on. That would be a last "check" to see if there are any other feed types that would be valuable to include.

maximearmstrong:
> @maximearmstrong do we need to update (END_TO_END.md)[https://github.com/MobilityData/gtfs-validator/blob/master/docs/END_TO_END.md]?

@isabelle-dr Sure, I wasn't aware of that file. Its content is outdated - I suggest we delete it. If you think there's is value in having our workflows documented, we can add `WORKFLOWS.md` in another PR to explain our workflows, including `test_package_doc.yml` and `end_to_end.yml`.

scripts/queue_runner.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash
 closing_curly_bracket="}"
-raw_queue_string=$*
+master="$1"
```
bdferris-v2, line 3: For this first command-line argument, the "true" vs "false" value doesn't give you much indication of how this flag is used at the call-site in the workflow.  I think this would be better if you specified the value as "--include-master" vs "--exclude-master" or something similar.  You don't actually need to add full command-line flag parsing to the script, you could just change line 18 to be `if [ "$master" = "--include-master" ]; then ...`

scripts/mobility-database-harvester/harvest_latest_versions.py
```diff
@@ -16,6 +16,7 @@
 import numpy as np
 import os
 from os import path
+import random
```
bdferris-v2, line 19: Is this still used?

maximearmstrong: Good point! It's not.

scripts/queue_runner.sh
maximearmstrong: Good point!

isabelle-dr:
@maximearmstrong sounds good, that's a good plan!

bdferris-v2 (approved):

Issue #1200, opened 2022-06-29
Sign the validator app with Apple and remove temporary fix


isabelle-dr:
A temporary fix has been made in #1186 to allow the validator app to work on Mac. 

When we finish the process of getting the app signed with Apple (July 2022), update and remove the temporary fix.

PR #1246, opened 2022-08-26
fix: Sign and Notarize MacOS App

bdferris-v2:
**Summary:**

As tracked in #1200, this PR adds support for code-signing and notarizing the MacOS app.  This will mean the user no longer has to ⌘-click the app to open it the first time.  Note that they will still get the following warning when running the app for the first time:

Screen Shot 2022-08-23 at 7 34 14 PM

But they can just click "Ok" and the app will run.  I don't think there is any way to avoid that dialog without distributing the app through the Mac App Store.

Closing #1200.


bdferris-v2:
A couple of high-level comments:
* This PR will ultimately need a number of new secrets configured in the repository, which I don't have permission to do.  @maximearmstrong or @isabelle-dr maybe we can setup some time to go through these, as it will take a few steps plus some discussion (example: which Apple Developer account should be used for the notarization request)
* I've tested this PR on my own repo fork (where I could configure secrets) but it's tricky to test as a pull-request because PRs don't have access to secrets.  I could potentially switch development to a branch within the main repo as an alternative, though I'm not sure I have permission to do that either.  Thoughts?

.github/workflows/package_installers.yml
```diff
@@ -33,35 +33,83 @@ jobs:
...
-        # unsign the app and repackage as a dmg.  This bug could be fixed in the future by upgrading
-        # to Java 18 or by officially signing the app.
-      - name: Unsign and package Mac OS app
+
+      # On MacOS, we now submit the app for "notarization", where Apple will scan the app for
+      # malware and other issues.  This step can take a few minutes or more; the action will wait
+      # until the report is available.
+      - name: "MacOS - Notarize Release Build"
+        if: matrix.os == 'macos-latest'
+        uses: devbotsxyz/xcode-notarize@v1
```
bdferris-v2, line 70: Apple requests that developers limit their # of notarization requests per day (~75/day is guidance).  I don't think we have that many pull requests per day, but I think it might be good to limit this either way.  That could mean restricting to just pushes+releases on the main master branch?

.github/workflows/package_installers.yml
```diff
@@ -33,35 +33,83 @@ jobs:
...
+        uses: devbotsxyz/import-signing-certificate@main
+        with:
+          # This should be a certificate + private key, exported as a p12 file and base64 encoded as
+          # a GitHub secret.  The certificate should be the:
+          # 'Developer ID Application: The International Data Organization For Transport (BF2U75HN4D)'
+          # certificate, locked with the specified passphrase.
+          certificate-data: ${{ secrets.DEVELOPER_ID_APPLICATION_CERTIFICATE_P12_BASE64 }}
+          certificate-passphrase: ${{ secrets.DEVELOPER_ID_APPLICATION_CERTIFICATE_PASSWORD }}
+          # The resulting keychain will be locked with this separate password.
+          keychain-password: ${{ secrets.KEYCHAIN_PASSWORD }}
```
bdferris-v2, line 58: Code-signing is a cheap operation (compare to notarization, as discussed below) but it still requires access to secrets.  We could restrict codesigning to just pushes+releases on the main master branch so that PRs would still run successfully?

app/pkg/build.gradle
```diff
@@ -109,8 +109,25 @@ jlink {
...
+                    // with the corresponding private key, if you wish to sign.  The certificate
+                    // should have a full name of:
+                    // 'Developer ID Application: The International Data Organization For Transport (BF2U75HN4D)'
+                    // jpackage will supply the 'Developer ID Application: ' prefix when calling the
+                    // codesign tool.
+                    '--mac-signing-key-user-name',
+                    'The International Data Organization For Transport (BF2U75HN4D)',
+                    // Entitlements are required for notarization of the app.
+                    '--mac-entitlements',
+                    "${projectDir}/src/main/mac-resources/entitlements.plist",
```
bdferris-v2, line 125: jpackage has support for codesigning the app, so I went with that option initially.  However, there is maybe an argument we should remove this change and make an explicit codesigning request in the `package_installers.yml` workflow instead.  That way, devs could still build the app locally without having to worry about codesigning setup.  Thoughts?

bdferris-v2:
I also left a number of open questions as comments to get your feedback on specific aspects of the change.

maximearmstrong:
> * This PR will ultimately need a number of new secrets configured in the repository, which I don't have permission to do.  @maximearmstrong or @isabelle-dr maybe we can setup some time to go through these, as it will take a few steps plus some discussion (example: which Apple Developer account should be used for the notarization request)

Both Isabelle and I can add secrets to the repo. I think it would be valuable to discuss the secrets, perhaps with @fredericsimard as well. @isabelle-dr, do you have thoughts? Could we plan a meeting for that?

> * I've tested this PR on my own repo fork (where I could configure secrets) but it's tricky to test as a pull-request because PRs don't have access to secrets.  I could potentially switch development to a branch within the main repo as an alternative, though I'm not sure I have permission to do that either.  Thoughts?

This is clearly not optimal for you! I've granted you maintainer permissions, so you can open branches and push them to the main repo. Isabelle or I would still need to add the secrets to facilitate your process.

fredericsimard:
> which Apple Developer account should be used for the notarization request
it@mobilitydata.org (the "owner" of the account)
> new secrets configured in the repository
I can do that as well, and I would do it with the account owner for GitHub, io@mobility just so it is not dependant/connected to any one employee.
> it's tricky to test as a pull-request because PRs don't have access to secrets
I found a possible alternative today: https://dev.to/petrsvihlik/using-environment-protection-rules-to-secure-secrets-when-building-external-forks-with-pullrequesttarget-hci
> Isabelle or I would still need to add the secrets to facilitate your process.
If you do, add them as either repo secrets or organization secrets limited to that repo.

.github/workflows/package_installers.yml
maximearmstrong: I think it's fair to restrict codesigning and notarization to just pushes+releases. `README.md` refers users to the release page to download the app, so I think most users won't need the PR app, only the contributors.

.github/workflows/package_installers.yml
maximearmstrong: > That could mean restricting to just pushes+releases on the main master branch?

I agree with this. I think it's fair solution since we won't be merging ~75 PR/day.

maximearmstrong (commented):
This is great. If testing goes well after we fix the secrets and restrict codesigning/notarization to just pushes+releases, I think we would be ready to merge.

app/pkg/build.gradle
fredericsimard: @bdferris-v2 I'll let @maximearmstrong approve this, if he has time. If not, I'll do it tomorrow before our meeting.

bdferris-v2: @fredericsimard not critical that we have this checked in beforehand.  We should be able to test it either way.

@maximearmstrong I took a slightly different approach here than I was originally thinking.  It turns out that the number of codesign commands run by jpackage is more than I initially thought.  As such, I stuck with the approach of letting jpackage sign the app but I wrapped the flags in a Property check for `sign-app`, which is disabled by default.  If someone needs to sign the app, including the workflow, they just have to add `-Psign-app=true` in the call to gradle.

bdferris-v2:
To assist with testing, I'm moving development over to a branch on the main repo.  You can track that work in the associated pull request: #1250

PR #1250, opened 2022-09-02
fix: Sign and Notarize MacOS App

bdferris-v2:
**Summary:**

As tracked in #1200, this PR adds support for code-signing and notarizing the MacOS app.  This will mean the user no longer has to ⌘-click the app to open it the first time.  Note that they will still get the following warning when running the app for the first time:

Screen Shot 2022-08-23 at 7 34 14 PM

But they can just click "Ok" and the app will run.  I don't think there is any way to avoid that dialog without distributing the app through the Mac App Store.

Closing #1200.

### DIFF
--- a/.github/workflows/package_installers.yml
+++ b/.github/workflows/package_installers.yml
@@ -2,7 +2,7 @@ name: Package Installers
 
 on:
   push:
-    branches: [ master, issue/1200/sign-mac-os ]
+    branches: [ master ]
     paths-ignore:
       - '**.md'
   pull_request:
--- a/.github/workflows/package_installers.yml
+++ b/.github/workflows/package_installers.yml
@@ -60,8 +60,9 @@ jobs:
       - name: "Package GUI app installer with Gradle"
         uses: gradle/gradle-build-action@v2
         with:
-          arguments: jpackage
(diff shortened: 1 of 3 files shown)


bdferris-v2:
Ok, made a last few tweaks.  This is now ready for final review and submission.

.github/workflows/package_installers.yml
```diff
@@ -2,7 +2,7 @@ name: Package Installers
 
 on:
   push:
-    branches: [ master ]
+    branches: [ master, issue/1200/sign-mac-os ]
```
maximearmstrong, line 5: Do we still need `issue/1200/sign-mac-os` here? I assume it was for testing purposes, but I may be wrong?

maximearmstrong (commented):
Great work. Thanks so much @bdferris-v2! One small question before we merge.

.github/workflows/package_installers.yml
bdferris-v2: That was added for testing and is no longer needed.  I have removed it.

maximearmstrong (approved):
LGTM, thanks @bdferris-v2

PR #1255, opened 2022-09-16
fix: "Validate" button disappears after app font-size change

### LINKED ISSUE #1244: "Validate" button disappearing from Desktop app
**Describe the bug**
This is a bug from Marcy Jaffe at the National RTAP. They are interested in adding the validator to their toolkit now that there is a Desktop app, but they have run into a problem: after they have selected the input and output directories, the "Validate" button disappears (see screenshot below), which can create confusion or even block some users.
When she made the interface bigger, the button appeared again. 

**How we reproduce the bug**
1. Download desktop app
2. Select input and output directories

**Expected behaviour**
Click on "Validate" to run the validator

**Observed behaviour**
The Validate button is not appearing anymore on the interface. When making the interface bigger, it appears again and the validator could run properly.

**Screenshots:** 
Screen Shot 2022-08-24 at 6 32 06 AM

**Environment versions**
- validator version: 3.1.1
- OS versions: Windows

bdferris-v2:
Fixes #1244, where in the "Validate" button sometimes disappears.  Per discussion in the bug, when a `JFileChooser` is shown on some platforms, it causes the apps default font size to increase.  This ultimately causes the app's layout to grow vertically, which can push the validate button off the bottom of the window.

While we don't know the root cause of the font change, we can address the symptom.  Namely, we introduce a `pack()` call after any use of the `JFileChooser`, which will cause the app to resize to match the preferred size of its content.  Most of the time, there will be no change, but if the font size has increased, the app will resize appropriately.  We already use a similar approach when showing and hiding the advanced app settings.

To make the re-pack slightly less jarring, we also specify a fixed column count for the text input fields, such that the app does not dynamically resize in response to strings of different lengths.  Finally, we set a minimum size of the app based on initial content size to keep the app from shrinking beyond that preferred minimum.


isabelle-dr:
Thank you for working on this Brian 🙏
I have sent this to one of the users that had this problem to see if this works.

isabelle-dr:
Do you know why we aren't getting the acceptance test message?

bdferris-v2:
You only get those if you are developing on a branch in the MobilityData/gtfs-validator repo itself, due to limitations from GitHub.  I've historically done all my development in a forked repo (bdferris-v2/gtfs-validator), though you all did recently grant me access to create branches on the main repo.

See also discussion at https://github.com/MobilityData/gtfs-validator/issues/1232#issuecomment-1210981009

maximearmstrong (approved):
LGTM! Thanks @bdferris-v2

> I have sent the app from this PR to one of the users that had this problem to see if this works.

@isabelle-dr I'd say we can merge if it works for the user.

isabelle-dr:
I received feedback from one of the users that had this problem: this fix works. 

Thanks for this fix @bdferris-v2, merging this PR!

PR #1228, opened 2022-08-05
feat: Initial support for Fares v2 validation.

bdferris-v2:
Per issue #1201, adds validation for files and fields added for Fares v2 base spec changes.

Please make sure these boxes are checked before submitting your pull request - thanks!

- [x] Run the unit tests with `gradle test` to make sure you didn't break anything
- [x] Format the title like "feat: [new feature short description]". Title must follow the Conventional Commit Specification(https://www.conventionalcommits.org/en/v1.0.0/).
- [x] Linked all relevant issues

PR #1261, opened 2022-10-03
feat: Validation for conditionally required/forbidden `fare_rules.txt` file.

bdferris-v2:
**Summary:**

Per updates introduced in the Fares v2 spec addition, we now check for the conditionally required/forbidden `fare_rules.txt` file.

### DIFF
--- a/RULES.md
+++ b/RULES.md
@@ -46,7 +46,7 @@ Each Notice is associated with a severity: `INFO`, `WARNING`, `ERROR`.
 | [`fare_transfer_rule_invalid_transfer_count`](#fare_transfer_rule_invalid_transfer_count)                         | A row from GTFS file `fare_transfer_rules.txt` has a defined `transfer_count` with an invalid value.                                                   |
 | [`fare_transfer_rule_missing_transfer_count`](#fare_transfer_rule_missing_transfer_count)                         | A row from `fare_transfer_rules.txt` has `from_leg_group_id` equal to `to_leg_group_id`, but has no `transfer_count` specified.                        |
 | [`fare_transfer_rule_with_forbidden_transfer_count`](#fare_transfer_rule_with_forbidden_transfer_count)           | A row from `fare_transfer_rules.txt` has `from_leg_group_id` not equal to `to_leg_group_id`, but has `transfer_count` specified.                       |
-| [`forbidden_file`](#forbidden_file)                                                                               | A conditionally forbidden file is present in the feed.                                                                                                 |
+| [`forbidden_file`](#forbidden_file)                                                                               | A forbidden file is present in the feed.                                                                                                 |
 | [`foreign_key_violation`](#foreign_key_violation)                                                                 | Wrong foreign key.                                                                                                                                     |
 | [`inconsistent_agency_timezone`](#inconsistent_agency_timezone)                                                   | Inconsistent Timezone among agencies.                                                                                                                  |
 | [`invalid_color`](#invalid_color)                                                                                 | A field contains an invalid color value.                                                                                                               |
@@ -195,7 +195,7 @@ Trips with the same block id have overlapping stop times.
 
 ### forbidden_file
 
-A conditionally forbidden file is present in the feed.
+A forbidden file is present in the feed.
 
 #### References
 * [GTFS terms definition](https://gtfs.org/reference/static/#term-definitions)
--- a/RULES.md
+++ b/RULES.md
@@ -32,6 +32,7 @@ Each Notice is associated with a severity: `INFO`, `WARNING`, `ERROR`.
 | Notice code                                                                                                       | Description                                                                                                                                            |
 |-------------------------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------|
 | [`block_trips_with_overlapping_stop_times`](#block_trips_with_overlapping_stop_times)                             | Block trips with overlapping stop times.                                                                                                               |
+| [`conditionally_forbidden_file`](#conditionally_forbidden_file)                                                   | A conditionally forbidden file is present in the feed.                                                                                                 |
 | [`csv_parsing_failed`](#csv_parsing_failed)                                                                       | Parsing of a CSV file failed.                                                                                                                          |
 | [`decreasing_shape_distance`](#decreasing_shape_distance)                                                         | Decreasing `shape_dist_traveled` in `shapes.txt`.                                                                                                      |
 | [`decreasing_or_equal_stop_time_distance`](#decreasing_or_equal_stop_time_distance)                               | Decreasing or equal `shape_dist_traveled` in `stop_times.txt`.                                                                                         |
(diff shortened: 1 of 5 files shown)


bdferris-v2:
@isabelle-dr per discussion, here is a separate PR for the Fares v1 file checks.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/FaresV1FileValidator.java
```diff
@@ -0,0 +1,33 @@
...
+      GtfsFareRuleTableContainer fareRuleTable) {
+    this.fareAttributeTable = fareAttributeTable;
+    this.fareRuleTable = fareRuleTable;
+  }
+
+  @Override
+  public void validate(NoticeContainer noticeContainer) {
+    if (!fareAttributeTable.isMissingFile() && fareRuleTable.isMissingFile()) {
+      noticeContainer.addValidationNotice(
+          new MissingRequiredFileNotice(fareRuleTable.gtfsFilename()));
```
bdferris-v2, line 27: @isabelle-dr I reused `MissingRequiredFileNotice` here, but wondering if I should introduce a new `MissingConditionallyRequiredFileNotice` to match `ConditionallyForbiddenFileNotice`?

isabelle-dr: Interesting! The most coherent from my perspective would be to do either:
1. Use `MissingRequiredFileNotice` and `FordiddenFileNotice` for everything. This would mean introducing `FordiddenFileNotice` in this PR instead.
2. Use `MissingRequiredFileNotice`, `MissingConditionallyRequiredFileNotice`, `FordiddenFileNotice`, `ConditionallyForbiddenFileNotice`. This would mean introducing the new `MissingConditionallyRequiredFileNotice` as you mentioned.

From a user perspective: there are often "conditions", even for fields that are tagged as Required (for example, `area_id` is required if the file is present).
If I had `missing_conditionally_required_file` and `missing_required_file` in the same validation report, I would be confused. Same thing for forbidden files, and for fields.

On another note, we have a different approach for **calendar.txt** and **calendar_dates.txt**: we have a custom rule ([`missing_calendar_and_calendar_date_files`](https://github.com/MobilityData/gtfs-validator/blob/master/RULES.md#missing_calendar_and_calendar_date_files)), which I think is the easiest for the user because we can provide the documentation they need & they don't have to go in the spec.

For these reasons, I would say first approach, or custom rules like I described [here](https://github.com/MobilityData/gtfs-validator/pull/1228#pullrequestreview-1116059614).

KClough (approved):
LGTM

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/FaresV1FileValidator.java
bdferris-v2: Ok, that all sounds reasonable to me.  For now, I went with option #1.  I agree that custom notices offer the most flexibility for linking to documentation, etc, but I'm wondering if there are ways we can get similar functionality without so many custom classes.  Or if nothing else, introduce something for automatically generating RULES.md from code, since updating it each time is tedious / error-prone :)

RULES.md
```diff
@@ -41,6 +41,7 @@ Each Notice is associated with a severity: `INFO`, `WARNING`, `ERROR`.
 | [`empty_column_name`](#empty_column_name)                                                                         | A column name is empty.                                                                                                                                |
 | [`empty_file`](#empty_file)                                                                                       | A CSV file is empty.                                                                                                                                   |
 | [`equal_shape_distance_diff_coordinates`](#equal_shape_distance_diff_coordinates)                                 | Two consecutive points have equal `shape_dist_traveled` and different lat/lon coordinates in `shapes.txt`.                                             |
+| [`forbidden_file`](#forbidden_file)                                                                               | A conditionally forbidden file is present in the feed.                                                                                                 |
```
isabelle-dr, line 44:
```suggestion
| [`forbidden_file`](#forbidden_file) | A forbidden file is present in the feed. |
```

RULES.md
```diff
@@ -184,6 +185,26 @@ Trips with the same block id have overlapping stop times.
 
 </details>
 
+<a name="ForbiddenFileNotice"/>
+
+### forbidden_file
+
+A conditionally forbidden file is present in the feed.
```
isabelle-dr, line 192:
```suggestion
A forbidden file is present in the feed.
```

isabelle-dr (approved):
2 minor changes in line. 
Can you resolve the merge conflicts? 
Then, I'll merge this PR!

RULES.md
bdferris-v2: Done.

bdferris-v2:
Pasting this in manually:

❌ Invalid acceptance test.
New Errors: 62 out of 1363 datasets (~5%) are invalid due to code change, which is above the provided threshold of 1%.
Dropped Errors: 0 out of 1363 datasets (~0%) are invalid due to code change, which is less than the provided threshold of 1%.
1 out of 1364 sources (~0 %) are corrupted.
Corrupted sources:
de-unknown-ulmer-eisenbahnfreunde-gtfs-1081
Commit: e0746f56cf5e668dac70f0594c561364a706a77b
Download the full acceptance test report [here](https://github.com/MobilityData/gtfs-validator/actions/runs/3184081737) (report will disappear after 90 days).
❌ Invalid acceptance test.

I was curious what this was going to trigger.  Taking a closer look.

isabelle-dr:
Isn't having `fare_rules.txt` going from Optional to "Required if `fare_attributes.txt ` is defined (which was a pre-existing file), a breaking change?
I'm not surprised that we have some existing datasets are becoming invalid.

bdferris-v2:
I'm a little confused by the change to the spec as well.

Looking back at the original Fares V1 example documentation (not really hosted anywhere anymore, but available at https://web.archive.org/web/20111207224351/https://code.google.com/p/googletransitdatafeed/wiki/FareExamples), having a `fare_attributes.txt` file with a single entry and no `fare_rules.txt` file was a supported use-case for systems that have a single base fare.  This was actually brought up in a [comment](https://github.com/google/transit/pull/286#issuecomment-1082480847) by @flocsy but I think his feedback was ignored?

And indeed, when I look through the acceptance report for this change, there are 62 feeds in the MobilityDatabase that have a `fare_attributes.txt` file but no `fare_rules.txt` file.  Of those, 39 have a single entry in `fare_attributes.txt`, indicating a default fare. 

That said, there are also 11 feeds where `fare_attributes.txt` has no entires (but still has a CSV row header).  I'm not sure how that should be interpreted?  A free fare?  There are additionally 10 feeds that specify multiple `fare_attributes.txt` entries where `fare_rules.txt` looks like should legitimately be specified.

I'm tempted to say "pause" on this PR and get some feedback from the #gtfs-fares group before proceeding.

bdferris-v2:
Ok, based off the [discussion in the slack channel](https://mobilitydata-io.slack.com/archives/C3FFFKX9C/p1665538998116809?thread_ts=1665087714.843639&cid=C3FFFKX9C), I've implemented the following validation logic:

* the following is flagged as an ERROR based on the spec
    * `stop_ids`: if the transfer type is 4 or 5, stations must not be used
    * `trip_ids`: required if the transfer type is 4 or 5
* the following is flagged as a WARNING so that our users can get the info
    * if the transfer type is 4 or 5, the stops linked must be the first & last of a trip in the documentation for this issue, we explain that if the intention is the use-case is mid-trip in-seat transfers, the warning can be ignored.

As part of this, I've also introduced a TransferDirection helper class that simplified some of the logic of dealing with continual `from_X` and `to_X` checks.  I updated some existing transfers.txt validation to use that logic as well.

@KClough and @isabelle-dr this is ready for another look

PR #1263, opened 2022-10-04
feat: Check validator subclasses are annotated with @GtfsValidator

### LINKED ISSUE #1262: Add unit-test to ensure all validator subclasses are annotated with @GtfsValidator
### Describe the problem

When implementing a new validation rule, I forgot to annotate my validator with the `@GtfsValidator` annotation.  This meant it wasn't picked up automatically by `ValidatorLoader` and not run as part of the default validation set, even though individual unit-tests for this validator ran successfully.

### Proposed solution

Add a unit-test that checks that any subclass of `SingleEntityValidator` or `FileValidator` is also annotated with `@GtfsValidator`.

We could alternatively rewrite `ValidatorLoader` to ignore the `@GtfsValidator` annotation and instead scan for subclasses directly.  I'll admit I wasn't around for the original design choice, but I could see an argument for having a single annotation type across all validators, regardless of subclass, especially if the validator subclasses grows in the future.  Either way, given current staffing on the project, I'm in favor of keeping the status quo for now.

### Additional context

_No response_

bdferris-v2:
Closes #1262

Add a unit-test that checks that any subclass of SingleEntityValidator or FileValidator is also annotated with @GtfsValidator.

### DIFF
--- a/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/GtfsValidatorAnnotationTest.java
+++ b/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/GtfsValidatorAnnotationTest.java
@@ -1,0 +1,49 @@
+package org.mobilitydata.gtfsvalidator.validator;
+
+import static com.google.common.truth.Truth.assertWithMessage;
+
+import com.google.common.collect.ImmutableList;
+import io.github.classgraph.AnnotationInfo;
+import io.github.classgraph.ClassGraph;
+import io.github.classgraph.ClassInfo;
+import io.github.classgraph.ScanResult;
+import org.junit.Test;
+import org.mobilitydata.gtfsvalidator.annotation.GtfsValidator;
+
+public class GtfsValidatorAnnotationTest {
+
+  /**
+   * This test ensures that any class that extends {@link SingleEntityValidator} or {@link
+   * FileValidator} is also annotated with {@link GtfsValidator}. Not including the annotation
+   * typically indicates an error, as the validator won't be picked up by scanning code in {@link
+   * ValidatorLoader}.
+   */
+  @Test
+  public void testAllValidatorSubclassesAreAnnotatedWithGtfsValidator() {
+    ImmutableList<String> validatorPackages =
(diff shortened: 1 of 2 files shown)


main/src/test/java/org/mobilitydata/gtfsvalidator/validator/GtfsValidatorAnnotationTest.java
```diff
@@ -0,0 +1,49 @@
...
+              .scan()) {
+        for (Class<?> validatorBaseClass : validatorBaseClasses) {
+          for (ClassInfo classInfo : scanResult.getSubclasses(validatorBaseClass)) {
+            AnnotationInfo annotationInfo = classInfo.getAnnotationInfo(GtfsValidator.class);
+            assertWithMessage(
+                    classInfo.getName()
+                        + ": "
+                        + validatorBaseClass.getSimpleName()
+                        + " validator subclass should be annotated with @GtfsValidator")
+                .that(annotationInfo != null)
```
asvechnikov2, line 42: you can probably use 
.that(annotationInfo).isNotNull();

asvechnikov2 (approved):
LGTM! Thanks!

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/GtfsValidatorAnnotationTest.java
bdferris-v2: I originally wrote it this way, but the resulting error message from Truth isn't particular helpful / readable in the context of this check.  With the isTrue() check, we get something like:

```
SometValidator: FileValidator validator subclass should be annotated with @GtfsValidator
expected to be true
```

With the alternate syntax, we get:

```
SometValidator: FileValidator validator subclass should be annotated with @GtfsValidator
expected not to be: null
```
I think the former is more understandable than the latter.

asvechnikov2: Thanks!

asvechnikov2 (approved):

PR #1273, opened 2022-10-14
Use smaller integer field types for GTFS entity classes

aababilov:
GTFS enums usually fit into one byte, so we don't need an int. GTFS tables usually less than 16 fields, so we need a byte or a short int for storing a bitmask of assigned fields.

Smaller fields allow us to save memory for the largest table - stop_times.txt that may have 30 M lines.

Instead of:
```
class GtfsStopTime {
  // ...
  private int pickupType;
  private int dropOffType;
  private int continuousPickup;
  private int continuousDropOff;
  private int timepoint;
  private int bitField0_;
}
```
we have now:
```
class GtfsStopTime {
  // ...
  private byte pickupType;
  private byte dropOffType;
  private byte continuousPickup;
  private byte continuousDropOff;
  private byte timepoint;
  private short bitField0_;
}
```
which is 5 * (4 - 1) + 2 = 17 bytes smaller. Java aligns classes by 8 bytes, so we actually save 16 bytes per line.

Total save: 0.5 GiB for 30 M lines in stop_times.txt

### DIFF
--- a/processor/src/main/java/org/mobilitydata/gtfsvalidator/processor/EntityImplementationGenerator.java
+++ b/processor/src/main/java/org/mobilitydata/gtfsvalidator/processor/EntityImplementationGenerator.java
@@ -26,6 +26,7 @@ import static org.mobilitydata.gtfsvalidator.processor.FieldNameConverter.hasMet
 import static org.mobilitydata.gtfsvalidator.processor.FieldNameConverter.setterMethodName;
 import static org.mobilitydata.gtfsvalidator.processor.GtfsEntityClasses.TABLE_PACKAGE_NAME;
 
+import com.google.common.collect.ImmutableMap;
 import com.google.common.geometry.S2LatLng;
 import com.squareup.javapoet.CodeBlock;
 import com.squareup.javapoet.FieldSpec;
@@ -36,9 +37,11 @@ import com.squareup.javapoet.TypeName;
 import com.squareup.javapoet.TypeSpec;
 import java.math.BigDecimal;
 import java.time.ZoneId;
+import java.util.List;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import javax.lang.model.element.Modifier;
+import javax.lang.model.type.DeclaredType;
 import javax.lang.model.type.TypeMirror;
 import org.mobilitydata.gtfsvalidator.annotation.FieldTypeEnum;
 import org.mobilitydata.gtfsvalidator.annotation.Generated;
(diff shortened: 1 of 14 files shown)


processor/src/main/java/org/mobilitydata/gtfsvalidator/processor/EntityImplementationGenerator.java
```diff
@@ -59,14 +62,44 @@ private enum ClassContext {
...
+    }
+    if (minValue >= Short.MIN_VALUE && maxValue <= Short.MAX_VALUE) {
+      return TypeName.SHORT;
+    }
+    return TypeName.INT;
   }
 
   private static int lastBitFieldNumber(int fieldCount) {
-    // Each bitField has 32 bits. We need bitField0_ to store 1..32 fields,
+    // All standard GTFS tables have less than 16 fields, so a bitmask fits into short.
```
asvechnikov2, line 100: The comment "All standard GTFS" will become obsolete relatively soon, what do you think about replacing it with "Most standard GTFS"?

processor/tests/src/test/java/org/mobilitydata/gtfsvalidator/processor/tests/EnumSizesSchemaTest.java
```diff
@@ -0,0 +1,92 @@
...
+    assertThat(enumSizes.byteEnum()).isEqualTo(ByteGtfs.UNRECOGNIZED);
+    // However, 10 can be still retrieved using byteEnumValue method.
+    assertThat(enumSizes.byteEnumValue()).isEqualTo(10);
+  }
+
+  @Test
+  public void testByteEnumOutOfBounds() {
+    // Short.MIN_VALUE is out of bounds for byte. It cannot be retrieved with byteEnumValue.
+    assertThat(new EnumSizes.Builder().setByteEnum((int) Short.MIN_VALUE).build().byteEnumValue())
+        .isEqualTo(0);
```
asvechnikov2, line 39: Should we throw any exception if values can't be retrieved? It looks like an undefined behaviour

processor/tests/src/main/java/org/mobilitydata/gtfsvalidator/processor/tests/ManyFieldsSchema.java
```diff
@@ -0,0 +1,26 @@
+package org.mobilitydata.gtfsvalidator.processor.tests;
+
+import org.mobilitydata.gtfsvalidator.annotation.GtfsTable;
+
+@GtfsTable("many_fields.txt")
+public interface ManyFieldsSchema {
```
asvechnikov2, line 6: Do you think it's worth testing an edge case when the number of fields is equal to the number of bits and the last field is handled correctly?

asvechnikov2 (commented):
Thanks for the tests!

f8full:
Nice

processor/src/main/java/org/mobilitydata/gtfsvalidator/processor/EntityImplementationGenerator.java
aababilov: Done.

processor/tests/src/test/java/org/mobilitydata/gtfsvalidator/processor/tests/EnumSizesSchemaTest.java
aababilov: We generate UnexpectedEnumValueNotice and store UNRECOGNIZED.

processor/tests/src/main/java/org/mobilitydata/gtfsvalidator/processor/tests/ManyFieldsSchema.java
aababilov: Changed to 8 fields.

aababilov:
> Nice

Thanks!

aababilov:
> wave Welcome back!

I missed GTFS Validator :)

asvechnikov2 (approved):
LGTM! Thanks!

PR #1274, opened 2022-10-14
Move GTFS field names from table loader to entity classes

aababilov:
E.g., now use `GtfsStop.STOP_ID_FIELD_NAME` instead of `GtfsStopTableLoader.STOP_ID_FIELD_NAME`.

Motivation

1. Table loader classes should be only used for loading data tables. They should not be used by validator classes.
2. Our final goal is to stop generating table loader classes and instead make them usual classes that use a GTFS schema as a parameter. GTFS field names have to be generated, so we move them to entity classes that are generated anyway.


asvechnikov2 (approved):
LGTM! Thanks!

aababilov:
Thanks for reviewing!

PR #1289, opened 2022-11-07
feat: Date validity - Calendar Expiration + Trip Coverage

### LINKED ISSUE #886: Best Practice: Date Validity (WARNING)
Per the best practices:
- Dataset should be valid for at least the next seven days [ from publication date? ]
- Dataset shouldn't include date ranges that have already ended [ No expired calendars ]

KClough:
**Summary:**

Resolves #886 

**Expected behavior:** 

**Note**: Marking as draft until I complete unit tests for the new validator.


main/src/main/java/org/mobilitydata/gtfsvalidator/validator/CalendarValidator.java
```diff
@@ -0,0 +1,58 @@
...
+
+  @Override
+  public void validate(NoticeContainer noticeContainer) {
+    LocalDate now = currentDateTime.getNow().toLocalDate();
+    GtfsDate currentDate = GtfsDate.fromLocalDate(now);
+
+    for (var calendar : calendarTable.getEntities()) {
+      if (calendar.endDate().isBefore(currentDate)) {
+        noticeContainer.addValidationNotice(
+            new CalendarValidator.CalendarValidatorExpiredCalendarNotice());
```
bdferris-v2, line 48: nit: I think you can remove the "CalendarValidator." prefix here?

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/CalendarValidator.java
```diff
@@ -0,0 +1,58 @@
...
+import org.mobilitydata.gtfsvalidator.annotation.GtfsValidator;
+import org.mobilitydata.gtfsvalidator.input.CurrentDateTime;
+import org.mobilitydata.gtfsvalidator.notice.NoticeContainer;
+import org.mobilitydata.gtfsvalidator.notice.SeverityLevel;
+import org.mobilitydata.gtfsvalidator.notice.ValidationNotice;
+import org.mobilitydata.gtfsvalidator.table.*;
+import org.mobilitydata.gtfsvalidator.type.GtfsDate;
+
+@GtfsValidator
+public class CalendarValidator extends FileValidator {
```
bdferris-v2, line 30: Maybe better to call this `ExpiredCalendarValidator`?

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/CalendarValidator.java
```diff
@@ -0,0 +1,58 @@
...
+    this.calendarTable = calendarTable;
+  }
+
+  @Override
+  public void validate(NoticeContainer noticeContainer) {
+    LocalDate now = currentDateTime.getNow().toLocalDate();
+    GtfsDate currentDate = GtfsDate.fromLocalDate(now);
+
+    for (var calendar : calendarTable.getEntities()) {
+      if (calendar.endDate().isBefore(currentDate)) {
```
bdferris-v2, line 46: While this technically works for simple cases, it might miss a case where a calendars.txt entry appears expired, but it is actually extended a few days by calendar_dates.txt.  Are you worried about this case?  An alternative approach would be to look at the combined dates in `DateTripsValidator` below.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/CalendarValidator.java
```diff
@@ -0,0 +1,58 @@
...
+        noticeContainer.addValidationNotice(
+            new CalendarValidator.CalendarValidatorExpiredCalendarNotice());
+      }
+    }
+  }
+
+  static class CalendarValidatorExpiredCalendarNotice extends ValidationNotice {
```
bdferris-v2, line 53: I think this should include the csvRowNumber and service id for the expired calendar?

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DateTripsValidator.java
```diff
@@ -0,0 +1,163 @@
...
+    GtfsDate maxEndDate = null;
+
+    GtfsDate serviceWindowStartDate = null;
+    GtfsDate serviceWindowEndDate = null;
+
+    LocalDate now = currentDateTime.getNow().toLocalDate();
+    GtfsDate currentDate = GtfsDate.fromLocalDate(now);
+
+    // Attempt to get Start Date and End Date from feedInfo
+    for (GtfsFeedInfo entity : this.feedInfoTable.getEntities()) {
```
bdferris-v2, line 74: feedInfoTable has a `getSingleEntity` method that might be more appropriate to use?

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/CalendarValidator.java
briandonahue: @bdferris-v2 @KClough I updated this to use the same service period mapping logic as `DateTripsValidator` and then just check the last service date per service Id.  I updated the tests to add a couple calendar date exception tests.
```
    final Map<String, SortedSet<LocalDate>> servicePeriodMap =
        CalendarUtil.servicePeriodToServiceDatesMap(
            CalendarUtil.buildServicePeriodMap(calendarTable, calendarDateTable));
```

bdferris-v2:
Apologies for the slow delay.  So looking at the logic for trip counting, I realize there is some additional complexity around handling things like frequency-based trips that the current code, as written, doesn't cover.

Digging into this on my side, I've gotten the go-ahead to open-source some utility code we have at Google that might help with this.  @briandonahue if it's ok with you, can I send you a PR with that code and talk about how it might simplify this PR?  Will send shortly.

bdferris-v2:
@briandonahue that support library is available in https://github.com/MobilityData/gtfs-validator/pull/1351

Namely, TripCalendarUtil provides methods for computing service coverage windows from daily trip counts, including support for frequency-based trips.  I think if we get that checked in, it can reduce the logic in your PR substantially.  Apologies for not bringing this up sooner, I had to unblock some things on my end first.

briandonahue:
@bdferris-v2 this has now been updated to use `TripCalendarUtil`. Thanks for that!

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DateTripsValidator.java
```diff
@@ -0,0 +1,109 @@
...
+    LocalDate now = currentDateTime.getNow().toLocalDate();
+
+    final Map<String, SortedSet<LocalDate>> servicePeriodMap =
+        CalendarUtil.servicePeriodToServiceDatesMap(
+            CalendarUtil.buildServicePeriodMap(calendarTable, calendarDateTable));
+
+    var tripCounts =
+        TripCalendarUtil.countTripsForEachServiceDate(
+            servicePeriodMap, tripContainer, frequencyTable);
+    var majorityServiceDates = TripCalendarUtil.computeMajorityServiceCoverage(tripCounts);
```
bdferris-v2, line 79: I think I'd prefer that `var` not be used here, as the return type is not obvious from context alone.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DateTripsValidator.java
```diff
@@ -0,0 +1,109 @@
...
+    LocalDate currentDatePlusSevenDays = now.plusDays(7);
+
+    if (majorityServiceDates != null) {
```
bdferris-v2, line 82: `majorityServiceDates` cannot be null, but it can be `Optional.empty`.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DateTripsValidator.java
```diff
@@ -0,0 +1,109 @@
...
+                GtfsDate.fromLocalDate(serviceWindowStartDate),
+                GtfsDate.fromLocalDate(serviceWindowEndDate)));
+      }
+    }
+  }
+
+  static class TripDataShouldBeValidForNext7DaysNotice extends ValidationNotice {
+    private final GtfsDate currentDate;
+    private final GtfsDate serviceWindowStartDate;
+    private final GtfsDate serviceWindowEndDate;
```
bdferris-v2, line 99: Could you add a comment to each field here to match the comments in RULES.md?  It would make my life easier regarding PR #1346.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DateTripsValidator.java
briandonahue: Gotcha, it's the C# developer in me.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DateTripsValidator.java
briandonahue: Good catch - my mistake.

bdferris-v2 (approved):

isabelle-dr (changes requested):
Thank you for the work here @briandonahue, and thank you for the review @bdferris-v2!
To follow our naming conventions, could you rename `trip_data_should_be_valid_for_next7_days` with `trip_expiration7_days`?

Also: what conditions are triggering the `expired_calendar` notice exactly? Do you have a test dataset for it?

bdferris-v2:
Naming conventions aside, I'm not sure `trip_expiration7_days` is a particularly good name for what this notice is validating. 
 Specifically, it's not trips that are expiring (whatever that might mean).  Maybe something like `majority_of_trip_coverage_not_active_for_next7_days` would be more descriptive and still match naming conventions?

briandonahue:
> @briandonahue what I meant is more: what is the pseudo-code here?

@isabelle-dr The `TripCalendarUtil` was imported by @bdferris-v2 from Google projects, as far as I understand,  and merged in #1351 so that I could use it here. The specific method being used to calculate "Majority Trip Service Coverage" is [here](https://github.com/MobilityData/gtfs-validator/blob/master/main/src/main/java/org/mobilitydata/gtfsvalidator/util/TripCalendarUtil.java#L87).  

I can try to come up with an English translation for the documentation - it looks like it looks for consecutive days with > 75% of trips to calculate the service window.  Open to suggestion for wording or any existing wording perhaps used by Google, @bdferris-v2?

isabelle-dr (commented):
Hello @briandonahue I've added a suggestion in the doc to be a little more descriptive about how this role works. 
Besides this small change, we are ready to merge!

briandonahue:
> @briandonahue can you update this branch with master? I will merge after

@isabelle-dr sorry, lost track of this!  Done!

PR #1297, opened 2022-11-21
fix: Drop DuplicateFareRuleZoneIdFieldsValidator and replace with multi-column @PrimaryKey for fare_rules.txt

### LINKED ISSUE #1296: False positives for duplicate_fare_rule_zone_id_fields in fare_rules.txt
### Describe the bug

This was flagged by [SDMTS](https://www.sdmts.com/).

False positives for the `duplicate_fare_rule_zone_id_fields` rule.
They only have (fare_id, route_id) fields defined, every row has a unique combination of both. 
The spec mentions that the primary key for `fare_rules.txt` is (*) which means: all fields ([reference](https://gtfs.org/schedule/reference/#fare_rulestxt))

They had `duplicate_fare_rule_zone_id_fields` triggered (which is an ERROR) and it looks like a problem with the way this rule is designed.

### Steps/Code to Reproduce

- download dataset provided
- run validator
- open report

### Expected Results

No issue with fare rules

### Actual Results

The notice `duplicate_fare_rule_zone_id_fields` is triggered 296 times.

### Screenshots

_No response_

### Files used

[https://www.sdmts.com/google_transit_files/google_transit.zip](https://linkprotect.cudasvc.com/url?a=https%3a%2f%2fwww.sdmts.com%2fgoogle_transit_files%2fgoogle_transit.zip&c=E,1,aLw0uc4wTOB7_PACNy8odDh4yHvww2XSr9wuZRtTptBblwDIYbfOczhpt_odyKlSWUxr213GdNd5ygJkLHuWd8sfM8uFor_7N_R2OJheTccuhreKAueT2vQ,&typo=1)

### Validator version

4.0.0

### Operating system

MacOS

### Java version

_No response_

### Additional notes

Now that we support multiple `@PrimaryKey` annotations, we don't need to have the [duplicate_fare_rule_zone_id_fields](https://github.com/MobilityData/gtfs-validator/blob/master/RULES.md#duplicate_fare_rule_zone_id_fields) rule in this validator at all. 
We can use the `@PrimaryKey` annotation in `fare_rules.txt` directly, and the notice [duplicate_key](https://github.com/MobilityData/gtfs-validator/blob/master/RULES.md#duplicate_key) will be triggered if we have duplicate records.

bdferris-v2:
Closes #1296.

Per discussion in #1296, `DuplicateFareRuleZoneIdFieldsValidator` says the combination of `route_id`, `origin_id`, `destination_id`, and `contains_id` must be unique in `fare_rules.txt`, when the spec says `fare_id` should be included (aka the primary key is all fields).  This PR drops the custom validator in favor of proper @PrimaryKey annotations for `fare_rules.txt`.

Per the acceptance tests, there are a number of feeds (~60) that trigger the existing`duplicate_fare_rule_zone_id_fields` notice.  Spot checking a few feeds, many of them are indeed specifying the same combination of fare rules with different fare ids.  The corresponding fare attributes often have a different price, probably indicating that the producer is attempting to model some other aspect of their fare system (different fare products?  different rider categories?) despite the fact that this is not really supported in Fares v1.  While it's maybe not obvious how feed consumers should productively use this data, other than showing the cheapest fare, the spec technically allows it, so the validator should as well.

The acceptance test shows that a number of feeds (~10) newly trigger a `duplicate_key` notice.  These are all feeds that really due have a duplicate combination of all fields in `fare_rules.txt`.  However, each of these feeds were already triggering  a `duplicate_fare_rule_zone_id_fields` notice, so there are no newly-invalidated feeds from this change.


isabelle-dr:
Thanks for working on this @bdferris-v2!
This makes sense.
Following our [process](https://github.com/MobilityData/gtfs-validator/blob/master/docs/ACCEPTANCE_TESTS.md#what-do-we-do-with-the-results) for the acceptance test results, I would consider that this change isn't generating additional errors for the reasons you just explained.

isabelle-dr (changes requested):
Can you remove the outdated rule from the documentation, please? 
Then we are good to merge this PR :)

bdferris-v2:
RULES.md has been updated.

isabelle-dr (approved):
Thanks! I think we are good to merge without an additional code review here.

PR #1300, opened 2022-12-14
Do not invoke single file validators if table indices were not generated

aababilov:
If a table has duplicated keys, then its indices won't be generated and the noticeContainer will contain errors. Single file validators may have undefined behaviour when called for such incomplete indices.


asvechnikov2 (approved):
Thanks!

bdferris-v2:
I gotta say, I have thoughts on this PR.  I hear @aababilov 's concern, but this seems a heavy-handed way of addressing it.  As implemented, a single-entity validation error will caused all single-file validators to skip execution, potentially missing out on a lot of useful validation.  I'd rather write validators to be resilient to underlying data that is malformed than disable them entirely.

What's more, multi-file validation rules are still run for feeds that potentially have duplicate key and foreign key validation errors as well.  Do we have to disable those as well?

aababilov:
Makes sense to me, Brian. Reverted in https://github.com/MobilityData/gtfs-validator/pull/1303

PR #1302, opened 2022-12-14
fix: byTranslationKey() for multi-key schemas without translations.txt support

bdferris-v2:
The `@PrimaryKey(translationRecordIdType)` field is used to indicated which keys, if any, for a given GTFS entity can be used for performing `translations.txt` lookups via `byTranslationKey()` methods generated in the table container.  Currently, code in `TableContainerIndexGenerator` will generate an implementation for `byTranslationKey()` even if none of the @PrimaryKey fields have been marked as the translation RECORD_ID.  This can lead to some weird behavior where the method will return a value even if none should be returned or generate code that does not actually compile.

This PR fixes this behavior by no longer generating a `byTranslationKey()` implementation if the entity type does not support.  Instead, empty is returned.


asvechnikov2 (approved):
LGTM!

PR #1311, opened 2023-01-12
Discover tables and validators in runtime using io.github.classgraph

aababilov:
This effectively reverts commit 7fb105e2b759d21b638badbc312108bec9f43b15 where registries of tables and validators were generated by the GtfsAnnotationProcessor. It is impossible to reliably generate a registry for multiple classes since there is an orthogonality rule for Java Annotation Processors [1]:

“The result of processing a given input is not a function of the presence or absence of other inputs (orthogonality).”

In other words, there is no way to guarantee that GtfsAnnotationProcessor receives all schema and validator classes: it may receive only a portion of them, so it cannot generate a global registry.

[1] https://docs.oracle.com/javase/7/docs/api/javax/annotation/processing/Processor.html


asvechnikov2 (approved):
LGTM!

core/src/main/java/org/mobilitydata/gtfsvalidator/table/ClassGraphTableDiscoverer.java
```diff
@@ -0,0 +1,46 @@
...
+import com.google.common.collect.ImmutableList;
+import com.google.common.flogger.FluentLogger;
+import io.github.classgraph.ClassGraph;
+import io.github.classgraph.ClassInfo;
+import io.github.classgraph.ScanResult;
+
+/*
+Discovers GtfsTableDescriptor subclasses in the default table package.
+ */
+public class ClassGraphTableDiscoverer {
```
bdferris-v2, line 12: I'm in favor of combing all the classpath discovery code into one impl class, combing Table and Validator scanning in one place.  I'd also move existing Notice discovery code here as well.  That way, all ClassGraph-based discovery code would be in one place, where it's easier to make sure it's being used in a consistent way.  I'd propose calling the combined class ClassGraphDiscovery or something and putting it in the validator/ subpackage.  Thoughts?

aababilov: Good idea, done.

Notice discovery code will be moved to the new ClassGraphDiscovery class in a separate commit.

PR #1315, opened 2023-01-19
Add a list of validators that meet the official GTFS requirements

aababilov:
* `BlockTripsWithConsistentTypeValidator`
  * Validates that trips that belong to the same block have consistent route types (e.g., bus does not transfer to rail).
  * Generated notice: `BlockTripsWithInconsistentRouteTypesNotice`.
* `FareAttributeAgencyIdValidator`
  * Checks that `fare_attributes.agency_id` field is defined for every row if there is more than 1 agency in the feed.
  * Generated notice: `MissingRequiredFieldNotice`.
* `FeedHasLanguageValidator`
  * Checks that a feed has language defined in either `feed_info.feed_lang` or `agency.agency_lang`. Note that `feed_lang` is a required field: if there is a `feed_info` entity, it must have `feed_lang` set. `agency_lang`, instead, is optional. The recommended way is to provide `feed_lang` and omit `agency_lang`.
  * Generated notice: `FeedHasNoLanguageNotice`.
* `StationUsageValidator`
  * Checks that every station (`location_type=1`) in `stops.txt` has child platforms (` location_type=0`). Note that a station that has child entrances (`location_type=2`) or generic nodes (`location_type=3`) is also reported.
  * Generated notice: `StationWithoutPlatformsNotice`.
* `StopLatLngRequiredValidator`
  * Checks that conditionally required fields `stop_lat` and `stop_lon` are set depending on `location_type`.
    *  `stop_lat` and `stop_lon` are required for stations (`location_type=0`), stops (`location_type=1`) and entrances (`location_type=2`) and optional for other types.
  * Generated notice: `MissingRequiredFieldError`.
* `StopTimeInSequenceOrderValidator`
  * Validates that fields `arrival_time`, `departure_time` and `shape_dist_traveled` are non-decreasing along each trip.
  * Generated notices:
    * `TripWithOutOfOrderArrivalTimeNotice`
    * `TripWithOutOfOrderDepartureTimeNotice`
    * `TripWithOutOfOrderShapeDistTraveledNotice`
* `TransferFrequencyBasedTripValidator`
  * Validates that trip-to-trip transfers do not involve frequency-based trips. 
  * Generated notice: `TransferForFrequencyBasedTripNotice`.
* `TripShapeDistTraveledValidator`
  * Validates that trips specify `stop_times.shape_dist_traveled` values properly. Please see generated notices for details.
  * Generated notices:
    * `TripWithShapeDistTraveledButNoShapeNotice`
[...]

PR #1318, opened 2023-01-24
fix 878: require agency_id when more than one agency, warn when only one

### LINKED ISSUE #878: Best Practice: Include agency_id in routes.txt (WARNING)
Including `agency_id` in `routes.txt` is a best practice and should generate a warning even though it is not required if there is only one. 

Implementatin thoughts:
Rather than returning here in this override, generate a WARNING.

https://github.com/MobilityData/gtfs-validator/blob/52434454b747bb4085038f27563e98fee6018ca5/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TripAgencyIdValidator.java#L46-L59

briandonahue:
**Summary:**

Closes #878

1. Added test for existing validation requiring `agency_id` if more than one agency in `agency.txt`
1. generate a **warning** if there is a single agency in `agency.txt` but `agency_id` is not provided
1. generate a **warning** if `agency_id` is missing in `routes.txt` and a single agency is defined in `agency.txt` with no `agency_id`
1. generate an **warning** if `agency_id` is missing in `routes.txt` but is included for single agency in `agency.txt`

**Expected behavior:** 

N/A

- [ ] Include screenshot(s) showing how this pull request works and fixes the issue(s) (N/A)


isabelle-dr:
See my latest comment [here](https://github.com/MobilityData/gtfs-validator/issues/878#issuecomment-1403821346) in case that helps :)

briandonahue:
@isabelle-dr I think this is ready for review. I've also commented on #1315 where I think there may be some redundancy with the `agency_id` check for `fare_attributes.txt`. This PR should address the conditional required & recommended notices as opposed to just the required notice.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/AgencyConsistencyValidator.java
```diff
@@ -52,6 +50,14 @@ public class AgencyConsistencyValidator extends FileValidator {
   public void validate(NoticeContainer noticeContainer) {
     final int agencyCount = agencyTable.entityCount();
     if (agencyCount < 2) {
+      GtfsAgency agency = agencyTable.getEntities().get(0);
```
bdferris-v2, line 53: Technically, there could be zero agencies.  I'd be defensive and iterate over the entries instead.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/RouteAgencyIdValidator.java
```diff
@@ -18,43 +18,50 @@
...
- * agency in the feed.
+ * agency in the feed, recommended if only 1 agency.
  *
  * <p>Generated notice: {@link MissingRequiredFieldNotice}.
+ *
+ * <p>Generated notice: {@link MissingRecommendedFieldNotice}.
  */
 @GtfsValidator
-public class TripAgencyIdValidator extends FileValidator {
+public class RouteAgencyIdValidator extends FileValidator {
```
bdferris-v2, line 33: Did you actually delete `TripAgencyIdValidator`?  Or just fork it?  It seems like `TripAgencyIdValidator` could be deleted entirely?

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/RouteAgencyIdValidatorTest.java
```diff
@@ -0,0 +1,71 @@
...
+import com.google.common.collect.ImmutableList;
+import org.junit.Test;
+import org.mobilitydata.gtfsvalidator.notice.MissingRecommendedFieldNotice;
+import org.mobilitydata.gtfsvalidator.notice.MissingRequiredFieldNotice;
+import org.mobilitydata.gtfsvalidator.notice.NoticeContainer;
+import org.mobilitydata.gtfsvalidator.table.*;
+
+public class RouteAgencyIdValidatorTest {
+
+  public static GtfsRoute createRoute(
```
bdferris-v2, line 14: Per advice in previous PR, I'd inline these helper functions and drop calls to setBlah(null) as appropriate.

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/RouteAgencyIdValidatorTest.java
```diff
@@ -0,0 +1,71 @@
...
+            ImmutableList.of(createAgency(1, null, "Agency with no ID")), noticeContainer);
+    GtfsRouteTableContainer routeTable =
+        GtfsRouteTableContainer.forEntities(
+            ImmutableList.of(createRoute(0, "route_0", null, "Route 0")), noticeContainer);
+    new RouteAgencyIdValidator(agencyTable, routeTable).validate(noticeContainer);
+    assertThat(noticeContainer.getValidationNotices())
+        .containsExactly(
+            new MissingRecommendedFieldNotice(
+                routeTable.gtfsFilename(), 0, GtfsRoute.AGENCY_ID_FIELD_NAME));
+  }
```
bdferris-v2, line 70: Can you add a no notice case as well?

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/RouteAgencyIdValidator.java
briandonahue: No, I'm not sure why it looks like that in the diff...  I dupblicated it because `agency_id` is required for `routes.txt`, `trips.txt` and `agency.txt` when there is more than one agency in `agency.txt` and recommended when there is only 1.

briandonahue: (there was already a validator for this for `trips.txt`, that is what `TripAgencyIdValidator` did.  I added it for `routes.txt`)

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/RouteAgencyIdValidatorTest.java
briandonahue: No problem doing this at all, just FYI this is a convention used in a many test fixtures, so I was copying those. Not sure if it's worth addressing elsewhere...

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/AgencyConsistencyValidator.java
bdferris-v2: Was this actually resolved?

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/RouteAgencyIdValidator.java
bdferris-v2: There is a TripAgencyIdValidator in the master branch: https://github.com/MobilityData/gtfs-validator/blob/master/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TripAgencyIdValidator.java

So looks like you renamed it and the old file won't be around, which is +1 from me.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/AgencyConsistencyValidator.java
briandonahue: This was my mistake. Since I was added as a contributor I tried to clean up my remotes locally, pushed this to the core repo instead of the original PR branch!  Above, I ignore 0 agencies, warn on 1, and error on > 1

isabelle-dr:
Sounds good, I will have a look shortly :)

bdferris-v2 (approved):

isabelle-dr (approved):
LGTM, thank you @briandonahue for working on this!

isabelle-dr:
Why aren't we getting the acceptance test report for this PR?

PR #1319, opened 2023-01-25
Validate that arrival_time, departure_time and shape_dist_traveled are non-decreasing along each trip

aababilov:
Validate that arrival_time, departure_time and shape_dist_traveled are non-decreasing along each trip

Generated notices:
* TripWithOutOfOrderArrivalTimeNotice
* TripWithOutOfOrderDepartureTimeNotice
* TripWithOutOfOrderShapeDistTraveledNotice

The new validator is more powerful that the removed one. The new validator can also properly handle cases when some shape_dist_traveled are missing.


isabelle-dr:
One question: would `TripWithOutOfOrderShapeDistTraveledNotice` replace the existing `DecreassingOrEqualStopTimeDistance`?

Issue #1180, opened 2022-06-03
Add a rule that checks the presence of stop_lat & stop_lon in stops.txt


isabelle-dr:
**What problem in GTFS datasets does this new rule address? Please describe.**
The spec in `stops.stop_lat` and `stops.stop_lon` mentions ([source](https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md#stopstxt)):
> Conditionally Required:
> - Required for locations which are stops (location_type=0), stations (location_type=1) or entrances/exits (location_type=2).
> - Optional for locations which are generic nodes (location_type=3) or boarding areas (location_type=4).

This check is missing from the validator. 

**Describe the new validation rule**
For each line in `stops.txt`, if either `stop_lat` or `stop_lon` (or both) are missing AND location_type = 0, 1 or 2, trigger a `missing_stop_coordinates` notice.

**Error vs warning**
Error

isabelle-dr:
@aababilov, @bdferris, do you have an implementation of this notice?

bdferris-v2:
I wasn't quite sure what you were asking here.  Are you asking if we want to contribute an implementation for this validation check?  Or if we had an existing implementation?

isabelle-dr:
If you had an existing implementation already :)

KClough:
@bdferris-v2 wanted to double check to see if you had an implementation already. If not, I'll gladly take this one.

isabelle-dr:
@KClough not sure where you are with this but we have a Pull Request solving it (#1315)

isabelle-dr:
This has been solved in #1321

PR #1321, opened 2023-02-01
fix: generate errors when missing lat/long for stops, stations, entrances

briandonahue:
**Summary:**

Generate error notice when Latitude and/or Longitude is not supplied for stops (location_type=0), stations (location_type=1) or entrances/exits (location_type=2).

Addresses [#1180](https://github.com/MobilityData/gtfs-validator/issues/1180)

**Expected behavior:** 

Should throw an error when missing geocoordinate data for stops, stations, entrances/exits

- [ ] Include screenshot(s) showing how this pull request works and fixes the issue(s) (N/A)

### DIFF
--- a/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopLatLongValidator.java
+++ b/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopLatLongValidator.java
@@ -1,0 +1,55 @@
+package org.mobilitydata.gtfsvalidator.validator;
+
+import org.mobilitydata.gtfsvalidator.annotation.GtfsValidator;
+import org.mobilitydata.gtfsvalidator.notice.NoticeContainer;
+import org.mobilitydata.gtfsvalidator.notice.SeverityLevel;
+import org.mobilitydata.gtfsvalidator.notice.ValidationNotice;
+import org.mobilitydata.gtfsvalidator.table.GtfsLocationType;
+import org.mobilitydata.gtfsvalidator.table.GtfsStop;
+
+/**
+ * Validates {@code stops.stop_lat} and {@code stops.stop_long} exist for a single {@code GtfsStop}
+ * of type Stop, Station, Entrance, and Exit.
+ *
+ * <p>Generated notices:
+ *
+ * <ul>
+ *   <li>{@link LatLongRequiredForStopType}
+ * </ul>
+ */
+@GtfsValidator
+public class StopLatLongValidator extends SingleEntityValidator<GtfsStop> {
+
+  @Override
+  public void validate(GtfsStop stop, NoticeContainer noticeContainer) {
+    if ((stop.locationType() == GtfsLocationType.STOP
+            || stop.locationType() == GtfsLocationType.STATION
+            || stop.locationType() == GtfsLocationType.ENTRANCE)
+        && !stop.hasStopLatLon()) {
+      noticeContainer.addValidationNotice(
+          new LatLongRequiredForStopType(stop.csvRowNumber(), stop.stopId(), stop.locationType()));
+    }
+  }
+
+  /**
+   * A {@code GtfsStop} is missing {@code stops.stop_lat} and/or {@code stops.stop_long}
+   *
+   * <p>"Required for locations which are stops (location_type=0), stations (location_type=1) or
+   * entrances/exits (location_type=2)." (http://gtfs.org/reference/static#stopstxt)
+   *
+   * <p>Severity: {@code SeverityLevel.ERROR}
+   */
+  static class LatLongRequiredForStopType extends ValidationNotice {
+
+    private final int csvRowNumber;
+    private final GtfsLocationType locationType;
+    private final String stopId;
+
+    LatLongRequiredForStopType(int csvRowNumber, String stopId, GtfsLocationType type) {
+      super(SeverityLevel.ERROR);
+      this.stopId = stopId;
+      this.csvRowNumber = csvRowNumber;
+      this.locationType = type;
+    }
+  }
+}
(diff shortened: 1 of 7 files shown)


main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopLatLongValidator.java
```diff
@@ -0,0 +1,55 @@
...
+ * of type Stop, Station, Entrance, and Exit.
+ *
+ * <p>Generated notices:
+ *
+ * <ul>
+ *   <li>{@link LatLongRequiredForStopType}
+ * </ul>
+ */
+@GtfsValidator
+public class StopLatLongValidator extends SingleEntityValidator<GtfsStop> {
```
bdferris-v2, line 21: I think it'd make sense to consolidate this logic into the existing StopLatLonValidator file.

briandonahue: @bdferris-v2 this is probably due to me being new to the project... I was only aware of the validators in the [validator](https://github.com/MobilityData/gtfs-validator/tree/master/main/src/main/java/org/mobilitydata/gtfsvalidator/validator) folder... but doing a class search, I see a generated `GtfsStopLatLonValidator` class that I assume is generated from by `LatLonValidatorGenerator`.  As you mentioned on the call earlier, it looks like maybe it already covers this scenario, as it seems to [check](https://github.com/MobilityData/gtfs-validator/blob/master/processor/src/main/java/org/mobilitydata/gtfsvalidator/processor/LatLonValidatorGenerator.java#L96) if either coordinate is `<= 1.0` 

So now I'm wondering if I'm misunderstanding something, or if this change is unnecessary?

bdferris-v2: Ah, duh, you are totally right, that's a generated validator :)  Sorry for the confusion, I was just doing a quick check.

bdferris-v2, line 21: As mentioned offline, I think this class name is a little too close to the name of the existing `GtfsStopLatLonValidator` that it's probably inviting confusion (case in point, I already managed to confuse myself once here already).  I'd like to propose an alternate name.  Thoughts on one of the following?
* `StopRequiredLocationValidator`
* `StopConditionallyRequiredLocationValidator`

Basically, capturing that this validator is checking the presence of the location data itself, as opposed to validating the values in some way.  Thoughts?

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopLatLongValidator.java
```diff
@@ -0,0 +1,55 @@
...
+
+  /**
+   * A {@code GtfsStop} is missing {@code stops.stop_lat} and/or {@code stops.stop_long}
+   *
+   * <p>"Required for locations which are stops (location_type=0), stations (location_type=1) or
+   * entrances/exits (location_type=2)." (http://gtfs.org/reference/static#stopstxt)
+   *
+   * <p>Severity: {@code SeverityLevel.ERROR}
+   */
+  static class LatLongRequiredForStopType extends ValidationNotice {
```
bdferris-v2, line 42: I'd name this `StopWithoutLocationNotice` for consistency with the existing `StopWithoutStopTimeNotice` and `StopWithoutZoneIdNotice`

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/StopLatLongValidatorTest.java
```diff
@@ -0,0 +1,153 @@
...
+import java.util.List;
+import javax.annotation.Nullable;
+import org.junit.Test;
+import org.mobilitydata.gtfsvalidator.notice.NoticeContainer;
+import org.mobilitydata.gtfsvalidator.notice.ValidationNotice;
+import org.mobilitydata.gtfsvalidator.table.GtfsLocationType;
+import org.mobilitydata.gtfsvalidator.table.GtfsStop;
+
+public class StopLatLongValidatorTest {
+  private static GtfsStop createStop(
```
bdferris-v2, line 14: In general, I'd prefer to inline this method and drop calls to setStopLat() and setStopLon() where stop location data is missing.  In tests, I tend to favor readability in the unit test itself over general DRY conventions for regular code.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopLatLongValidator.java
briandonahue: @bdferris-v2 Sure - I'm happy to change it.  No strong preference unless we want to get very specific (and long) with something like `LocationRequiredForStopStationEntranceValidator` or `StopStationEntranceLocationRequiredValidator` 😬Whatever feels more in line with the current naming convention...

bdferris-v2: I'm not a Java long naming absolutist :)  I'd just go with `StopRequiredLocationValidator`

briandonahue:
@bdferris-v2 I believe all your suggestions are now implemented, thanks for reviewing!

bdferris-v2 (approved):
LGTM

bdferris-v2:
@briandonahue one other issue: I think the code formatting check is failing on your PR.  Can you run the Gradle `goJF` command on your code to format it properly and update the PR?  Thanks!

isabelle-dr:
@briandonahue and @bdferris-v2 sorry to arrive late on this PR, the documentation is missing for this new notice. 
I know we are discussing generating [RULES.md](https://github.com/MobilityData/gtfs-validator/blob/master/RULES.md) automatically, but in the meantime, we need it 😊.
The HTML report links directly to RULES.md to tell users what the notice is, and how to fix it.

Also, I acknowledge that a review by a core developer is enough to unblock the merge (thank you @bdferris-v2), but please wait for a MobilityData staff to review as well before merging. I know our capacity doesn't allow us to review right away, but we need to acknowledge what has been merged on this project so that we can:
- close the issue if needed, open follow-up issues
- coordinate with other contributors (for example, PR #1315 contains the same fix)
- make sure dependencies (like the docs) have been updated
- answer questions
Many thanks 🙏

PR #1326, opened 2023-02-03
Manual dispatch of workflows

bdferris-v2:
(This is a test)

### DIFF
--- a/.github/workflows/acceptance_test.yml
+++ b/.github/workflows/acceptance_test.yml
@@ -13,6 +13,7 @@ on:
       - 'test_pack_dock.yml'
       - 'triage.yml'
       - 'end_to_end.yml'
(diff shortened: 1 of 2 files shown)


bdferris-v2:
@isabelle-dr this might be a better alternative to manual PRs like #1347 when you want to run the acceptance workflows

isabelle-dr:
Thank you! 
Definitely a better alternative 🙃

bdferris-v2:
So I think there is still a bit of code I would write to help you here.  Specifically, I would output the mapping for validation notices => feeds in the output comparator as an additional output.  Then you could see the set of notices as generated against the `master` branch for each run.

(I had thought something like this already existed in the output, but after looking closer, it doesn't output quite what you need).

PR #1330, opened 2023-02-10
fix: update RULES.md with missing rule

briandonahue:
**Summary:**

Closes #1329. Missing rule definition from PR #1321 

**Expected behavior:** 

Rule should have been added to RULES.md

Please make sure these boxes are checked before submitting your pull request - thanks!

- [x] Run the unit tests with `gradle test` to make sure you didn't break anything
- [x] Format the title like "feat: [new feature short description]". Title must follow the Conventional Commit Specification(https://www.conventionalcommits.org/en/v1.0.0/).
- [x] Linked all relevant issues
- [ ] Include screenshot(s) showing how this pull request works and fixes the issue(s)

Issue #1324, opened 2023-02-02
[feature request / refactor] Produce structured data containing human-readable code descriptions (refactor RULES.md)


lauriemerrell:
### Describe the problem

Cal-ITP produces https://reports.calitp.org/, where we report on various aspects of GTFS data quality. One of the things we currently display on the site is a grid of validator notices output for a given feed in a given month. We like to display a human-readable notice description so that the notice can be understood by agencies and the general public, who may not be familiar with validator code names. 

Currently, to update those human readable descriptions, we have to manually scrape the data from RULES.md for each validator version and turn it into a CSV that we can import through our pipeline. 

To make the CSV, I:
* Regex'd the .md file to extract the code with its simple description 
* Manually annotated with rule severity, because the current format doesn't actually contain a table with `code, description, severity` in one place (the severity is just indicated in the *title* of the table, which makes it harder to scrape)
* Manually removed Markdown and HTML (RULES.md uses an inconsistent mixture of both) 

This also opens up issues like #1322 because RULES.md is maintained separately as a text file and not related to the actual validator code.

It would be nice if the human readable description about rule implementation were available as structured data (CSV or JSON) and could be output by the validator itself, rather than requiring reference to the RULES.md file (analogous to the new `notice_schema.json` file that can be output by the JAR). 

### Proposed solution

Rule descriptions could be attributes within the rule implementation itself, and then RULES.md could be programmatically generated based on those attributes, rather than RULES.md being the source of truth but maintained separately. 

### Additional context

It would be really nice to have something like `code, severity, short_desc, detailed_desc, formatted_desc` where `formatted` could contain Markdown (RULES.md has a shorter rule description in the tables at the top and then a slightly longer description below.)

lauriemerrell:
I think that more structured of this type would also be useful for the web-based validator UI, which currently just links to RULES.md to explain rule meaning. If the rule descriptions were in the rule code, presumably the web UI could display them more easily. But I defer to @KClough et al. on those considerations.

themightychris:
The [`index.js` file in this PR](https://github.com/MobilityData/gtfs-validator/pull/1088/files#diff-62afa71bbd7b0a66a50b65ec952daf22ec73b11bf2b70e07ce475fb586ec3f05) (click to expand the collapsed-by-default file) contained some improved language describing a lot of errors. It's been in the collective mental backlog to capture those somewhere. If there's an effort to capture the rules into a more machine-readable format that might be a good moment to gobble those up

[...]

PR #1331, opened 2023-02-11
feat: Add unit test to verify that every Notice has an entry in RULES.md

bdferris-v2:
**Summary:**

When writing new validation rules, we often forget to update `RULES.md` with updated documentation.  While #1324 has some ideas on addressing that in the long term, in the short term, we can add a unit-test that will fail if there is an entry in `RULES.md` for a newly added notice.

For example, this test should fail initially because `StopWithoutLocationNotice` is missing from `RULES.md` per #1330.

### DIFF
--- a/core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NoticeSchemaGenerator.java
+++ b/core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NoticeSchemaGenerator.java
@@ -98,7 +98,7 @@ public class NoticeSchemaGenerator {
    *
    * <pre>{@code
    * {
-   *   "AttributionWithoutRoleNotice": {
+   *   "attribution_without_role": {
    *     "attributionId": String.class,
    *     "csvRowNumber": Long.class,
    *   }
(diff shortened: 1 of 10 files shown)


isabelle-dr:
Thank you for adding this test @bdferris-v2 !

>  unit-test that will fail if there is an entry in RULES.md for a newly added notice.
You mean that will fail if there is **_no_** entry in RULES.md for a newly added notice, right?

Why is the `test_pack_doc` workflow failing in this PR?

bdferris-v2:
@isabelle-dr the test is correctly identifying that StopWithoutLocationNotice is not properly documented in RULES.md and failing as result :)

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NoticeSchemaGenerator.java
```diff
@@ -123,8 +114,9 @@ static Map<String, Map<String, Class<?>>> contextFieldsInPackages(List<String> p
       throws IOException {
     // Return a sorted TreeMap for stable results.
     Map<String, Map<String, Class<?>>> contextFieldsByNotice = new TreeMap<>();
-    for (Class<Notice> noticeClass : findNoticeSubclasses(packages)) {
-      contextFieldsByNotice.put(noticeClass.getSimpleName(), contextFieldsForNotice(noticeClass));
+
+    for (Class<Notice> noticeClass : ClassGraphDiscovery.discoverNoticeSubclasses(packages)) {
+      contextFieldsByNotice.put(Notice.getCode(noticeClass), contextFieldsForNotice(noticeClass));
```
asvechnikov2, line 119: AFAIU the returned Map has notice code as its instead of class name, could you please update the method comment?

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/NoticeDocumentationTest.java
```diff
@@ -0,0 +1,89 @@
...
+import org.junit.runners.JUnit4;
+import org.mobilitydata.gtfsvalidator.notice.Notice;
+
+@RunWith(JUnit4.class)
+public class NoticeDocumentationTest {
+  private static Pattern MARKDOWN_NOTICE_SIMPLE_CLASS_NAME_ANCHOR_PATTERN =
+      Pattern.compile("^<a name=\"(\\w+(Error|Notice))\"/>");
+
+  private static Pattern MARKDOWN_NOTICE_CODE_HEADER_PATTERN =
+      Pattern.compile("^### (([a-z0-9]+_)*[a-z0-9]+)");
```
asvechnikov2, line 26: Do we allow numbers in notice names? (or do want to allow?)

asvechnikov2 (approved):
A couple of questions, otherwise, looks good, thank you!

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/NoticeDocumentationTest.java
bdferris-v2: We do because we have notices with numbers in their name; `feed_expiration_date7_days` and `feed_expiration_date30_days` to be specific.

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NoticeSchemaGenerator.java
bdferris-v2: Done

bdferris-v2:
@asvechnikov2 thanks for the review!

PR #1340, opened 2023-02-16
feat: Add MixedCase validation (#881)

### LINKED ISSUE #881: Best Practice: stop names should be in mixed case (WARNING)
Provide stop names in mixed case, following local conventions, as per recommendation for all customer-facing text fields.

briandonahue:
**Summary:**

Closes #881 

**Expected behavior:** 


processor/src/main/java/org/mobilitydata/gtfsvalidator/processor/MixedCaseValidatorGenerator.java
```diff
@@ -0,0 +1,92 @@
...
+      validateMethod
+          .beginControlFlow(
+              "if (entity.$L())", FieldNameConverter.hasMethodName(mixedCaseField.name()))
+          .addStatement("$T value = entity.$L()", String.class, mixedCaseField.name())
+          .beginControlFlow("if (!(value.matches(\".*[a-z].*\") && value.matches(\".*[A-Z].*\")))")
+          .addStatement(
+              "noticeContainer.addValidationNotice(new $T(\"$L\", \"$L\", entity.csvRowNumber()))",
+              MixedCaseNotice.class,
+              fileDescriptor.filename(),
+              CaseFormat.UPPER_CAMEL.to(CaseFormat.LOWER_UNDERSCORE, mixedCaseField.name()))
```
briandonahue, line 83: @bdferris-v2 This is what I added to get my tests to pass, but I'm not sure it's a great solution... I don't see an obvious way to validate a column header in the CSV vs. a field name in the Schema.  E.g. for `agency.txt` I don't see anywhere that the `agencyName` field is specified in code or validated that the column header is `agency_name`.  It seems it could be arbitrary. In `RowParser` we have a `CsvHeader` object where the correct header can be retrieved from for things like `MissingRequiredFieldNotice` but we don't have that in the generated validators.  

Let me know if you have any recommendations...

core/src/main/java/org/mobilitydata/gtfsvalidator/testing/LoadingHelper.java
```diff
@@ -63,6 +63,7 @@ public <X extends GtfsEntity, Y extends GtfsTableContainer<X>> Y load(
       GtfsTableDescriptor<X> tableDescriptor, String... lines) throws ValidatorLoaderException {
     String content = Arrays.stream(lines).collect(Collectors.joining("\n"));
     InputStream in = new ByteArrayInputStream(content.getBytes(StandardCharsets.UTF_8));
+    noticeContainer = new NoticeContainer(); // reset notices if reloading helper
```
briandonahue, line 66: @bdferris-v2 just FYI, I added this here to clear notices when calling `load` so that I could call it multiple times in one test.  I don't think it it's a problem, and all tests pass, but let me know if you have any concerns.

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/MixedCaseNotice.java
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * A string field has a value that does not contain mixed case.
+ *
+ * <p>Severity: {@code SeverityLevel.WARNING}
+ *
+ * @see org.mobilitydata.gtfsvalidator.annotation.MixedCase
+ */
+public class MixedCaseNotice extends ValidationNotice {
+  private final String fileName;
```
bdferris-v2, line 11: Make this "filename" for consistency with other notices.

core/src/main/java/org/mobilitydata/gtfsvalidator/annotation/MixedCase.java
```diff
@@ -0,0 +1,39 @@
...
+
+package org.mobilitydata.gtfsvalidator.annotation;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Generates a validation that a string field is in mixed case.
```
bdferris-v2, line 25: I found this string a little unclear.  Maybe something like:

> Specifies a string field should have a mixed-case value.  A validator will be automatically generated.

core/src/main/java/org/mobilitydata/gtfsvalidator/testing/LoadingHelper.java
bdferris-v2: I think that's probably fine.

main/src/main/java/org/mobilitydata/gtfsvalidator/table/GtfsAgencySchema.java
```diff
@@ -34,6 +29,7 @@ public interface GtfsAgencySchema extends GtfsEntity {
   String agencyId();
 
   @Required
+  @MixedCase
```
bdferris-v2, line 32: I know this annotation is intended for more than this field, but it's going to be interesting to see how this actually validates for agency name.  For example, if the agency name is "SFMTA" or "MARTA" (aka an acronym) it won't actually be mixed-case.

processor/src/main/java/org/mobilitydata/gtfsvalidator/processor/MixedCaseValidatorGenerator.java
bdferris-v2: I think what you is:

```
FieldNameConverter.gtfsColumnName(mixedCaseField.name())
```

It's probably a bug that this conversion is missing in `CurrencyAmountValidatorGenerator`.

processor/tests/src/main/java/org/mobilitydata/gtfsvalidator/processor/tests/MixedCaseSchema.java
```diff
@@ -0,0 +1,12 @@
...
+
+import org.mobilitydata.gtfsvalidator.annotation.GtfsTable;
+import org.mobilitydata.gtfsvalidator.annotation.MixedCase;
+import org.mobilitydata.gtfsvalidator.annotation.Required;
+
+@GtfsTable("mixed_case.txt")
+public interface MixedCaseSchema {
+  @Required
+  @MixedCase
+  String mixedCase();
```
bdferris-v2, line 11: I would actually name this field something different.  Like "someField".  I think having the field named "mixedCase" causes just a bit of confusion in the unit-test about when "mixedCase" is being used as field name or in some other context.

processor/tests/src/test/java/org/mobilitydata/gtfsvalidator/processor/tests/MixedCaseSchemaTest.java
```diff
@@ -0,0 +1,88 @@
...
+    tableDescriptor = new MixedCaseTableDescriptor();
+    helper = new LoadingHelper();
+    helper.setValidatorLoader(
+        ValidatorLoader.createForClasses(ImmutableList.of(MixedCaseMixedCaseValidator.class)));
+  }
+
+  @Test
+  public void testValidMixedCase() throws ValidatorLoaderException {
+
+    helper.load(tableDescriptor, MixedCase.MIXED_CASE_FIELD_NAME, "MixedCase");
```
bdferris-v2, line 50: Ditto here.  "MixedCase" -> "MixedCaseValue" or "SomeValue"

processor/tests/src/main/java/org/mobilitydata/gtfsvalidator/processor/tests/MixedCaseSchema.java
bdferris-v2: I also found it a little confusing that "MixedCase" is both used as the annotation and the entity class names in the tests below.

briandonahue: @bdferris-v2 I modeled the `CurrencyAmount` example, so it has the same issue. I renamed it to `MixedCaseTestSchema`, though not sure it's a major improvement in readability. Let me know what you think.

bdferris-v2: Ha, it's true.  Maybe it was less noticeable in that test because I inlined all the file and field names in the test, as opposed to referencing CurrencyAmount.FILENAME, etc?  Sorry for the conflicting advice.

bdferris-v2 (approved):

isabelle-dr:
We generally name the notices with words that describe the problem (equal_shape_distance_diff_coordinates, invalid_currency, missing_required_file). 
I wonder if we could find another name for this one that would follow the same pattern. Maybe non_mixed_case_field?

briandonahue:
> I wonder if we could find another name for this one that would follow the same pattern. Maybe non_mixed_case_field?

@isabelle-dr How about `mixed_case_required` or `mixed_case_recommended` since it's a warning / best practice?

isabelle-dr:
I think`mixed_case_recommended_field` would be good :)

briandonahue:
> I think`mixed_case_recommended_field` would be good :)

@isabelle-dr done!

isabelle-dr:
@briandonahue one last comment before we merge, based on Brian's comment [here](https://github.com/MobilityData/gtfs-validator/pull/1340#discussion_r1112105456) and after reading the Best Practice.
I think we need to explain why this is recommended in a little mode in detail.

Can we replace the long description of this notice with the following:

> This field contains customer-facing text and should use Mixed Case (upper and lower case letters) to ensure good readability when displayed to riders. Avoid the use of abbreviations throughout the feed (e.g. St. for Street) unless a location is called by its abbreviated name (e.g. “JFK Airport”). Abbreviations may be problematic for accessibility by screen reader software and voice user interfaces.

Lastly, can you remove the extra "." character in the short description (in the table)?

After this, we are good to go!

briandonahue:
@isabelle-dr updated!

isabelle-dr (approved):
LGTM 🚀

isabelle-dr:
It looks like we have a problem with [test pack doc](https://github.com/MobilityData/gtfs-validator/actions/runs/4366414248/jobs/7638022895) 🤔

bdferris-v2:
@isabelle-dr it was a transient failure.  I re-ran the failing test and it's now run cleanly.

isabelle-dr:
Ok, merging now :)

Issue #1343, opened 2023-02-28
Clarify @Required annotations for validation


briandonahue:
### Describe the problem

Currently the [`@Required`](https://github.com/MobilityData/gtfs-validator/blob/master/core/src/main/java/org/mobilitydata/gtfsvalidator/annotation/Required.java#L25) annotation is used on the schema files both at a Class level (for required **files**) and the Property level (for required **fields**). This can be confusing and recently, we in #1198 we had a scenario where a *column* was required, but it was not required that the *field* have data for the column.   We discussed adding a new annotation when a column **Header**  is required, but the column data can be empty (not required), which led to discussion of the existing `@Required` annotation.

### Proposed solution

I think for additional clarity, it would make sense to break the original `@Required` header into two:
`@RequiredFile` - when the file can not be missing from the feed (valid on the schema Class)
`@RequiredValue` - when a value is required for the field (
`@RequiredColumn` - when the column (header) is required, but the field (value) is allowed to be empty

### Alternatives you've considered

Leave `@Required` as-is, and just add the new `@ RequiredColumn` annotation for the new case.

briandonahue:
On further review, the `@Required` annotation is **not** used to validate field data, but is actually only used to validate the column header exists.  It is also used on the class to validate the file exists. While I think the dual use is not ideal, I realize that a third annotation is not necessary which makes it less of an issue.  

Required fields are specified in the `GtfsColumnDescriptor` for the column, which is created in the `GtfsTableDescriptor<T>` for that entity type.

briandonahue:
@isabelle-dr mentioned another scenario in #1350 regarding a warning for a `RecommendedColumn`.  I think adding several more specific annotations may be useful.  

Currently we have the following:

### Current:

| Rule                    | Annotation        | Target     | Example        | Type  | Occurs        | Notes                                       |
| ----------------------- | ----------------- | ---------- | -------------- | ----- | ------------- | ------------------------------------------- |
| File Required           | `@Required`       | Class    | `Agency`       | Error | Feed          | Dual purpose annotation may cause confusion |
| Column & Value Required | `@Required`       | Field | `agencyName`   | Error | File & Record | Dual purpose annotation may cause confusion |
| Column & Value Optional | `@RequiredColumn` | Field | `transferType` | Error | File          |                                             |

[...]

PR #1344, opened 2023-02-28
Fix 1198: separate required column from required value

### LINKED ISSUE #1198: False positives of missing_required_field when encountering empty transfer_type values in transfers.txt
## **Bug report** 

The GTFS Validator is generating false positives of the `missing_required_field` when it encounters empty `transfers.txt#transfer_type` field values.

**Describe the bug**

Given an empty value within the `transfers.txt#transfer_type` field (which is valid [according to the spec](https://gtfs.org/schedule/reference/#:~:text=recommended%20transfer%20point%20between%20routes)), when the validator checks the file, then it will falsely raise a `missing_required_field` notice.

**How we reproduce the bug**

Example feed: [Archive.zip](https://github.com/MobilityData/gtfs-validator/files/9002895/Archive.zip)

In this feed in the `transfers.txt` file, the following use case is shown:

```csv
from_stop_id,to_stop_id,transfer_type,min_transfer_time
1,2,,
```

This is a valid use case in that the `transfer_type` value is empty which indicates a "Recommended transfer point between routes".

**Expected behaviour**

The validator should not flag this as an error.

**Observed behaviour**

The GTFS Validator flagged the above-mentioned use case as an error.

**Screenshots:** 

See relevant JSON output for the above-linked feed.

```json
{
  "notices": [{
    "code": "missing_required_field",
    "severity": "ERROR",
    "totalNotices": 1,
    "sampleNotices": [{
      "filename": "transfers.txt",
      "csvRowNumber": 2.0,
      "fieldName": "transfer_type"
    }]
  },
...
```

**Environment versions**
- validator version: v2, v3.0.1, v3.1.0
- Java version: 
- OS versions:

briandonahue:
**Summary:**

Closes #1198. Added a new annotation `@RequiredColumn` to allow for scenario where a column must exist in the file, but values for that column may be empty.

It may be useful in the future to separate the existing `@Required` annotation into `@RequiredFile` and `@RequiredValue` annotations. I've opened #1343 for discussion of that topic.

**Expected behavior:** 

Given the following `transfers.txt` file, the following should not throw an error:

```
from_stop_id,to_stop_id,transfer_type,min_transfer_time
1,2,,
```

This is a valid use case [according to the spec](https://gtfs.org/schedule/reference/#:~:text=recommended%20transfer%20point%20between%20routes) in that the transfer_type value is empty which indicates a "Recommended transfer point between routes".

- (N/A) Include screenshot(s) showing how this pull request works and fixes the issue(s)

### DIFF
--- a/core/src/main/java/org/mobilitydata/gtfsvalidator/annotation/Required.java
+++ b/core/src/main/java/org/mobilitydata/gtfsvalidator/annotation/Required.java
@@ -24,7 +24,7 @@ import java.lang.annotation.Target;
 /**
  * Adds a validation that the field or a file is required.
  *
- * <p>In the case of a field, both the column header, and a value for each row is required.
+ * <p>In the case of a field, both the column header and a value for each row is required.
  *
  * <p>Example.
  *
--- a/core/src/main/java/org/mobilitydata/gtfsvalidator/annotation/RequiredColumn.java
+++ b/core/src/main/java/org/mobilitydata/gtfsvalidator/annotation/RequiredColumn.java
@@ -22,9 +22,8 @@ import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
 /**
- * Adds a validation that the field or a file is required.
- *
- * <p>In the case of a field, both the column header, and a value for each row is required.
+ * Adds a validation that the column and header must be present. A value for the field may be
+ * optional.
  *
  * <p>Example.
  *
@@ -35,10 +34,10 @@ import java.lang.annotation.Target;
  *       {@literal @}PrimaryKey
  *       String agencyId();
  *
- *       {@literal @}Required String agencyName();
+ *       {@literal @}RequiredColumn String agencyName();
  *   }
  * </pre>
  */
 @Target({ElementType.METHOD, ElementType.TYPE})
 @Retention(RetentionPolicy.SOURCE)
-public @interface Required {}
+public @interface RequiredColumn {}
--- a/core/src/main/java/org/mobilitydata/gtfsvalidator/annotation/RequiredColumn.java
+++ b/core/src/main/java/org/mobilitydata/gtfsvalidator/annotation/RequiredColumn.java
@@ -28,14 +28,23 @@ import java.lang.annotation.Target;
  * <p>Example.
  *
  * <pre>
- *   {@literal @}GtfsTable("agency.txt")
- *   public interface GtfsAgencySchema extends GtfsEntity {
- *       {@literal @}FieldType(FieldTypeEnum.ID)
- *       {@literal @}PrimaryKey
- *       String agencyId();
+ *    {@literal @}GtfsTable("transfers.txt")
+ *    public interface GtfsTransferSchema extends GtfsEntity {
+ *        {@literal @}FieldType(FieldTypeEnum.ID)
+ *        {@literal @}Required
+ *        {@literal @}ForeignKey(table = "stops.txt", field = "stop_id")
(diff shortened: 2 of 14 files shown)


briandonahue:
@bdferris-v2 As this uses code generation, I was unsure if/how we could add tests for this. Please let me know if you have thoughts.

core/src/main/java/org/mobilitydata/gtfsvalidator/annotation/RequiredColumn.java
```diff
@@ -0,0 +1,43 @@
...
+ * <p>Example.
+ *
+ * <pre>
+ *   {@literal @}GtfsTable("agency.txt")
+ *   public interface GtfsAgencySchema extends GtfsEntity {
+ *       {@literal @}FieldType(FieldTypeEnum.ID)
+ *       {@literal @}PrimaryKey
+ *       String agencyId();
+ *
+ *       {@literal @}RequiredColumn String agencyName();
```
bdferris-v2, line 37: It might make more sense to use transfers.txt as the example here?

core/src/main/java/org/mobilitydata/gtfsvalidator/validator/DefaultTableHeaderValidator.java
```diff
@@ -34,16 +34,16 @@ public class DefaultTableHeaderValidator implements TableHeaderValidator {
   public void validate(
       String filename,
       CsvHeader actualHeader,
-      Set<String> supportedColumns,
-      Set<String> requiredColumns,
+      Set<String> supportedHeaders,
+      Set<String> requiredHeaders,
```
bdferris-v2, line 38: Does it make sense to change the naming from "columns" to "headers" when the rest of the method still uses "column" as its naming convention?

bdferris-v2:
Regarding testing, could you introduce a test case in [processor/tests](https://github.com/MobilityData/gtfs-validator/tree/master/processor/tests) with a test schema that exercises the @RequiredColumn behavior?

core/src/main/java/org/mobilitydata/gtfsvalidator/validator/DefaultTableHeaderValidator.java
briandonahue: I reverted the change.  I think Header *may* be more accurate as the class is validating headers, but since the notice uses the term "column" and the terms are somewhat mixed and matched, I figured I'd go for consistency in this class.

core/src/main/java/org/mobilitydata/gtfsvalidator/annotation/Required.java
```diff
@@ -24,6 +24,8 @@
 /**
  * Adds a validation that the field or a file is required.
  *
+ * <p>In the case of a field, both the column header, and a value for each row is required.
```
bdferris-v2, line 27: nit: remove comma after "column header"

bdferris-v2 (approved):

isabelle-dr (approved):

isabelle-dr:
Just a side thought: We have a [missing_required_column](https://github.com/MobilityData/gtfs-validator/blob/master/RULES.md#missing_required_column) in this validator, and it is triggered when the whole column is missing (as opposed to a missing required field in a given row).
I don't think it necessarily interacts with the work in this PR, just something to keep in mind. 

Merging this PR, thank you @briandonahue!

isabelle-dr:
@briandonahue could you resolve the conflicts? :)

briandonahue:
> @briandonahue could you resolve the conflicts? :)

@isabelle-dr done!

isabelle-dr:
Thank you for this awesome contribution @briandonahue! ✨

PR #1345, opened 2023-03-01
feat: Add unit-test to check consistency of Notice field names.

bdferris-v2:
While working on better alignment of code and documentation as part of #1324, I fixed a few inconsistent field names (e.g. `filename` vs `fileName`).  But I also wonder if we could catch these inconsistencies earlier.  This PR provides a unit-test that checks all `Notice` field names against a manually curated allow-list.

Any time a new field name is added, this unit-test will fail.  When that happens, we hope the author will do one of the following:
1) See there is an existing field name that might be more appropriate to use.
2) If not, add their new field to the allow-list.

### DIFF
--- a/.github/workflows/test_pack_doc.yml
+++ b/.github/workflows/test_pack_doc.yml
@@ -3,12 +3,8 @@ name: Test Package Document
 on:
   push:
     branches: [ master ]
-    paths-ignore:
-      - '**.md'
   pull_request:
     branches: [ master ]
-    paths-ignore:
-      - '**.md'
   release:
     types: [ prereleased, released ]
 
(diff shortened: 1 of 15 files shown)


bdferris-v2:
@isabelle-dr curious to get your thoughts on this

isabelle-dr:
Great addition! Can we see what it looks like when the test fails?

bdferris-v2:
Ideally, a developer would run `gradlew test` before submitting their PR (per the instructions) and get a failing unit test that would look like:

Screenshot 2023-03-06 at 4 03 22 PM

In this example, `csvRowNumberA` and `csvRowNumberB` are the new unexpected fields (artificial example).  At that point, the developer would hopefully go look at `NoticeFieldsTest. testNoticeClassFieldNames`, see the instructions there, and either change their field names or update the test.

davidgamez (approved):
LGTM!

Issue #1349, opened 2023-03-06
Acceptance Tests only report changes in ERROR notices, not WARNINGS or INFOs


bdferris-v2:
Currently, the Acceptance Test workflow only reports on changes to notices of severity level `ERROR`, per logic in the [output comparator](https://github.com/MobilityData/gtfs-validator/blob/master/output-comparator/src/main/java/org/mobilitydata/gtfsvalidator/outputcomparator/cli/Main.java).  I think it would be useful to also report on changes of other severity levels, especially `WARNING` but probably `INFO` as well for completeness.

bdferris-v2:
@isabelle-dr per your question about acceptance test, I also wanted to call out this behavior regarding errors vs warnings.

isabelle-dr:
That's right, thank you for opening this issue

bdferris-v2:
I've got a PR to address this.  Let's see if we can get it in.

bdferris-v2:
Re-opening this to ask a question to @isabelle-dr:

Do you want acceptance tests to fail if the number of WARNINGS generated changes in a significant way, similar to the existing logic for ERRORs?  PR #1354 add the reporting piece for warnings but it did not go so far as to fail if the warning count change is suspicious.

isabelle-dr:
Thanks for asking @bdferris-v2 
Yes, I'd like this test to fail if there is a significant change in the WARNINGS. It could mean a problem with the code, and if not, it needs additional attention (such as contacting the agencies affected).

isabelle-dr:
Re-opening because I believe we need https://github.com/MobilityData/gtfs-validator/blob/master/docs/ACCEPTANCE_TESTS.md to be updated to call this done.

PR #1352, opened 2023-03-13
feat: Refactor JSON report comparison for better diffing in unit-tests.

bdferris-v2:
Refactor JSON test artifact checks in the acceptance test `output-comparison` tool to make it easier to understand diffs when tests fail or or a test is updated.  Currently,  the JSON output report is tested in the unit-test as a single-line, non-pretty-printed string, which makes it really hard to tell what changed when the test fails or is updated.  This PR refactors the test to check a pretty-printed version of the JSON output for better diffing.  The expected content of the report is stored as a resource in the test/resources directory, since inlining a multi-line string is still pretty ugly in Java.

This refactor is in support of issue #1349 where some changes to the output comparator are coming in a follow-up PR.


output-comparator/src/test/java/org/mobilitydata/gtfsvalidator/outputcomparator/cli/MainTest.java
```diff
@@ -68,7 +69,16 @@ private void writeFile(JsonObject fileData, Path path) throws IOException {
...
 
   private static String retrieveReportString(Path path) throws IOException {
-    return Files.readString(path);
+    String content = Files.readString(path);
+    Gson gson = new GsonBuilder().setPrettyPrinting().create();
+    JsonObject obj = gson.fromJson(content, JsonObject.class);
+    return gson.toJson(obj);
+  }
+
+  private static String retrieveResource(String name) throws IOException {
```
davidgamez, line 78: nit[non-blocking]: We can define this method at instance level rather than `static`. The same change can be done to `retrieveReportString` to keep them aligned.

davidgamez (approved):
Nice! LGTM

output-comparator/src/test/java/org/mobilitydata/gtfsvalidator/outputcomparator/cli/MainTest.java
bdferris-v2: Any particular reason to prefer an instance method here as opposed to a static method?

davidgamez: This method is not accessing any other static method and can be see as an independent unit of processing.  As it is a private method is not needed to be shared across classes, only internal consumers.

bdferris-v2: Maybe I'm not totally clear on what you are proposing.  When you said "define this method at instance level" rather than static, I assumed you were proposing to remove the `static` qualifier.  Is that correct?  Or were you proposing something different?

davidgamez: Sorry for the misunderstanding, remove the static qualifier.

PR #1354, opened 2023-03-13
feat: Update acceptance test report to include changes in WARNINGS.

### LINKED ISSUE #1349: Acceptance Tests only report changes in ERROR notices, not WARNINGS or INFOs

bdferris-v2:
Refactor the acceptance test validation report comparison logic to consider changes to WARNING-level notices, not just ERRORs.  We could more easily expand this to include INFO-level notices as well, if we so desire.  Currently, this change does not cause automatic blocking of a change if the # of warnings changes significantly, but we could change that as well.

As part of this change, I refactored a bunch of logic out of `ValidationReport` and moved it into a new `NoticeSet` to be separate some concerns.  I also refactored a bunch of references to "errors" to more generic "notices".

Closes #1349.


davidgamez (approved):
LGTM!

isabelle-dr:
Do we need to update https://github.com/MobilityData/gtfs-validator/blob/master/docs/ACCEPTANCE_TESTS.md as a result of this PR?

PR #1361, opened 2023-03-23
feat: Initial entry for @GtfsValidationNotice annotation

bdferris-v2:
As part of on-going work for #1324 for automatic generation of notice documentation from source code, this PR introduces @GtfsValidationNotice, an annotation that will be used to configure and document validation notices in source code.

As generally described in https://bit.ly/gtfs-validator-notice-documentation, this annotation will be applied to all ValidationNotice classes in a follow-up PR.  It will then be used as run-time to resolve a notice's severity and also to generate documentation for the notice.

I'm introducing the annotation in its own PR so that we can confirm the API before I do the large-scale transformation of the code-base to add it.  I do include documentation for one notice as an example of its usage.


core/src/main/java/org/mobilitydata/gtfsvalidator/annotation/GtfsValidationNotice.java
```diff
@@ -0,0 +1,80 @@
...
+import org.mobilitydata.gtfsvalidator.notice.SeverityLevel;
+import org.mobilitydata.gtfsvalidator.table.GtfsEntity;
+
+/**
+ * An annotation used to identify, configure, and document a validation notice. All notices should
+ * be annotated with this type.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+public @interface GtfsValidationNotice {
```
davidgamez, line 16: question: Do you think there is value in adding the validator class as a field in the annotation? Add it might help us to identify programmatically easier the relation between validators and the notices.

bdferris-v2: Perhaps?  I know some generic Notices are used by more than one validator.  That said, most Notices are defined as static inner classes of the validator itself.  In the spirit of preferring convention over configuration, we can probably extract that relationship automatically to determine the linkage between validator and notice.

davidgamez: The `convention over configuration` is a great point! Thanks!

davidgamez (approved):
LGreatTM!

bdferris-v2:
As an FYI to @isabelle-dr in case you have any feedback on the proposed design.

bdferris-v2:
I discussed this offline with @davidgamez but I have proposed a refactor to where some key annotations and entity interface are stored (#1362) that will breakup a potential circular dependency.  Critically, we want to use the new @GtfsValidationNotice annotation for notice processing, but there are already notices defined in `core`.  By putting the annotation there as well, it limits our ability to do that processing from a Gradle module standpoint.

I have proposed moving the annotations into their own module.  I think this has some nice properties down the road for splitting up our model API and our business logic, so I think it's a net win.  I will update this PR based on the refactor PR.

PR #1365, opened 2023-03-27
feat: Extract notice docs from source-code for in-app use and schema export

bdferris-v2:
**Summary:**

Per discussion in #1324, this PR adds support for extracting documentation from notice source code, to be included as JSON resources bundles in the compiled application.  These will be included in the now-expanded JSON schema export feature, including comments, reference links, and hopefully enough info to generate RULES.md from source.

See https://bit.ly/gtfs-validator-notice-documentation for initial design discussion.


core/src/main/java/org/mobilitydata/gtfsvalidator/notice/CsvParsingFailedNotice.java
```diff
@@ -52,7 +56,7 @@ public class CsvParsingFailedNotice extends ValidationNotice {
    * @param exception the exception thrown
    */
   public CsvParsingFailedNotice(String filename, TextParsingException exception) {
-    super(SeverityLevel.ERROR);
+    super(ERROR);
```
davidgamez, line 59: Are we planning to remove the severity field in the future?

davidgamez:
@bdferris-v2 let's connect to discuss where the generated documentation can live to make it easily accessible to consumers.

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/CsvParsingFailedNotice.java
bdferris-v2: Yes.  That will be a follow-up refactor.

PR #1366, opened 2023-03-27
feat: Add @GtfsValidationNotice annotations to all notice classes

bdferris-v2:
**Summary:**

This is an automated transform of the existing source code, extracting documentation references from RULES.md where appropriate.  This part of #1324 and our effort to generate documentation from source code.

This PR also includes a small change to @GtfsValidationNotice itself to handle GTFS reference section links.


core/src/main/java/org/mobilitydata/gtfsvalidator/notice/DuplicateKeyNotice.java
```diff
@@ -15,13 +15,26 @@
...
+import org.mobilitydata.gtfsvalidator.annotation.GtfsValidationNotice.UrlRef;
 
 /**
  * The values of the given key and rows are duplicates.
  *
  * <p>Severity: {@code SeverityLevel.ERROR}
  */
+@GtfsValidationNotice(
+    severity = ERROR,
+    sections = @SectionRefs({"file-requirements"}),
```
davidgamez, line 32: Is it worthwhile having the section ref as enums?

bdferris-v2: I did consider that option.  My thinking at the time was that keeping it as a string would allow easier addition of new section refs without a code change, but maybe in reality that is overblown, since by definition if you are referencing a new section in a @GtfsVaildationNotice, you are making a code change in the Notice file, so updating an additional enum wouldn't be too much additional work.

Thoughts?

davidgamez: I'm concerned that we can have typos in the ref sections, and this can be rendered out to the documentation without notice.

bdferris-v2: Fair enough.  Let me make a change in a separate PR and then update this one to use it (since this PR is an automated change).

bdferris-v2: Actually, I take that back, I made that change locally.  So I've fixed it in this PR.  Please take another look.

davidgamez (approved):
LGTM!

PR #1367, opened 2023-03-30
ci: build & publish linux/arm64 Docker images

derhuerst:
On an ARM64 system, I would like to use native Docker images in order not to run an emulation layer.

This PR extends the Docker build & publishing CI workflow to also build `linux/arm64` images in addition to `linux/amd64` images (implicit default).

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -2,7 +2,10 @@ name: Docker image
 
 on:
   push:
-    branches: [ master ]
+    branches:
+      - master
+      # todo: revert
+      - docker-arm64
   release:
     types: [ prereleased, released ]
 
@@ -114,7 +117,8 @@ jobs:
           file: ./Dockerfile
           build-args: |
             VERSION_TAG=${{ steps.prep.outputs.version }}
-          push: true
+          # todo: revert
+          push: false
           platforms: linux/amd64,linux/arm64
           tags: ${{ steps.prep.outputs.tags }}
           labels: |
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -98,6 +98,7 @@ jobs:
           build-args: |
             VERSION_TAG=${{ steps.prep.outputs.version }}
           load: true
+          platforms: linux/amd64,linux/arm64
           tags: ${{ steps.prep.outputs.tags }}
           labels: |
             org.opencontainers.image.source=https://github.com/${GITHUB_REPOSITORY,,}.git
@@ -114,6 +115,7 @@ jobs:
           build-args: |
             VERSION_TAG=${{ steps.prep.outputs.version }}
           push: true
+          platforms: linux/amd64,linux/arm64
           tags: ${{ steps.prep.outputs.tags }}
           labels: |
             org.opencontainers.image.source=https://github.com/${GITHUB_REPOSITORY,,}.git


derhuerst:
I think it would make sense to temporarily run the Docker build in this PR too, without publishing of course. What do you think?

bdferris-v2:
@derhuerst do you still want to get this checked in?  You'll need to sign the CLA before we can do so.

Regarding running the Docker build against this PR.  I'll admit I'm not totally clear on how to do that, since the Docker workflow doesn't run for PRs by default.  That said, this change seems mostly reasonable, so we could probably check it in as-is and rollback if it causes some issue with the Docker build?  Or did you have a different concern in mind?

.github/workflows/docker.yml
```diff
@@ -2,7 +2,10 @@ name: Docker image
 
 on:
   push:
-    branches: [ master ]
+    branches:
+      - master
+      # todo: revert
+      - docker-arm64
```
derhuerst, line 8: This change should be reverted before merging.

.github/workflows/docker.yml
```diff
@@ -132,7 +136,9 @@ jobs:
           file: ./Dockerfile
           build-args: |
             VERSION_TAG=${{ steps.prep.outputs.version }}
-          push: true
+          # todo: revert
+          push: false
```
derhuerst, line 140: This change should be reverted before merging.

bdferris-v2:
Am I reading correctly that this failed to run on your repo per https://github.com/derhuerst/gtfs-validator/actions/runs/4741545166/jobs/8418790748 ?  Any thoughts?

derhuerst:
Note sure 🤷

bdferris-v2:
There is a long running bug with Docker (https://github.com/docker/buildx/issues/59) where multi-platform builds don't work with a `load`, only a `push`.   I think you can see that in the error from your Docker run:

> ERROR: docker exporter does not currently support exporting manifest lists

I think the solution here would be to only build a single platform for the test Docker build but then enable multi platforms for the push.  Thoughts?

derhuerst:
> I think the solution here would be to only build a single platform for the test Docker build but then enable multi platforms for the push. Thoughts?

Sounds good, as I'm mainly interested in having a native Docker image to pull and use.

davidgamez:
@derhuerst I created a draft PR with an alternative solution based on @bdferris-v2 [comment](https://github.com/MobilityData/gtfs-validator/pull/1367#issuecomment-1516714791). Can I get some feedback on #1528 ? Thanks!

.github/workflows/docker.yml
```diff
@@ -108,6 +111,7 @@ jobs:
           build-args: |
             VERSION_TAG=${{ steps.prep.outputs.version }}
           load: true
+          platforms: linux/amd64,linux/arm64
```
jcfj, line 114: To avoid getting 
```
ERROR: docker exporter does not currently support exporting manifest lists
```
you'll have to either delete this line, or create a third copy of the `docker/build-push-action@v3`, (one with load, one with platforms, one with push).

jcfj:
Btw, why not set `FROM --platform=$BUILDPLATFORM gradle:7-jdk11-alpine AS build` in the `Dockerfile` so the actual building only has to be done once.

davidgamez:
Sorry, wrong PR...  This is the right one, [link](https://github.com/MobilityData/gtfs-validator/pull/1528#issuecomment-4306163019).

davidgamez:
Addressed by #1528

PR #1380, opened 2023-04-06
chore: Add unit tests to AnyTableLoader 

davidgamez:
**Summary:**

Add unit tests to [AnyTableLoader](core/src/main/java/org/mobilitydata/gtfsvalidator/table/AnyTableLoader.java) preparing the class for future refactoring.

**Expected behavior:** 

Unit tests cover error scenarios and a happy path.


core/src/test/java/org/mobilitydata/gtfsvalidator/table/AnyTableLoaderTest.java
```diff
@@ -0,0 +1,216 @@
+package org.mobilitydata.gtfsvalidator.table;
+
+import static com.google.common.truth.Truth.assertThat;
+import static org.junit.Assert.*;
```
bdferris-v2, line 4: While I'll admit it's not fully consistent, the majority of tests in the project use Truth alone and none of the junit assertions.  I'd argue for sticking to Truth here and using its assertion exclusively, instead of a mix.

core/src/test/java/org/mobilitydata/gtfsvalidator/table/AnyTableLoaderTest.java
```diff
@@ -0,0 +1,216 @@
...
+    AnyTableLoader.load(
+        testTableDescriptor,
+        mock(ValidatorProvider.class),
+        retrieveInputStream("_not_a_valid_file_"),
+        loaderNotices);
+
+    assertTrue(loaderNotices.hasValidationErrors());
+
+    assertValidationNotice(loaderNotices, CsvParsingFailedNotice.class);
+    assertEquals(1, loaderNotices.getValidationNotices().size());
```
bdferris-v2, line 55: In general, I'm not a big fan of test assertion helpers where the assertion failure is in a test helper method.  I recognize why it's tricky to validation the notices directly here, but I'd propose a helper like the following:

```
// In TestUtils...
public static List<Class<? extends ValidationNotice>> validationNoticeTypes(NoticeContainer notices) {
  return notices.getValidationNotices().stream().map(Object::getClass).collect(toList());
}
```

And then the three test lines on 52, 54, and 55 could replaced with a single line:

```
assertThat(validationNoticeTypes(loaderNotices)).containsExactly(CsvParsingFailedNotice.class);
```

that would be more expressive in the failure case and trigger within the test itself.

I think

core/src/test/java/org/mobilitydata/gtfsvalidator/table/AnyTableLoaderTest.java
```diff
@@ -0,0 +1,216 @@
...
+    verify(testTableDescriptor, times(0)).createContainerForHeaderAndEntities(any(), any(), any());
```
bdferris-v2, line 56: I think line is maybe overkill.  I'd suggest instead you do something like:

```
GtfsTableContainer mockContainer = mock(GtfsTableContainer.class);
when(testTableDescriptor.createContainerForInvalidStatus(
              GtfsTableContainer.TableStatus.INVALID_HEADERS))
    .thenReturn(mockContainer);
```

and then verify the return value of `AnyTableLoader.load()` ala `assertThat(loadedContainer).isEqualTo(mockContainer);`

This has the benefit verify that the correct type of table was created for this situation.

core/src/test/java/org/mobilitydata/gtfsvalidator/TestUtils.java
```diff
@@ -0,0 +1,33 @@
...
+
+public class TestUtils {
+
+  protected TestUtils() {
+    //        Hiding default constructor to avoid instantiation
+  }
+
+  public static InputStream retrieveInputStream(String path) {
+    return TestUtils.class.getResourceAsStream(path);
+  }
```
bdferris-v2, line 19: I think this is only used in one-place and for a file that DNE?  I'd get rid of this helper as such.

core/src/test/java/org/mobilitydata/gtfsvalidator/table/AnyTableLoaderTest.java
```diff
@@ -0,0 +1,216 @@
...
+  }
+
+  @Test
+  public void emptyInputStream() {
+    var testTableDescriptor = mock(GtfsTableDescriptor.class);
+    when(testTableDescriptor.gtfsFilename()).thenReturn("filename");
+    NoticeContainer loaderNotices = new NoticeContainer();
+    InputStream csvInputStream = new ByteArrayInputStream("".getBytes(StandardCharsets.UTF_8));
```
bdferris-v2, line 64: Do you want to use your string-to-InputStream helper here?

core/src/test/java/org/mobilitydata/gtfsvalidator/table/AnyTableLoaderTest.java
```diff
@@ -0,0 +1,216 @@
...
+    verify(validatorProvider, times(0)).createSingleFileValidators(any());
+  }
+
+  @Test
+  public void invalidHeaders() {
+    var testTableDescriptor = mock(GtfsTableDescriptor.class);
+    when(testTableDescriptor.gtfsFilename()).thenReturn("filename");
+    GtfsColumnDescriptor[] columns =
+        new GtfsColumnDescriptor[] {
+          AutoValue_GtfsColumnDescriptor.builder()
```
bdferris-v2, line 83: It's usually a bad sign if an `AutoValue_` impl class is used outside of the class marked with `@AutoValue` itself.  Prefer to use `GtfsColumDescriptor.builder()` instead.

core/src/test/java/org/mobilitydata/gtfsvalidator/table/AnyTableLoaderTest.java
```diff
@@ -0,0 +1,216 @@
...
+        .containsExactly(new EmptyFileNotice("filename"));
+    verify(testTableDescriptor, times(0)).createContainerForHeaderAndEntities(any(), any(), any());
+    verify(validatorProvider, times(0)).createSingleFileValidators(any());
+  }
+
+  @Test
+  public void invalidHeaders() {
+    var testTableDescriptor = mock(GtfsTableDescriptor.class);
+    when(testTableDescriptor.gtfsFilename()).thenReturn("filename");
+    GtfsColumnDescriptor[] columns =
```
bdferris-v2, line 81: Just use `ImmutableList.of(...)` instead here?

core/src/test/java/org/mobilitydata/gtfsvalidator/table/AnyTableLoaderTest.java
```diff
@@ -0,0 +1,216 @@
...
+              .setFieldLevel(FieldLevelEnum.REQUIRED)
+              .setIsMixedCase(false)
+              .setIsCached(false)
+              .build()
+        };
+    when(testTableDescriptor.getColumns()).thenReturn(ImmutableList.copyOf(columns));
+    ValidatorProvider validatorProvider =
+        spy(
+            new DefaultValidatorProvider(
+                mock(ValidationContext.class), mock(ValidatorLoader.class)));
```
bdferris-v2, line 95: Why do you need to spy on this?

core/src/test/java/org/mobilitydata/gtfsvalidator/table/AnyTableLoaderTest.java
```diff
@@ -0,0 +1,216 @@
...
+    var testTableDescriptor = mock(GtfsTableDescriptor.class);
+    when(testTableDescriptor.gtfsFilename()).thenReturn("filename");
+    NoticeContainer loaderNotices = new NoticeContainer();
+    InputStream csvInputStream = new ByteArrayInputStream("".getBytes(StandardCharsets.UTF_8));
+    ValidatorProvider validatorProvider = mock(ValidatorProvider.class);
+
+    AnyTableLoader.load(testTableDescriptor, validatorProvider, csvInputStream, loaderNotices);
+
+    assertTrue(loaderNotices.hasValidationErrors());
+    assertEquals(1, loaderNotices.getValidationNotices().size());
```
bdferris-v2, line 70: These two test lines don't really add anything beyond what is tested on line 71.  I'd drop them.

core/src/test/java/org/mobilitydata/gtfsvalidator/testgtfs/GtfsStop.java
```diff
@@ -16,13 +16,296 @@
...
 package org.mobilitydata.gtfsvalidator.testgtfs;
 
+import com.google.common.geometry.S2LatLng;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 import org.mobilitydata.gtfsvalidator.table.GtfsEntity;
+import org.mobilitydata.gtfsvalidator.table.GtfsEntityBuilder;
 
 /** Test class to avoid dependency on the real GtfsStop and annotation processor. */
 public class GtfsStop implements GtfsEntity {
```
bdferris-v2, line 26: Why did you need to expand GtfsStop to contain all fields?

In general, I'd actually be in favor of renaming these `GtfsStop` test classes to something like `GtfsTestEntity` (with a minimal set of test fields) to make it clearer that they are not to be confused with the actual `GtfsStop` class.

core/src/test/java/org/mobilitydata/gtfsvalidator/table/AnyTableLoaderTest.java
davidgamez: Good suggestion!

core/src/test/java/org/mobilitydata/gtfsvalidator/TestUtils.java
davidgamez: I added it with the idea of using it to load _tests cases_  files but ended up loading the content using the string version of the files inline as a better alternative. Removing it for now.

core/src/test/java/org/mobilitydata/gtfsvalidator/testgtfs/GtfsStop.java
davidgamez: I expanded this class with some fields, not all of them, to be able to parse a valid file without having to mock a lot of methods. Too many mocks increase the efforts of maintaining the test code. I agree that the class name can be misleading, I'll change it to a more appropriate name like `GtfsTestEntity`. I'll also reduce the fields and code within the class.

core/src/test/java/org/mobilitydata/gtfsvalidator/table/AnyTableLoaderTest.java
davidgamez: I agree with the change requested. I'm leaving one assertion as it's testing the invocation of the validators https://github.com/MobilityData/gtfs-validator/blob/0d62e9450d93348149e2fded4da9894742766265/core/src/test/java/org/mobilitydata/gtfsvalidator/table/AnyTableLoaderTest.java#L199. I'm open to suggestions if there is a better way, in this case, to guarantee that the execution of the validators.

core/src/test/java/org/mobilitydata/gtfsvalidator/table/AnyTableLoaderTest.java
```diff
@@ -0,0 +1,239 @@
...
+    when(testTableDescriptor.createContainerForInvalidStatus(
+            GtfsTableContainer.TableStatus.INVALID_HEADERS))
+        .thenReturn(mockContainer);
+    NoticeContainer loaderNotices = new NoticeContainer();
+
+    var loadedContainer =
+        AnyTableLoader.load(
+            testTableDescriptor, mock(ValidatorProvider.class), null, loaderNotices);
+
+    assertThat(loaderNotices.hasValidationErrors()).isTrue();
```
bdferris-v2, line 55: Echoing my previous comment, I'd still argue for dropping this `hasValidationErrors` check, both here and below.  It's duplicative of the next check, but if it fails, it doesn't give you as much information as the `validationNoticeTypes` check on the next line.  Thus, favor the next check.

core/src/test/java/org/mobilitydata/gtfsvalidator/table/AnyTableLoaderTest.java
bdferris-v2: So no longer necessary to mock it?

Also, you seem to use different combinations of mocked ValidatorProvider vs DefaultValidatorProvider here and in other tests.  Possible just to use DefaultValidatorProvider, possibly created in @Before section along with context?

core/src/test/java/org/mobilitydata/gtfsvalidator/table/AnyTableLoaderTest.java
```diff
@@ -0,0 +1,239 @@
...
+
+  @Test
+  public void invalidRowLengthNotice() {
+    var testTableDescriptor = spy(new GtfsTestTableDescriptor());
+    GtfsTableContainer mockContainer = mock(GtfsTableContainer.class);
+    when(testTableDescriptor.createContainerForInvalidStatus(
+            GtfsTableContainer.TableStatus.UNPARSABLE_ROWS))
+        .thenReturn(mockContainer);
+    ValidatorProvider validatorProvider =
+        spy(new DefaultValidatorProvider(validationContext, ValidatorLoader.createEmpty()));
```
bdferris-v2, line 127: Again, does this need to be a spy?

core/src/test/java/org/mobilitydata/gtfsvalidator/table/AnyTableLoaderTest.java
```diff
@@ -0,0 +1,239 @@
...
+        AnyTableLoader.load(testTableDescriptor, validatorProvider, inputStream, loaderNotices);
+
+    assertThat(loaderNotices.hasValidationErrors()).isTrue();
+    assertThat(loaderNotices.getValidationNotices())
+        .containsExactly(new InvalidRowLengthNotice("filename.txt", 2, 1, 2));
+    assertThat(loadedContainer).isEqualTo(mockContainer);
+  }
+
+  @Test
+  public void missingRequiredColumn() {
```
bdferris-v2, line 141: This seems like a unit-test of `DefaultTableHeaderValidator` and its presence in DefaultValidatorProvider.  While that may be a reasonable thing to test, does it belong in `AnyTableLoader` unit test?

core/src/test/java/org/mobilitydata/gtfsvalidator/table/AnyTableLoaderTest.java
```diff
@@ -0,0 +1,239 @@
...
+        AnyTableLoader.load(testTableDescriptor, validatorProvider, inputStream, loaderNotices);
+
+    assertThat(loaderNotices.hasValidationErrors()).isTrue();
+    assertThat(loaderNotices.getValidationNotices())
+        .containsExactly(new MissingRequiredColumnNotice("filename.txt", "code"));
+    assertThat(loadedContainer).isEqualTo(mockContainer);
+  }
+
+  @Test
+  public void parsableTableRows() {
```
bdferris-v2, line 179: Ditto comment above about `DefaultTableHeaderValidator`.  I think if you want to mock `ValidatorProvider` entirely and verify that different types of validators are created and called (like you do with the mocked `GtfsTestFileValidator` in this test), that's reasonable but I wouldn't mix in the behavior of `DefaultValidatorProvider`, which could change independently of `AnyTableLoader`.

core/src/test/java/org/mobilitydata/gtfsvalidator/validator/ValidatorLoaderTest.java
```diff
@@ -44,10 +43,10 @@ public class ValidatorLoaderTest {
 
   @Test
   public void createValidatorWithContext_injectsContext() throws ReflectiveOperationException {
-    GtfsStopTableContainer stopTable =
-        new GtfsStopTableContainer(TableStatus.EMPTY_FILE, CsvHeader.EMPTY);
-    StopEntityValidator validator =
-        ValidatorLoader.createValidatorWithContext(StopEntityValidator.class, VALIDATION_CONTEXT);
+    //    GtfsTestTableContainer table = new GtfsTestTableContainer(TableStatus.EMPTY_FILE);
```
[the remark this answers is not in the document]
davidgamez: I noticed that was not in use and commented on the line to test it, I forgot to come back and delete the line.

core/src/test/java/org/mobilitydata/gtfsvalidator/table/AnyTableLoaderTest.java
davidgamez: Removed the test in favor of a more generic test _invalidHeaders_. Also, removed all dependencies with DefaultTableHeaderValidator and DefaultValidatorProvider.

core/src/test/java/org/mobilitydata/gtfsvalidator/table/AnyTableLoaderTest.java
davidgamez: Removed all dependencies with DefaultTableHeaderValidator and DefaultValidatorProvider.

bdferris-v2 (approved):

PR #1389, opened 2023-04-17
feat: Add metadata to report

### LINKED ISSUE #1363: Add metadata in the validator report
### Describe the problem

Metadata is a piece of key information, it can provide a quick overview of what is included in the dataset:
- the size: number of stops, routes, trips
- the files that are present (if `frequency.txt` is present -> it's a [frequency-based](https://gtfs.org/schedule/examples/frequencies/) service)
- information about the agency (name, phone number & e-mail if provided, and the publisher)
- the spec "features" present (fares v1, fares v2, pathways)

It can be used to answer:
- How much work is it going to take to ingest this data?
- How many datasets contain Fares v2? (assumed if running this validator on a group of datasets)
- Is this a frequency-based trip? 
- Are shapes included?
- etc

### Proposed solution

Add a section in the report that contains metadata (it could have a less jargon-y name, such as "Overview").

For efficiency when the validator is used via the command line, We could allow users to run only the portion that triggers the list of notices, or the portion that generates the metadata.

The content I am proposing for this section is inspired by a series of user interviews, and by looking at other validators that exist. This will evolve with time, but I suggest we start with:

1. file presence: what files are present in the dataset?

2. "counts": number of agencies, routes, trips, stops, shapes, blocks (using the `id` fields)
  
3. Feed info (from `feed_info.txt`, which is optional so not always provided)
  - start and end date (when provided)
  - publisher name (required)
  - publisher URL (required)
  - language (required)

4. information about the agencies (from `agency.txt`, which is required)
  - name (required)
  - url (required)
  - phone number (when provided)
  - email (when provided)

5. presence of spec features (based on [this script](https://github.com/MobilityData/mobility-database-catalogs/blob/main/update_gtfs_schedule_sources.py) from Mobility Database)
  - fares-v1 
  - fares-v2
  - flex-v1
  - flex-v2
  - pathways

### Other information could be added, where additional research is needed.

6. More "features" (I want to confirm how to define these ones, how many we need, and how they relate to fares-v1 and fares-v2 concepts).
  - wheelchair-accessible fields
  - bike-allowed field
  - etc

7. More calendar information
  - first and last day of service 

8. Provide detailed information per agency
  -  for example: the first and last service day for each agency in the dataset. I am especially inspired by how transport.data.gouv displays this, here is what it looks like for MTC's dataset that contains many agencies.

Screenshot 2023-03-24 at 5 46 14 PM

Add this type of thing as an `INFO` validator notice. 
It would have been a reasonable thing to do also, especially because `INFO` isn't well defined, but I think `INFO` gives us a place to put the "not technically a spec violation but FYI you might want to have a look" type of things, such as unknown files and columns, or unused stations.

I think this type of information deserves a new section. Happy to be challenged if anyone thinks this should belong alongside the validator notices.

### Additional context

_No response_

briandonahue:
**Summary:**

Closes #1363.

Add a metadata / overview section to the HTML report.

**Expected behavior:** 


briandonahue:
@isabelle-dr This is still a work in progress, and obviously the report needs better organization, but wanted to show you the data as it's currently being output and see if you had any specific requests for organizing it on the report:

image

For the "presence of spec features" section, how do you envision that looking?

briandonahue:
Updated Screen shot: 
<a href="https://user-images.githubusercontent.com/40084/232911218-d3748ff6-3575-4e11-b5c4-77a94e5f53bc.png" target="_blank">image</a>

briandonahue:
@bdferris-v2 @davidgamez Leaving this a draft for now, but looking for any feedback/suggestions on the approach and/or the report organization/layout

isabelle-dr:
As linked to the GitHub issue, [here is a scoping doc](https://docs.google.com/document/d/157X9lMTkZHH7oYxK1fYPiY-rMh_VWyPMW_s9wImOP-I/edit#) for this PR.

briandonahue:
Getting a little closer.  May need to call in some CSS support for a more responsive design :) 
image

isabelle-dr:
This is beautiful (and why so many shapes in this dataset? 😅)

briandonahue:
@isabelle-dr 

> why so many shapes in this dataset? 😅 is 1171 the number of unique shape_ids or the number of records?

Yes!  I fixed it.  Are all the entities under `Counts` supposed to count unique IDs?  Right now only Blocks and Shapes are doing so.

> You can remove the section at the end (starting at "parameters used")

Yes!


I also have updated the HTML with the help of my BFF ChatGPT. The bottom (original) compliance report is no longer affected by the Summary CSS, and the Summary columns stack as you narrow the browser window.  

Full width:

image

Narrower width:

image

It is even pretty viewable on a mobile-size width, but the bottom table currently does not collapse so it will require horizontal scrolling 😄 

I think this is now ready for review!

isabelle-dr:
How nice @briandonahue (and thank you Chat GPT)!

> Are all the entities under Counts supposed to count unique IDs? Right now only Blocks and Shapes are doing so.

Yes, we want the # of unique ID fields for the counts. It shouldn't make a difference for the others because each record should have a unique ID, but we could have duplicates so I'd rather look at the unique IDs all the time. 

Why are we having all these corrupted sources in the acceptance test report all of a sudden? cc @davidgamez

briandonahue:
> Yes, we want the # of unique ID fields for the counts. It shouldn't make a difference for the others because each record should have a unique ID, but we could have duplicates so I'd rather look at the unique IDs all the time.

@isabelle-dr I've updated it to always count unique IDs and simplified the code a bit to use generics (@davidgamez open to feedback on this if there is a cleaner, more concise approach).

isabelle-dr:
@briandonahue we're getting there!
In order to test this PR, could you share the dataset you've been using to generate these reports?

Another question I have is: under the "Files included", are you listing all of them or only the official GTFS files? 
Maybe it wasn't clear in the Requirements doc, we want all files listed there.

main/src/main/java/org/mobilitydata/gtfsvalidator/report/model/FeedMetadata.java
```diff
@@ -0,0 +1,131 @@
...
+  private Map<String, TableMetadata> tableMetaData;
+  public Map<String, Integer> counts = new TreeMap<>();
+
+  public Map<String, String> feedInfo = new LinkedHashMap<>();
+
+  public Map<String, String> specFeatures = new LinkedHashMap<>();
+
+  public ArrayList<AgencyMetadata> agencies = new ArrayList<>();
+
+  public static FeedMetadata from(GtfsFeedContainer feedContainer) {
```
davidgamez, line 17: suggestion: this method populates all the necessary fields and counts. I suggest making the class-default constructor protected to avoid instantiation and prevent unexpected behaviors while calling methods in the class without proper field initialization.

main/src/main/java/org/mobilitydata/gtfsvalidator/runner/ValidationRunner.java
```diff
@@ -121,14 +118,17 @@ public Status run(ValidationRunnerConfig config) {
       feedContainer =
           loadAndValidate(
               validatorLoader, feedLoader, noticeContainer, gtfsInput, validationContext);
+      // TODO: collect metadata from feedContainer and include in report.
+
```
davidgamez, line 122: question: is this been addressed in line 127?

davidgamez (approved):
LGTM!

briandonahue:
> @briandonahue we're getting there! In order to test this PR, could you share the dataset you've been using to generate these reports?

Yes, these were 2 I was testing with:
[Westchester_County_Bee_Line.zip](https://github.com/MobilityData/gtfs-validator/files/11490480/Westchester_County_Bee_Line.zip) 
[alhambra-ca-us.zip](https://github.com/MobilityData/gtfs-validator/files/11490473/alhambra-ca-us.zip)

> Another question I have is: under the "Files included", are you listing all of them or only the official GTFS files? Maybe it wasn't clear in the Requirements doc, we want all files listed there.

I was not, but I've updated it now to do so.

/cc @davidgamez

isabelle-dr:
@briandonahue can you share a screenshot of your latest version for [alhambra-ca-us.zip](https://github.com/MobilityData/gtfs-validator/files/11490473/alhambra-ca-us.zip)?

briandonahue:
> @briandonahue can you share a screenshot of your latest version for [alhambra-ca-us.zip](https://github.com/MobilityData/gtfs-validator/files/11490473/alhambra-ca-us.zip)?

@isabelle-dr Here you go 😄 

image

isabelle-dr:
Perfect! Merging

PR #1393, opened 2023-04-20
fix: add field value to mixed case notice

### LINKED ISSUE #1392: Mixed case validation notice schema is incomplete
### Describe the problem

Mixed case validation was introduced in #1340.
In the validation report, we don't see the record values that trigger this notice, only the number of the row (see image below).
Seeing the actual value here would encourage the user to fix the issue because they would see what the rider would see.
Additionally, the `filename` name doesn't follow our naming conventions, it should be `fileName`.

Screenshot 2023-04-19 at 6 42 11 PM

### Proposed solution

I recommend adding a `FieldValue` in the schema that displays the value of the faulty record, and changing `filename` to `fileName` to follow our naming conventions.

 

Status quo

### Additional context

_No response_

briandonahue:
**Summary:**

Closes #1392

Add field value to notice.  Note, does not change casing of `filename` field as everywhere else is using the lowercase version.  

**Expected behavior:** 

Should show the offending value.

### DIFF
--- a/RULES.md
+++ b/RULES.md
@@ -2189,11 +2189,12 @@ This field contains customer-facing text and should use Mixed Case (upper and lo
 
 
 #### Notice fields description
-| Field name     	| Description                           	| Type   	|
-|----------------	|---------------------------------------	|--------	|
-| `csvRowNumber` 	| The row number of the faulty record.  	| Long   	|
-| `filename`    	| Name of the faulty file.              	| String 	|
-| `fieldName`    	| Name of the faulty field.              	| String 	|
+| Field name     	 | Description                           	| Type   	|
+|------------------|---------------------------------------	|--------	|
+| `csvRowNumber`   | The row number of the faulty record.  	| Long   	|
+| `filename`       | Name of the faulty file.              	| String 	|
+| `fieldName`      | Name of the faulty field.              	| String 	|
+| `fieldValue`     | Name of the faulty field.              	| String 	|
 
 #### Affected files & fields
 * [`agency.agency_name`](https://gtfs.org/schedule/reference/#agencytxt)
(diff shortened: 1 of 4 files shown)


briandonahue:
Leaving this as a draft for now as it has some uncommitted changes that will be added in another PR, that will fix #1394

briandonahue:
@isabelle-dr since that bug is resolved, this is ready for review. Let me know your thoughts on the `filename` casing as mentioned [here](https://github.com/MobilityData/gtfs-validator/issues/1392#issuecomment-1515555085)

isabelle-dr:
Will have a look shortly, thank you. 
We'll also need a review from @davidgamez :)

web/service/build.gradle
```diff
@@ -49,7 +49,7 @@ dependencies {
 	implementation 'com.google.cloud:spring-cloud-gcp-starter-storage'
   implementation 'org.projectlombok:lombok:1.18.22'
   implementation 'org.projectlombok:lombok:1.18.22'
-  implementation 'io.sentry:sentry-spring:1.7.30'
+  implementation 'io.sentry:sentry-spring-boot-starter:6.17.0'
```
davidgamez, line 52: question: do we need this change as part of the PR? It seems unrelated to the scope.

web/service/src/main/java/org/mobilitydata/gtfsvalidator/web/service/Main.java
```diff
@@ -17,24 +17,11 @@
...
-
-  @Bean
-  public HandlerExceptionResolver sentryExceptionResolver() {
-    return new io.sentry.spring.SentryExceptionResolver();
-  }
-
-  @Bean
-  public ServletContextInitializer sentryServletContextInitializer() {
-    return new io.sentry.spring.SentryServletContextInitializer();
-  }
```
davidgamez, line 39: Same as the above comment. It looks unrelated to the scope of the PR.

briandonahue:
@davidgamez I think this is fixed now. I think I had to pull in some of Kevin's changes to fix an issue I was having, but then afterward could not replicate the issue, so now I've removed the changes, and this should just be the original PR changes.

davidgamez:
> @davidgamez I think this is fixed now. I think I had to pull in some of Kevin's changes to fix an issue I was having, but then afterward could not replicate the issue, so now I've removed the changes, and this should just be the original PR changes.

Great, thanks! Approving now!

davidgamez (approved):

PR #1396, opened 2023-04-21
feat: integrate Cypress for E2E testing

mijogu:


web/client/cypress/e2e/validator.cy.js
```diff
@@ -0,0 +1,93 @@
...
+  });
+
+  it('Validate GTFS .zip file', () => {
+    const url = 'https://developers.google.com/static/transit/gtfs/examples/sample-feed.zip';
+    const jobId = '8f6be6fb-1fee-41f8-b401-b2b4b552e177';
+
+    // Setup intercept aliases
+    cy.intercept(
+        'POST',
+        'https://gtfs-validator-web-mbzoxaljzq-ue.a.run.app/create-job',
```
themightychris, line 17: Can we use a wildcard here like `https://*/create-job`? This hostname looks unstable and there probably aren't other requests to `/create-job` on this page that we'd need to disambiguate from with the hostname

web/client/cypress/e2e/validator.cy.js
```diff
@@ -0,0 +1,93 @@
...
+        (req) => {
+          req.headers['x-custom-headers'] = 'added by cy.intercept';
+          req.reply({
+            statusCode: 200,
+            statusMessage: "OK",
+            headers: {
+              "x-custom-header": 'mgunn',
```
themightychris, line 24: Does this need to remain here?

web/client/cypress/e2e/validator.cy.js
```diff
@@ -0,0 +1,93 @@
...
+            }
+          });
+        }
+      )
+      .as('createJob');
+
+
+    cy.intercept(
+      'HEAD',
+      'https://gtfs-validator-results.mobilitydata.org/*/report.html',
```
themightychris, line 37: Let's use `${jobId}` here instead of `*` so we're asserting that the `jobId` being returned by `/create-job` is the one the client is going after

web/client/cypress/e2e/validator.cy.js
mijogu: removed

web/client/cypress/e2e/validator.cy.js
mijogu: i was running into an issue here bc the first 404 HEAD response doesn't include the `jobId`, it includes `_waiting_`. But i've updated the test to assert that the `jobId` is found where expected.

.github/workflows/cypress_e2e_tests.yml
```diff
@@ -0,0 +1,23 @@
...
+
+jobs:
+  cypress-run:
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Install Cypress
+        run: npm i cypress@^12.10
```
davidgamez, line 17: Do we need this command here if the `package.json` contains the `cypress` dependency? Using the `package.json` is preferable to avoid conflicts in the future.

web/client/cypress/support/commands.js
```diff
@@ -0,0 +1,37 @@
+/// <reference types="cypress" />
```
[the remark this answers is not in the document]
davidgamez: I see. To me, commented code can quickly become obsolete with every library upgrade. What is valuable information today can be noisy a year from now. If there is no immediate plan to use it, I prefer to clean up the repository. We can always return to the official documentation to achieve a particular task.

.github/workflows/cypress_e2e_tests.yml
mijogu: Removed that and installing packages from package.json instead.

davidgamez (approved):
Great Work! 💪

web/client/cypress/support/commands.js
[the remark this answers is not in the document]
mijogu: Removed everything in this file. I did try deleting both `e2e.js` and `commands.js`, but Cypress requires the `e2e.js` file, so I kept it importing the empty `commands.js` file.

PR #1413, opened 2023-04-29
feat: Add unit-test check for @GtfsValidationNotice annotation and notices that I missed

bdferris-v2:
This adds a new `NoticeAnnotationTest` to make sure each validation notice is annotated with @GtfsValidationNotice.  It also adds @GtfsValidationNotices to a number of notices that I missed in the first pass:
* SystemErrors
* Files with more than one child Notice class (my script only caught the first notice in the file)


davidgamez (approved):
LGTM!

PR #1427, opened 2023-05-02
feat: Add a few missing "Affected files" references

bdferris-v2:
Reliably parsing `RULES.md` is an inexact science and I missed a few "Affected files" references in my first automated pass due to inconsistent formatting in `RULES.md`.  This PR adds them in.  Part of #1324.

### DIFF
--- a/core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NumberOutOfRangeNotice.java
+++ b/core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NumberOutOfRangeNotice.java
@@ -15,6 +15,7 @@
  */
 package org.mobilitydata.gtfsvalidator.notice;
 
+import static org.mobilitydata.gtfsvalidator.annotation.GtfsValidationNotice.SectionRef.DATASET_FILES;
 import static org.mobilitydata.gtfsvalidator.annotation.GtfsValidationNotice.SectionRef.FILED_TYPES;
 import static org.mobilitydata.gtfsvalidator.annotation.GtfsValidationNotice.SectionRef.FILE_REQUIREMENTS;
 import static org.mobilitydata.gtfsvalidator.notice.SeverityLevel.ERROR;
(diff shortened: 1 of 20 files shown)


isabelle-dr:
One question that is probably out of scope for this PR: 
During the contributor meeting, we talked about merging the two concepts of "files affected" and "references" into one category. 
In this PR it looks like we are going to keep the two categories via `@FileRefs` and `@UrlRef, correct?

bdferris-v2:
@isabelle-dr a couple points:

* `@FileRefs` is used to capture all GTFS file links, whether they were in "files affected" or "references".
* `@UrlRef` is used to capture links to arbitrary webpages not already covered by the existing spec.  You can see it used in this PR for a link to the "Original Python validator implementation".  We don't want someone using it to link to a section of the GTFS spec or a GTFS file reference (we could probably enforce that with a unit-test?)

Does that make sense?

There is a bit of an edge-case where for the "https://gtfs.org/best-practices/#dataset-publishing--general-practices" link.  I suppose I could update that to be a section ref?

As for @FileRefs vs @UrlRef, the difference is intentional.

Namely, @FileRefs can contain an array of values, such that you can specify:

```
files = @FileRefs({GtfsStopTimeSchema.class, GtfsStopSchema.class})
```

I didn't want to go with `@FileRef` here because it seemed more verbose to have to do something like:

```
files = {@FileRef(GtfsStopTimeSchema.class), @FileRef(GtfsStopSchema.class)}
```

Compare that to `@UrlRef`, which has two distinct arguments:

```
@UrlRef(
            label = "Original Python validator implementation",
            url = "https://github.com/google/transitfeed")
```

It doesn't make as much sense to do `@UrlRefs` because it's not possible to construct multiple different URLs in the specification.  Thus, I did go with individual `@UrlRef` instead.

Looking back, I think I had imagined adding more fields to `@FileRefs` but instead, that info ended up in `@SectionRefs` instead.  We could probably refactor the code to look something like:

```
files = {GtfsStopTimeSchema.class, GtfsStopSchema.class}
```

getting rid of `@FileRefs` entirely.

Issue #1423, opened 2023-05-01
Only warn if feed_contact_email or url are both unset.


takohei:
### Describe the bug

Best Practice recommends that at least one of `feed_contact_email` and `feed_contact_url` be provided in `feed_info.txt`.
https://gtfs.org/schedule/best-practices/#feed_infotxt

However, Validator raise a `missing_recommended_field` warning if either the email or the url is not provided.

### Steps/Code to Reproduce

Validate test feed.
https://github.com/MobilityData/gtfs-validator/files/11365074/akocity_with_email.zip

Tested version: 4.0.0

### Expected Results

No `missing_recommended_field` warning.

It is inappropriate to recommend that both be set, since some agencies only publish either an email address or a web form as their contact information.

### Actual Results

A "missing_recommended_field" raised as follows

filename | csvRowNumber
-- | --
"feed_info.txt" | 2

### Files used

[akocity_with_email.zip](https://github.com/MobilityData/gtfs-validator/files/11365074/akocity_with_email.zip)

### Validator version

4.0.0

### Operating system

Windows

### Additional notes

Warnings for missing recommended fields in `feed_info.txt` was requested. #885 

In response to that request, the recommended annotation feature (#1149) and the `missing_recommended_field` warning (#1157) were added. 

At that time, recommended annotations were added to `feed_contact_email` and `feed_contact_url`, so that each is checked independently. #1156

It would be better to remove the recommended annotations for `feed_contact_email` and `feed_contact_url` and then add a check that at least one of them is set.

isabelle-dr:
Hello @takohei, thank you for opening this issue and providing all this context! 

You're right, at least one of `feed_contact_email` and `feed_contact_url` should be provided in `feed_info.txt`, the validator should only give a warning if the two are missing. 
Thank you for flagging this, our team will look at this issue shortly.

l646505418:
Hi, @isabelle-dr ,for this issue, maybe can add a specific function after `load` 
![issue](https://user-images.githubusercontent.com/50475131/236162760-d17bcf14-ec37-4456-b490-28f6e2efbebb.png)
to check the `NoticeContainer` and adjust `validationNotice` accordingly.

l646505418:
https://github.com/MobilityData/gtfs-validator/pull/1434#issue-1707725267

PR #1434, opened 2023-05-12
Issue/1423/only warn both unset

l646505418:
**Summary:**
related issue (https://github.com/MobilityData/gtfs-validator/issues/1423)
Add a `NoticeFilter` interface that can be implemented to filter validation notices.
Add a `filterNotice` method in `NoticeContainer` to filter notices that are produced after validation.
**Expected behavior:** 
Only `feed_contact_email` and `feed_contact_url` are all unset will warn `missing_recommended_field` for `feed_info.txt`.


davidgamez:
Hi @l646505418, thanks for your contribution. Reading #1423 and the best practices, re-using the _MissingRecommendedFieldNotice_ is not clear enough in this case.  The [_MissingRecommendedFieldNotice_](https://github.com/MobilityData/gtfs-validator/blob/master/core/src/main/java/org/mobilitydata/gtfsvalidator/notice/MissingRecommendedFieldNotice.java) only covers one field within a file. I suggest creating a new notice to cover multiple fields(or just two). To accomplish this, we can create a validator's class inheriting from [SingleEntityValidator](https://github.com/MobilityData/gtfs-validator/blob/master/core/src/main/java/org/mobilitydata/gtfsvalidator/validator/SingleEntityValidator.java) interface to trigger the notice when both fields are missing. This can give the consumer more insight into the issue and how to fix it. An example of a validator implementation that looks for two fields within an _entity_ can be found here, [ RouteColorContrastValidator](https://github.com/MobilityData/gtfs-validator/blob/master/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/RouteColorContrastValidator.java)

qcdyx:
Hello, thanks for your contribution. After discussing with David, we decided to implement a FeedContactValidator class which has a static class member MissingFeedContactEmailAndUrlNotice. Please see details in this PR:
https://github.com/MobilityData/gtfs-validator/pull/1514

PR #1438, opened 2023-05-15
fix: Use correct import for @Nullable annotation.

bdferris-v2:
I accidentally included the wrong @Nullable annotation on a new class (used `org.jetbrains.annotations.Nullable` when I should have used `javax.annotation.Nullable`).  We should be consistent with the rest of the codebase.

### DIFF
--- a/model/build.gradle
+++ b/model/build.gradle
@@ -6,6 +6,7 @@ dependencies {
     annotationProcessor 'com.google.auto.value:auto-value:1.7.4'
     compileOnly 'com.google.auto.value:auto-value-annotations:1.7.4'
     implementation 'org.jetbrains:annotations:20.1.0'
+    implementation 'com.google.code.findbugs:jsr305:3.0.2'
 }
 
 compileJava {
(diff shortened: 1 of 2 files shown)


model/build.gradle
```diff
@@ -6,6 +6,7 @@ dependencies {
     annotationProcessor 'com.google.auto.value:auto-value:1.7.4'
     compileOnly 'com.google.auto.value:auto-value-annotations:1.7.4'
     implementation 'org.jetbrains:annotations:20.1.0'
+    implementation 'com.google.code.findbugs:jsr305:3.0.2'
```
davidgamez, line 9: question: why do we need to import this library?

bdferris-v2: The @Nullable annotation is not included in the JDK by default so it has to be included via a dependency.  `com.google.code.findbugs:jsr305` is actually where we've been getting this annotation in `core` and `main`, but it was a transitive dependency via `com.google.guava:guava`.  Since `model` doesn't need Guava, I just imported the dependency directly.

davidgamez (approved):

PR #1439, opened 2023-05-15
fix: Modified to accept mime type encountered on Windows

### LINKED ISSUE #1404: Error "Sorry, only ZIP files are supported at this time." when uploading zip file
### Describe the bug

When uploading a valid GTFS file, the file is rejected as it is not considered a GTFS file.

### Steps/Code to Reproduce

- Navigate to https://gtfs-validator.mobilitydata.org/
- Either use the upload button or drag-and-drop a zip file (both produce the same result)
- The javascript console shows 'file type error application/x-zip-compressed'
- The user interface shows "Sorry, only ZIP files are supported at this time."

### Expected Results

The file is accepted

### Actual Results

The file is rejected as "not a zip file"

### Screenshots

Screenshot

### Files used

[skane.zip](https://github.com/MobilityData/gtfs-validator/files/11354682/skane.zip)

### Validator version

1680987682907

### Operating system

Windows 10

### Java version

_No response_

### Additional notes

Using Firefox 112.0.2 with uBlock Origin installed. Disabling uBlock Origin does not make any difference.

jcpitre:
Fixes #1404 On Windows browsers the mime-type of a zip file is _application/x-zip-compressed_, and we were just allowing mime-type _application/zip_. Modified to allow the  _application/x-zip-compressed_

A note about how I tested this PR:

- Followed https://github.com/MobilityData/gtfs-validator/blob/master/web/client/README.md
- Since only the client is to be tested, no need to change apiRoot, meaning the service will be the production service
- I could not install this on a Windows machine, so I just used the Windows browser to load the page residing on my development machine. To do that you need to tell npm to allow access from a machine on the network instead of only localhost. Use
   - `npm run dev -- --host `


emmambd:
This is great work @jcpitre! Let's discuss next standup how we can set up testing to catch these kinds of issues in the future. cc @davidgamez (I tested uploading a ZIP file in the interface on a Windows instance when we were reviewing the PRs, but all the GTFS feeds I was using were initially downloaded onto my Mac, which is why this bug wasn't spotted.)

PR #1453, opened 2023-05-26
fix(web/service): change permissions for application.properties, service acount json, & newrelic.yml

lottspot:
The gcloud-secrets CLI will create these files with 0600 permissions, and does not document an option to change the mode. This patch uses the chown/chmod CLI utilities rather than --chown and --chmod flags because their support is not universal across docker versions or distributions which do not include buildkit (the cloudbuild environment provides one such distribution).

### DIFF
--- a/web/service/Dockerfile
+++ b/web/service/Dockerfile
@@ -4,19 +4,26 @@ RUN addgroup -S spring && adduser -S spring -G spring
 RUN mkdir -p /newrelic/logs
 RUN chown -R spring:spring /newrelic/logs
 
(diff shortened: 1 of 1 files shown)


KClough (approved):

lottspot:
This change was validated by completing the following build successfully: https://console.cloud.google.com/cloud-build/builds;region=global/9a0cc20f-7604-4c23-a9a4-d76197f17e46?&project=web-based-gtfs-validator

davidgamez (approved):
🚀

PR #1462, opened 2023-05-31
fix: Add back support for @GtfsTable.maxCharsPerColumn.

### LINKED ISSUE #1461: Add back support for @GtfsTable.maxCharsPerColumn
In PR #1228 I added support for `@GtfsTable.maxCharsPerColumn` to support files with especially large column values.  Unfortunately, this feature was accidentally removed in PR #1284.  I'd like to add it back.

bdferris-v2:
Closes #1461.

In PR #1228 I added support for @GtfsTable.maxCharsPerColumn to support files with especially large column values. Unfortunately, this feature was accidentally removed in PR #1284.  This PR adds it back.

### DIFF
--- a/core/src/main/java/org/mobilitydata/gtfsvalidator/table/AnyTableLoader.java
+++ b/core/src/main/java/org/mobilitydata/gtfsvalidator/table/AnyTableLoader.java
@@ -4,9 +4,11 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.flogger.FluentLogger;
 import com.univocity.parsers.common.TextParsingException;
+import com.univocity.parsers.csv.CsvParserSettings;
 import java.io.InputStream;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Optional;
 import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 import org.mobilitydata.gtfsvalidator.notice.CsvParsingFailedNotice;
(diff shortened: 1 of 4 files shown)


bdferris-v2:
@davidgamez I wanted to explain the acceptance test failures on this PR.  Specifically, these changes are expected.

The maxCharsPerColumn field is only used by one existing table: [GtfsAreaSchema](https://github.com/MobilityData/gtfs-validator/blob/master/main/src/main/java/org/mobilitydata/gtfsvalidator/table/GtfsAreaSchema.java).  Per the comment there:

    // We specify no limit (-1) here to support the experimental 'wkt' column, which encodes
    // an area polygon as a WKT string, which can grow quite large.  While this column has not yet
    // been added to the spec, it is present in many feeds (e.g. those produced by Trillium), which
    // can cause issues for the CSV parser.

Indeed, there is a large set of around ~56 feeds in the acceptance test corpus that currently fail CSV parsing because the CSV parser by default cannot handle the large `wkt` column in `areas.txt`.  By re-enabling support for maxCharsPerColumn, this PR allows those feeds to be parsed, which explains the drop in errors (56 `csv_parsing_failed` errors dropped to be specific).

So what's with the new errors and warnings?  Well, now that we can actually parse all those feeds, they have other errors and warnings that we previously weren't seeing that are now visible.  So as I said, this is actually a good thing / works as intended.

davidgamez (approved):
LGTM!

PR #1468, opened 2023-05-31
feat: separate and clarify file, column, value annotations

### LINKED ISSUE #1350: Create a @RecommendedColumn annotation
## Context
The spec requires certain columns, but having an empty value in them is valid (example [transfers.transfer_type](https://gtfs.org/schedule/reference/#transferstxt)).
The fact that we only had the `@Required` annotation, used in the schema files created [false positives](https://github.com/MobilityData/gtfs-validator/issues/1198).
```
  @Required
  GtfsTransferType transferType();
```
This is why the awesome @briandonahue created a new @RequiredColumn annotation in #1344, designed especially for the case where the column itself is required, but some of the fields can be empty.
```
  @RequiredColumn
  GtfsTransferType transferType();
```

## Proposition
We have a similar situation with `stop_times.timepoint`, except it's not a requirement but rather a recommendation. The GTFS Best Practice mentions (EDIT: they mentioned, it looks like we had another accidental removal in https://github.com/MobilityData/GTFS_Schedule_Best-Practices/pull/44, I am working on it 😅) that the timepoint should be provided, although the values can be empty under certain conditions.

In the validator, we have a [missing_timepoint_column](https://github.com/MobilityData/gtfs-validator/blob/master/RULES.md#missing_timepoint_column) notice that looks for the presence of the column
For consistency with the work done in #1344, we could replace this notice with a `missing_recommended_column` using a `@RecommendedColumn`. 

## Alternatives considered
Keep the validator as-is. I acknowledge this item doesn't add much to the user experience but figured this validator might be easier to maintain if we made this consistent (developers, I would love your POV). 
Additionally, we might want to use the `@RecommendedColumn` for other fields than `stop_times.timepoint` in the future.

briandonahue:
**Summary:**

Closes #1350 and #1344 

Create more purpose-focused annotations. Break `@Required` into `@RequiredFile`, and `@RequiredValue`, and `@Recommended` in to `@RecommendedFile` and `@RecommendedValue`. `@RequiredColumn` remains.  

NOTE: I excluded `@RecommendedColumn` because I don't see a place where it would be useful to recommend the column header is included, but have the value be optional.  

This implements most of the recommendation from #1344:

| Annotation           | Target   | Example        | Type    | Notes                                                |
| -------------------- | -------- | -------------- | ------- | ---------------------------------------------------- |
| `@RequiredFile`      | Class    | `Agency`       | Error   |                                                      |
| `@RequiredColumn`    | Field    | `transferType` | Error   | Column must exist, values are optional               |
| `@RequiredValue`     | Field    | `agencyName`   | Error   | Column must exist, value is required for each record |
| `@RecommendedFile`   | Class    | `FeedInfo`     | Warning |                                                      |
| `@RecommendedValue`  | Field    | `feedStartDate` | Warning | Recommended to include column and value              |
| `@RecommendedColumn` - EXCLUDED | Field    | DOES NOT OCCUR | Warning | Recommended to include column, value is optional     |

**Expected behavior:** 

Annotations should be more clearly defined and allow for all the scenarios required by the spec.

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -118,7 +118,7 @@ import org.mobilitydata.gtfsvalidator.annotation.FieldType;
 import org.mobilitydata.gtfsvalidator.annotation.FieldTypeEnum;
 import org.mobilitydata.gtfsvalidator.annotation.GtfsTable;
 import org.mobilitydata.gtfsvalidator.annotation.PrimaryKey;
-import org.mobilitydata.gtfsvalidator.annotation.Required;
+import org.mobilitydata.gtfsvalidator.annotation.RequiredFile;
 import org.mobilitydata.gtfsvalidator.type.GtfsDate;
 
 @GtfsTable("calendar.txt")
--- a/main/src/main/java/org/mobilitydata/gtfsvalidator/table/GtfsAgencySchema.java
+++ b/main/src/main/java/org/mobilitydata/gtfsvalidator/table/GtfsAgencySchema.java
@@ -21,22 +21,22 @@ import java.util.Locale;
 import org.mobilitydata.gtfsvalidator.annotation.*;
 
 @GtfsTable("agency.txt")
-@Required
+@RequiredFile
 public interface GtfsAgencySchema extends GtfsEntity {
   @FieldType(FieldTypeEnum.ID)
   @PrimaryKey
   @ConditionallyRequired
   String agencyId();
 
-  @Required
+  @RequiredValue
   @MixedCase
   String agencyName();
 
   @FieldType(FieldTypeEnum.URL)
-  @Required
+  @RequiredValue
   String agencyUrl();
 
-  @Required
+  @RequiredValue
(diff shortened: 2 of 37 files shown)


isabelle-dr:
@briandonahue to your question:

> But in the case of stop_times.timepoint is the goal to recommend the column be included, but a value is not required for every row?

This is exactly it :). An empty value is valid and actually treated as a `1`. The spec mentions "1 or empty - Times are considered exact."
So, a `missing_recommended_column` would replace the existing `missing_timepoint_column`.

isabelle-dr:
We are going to need some time to review this PR. The idea makes sense, but because this is a refactor with no effect on the end user (I expect `missing_required_file` and `missing_required_field` to behave as-is), I suggest we move forward with #1470, and use this PR to discuss breaking @Required into @RequiredFile, and @RequiredValue, and @Recommended in to @RecommendedFile and @RecommendedValue.

davidgamez:
Closing it was addressed by #1470

PR #1470, opened 2023-06-01
feat: Added missing_recommended_column notice 

### LINKED ISSUE #1350: Create a @RecommendedColumn annotation

### LINKED ISSUE #1455: Replace missing_timepoint_column with missing_recommended_column
After we finish adding a @RecommendedColumn annotation in this validator (issue #1350), we should remove [missing_timepoint_column](https://github.com/MobilityData/gtfs-validator/blob/master/RULES.md#missing_timepoint_column) from the validator and replace by a missing_recommended_column notice using the @RecommendedColumn in `stop_times.timepoint`.

This goes along the lines of the [missing_required_column](https://github.com/MobilityData/gtfs-validator/blob/master/RULES.md#missing_required_column) notice triggered by the @RequiredColumn annotation.

This is more consistent with the rest of the validator and allows us to re-use this notice for other similar use cases that might appear in the future.

jcpitre:
Also removed missing_timepoint_column notice that it replaces.

**Summary:**
Closes #1350 
Closes #1455 

In the report, we should not have a missing_recommended_column if the timepoint column if absent from the stop_times.txt fille.


isabelle-dr (approved):
This PR works from a functionality standpoint. I have used one dataset that does not have the `timepoint` column in `stop_times.txt` on the current version and the version in this PR and it works as expected.

Current version: missing_timepoint_column notice.
Screenshot 2023-06-05 at 2 26 32 PM

New version: missing_recommended_column notice.
Screenshot 2023-06-05 at 2 25 12 PM

We will still need a code review for this PR!

core/src/test/java/org/mobilitydata/gtfsvalidator/validator/TableHeaderValidatorTest.java
```diff
@@ -71,13 +75,32 @@ public void missingRequiredColumnShouldGenerateNotice() {
...
+            new CsvHeader(new String[] {"stop_name"}),
+            Set.of("stop_id", "stop_name"),
+            Set.of(),
+            Set.of("stop_id"),
+            container);
+
+    assertThat(container.getValidationNotices())
+        .containsExactly(new MissingRecommendedColumnNotice("stops.txt", "stop_id"));
+    assertThat(container.hasValidationErrors()).isFalse();
+    assertThat(container.hasValidationWarnings()).isTrue();
```
davidgamez, line 101: These conditions are already tested by line 99, as the container only holds the MissingRecommendedColumnNotice

jcpitre: Right. I might as well remove something similar in missingRequiredColumnShouldGenerateNotice that I just copied and slightly modified. Line 83.

davidgamez (approved):
Nice addition!

PR #1474, opened 2023-06-02
feat: add StopTimeTripWithoutTimes validator

### LINKED ISSUE #1089: `BlockTripsWithOverlappingStopTimesNotice` error not being caught
## **Bug report** 

**Describe the bug**
 Trillium Solutions is building a UI for the gtfs-validator. During our QA process, we encountered a potential bug in which the validator tool doesn't catch overlapping blocks and the `BlockTripsWithOverlappingStopTimesNotice` rule doesn't get triggered.  We've created a GTFS dataset with purposely "bad" data to test the validator.

**How we reproduce the bug**
1. Find a GTFS dataset with known block overlaps or use the file we created 
([gtfs-validator-test-us.zip](https://github.com/MobilityData/gtfs-validator/files/7954558/gtfs-validator-test-us.zip)).
2. Run your GTFS dataset of choice through the gtfs-validator.
3. Read through the output searching for the `BlockTripsWithOverlappingStopTimesNotice` error but not find the error message.

**Expected behaviour**
We would expect the `BlockTripsWithOverlappingStopTimesNotice` error message to appear with all necessary identifying data points to allow a user to make corrections to the dataset.

**Observed behaviour**
`BlockTripsWithOverlappingStopTimesNotice` doesn't appear to be working properly.

**Screenshots:** 
Screen Shot 2022-01-27 at 15 47 08
_Image is from Trillium Solutions GTFS Manager software_
There are six block overlaps that our own software is catching but gtfs-validator isn't.

We think investigating this possible bug at this file location would be a good start `main/src/test/java/org/mobilitydata/gtfsvalidator/validator/` and adding more test cases [here](https://github.com/Trillium-Solutions/gtfs-validator/blob/html_output/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/BlockTripsWithOverlappingStopTimesValidatorTest.java).

**Environment versions**
- validator version: 3.0
- Java version: 11
- OS versions: macOS 10.15.7

Please let me know if there's any other info y'all need.

KClough:
**Summary:**

Partially resolves #1089 

**Expected behavior:** 

Adds a new validator that reports an error if a trip includes stop time that are missing either arrival time or departure time.

### DIFF
--- a/main/src/main/java/org/mobilitydata/gtfsvalidator/report/model/FeedMetadata.java
+++ b/main/src/main/java/org/mobilitydata/gtfsvalidator/report/model/FeedMetadata.java
@@ -85,9 +85,8 @@ public class FeedMetadata {
   }
 
   private void loadSpecFeatures(GtfsFeedContainer feedContainer) {
-    var pathwaysTable = feedContainer.getTableForFilename(GtfsPathway.FILENAME);
-    boolean pathways = pathwaysTable.isPresent() && pathwaysTable.get().entityCount() > 0;
-    specFeatures.put("Pathways", pathways ? "Yes" : "No");
+    specFeatures.put(
+        "Pathways", hasOneRecordOfComponent(feedContainer, GtfsPathway.FILENAME) ? "Yes" : "No");
     var fareAttributesTable = feedContainer.getTableForFilename(GtfsFareAttribute.FILENAME);
     boolean faresV1 =
         fareAttributesTable.isPresent() && fareAttributesTable.get().entityCount() > 0;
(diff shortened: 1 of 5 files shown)


briandonahue (approved):
LGTM

isabelle-dr:
If I understand properly, this notice is different from the other ones in the sense that it informs that a notice could not have been checked. 

I saw that this was briefly discussed in the issue related, but this is a broader question of what to do when portions of this validator don't run (and a notice could not have been checked). This is the case for plenty of these validator notices and I think they should be treated differently since it's not technically a spec violation.

cswilson252:
is there anything here that can be salvaged into a new PR?

PR #1490, opened 2023-06-06
refactor(ci): rename cypress workflow to be consistent

themightychris:
Minor change to promote consistency. This also highlights that maybe this workflow could/should be merged into `.github/workflows/web_client_pr.yml` if we're factoring workflows more around lifecycle stage than outcome


davidgamez (approved):
LGTM!

PR #1497, opened 2023-06-08
refactor: remove web merge actions

### LINKED ISSUE #1486: Deploy web client+server to production only on release
Currently, the production deployment is updated each time `master` is pushed to / merged into.

We want to change this so that the production deployment is only updated when named releases are created, for two reasons:

1. We don't want production running un-released code
2. We want to eventually show on the homepage what version of the validator will be used

lottspot:
Resolves #1486


davidgamez (approved):
LGTM!

PR #1531, opened 2023-07-10
feat: Add overview section to the Json report

### LINKED ISSUE #1503: Add the Overview section to the JSON Report
### Describe the problem

A "Summary" section has been added to the validator in https://github.com/MobilityData/gtfs-validator/pull/1389. 
We want the JSON Report and the HTML report to have parity in terms of the information accessible. 
This was mentioned in the [Product Required Document](https://docs.google.com/document/d/157X9lMTkZHH7oYxK1fYPiY-rMh_VWyPMW_s9wImOP-I/edit), but this detail was missed at implementation.

In addition, we want the JSON Report to include the [user-defined parameters](https://github.com/MobilityData/gtfs-validator/blob/master/docs/USAGE.md#via-cli-app).
Note that this last part was delegated to a new issue: https://github.com/MobilityData/gtfs-validator/issues/1533

### Proposed solution

Add the following information to the JSON Report (this is not a suggestion for the new JSON structure, just a list of info to appear). 
- Settings and version
  - validator version
  - validation date and time
  - dataset name (name of the URL or the ZIP file)
  - country code
  - number of threads
  - output directory
  - validation report name
- The summary section
  - Agencies included
  - Feed info
  - Files included
  - Counts
  - GTFS Components included

_No response_

### Additional context

_No response_

jcpitre:
**Summary:**
Closes #1503 
Added an overview section to the JSON report according to https://docs.google.com/document/d/157X9lMTkZHH7oYxK1fYPiY-rMh_VWyPMW_s9wImOP-I/edit

**Expected behavior:** 
There should be this summary section in the JSON report:

```json  
  "summary": {
    "validatorVersion": "4.1.1-SNAPSHOT",
    "validatedAt": "2023-07-17 at 22:10:15 EDT",
    "gtfsInput": "file:///Users/jcpitre/datasets/test.zip",
    "threads": 1,
    "outputDirectory": "qaz",
    "systemErrorsReportName": "system_errors.json",
    "validationReportName": "report.json",
    "htmlReportName": "report.html",
    "countryCode": "ZZ",
    "feedInfo": {
      "publisherName": "Otwarte Dane GTFS przyjazdy.pl",
      "publisherUrl": "https://przyjazdy.pl/gtfs",
      "feedLanguage": "Polish"
    },
    "agencies": [
      {
        "name": "MKS Krosno",
        "url": "http://www.mks-krosno.pl",
        "phone": "N/A",
        "email": "N/A"
      }
    ],
    "files": [
      "agency.txt",
      "calendar.txt",
      "calendar_dates.txt",
      "feed_info.txt",
      "routes.txt",
      "stop_times.txt",
      "stops.txt",
      "trips.txt"
    ],
    "counts": {
      "Shapes": 0,
      "Stops": 514,
      "Routes": 18,
      "Trips": 659,
      "Agencies": 1,
      "Blocks": 1
    },
    "gtfsComponents": [
      "Feed Information",
      "Route Names",
      "Headsigns",
      "Wheelchair Accessibility",
      "Bikes Allowance"
    ]
 
```
Note that the summary section is not used for validation result comparison.


main/src/main/java/org/mobilitydata/gtfsvalidator/report/JsonReportSummary.java
```diff
@@ -0,0 +1,75 @@
...
+import org.mobilitydata.gtfsvalidator.report.model.FeedMetadata;
+import org.mobilitydata.gtfsvalidator.runner.ValidationRunnerConfig;
+import org.mobilitydata.gtfsvalidator.util.VersionInfo;
+
+/**
+ * Used to generate the summary of json report using the built-in class serialization of GSON.
+ * Fields names used here will be the same as in the generated json file.
+ */
+@SuppressWarnings("unused") // The fields of this class are only read by Gson when serializing.
+public class JsonReportSummary {
```
bdferris-v2, line 17: Hey, I had a macro comment on this class.

This class defines the exported API for the JSON report summary.  As such, identifiers used here should be stable, since folks external users will be depending on this API.  However, this summary depends on a number of identifiers export from `FeedMetadata` and `AgencyMetadata`, two classes that were written primarily as UI presentation model classes.  As such, I think there is a risk someone might make a change to one of those UI presentation classes, not realizing it would impact the exported JSON api.

For example, the key names in `feedInfo` below, any of the data in `AgencyMetadata`, the keys in `counts`, and the identifiers in `gtfsComponents`.  They'd all ideally be defined fields or enums, where it's very clear that changing a value would change the exported API.

Does that feedback make sense?

main/src/test/java/org/mobilitydata/gtfsvalidator/report/model/JsonReportGeneratorTest.java
```diff
@@ -0,0 +1,129 @@
...
+    ValidationRunnerConfig config = ValidationRunnerConfig.builder().build();
+
+    JsonReport report =
+        reportGenerator.generateReport(null, noticeContainer, config, VersionInfo.empty(), "");
+
+    return report;
+  }
+
+  @Test
+  public void noticeReportsTest() throws IOException {
```
qcdyx, line 59: Shall we remove the empty test and commented tests?

main/src/main/java/org/mobilitydata/gtfsvalidator/runner/ValidationRunner.java
```diff
@@ -229,18 +233,31 @@ public static void exportReport(
             "Error creating output directory: %s", config.outputDirectory());
       }
     }
+    SimpleDateFormat formatter = new SimpleDateFormat("yyyy-MM-dd 'at' HH:mm:ss z");
```
qcdyx, line 236: I see here you are using SimpleDateFormat. I recommend DateTimeFormatter. DateTimeFormatter is a replacement for the old SimpleDateFormat that is thread-safe and provides additional functionality. I recommend that the new Date Time API introduced in Java 8.

qcdyx (commented):
I used ae-dubayy-road-and-transport-authority-rta-gtfs-904.zip to generate the report. I saw that the value of agency email is N/A. Question: shall we keep it in this way? Or, remove email from Summary section since it's N/A?
Screenshot 2023-07-13 at 12 07 04 PM

[ae-dubayy-road-and-transport-authority-rta-gtfs-904.zip](https://github.com/MobilityData/gtfs-validator/files/12041474/ae-dubayy-road-and-transport-authority-rta-gtfs-904.zip)

main/src/test/java/org/mobilitydata/gtfsvalidator/report/model/JsonReportSummaryTest.java
```diff
@@ -0,0 +1,156 @@
...
+
+@RunWith(JUnit4.class)
+public class JsonReportSummaryTest {
+
+  private static VersionInfo versionInfo =
+      VersionInfo.create(Optional.of("1.0"), Optional.of("1.1"));
+
+  private static Gson gson = new Gson();
+
+  private static JsonReportSummary generateJsonReportSummary() throws Exception {
```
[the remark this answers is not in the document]
jcpitre: Removed.

main/src/main/java/org/mobilitydata/gtfsvalidator/runner/ValidationRunner.java
jcpitre: This piece of code already existed and was simply moved from main/src/main/java/org/mobilitydata/gtfsvalidator/report/HtmlReportGenerator.java.
I would prefer at this point to keep it as is.

main/src/main/java/org/mobilitydata/gtfsvalidator/report/JsonReportSummary.java
jcpitre: Created classes devoted to the json generation with controlled fields and copied values from the existing classes. 
Also added some string constants used in the Html and Json generation code so it's obvious changing the string could have an effect in the Json report code.

Continued using gtfsComponents directly. We will have to be avoid changing the component names or document it properly.

main/src/main/java/org/mobilitydata/gtfsvalidator/runner/ValidationRunner.java
qcdyx: I see that the GtfsDate is already using LocalDate which is introduced with DateTimeFormatter in the new java 8 date time API. I'm ok with keeping it for now for this issue, but shall we open an issue about updating to the java 8 data time API since it's more immutable and threadsafe than the old ones? @davidgamez

qcdyx (approved):
LGTM

jcpitre:
@isabelle-dr If you have time could you take a look at the exemple summary in the first comment of this PR: https://github.com/MobilityData/gtfs-validator/pull/1531#issue-1795739874 
And maybe tell us if it looks like what you expected.

Also @qcdyx mentioned this in her review https://github.com/MobilityData/gtfs-validator/pull/1531#pullrequestreview-1528779243:
> I used ae-dubayy-road-and-transport-authority-rta-gtfs-904.zip to generate the report. I saw that the value of agency email is N/A. Question: shall we keep it in this way? Or, remove email from Summary section since it's N/A? 

Do you have an opinion on this? The N/A is also present in the json report summary.

emmambd:
@jcpitre Hello! Isabelle's off for the rest of the week so I can share comments here to unblock this. I downloaded the JAR file from this branch's [Test Package Document](https://github.com/MobilityData/gtfs-validator/actions/runs/5582156215) action to test this on a few reports. 

1) The JSON report adheres to [the original design that you linked](https://docs.google.com/document/d/157X9lMTkZHH7oYxK1fYPiY-rMh_VWyPMW_s9wImOP-I/edit) so this looks good to me! 
2) Re: omitting or including N/As, I think this is outside the scope of this issue, since as you mentioned, it would need to be included or omitted in both report versions. We can discuss this next week with @isabelle-dr for design history/context.

main/src/main/java/org/mobilitydata/gtfsvalidator/report/JsonReportSummary.java
bdferris-v2: These changes look good.  Thanks!

Doesn't have to be in this PR, but long-term, I think it would be more valuable to have GTFS Components modeled as an enum internally, not a string.  It might ultimately be rendered as a string, but I think a enum would be more practical for long-term programatic access.

main/src/main/java/org/mobilitydata/gtfsvalidator/runner/ValidationRunner.java
jcpitre: I was willing to give it a go, but DateTimeFormatter had trouble interpreting the part of the pattern related to time zone. 
Since like I said the date related code in this PR is actually some code that was moved and not modified, I prefer to keep it the way it is.

jcpitre: In reality I don't think you should open an issue about this.
The fact that it's thread safe is irrelevant here because we are creating and using the formatter in the same thread. As for extra functionalities, we don't need them here.

Issue #1446, opened 2023-05-23
Run validator correctly when folder is compressed


evansiroky:
### Describe the bug

With some specific zip files, like [this one](https://github.com/MobilityData/gtfs-validator/files/11548104/auburntransit-ca-us-04132023.2.zip), both the desktop and web validator say the required files are missing. If I unzip on my local machine the files appear to be in the folder that got unzipped.

### Steps/Code to Reproduce

With either the web validator or desktop validator, attempt to upload certain files (like [this one](https://github.com/MobilityData/gtfs-validator/files/11548104/auburntransit-ca-us-04132023.2.zip)).

### Expected Results

Validator checks various things and is able to detect required files.

### Actual Results

Validator flags that required files are missing.

### Validator version

4.0.0

### Operating system

MacOS and web

isabelle-dr:
Closing, I am assuming this was the problem.
@evansiroky please let us know if you try zipping the GTFS folder again and the problem persists

emmambd:
Reopening this as a validator enhancement to run the validator properly when the folder is compressed instead of the folder contents.

sylvansson:
@emmambd Sure, I'll have a pull request by the end of the week :-)

PR #1543, opened 2023-07-24
feat: 1532 support mac created zip files

### LINKED ISSUE #1532: Support Mac created Zip files
### Describe the problem

Mac created zip files reproduce the entire folder, for instance, test.zip contains a test/ directory that contains the GTFS files and the validator generates missing_required_file notice.

name: test/
isDirectory: true
name: test/fare_attributes.txt
isDirectory: false
name: __MACOSX/test/._fare_attributes.txt
isDirectory: false
name: test/transfers.txt
isDirectory: false

### Proposed solution

For a Mac created zip file, the validator is able to omit the generated directory and only check if the real subfolders that contain gtfs files(test/subfolder) for generating invalid_input_files_in_subfolder error notice purposes.

Right now, gtfs validator generated missing_required_file and missing_recommended_file even if they are already present in the zip file

image

_No response_

### Additional context

https://github.com/MobilityData/gtfs-validator/issues/1504

qcdyx:
**Summary:**

Closes #1532 

**Expected behavior:** 
The screenshots shows that the validator does not generate missing_required_file and missing_recommended file for the gtfs files in the root directory folder using maczip.zip.

[maczip.zip](https://github.com/MobilityData/gtfs-validator/files/12148500/maczip.zip)

Screenshot 2023-07-24 at 11 16 50 AM


core/src/main/java/org/mobilitydata/gtfsvalidator/input/GtfsFiles.java
```diff
@@ -0,0 +1,37 @@
+package org.mobilitydata.gtfsvalidator.input;
```
jcpitre, line 1: In general there are changes here an in 1504-... branch that touch the same files.
This will likely lead to conflicts.
I suggest to wait for 1504-... to be merged in master, then merge master back in this 1532-... branch and resolve the conflicts.

core/src/main/java/org/mobilitydata/gtfsvalidator/input/GtfsInput.java
```diff
@@ -44,6 +44,8 @@ public abstract class GtfsInput implements Closeable {
   public static final String invalidInputMessage =
       "At least 1 GTFS file is in a subfolder. All GTFS files must reside at the root level directly.";
 
+  public static boolean isMacZip = false;
```
jcpitre, line 47: This variable should not be static. 
If we call hasSubfolderWithGtfsFile twice, once with a mac zip file and the second time with a non-mac zip file, isMacZip will be set to true for the first file, and will keep being true for the second file. But it's not a mac zip file!
Can't you determine if it's a mac zip file directly in the GtfsZipFileInput constructor instead of doing it in  containsGtfsFileInSubfolder? That way the isMacZip variable will be confined in the GtfsZipFileInput class.

qcdyx: This is a very good point. I've made the changes.

core/src/main/java/org/mobilitydata/gtfsvalidator/input/GtfsInput.java
```diff
@@ -61,85 +59,81 @@ public static GtfsInput createFromPath(Path path, NoticeContainer noticeContaine
...
-          }
-        }
-      }
-    } catch (IOException ioException) {
-      logger.atSevere().withCause(ioException).log(
-          "GtfsInput containsSubfolderWithGtfsFile throws IOException");
-    }
-    return containsGtfsFileInSubfolder;
+  public static boolean hasSubfolderWithGtfsFile(Path path) throws IOException {
+    String fileName = path.getFileName().toString().replace(".zip", "");
```
jcpitre, line 90: Unused variable

core/src/main/java/org/mobilitydata/gtfsvalidator/input/GtfsZipFileInput.java
```diff
@@ -32,28 +33,43 @@
 public class GtfsZipFileInput extends GtfsInput {
   private final ImmutableSet<String> filenames;
   private final ZipFile zipFile;
+  private boolean isMacZip;
 
-  public GtfsZipFileInput(ZipFile zipFile) {
+  public GtfsZipFileInput(ZipFile zipFile, String zipFileName) {
     this.zipFile = zipFile;
-
+    StringBuilder strBuilder = new StringBuilder();
```
jcpitre, line 40: Using StringBuilder is an overkill for that. you should use se a simple string concatenation instead:
```
String macDirectory = zipFileName + "/";
```

core/src/main/java/org/mobilitydata/gtfsvalidator/input/GtfsZipFileInput.java
```diff
@@ -32,28 +33,43 @@
...
     for (Enumeration<ZipArchiveEntry> entries = zipFile.getEntries(); entries.hasMoreElements(); ) {
       ZipArchiveEntry entry = entries.nextElement();
-      if (!isInsideZipDirectory(entry)) {
-        filenamesBuilder.add(entry.getName());
+      String entryName = entry.getName();
+      // check if the first entry is a directory with the name of the zip file
+      if (entry.isDirectory() && zipFileName.equals(entryName.replaceFirst("/", ""))) {
+        isMacZip = true;
+      }
+      Pair<String, Boolean> pair = isInsideZipDirectory(entry, macDirectory, entryName);
```
jcpitre, line 52: I would not use a pair for that, just a boolean, and move this test inside the isInsideZipDirectory method.
I would also put a comment in the code as to why we test for ".DS_Store"

core/src/main/java/org/mobilitydata/gtfsvalidator/input/GtfsZipFileInput.java
```diff
@@ -32,28 +33,43 @@
...
     // All slashes MUST be forward slashes '/' as opposed to backwards slashes '\' for compatibility
     // with Amiga and
     // UNIX file systems etc.
     //
     // Directory names in end with '/'.
-    return entry.getName().contains("/");
+    if (isMacZip) {
+      entryName = entry.getName().replaceFirst(macDirectory, "");
+    }
+    Pair<String, Boolean> pair = Pair.of(entryName, entryName.contains("/"));
```
jcpitre, line 71: To continue on the comment above, remove the Pair and replace with somethning like:
```
// Check for .DS_Store because ....
return entryName.isBlank() || entryName.contains("/") || entryName.equals(".DS_Store");
```

qcdyx: good point, just add the comments in code.

core/src/main/java/org/mobilitydata/gtfsvalidator/input/GtfsZipFileInput.java
qcdyx: Ok, make sense, I had this constraint of keeping isInsideZipDirectory when I coded the Pair stuff - it is removed now.

core/src/main/java/org/mobilitydata/gtfsvalidator/input/GtfsZipFileInput.java
jcpitre: Fine. 
I just read that the java compiler does a StringBuilder under the hood when you use +. So the two forms are equivalent. 
I just feel that 
```
String macDirectory = zipFileName + "/";
```
Is more readable.
But that's a question of style.
I learn something new, that's good.

core/src/main/java/org/mobilitydata/gtfsvalidator/input/GtfsZipFileInput.java
```diff
@@ -46,32 +47,24 @@ public GtfsZipFileInput(ZipFile zipFile, String zipFileName) {
       ZipArchiveEntry entry = entries.nextElement();
       String entryName = entry.getName();
       // check if the first entry is a directory with the name of the zip file
-      if (entry.isDirectory() && zipFileName.equals(entryName.replaceFirst("/", ""))) {
+      String firstEntryName = entryName.replaceFirst("/", "");
+      if (entry.isDirectory() && zipFileName.equals(firstEntryName)) {
         isMacZip = true;
```
jcpitre, line 52: Now that the two usages of isMacZip is confined to the constructor, it seems to it could be a local variable. No?

jcpitre (approved):

emmambd (commented):
The following use case works correctly: 

- Take a pre-existing GTFS feed
- Unzip it 
- Open the folder 
- Compress the folder contents (not the folder itself) 
- Run the validator on the new zip created from the folder contents 

It still doesn't work when I 

- Take a pre-existing GTFS feed
- Unzip it 
- Open the folder
- Compress the folder 
- Run the validator on the new zip 

See https://github.com/MobilityData/gtfs-validator/issues/1446 for the second use case.

PR #1555, opened 2023-08-08
feat: 1520 gtfs validator desktop app deprecation

### LINKED ISSUE #1520: GTFS-Validator Desktop App Deprecation
### Describe the problem

Now that [the validator is available as a web service](https://gtfs-validator.mobilitydata.org/), we want to consider deprecating the [GUI desktop app](https://github.com/MobilityData/gtfs-validator#using-the-desktop-app). We suspect that the desktop app is significantly less used than the web service. 

- Managing releases for both the desktop app and web service requires more time and resources, and may be a poor use of our time if usage of the desktop app is low. 
- We want to focus our efforts on what's of the highest value to the community, such as PR review, Mobility Database API, etc.

### Proposed solution

 To better understand what's the highest value, we need feedback! We want to know

1. Are you using the desktop app currently? If so, how?  
2. What value is there in having both a desktop app and a web service? 
3. Any special considerations with deprecating the desktop app?
4. What communication channels are critical to sharing the deprecation?

_No response_

### Additional context

_No response_

qcdyx:
**Summary:**
Closes #1520 

**Expected behavior:** 

image


web/service/src/main/java/org/mobilitydata/gtfsvalidator/web/service/controller/ValidationController.java
```diff
@@ -121,6 +127,22 @@ public ResponseEntity runValidator(
     }
   }
 
+  @GetMapping("/version")
+  public ResponseEntity currentVersion() {
```
davidgamez, line 131: Let's annotate the ResponseEntity with the right response class. This change will propagate some changes in the return statement. 
```suggestion
  public ResponseEntity<String> currentVersion() {
```

web/service/src/main/java/org/mobilitydata/gtfsvalidator/web/service/controller/ValidationController.java
```diff
@@ -121,6 +127,22 @@ public ResponseEntity runValidator(
...
+    // https://gtfs-validator-web-mbzoxaljzq-ue.a.run.app/version
+    StringBuilder stringBuilder = new StringBuilder();
+    VersionResolver checker = new VersionResolver();
+    VersionInfo versionInfo = checker.getVersionInfoWithTimeout(TIMEOUT);
```
davidgamez, line 135: If we are only interested in the current version, I suggest instead of using `getVersionInfoWithTimeout` make the [VersionResolver#resolveCurrentVersion](https://github.com/MobilityData/gtfs-validator/blob/ccfdd65b08187f58caed823c7765366de1abebbd/main/src/main/java/org/mobilitydata/gtfsvalidator/util/VersionResolver.java#L114) method public and call it directly. This will save a network call that is executed inside `getVersionInfoWithTimeout`.

web/service/src/main/java/org/mobilitydata/gtfsvalidator/web/service/controller/ValidationController.java
```diff
@@ -121,6 +127,22 @@ public ResponseEntity runValidator(
...
+    stringBuilder.append(versionInfo.currentVersion());
```
davidgamez, line 136: The current version is of class `Optional`. Let's return the string if available. If not available, we can throw an exception with an explanatory message; this should not happen with a properly packaged jar.

web/service/src/main/java/org/mobilitydata/gtfsvalidator/web/service/controller/ValidationController.java
```diff
@@ -121,6 +127,22 @@ public ResponseEntity runValidator(
     }
   }
 
+  @GetMapping("/version")
+  public ResponseEntity currentVersion() {
+    // https://gtfs-validator-web-mbzoxaljzq-ue.a.run.app/version
+    StringBuilder stringBuilder = new StringBuilder();
+    VersionResolver checker = new VersionResolver();
```
davidgamez, line 134: We should use here the autowired bean instead of instantiating the class.

web/service/src/main/java/org/mobilitydata/gtfsvalidator/web/service/controller/ValidationController.java
```diff
@@ -43,6 +46,9 @@ public class ValidationController {
   @Autowired private StorageHelper storageHelper;
   @Autowired private ValidationHandler validationHandler;
 
+  @Autowired private VersionResolver checker;
```
davidgamez, line 49: Rename it to increase readability
```suggestion
  @Autowired private VersionResolver versionResolver;
```

davidgamez (changes requested):

web/service/src/main/java/org/mobilitydata/gtfsvalidator/web/service/controller/ValidationController.java
```diff
@@ -121,6 +127,22 @@ public ResponseEntity runValidator(
     }
   }
 
+  @GetMapping("/version")
+  public ResponseEntity<String> currentVersion() {
+    // https://gtfs-validator-web-mbzoxaljzq-ue.a.run.app/version
```
davidgamez, line 132: Please remove this commented link.

web/service/src/main/java/org/mobilitydata/gtfsvalidator/web/service/controller/ValidationController.java
```diff
@@ -43,6 +45,9 @@ public class ValidationController {
   @Autowired private StorageHelper storageHelper;
   @Autowired private ValidationHandler validationHandler;
 
+  @Autowired private VersionResolver versionResolver;
+  private static final Duration TIMEOUT = Duration.ofSeconds(5);
```
davidgamez, line 49: Please remove the TIMEOUT field, as it is not being used anymore.

davidgamez (approved):
LGTM!

web/service/src/main/java/org/mobilitydata/gtfsvalidator/web/service/controller/ValidationController.java
qcdyx: removed

KClough (approved):

PR #1565, opened 2023-09-07
feat: 1537 get list of validators that couldnt run because of a parsing problem

### LINKED ISSUE #1537: Get list of validators that couldn't run because of a parsing problem
### Describe the problem
Currently this tool logs all validators that **didn't** run. We want to log a list of validators that **couldn't** run (because of a parsing problem. Should exclude validators that are optional and should not be run on the specific feed.
In order to achieve https://github.com/MobilityData/gtfs-validator/issues/1536 and then https://github.com/MobilityData/gtfs-validator/issues/1485

I need to see the notices that were not triggered but could be present in my dataset.
- **Users should only see validators in the parsing [error log](https://github.com/MobilityData/gtfs-validator/issues/1193#issuecomment-1618954067) that are relevant to their feed**. If I don’t have any GTFS fields that are of the [Color field type](https://gtfs.org/schedule/reference/#field-types), then the [route_color_contrast](https://github.com/MobilityData/gtfs-validator/blob/master/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/RouteColorContrastValidator.java) is irrelevant. It doesn’t run, and I don’t need to even know about it.
- If I do have GTFS fields that are of the Color field type (e. g. route_color and route_text_color), and they are properly formatted, then either I do get the route_color_contrast notice, or I don’t. If I do, my dataset is violating a GTFS recommendation.
- Same situation than above, but my color fields are not properly formatted. Then I will get the invalid_color in my errors.
:warning: But I also want to know that if the route_color_contrast is not in my list of notices, it’s not because my data is good. My dataset might or might not violate this GTFS recommendation. This notice could be present, we can’t ensure that it’s not.

User should only see the list of validators that are relevant to their feed. For example, 

### Proposed solution

To do!

- Related PR:  https://github.com/MobilityData/gtfs-validator/pull/1496

### Additional context

_No response_

qcdyx:
**Summary:**

Closes #1537 
My solution goes thru both SingleEntityValidators and SingleFileValidators, gathers validators that have dependencies with parsing errors. From my perspectives, we should decouple this problem into two parts. First, get the list of validators that couldn't run due to parsing errors. Second, a new logic could be added into the gtfs-validator - if we don't have this and that text files, then validator A B C should not run. @emmambd I request all dev team for code review. Let me know your opinion on the idea of decoupling this issue.

**Expected behavior:** 

You will see the following info printing out if there's validators that couldn't run due to parsing errors
 ------------------------------------------------------------------------- 
|   Some validators were skipped due to parsing problems.   |
 ------------------------------------------------------------------------- 
 Validators with Parsing Errors: BikesAllowanceValidator,GtfsAttributionRouteIdForeignKeyValidator,GtfsFareLegRuleNetworkIdForeignKeyValidator,GtfsFareRuleRouteIdForeignKeyValidator,GtfsRouteAgencyIdForeignKeyValidator,GtfsTransferFromRouteIdForeignKeyValidator,GtfsTransferToRouteIdForeignKeyValidator,GtfsTripRouteIdForeignKeyValidator,RouteAgencyIdValidator,ShapeToStopMatchingValidator,StopTimeTravelSpeedValidator,TranslationFieldAndReferenceValidator,UrlConsistencyValidator

### DIFF
--- a/core/src/main/java/org/mobilitydata/gtfsvalidator/table/AnyTableLoader.java
+++ b/core/src/main/java/org/mobilitydata/gtfsvalidator/table/AnyTableLoader.java
@@ -30,10 +30,18 @@ import org.mobilitydata.gtfsvalidator.validator.ValidatorUtil;
 public final class AnyTableLoader {
 
   private static final FluentLogger logger = FluentLogger.forEnclosingClass();
-  private static final List<Class<? extends FileValidator>> validatorsWithParsingErrors = new ArrayList<>();
+  private static final List<Class<? extends FileValidator>> validatorsWithParsingErrors =
+      new ArrayList<>();
+
+  private static final List<Class<? extends SingleEntityValidator>>
+      singleEntityValidatorsWithParsingErrors = new ArrayList<>();
 
   public List<Class<? extends FileValidator>> getValidatorsWithParsingErrors() {
     return Collections.unmodifiableList(validatorsWithParsingErrors);
+  }
+
+  public List<Class<? extends SingleEntityValidator>> getSingleEntityValidatorsWithParsingErrors() {
+    return Collections.unmodifiableList(singleEntityValidatorsWithParsingErrors);
   }
 
   public static GtfsTableContainer load(
@@ -91,7 +99,8 @@ public final class AnyTableLoader {
     final List<GtfsEntity> entities = new ArrayList<>();
     boolean hasUnparsableRows = false;
     final List<SingleEntityValidator<GtfsEntity>> singleEntityValidators =
-        validatorProvider.createSingleEntityValidators(tableDescriptor.getEntityClass());
+        validatorProvider.createSingleEntityValidators(
+            tableDescriptor.getEntityClass(), singleEntityValidatorsWithParsingErrors::add);
     try {
       for (CsvRow row : csvFile) {
         if (row.getRowNumber() % 200000 == 0) {
@@ -137,7 +146,8 @@ public final class AnyTableLoader {
     GtfsTableContainer table =
         tableDescriptor.createContainerForHeaderAndEntities(header, entities, noticeContainer);
     ValidatorUtil.invokeSingleFileValidators(
-        validatorProvider.createSingleFileValidators(table, validatorsWithParsingErrors::add), noticeContainer);
+        validatorProvider.createSingleFileValidators(table, validatorsWithParsingErrors::add),
+        noticeContainer);
     return table;
   }
 
@@ -201,7 +211,8 @@ public final class AnyTableLoader {
       noticeContainer.addValidationNotice(new MissingRequiredFileNotice(gtfsFilename));
     }
     ValidatorUtil.invokeSingleFileValidators(
-        validatorProvider.createSingleFileValidators(table, validatorsWithParsingErrors::add), noticeContainer);
+        validatorProvider.createSingleFileValidators(table, validatorsWithParsingErrors::add),
+        noticeContainer);
     return table;
   }
 }
--- a/core/src/main/java/org/mobilitydata/gtfsvalidator/table/AnyTableLoader.java
+++ b/core/src/main/java/org/mobilitydata/gtfsvalidator/table/AnyTableLoader.java
@@ -46,6 +46,7 @@ public final class AnyTableLoader {
       csvFile = new CsvFile(csvInputStream, gtfsFilename, settings);
     } catch (TextParsingException e) {
       noticeContainer.addValidationNotice(new CsvParsingFailedNotice(gtfsFilename, e));
+      //todo
       return tableDescriptor.createContainerForInvalidStatus(
           GtfsTableContainer.TableStatus.INVALID_HEADERS);
     }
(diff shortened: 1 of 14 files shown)


main/src/main/java/org/mobilitydata/gtfsvalidator/runner/ValidationRunner.java
```diff
@@ -159,6 +164,33 @@ public static void printSummary(
           skippedValidators.stream().map(Class::getSimpleName).collect(Collectors.joining(",")));
       logger.atSevere().log(b.toString());
     }
+
+    List<Class<? extends FileValidator>> validatorsWithParsingErrors =
```
jcpitre, line 168: Shouldn't the 3 "skipped" lists (skippedValidators, validatorsWithParsingErrors, singleEntityValidatorsWithParsingErrors) all be listed together?
I thought all the validators in these lists were not run because of dependency problems.

core/src/main/java/org/mobilitydata/gtfsvalidator/table/AnyTableLoader.java
```diff
@@ -21,13 +22,27 @@
...
 import org.mobilitydata.gtfsvalidator.parsing.RowParser;
+import org.mobilitydata.gtfsvalidator.validator.FileValidator;
 import org.mobilitydata.gtfsvalidator.validator.SingleEntityValidator;
 import org.mobilitydata.gtfsvalidator.validator.ValidatorProvider;
 import org.mobilitydata.gtfsvalidator.validator.ValidatorUtil;
 
 public final class AnyTableLoader {
 
   private static final FluentLogger logger = FluentLogger.forEnclosingClass();
+  private static final List<Class<? extends FileValidator>> validatorsWithParsingErrors =
```
jcpitre, line 33: You should consider renaming `validatorsWithParsingErrors` to `singleFileValidatorsWithParsingErrors`, just to be consistent with `singleEntityValidatorsWithParsingErrors`

core/src/main/java/org/mobilitydata/gtfsvalidator/table/AnyTableLoader.java
```diff
@@ -21,13 +22,27 @@
...
+      new ArrayList<>();
```
jcpitre, line 34: Also it might be a good idea to rename the `skippedValidators` list used in existing code to `multiFileValidatorsWithParsingErrors`. Your way of naming is more descriptive (`skippedValidators` does not tell us why it was skipped, whereas `singleEntityValidatorsWithParsingErrors` for example does)
This change would not be in your code, so I realize it's a bit more touchy.

core/src/main/java/org/mobilitydata/gtfsvalidator/table/AnyTableLoader.java
qcdyx: Renamed!

core/src/main/java/org/mobilitydata/gtfsvalidator/table/AnyTableLoader.java
qcdyx: Renaming Done.

core/src/test/java/org/mobilitydata/gtfsvalidator/validator/DefaultValidatorProviderTest.java
```diff
@@ -35,13 +35,22 @@ public void testCreateValidators() throws ValidatorLoaderException {
...
         .containsExactly(GtfsTestEntityValidator.class);
 
-    assertThat(provider.createSingleFileValidators(tableContainer).stream().map(Object::getClass))
+    List<Class<? extends FileValidator>> validatorsWithParsingErrors = new ArrayList<>();
+    assertThat(
+            provider
+                .createSingleFileValidators(tableContainer, validatorsWithParsingErrors::add)
+                .stream()
+                .map(Object::getClass))
         .containsExactly(GtfsTestFileValidator.class);
```
jcpitre, line 55: You should consider adding tests to the `testCreateValidators_skippedValidators` method below, to test the cases where there are validatorsWithParsingErrors

main/src/main/java/org/mobilitydata/gtfsvalidator/runner/ValidationRunner.java
qcdyx: Changes are pushed.

jcpitre (approved):

jcpitre:
I created a dataset where the agency.txt file did not have an agency name. Here is what is printed:
```
---------------------------------------------------------------------------------------- 
| Some validators were skipped because the GTFS files they rely on could not be parsed | 
---------------------------------------------------------------------------------------- 
Multi-file validators:
   FareAttributeAgencyIdValidator
   GtfsAttributionAgencyIdForeignKeyValidator
   GtfsFareAttributeAgencyIdForeignKeyValidator
   GtfsRouteAgencyIdForeignKeyValidator
   MatchingFeedAndAgencyLangValidator
   RouteAgencyIdValidator
   TranslationFieldAndReferenceValidator
   UrlConsistencyValidator
```

core/src/test/java/org/mobilitydata/gtfsvalidator/validator/DefaultValidatorProviderTest.java
qcdyx: I went thru your commit and it looks good to me. I like the method comments - pretty helpful. Also, the printed message "Some validators were skipped because the GTFS files they rely on could not be parsed" is much clearer!
You can go ahead merge it.

PR #1571, opened 2023-09-13
Fix for deprecated of `set-output` in GH actions

fredericsimard:
GitHub has deprecated set-output and will eventually make it unusable, so this is a preventative fix. See more info here: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

### DIFF
--- a/.github/workflows/acceptance_test.yml
+++ b/.github/workflows/acceptance_test.yml
@@ -33,7 +33,7 @@ jobs:
       - name: "[Pull Request] Get commit message"
         if: github.event_name == 'pull_request'
         id: pr_get_commit_message
-        run: echo "pr_commit_message=$(git log --format=%B -n 1 HEAD^2)" >> $GITHUB_OUTPUT
+        run: echo "pr_commit_message=\"$(git log --format=%B -n 1 HEAD^2)\"" >> $GITHUB_OUTPUT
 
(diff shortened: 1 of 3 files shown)


.github/workflows/acceptance_test.yml
```diff
@@ -33,7 +33,7 @@ jobs:
       - name: "[Pull Request] Get commit message"
         if: github.event_name == 'pull_request'
         id: pr_get_commit_message
-        run: echo ::set-output name=pr_commit_message::$(git log --format=%B -n 1 HEAD^2)
+        run: echo "pr_commit_message=$(git log --format=%B -n 1 HEAD^2)" >> $GITHUB_OUTPUT
```
davidgamez, line 36: The Github action is failing. Try the line below to add quotes to the value in case the commit message contains multiple words: 
```suggestion
        run: echo "pr_commit_message=\"$(git log --format=%B -n 1 HEAD^2)\"" >> $GITHUB_OUTPUT
```

fredericsimard: Ah! Shoot you're right, didn't see the ".

fredericsimard: Looks ok now: https://github.com/MobilityData/gtfs-validator/actions/runs/6177475232

davidgamez (approved):

PR #1586, opened 2023-09-25
feat: Resolve release version from new /version endpoint.

bdferris-v2:
Per discussion in #1520, we'd like to pull the current release version of the app from the new /version endpoint.

### DIFF
--- a/app/gui/src/main/java/org/mobilitydata/gtfsvalidator/app/gui/Main.java
+++ b/app/gui/src/main/java/org/mobilitydata/gtfsvalidator/app/gui/Main.java
@@ -25,6 +25,7 @@ import java.util.logging.LogManager;
 import java.util.logging.Logger;
 import javax.swing.SwingUtilities;
 import javax.swing.UIManager;
+import org.mobilitydata.gtfsvalidator.runner.ApplicationType;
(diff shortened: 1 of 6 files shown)


jcpitre:
@bdferris-v2 I'm a bit late to the party on this one. 
If I understand correctly you want to call a version endpoint so you can count how many times the software is used?
With the ultimate goal of deprecating the desktop GUI version if we realize it's not really used.
I did not see how we distinguish between runs of the CLI version, the GUI version and the Web version.
Also will the runs of the app done during development be counted also?

bdferris-v2:
@jcpitre that's a great point.  Technically, the web validator doesn't check for the latest release version, so it wouldn't be counted here.  But it's reasonable to want to distinguish between CLI and Desktop App traffic.  It's also reasonable to distinguish between development vs prod traffic.

To that end, I've add two parameters to the /version api call:
* `application_type`: an enum that will be CLI, DESKTOP, or WEB based on the application
* `current_version`: the current application version.  Will have a `-SNAPSHOT` suffix for dev builds.

I've also verified with @davidgamez that it's possible to see these parameters in the logs.

I think that should be sufficient to distinguish between the various type of traffic.  Thoughts?

jcpitre (approved):
LGTM

PR #1596, opened 2023-10-19
feat: 1588 publish to maven central

### LINKED ISSUE #1588: Jitpack as artifact hosting is unreliable, consider using Maven Central or GitHub Packages instead
### Describe the problem

It's all in the title, really. Jitpack seems to have intermittent issues especially with GitHub Actions (_I assume IP based throttling/automatic blocking, but can't be sure_) and instead of having broken builds randomly for half a day at the time I'd rather just have the artifacts in tried and true location instead.

This happens with proper caching setup as well, because it seems our runner had a cache miss at a critical point and now it can't download the dependency into the cache because of who-knows-what, so things are stuck. [Jitpack's status page](https://status.jitpack.io/) gives about 96,8 % uptime for "GitHub", which I'm not entirely sure what it is supposed to indicate, but that's well below the kind of numbers a reliable service should have.

### Proposed solution

Migrate artifact hosting to either Maven Central or GitHub Packages.

I replaced Jitpack repository configuration with local file based repository contained within the project source tree and copied the JARs + metadata there.

### Additional context

_No response_

jcpitre:
**Summary:**
closes #1588 
Modified the build.gradle of the modules to upload (main, core and model) to add signing and publishing.
Added a Github action to publish the released artifact. 

**Expected behavior:** 
To publish, you need to have a git tag directly in the commit, with semantic versioning format.
The GH action can be triggered by hand or for a prerelease or release.

If successful, there will be a new staging repo in https://s01.oss.sonatype.org/index.html#stagingRepositories containing the artefacts (e.g. https://s01.oss.sonatype.org/content/repositories/orgmobilitydata-1029/. This repo will likely be deleted by the time you read this, but it should be similar)
The artefacts should be signed (look for .sha1, .sha256 etc)

To test the deployment, you have to manually close the repo on sonatype. This will trigger acceptance tests for maven (artefacts are signed, etc).
If successful the staging repo will be available to use for testing. You will need to add a repository with the URL mentioned above (or similar) to your pom.xml or build.gradle.

Once the testing is done, the repo can be manually promoted to maven central.


.github/workflows/release_jars.yml
```diff
@@ -0,0 +1,48 @@
...
+# See https://s01.oss.sonatype.org for accessing these repos.
+# At this point it should manually be closed, which will trigger acceptance tests for maven central (but not transfer yet)
+# Once closed, thew repo is available for testing.
+# After testing, it can be manually promoted on the sonatype site, which will then publish to maven central.
+name: Upload Release Jars to Sonatype
+
+on:
+  release:
+    types: [ prereleased, released ]
+    workflow_dispatch:
```
davidgamez, line 15: It might be more generic to trigger the action by listening to tag pushes so that any new tag will be published. 
Example:
```
on:
  push:
    tags:
      - '*'
```

.github/workflows/release_jars.yml
```diff
@@ -0,0 +1,48 @@
+# This action will publish jars that can be used by third parties.
+# Pretty much everything is done through gradle.
+# The version used will be according to the axion-release-plugin, meaning it will take a tag if present,
+# if not it will create a SNAPSHOT version.
+# gradle will compile, sign then upload artifacts to a Sonatype staging repo.
+# See https://s01.oss.sonatype.org for accessing these repos.
+# At this point it should manually be closed, which will trigger acceptance tests for maven central (but not transfer yet)
+# Once closed, thew repo is available for testing.
+# After testing, it can be manually promoted on the sonatype site, which will then publish to maven central.
+name: Upload Release Jars to Sonatype
```
davidgamez, line 10: What do you think if we  add the jobs implemented here to https://github.com/MobilityData/gtfs-validator/blob/85beb70856203dd8edc7f4b06ef5b89a840e5908/.github/workflows/publish_assets.yml#L1 to avoid sync issues between what we have in the Github repository and what is published in maven

.github/workflows/release_jars.yml
jcpitre: Sure. But it should be restricted to tags that have the proper syntax (e.g. v4.2.1), If not the axion-release plugin will not work. I initially was using an uppercase V at the beginning of the version, and it did not work! It came with some headaches.

.github/workflows/release_jars.yml
jcpitre: So you want to combine the two actions?
Or have one trigger the other

davidgamez: Combining the two actions seems to be a reasonable solution.

jcpitre: Combining is possible, but having a tag push as a trigger instead of a release event will not work since publish_release uploads to a release and needs the release event to obtain the release url.

davidgamez: I see your point. My bad, I was looking at the default example from `upload-release-asset` action and the implementation is the other way around. They create a release as part of the tag push: https://github.com/actions/upload-release-asset#example-workflow---upload-a-release-asset. 
To conclude, let's leave the pre-release and release triggers.

.github/workflows/release_jars.yml
jcpitre: Did not put tag as a trigger in the end since the other part of publish_assets.yml needs a release trigger.

.github/workflows/release_jars.yml
jcpitre: Still kept them combined but added a workflow_dispatched trigger that is used only by the sonatype upload.

davidgamez:
This looks good to me. As a last ask, can you update the [release documentation](https://github.com/MobilityData/gtfs-validator/blob/master/docs/RELEASE.md), adding the steps to publish to maven? Thanks!

davidgamez (approved):

jcpitre:
You can now use Maven Central as repository for released versions of GTFS-validator artifacts.
For the dependency in pom.xml or build.gradle, the groupId is now 
     `org.mobilitydata.gtfs-validator` 
(the version built with Jitpack had groupId com.github.MobilityData.gtfs-validator)

PR #1607, opened 2023-11-22
feat: Support injection of arbitrary types via ValidationContext

aababilov:
ValidationContext is used to inject params to constructors of *Validator classes. ValidationContext initially supported only CountryCode and CurrentDateTime, e.g.:

  @Inject
  ExpiredCalendarValidator(
      CurrentDateTime currentDateTime, // Injected from ValidationContext
      GtfsCalendarTableContainer calendarTable,
      GtfsCalendarDateTableContainer calendarDateTable) {...}

This change allows the user to inject any types required for custom validation logic.

### DIFF
--- a/core/src/main/java/org/mobilitydata/gtfsvalidator/validator/ValidationContext.java
+++ b/core/src/main/java/org/mobilitydata/gtfsvalidator/validator/ValidationContext.java
@@ -16,7 +16,7 @@
 
 package org.mobilitydata.gtfsvalidator.validator;
 
-import com.google.auto.value.AutoValue;
+import com.google.common.collect.ImmutableMap;
 import org.mobilitydata.gtfsvalidator.input.CountryCode;
 import org.mobilitydata.gtfsvalidator.input.CurrentDateTime;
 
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/mobilitydata/gtfsvalidator/validator/ValidationContext.java
```diff
@@ -16,26 +16,33 @@
...
+  private final ImmutableMap<Class<?>, Object> context;
+
+  private ValidationContext(ImmutableMap<Class<?>, Object> context) {
+    this.context = context;
   }
 
   /**
    * Represents a name of a GTFS feed, such as "nl-openov".
    *
    * @return the @code{GtfsFeedName} representing the feed's name
```
bdferris-v2, line 41: Unrelated to your PR, but this method comment seems incorrect.  Understood if you don't want to fix it here, but if you were willing.

bdferris-v2 (approved):

bdferris-v2:
Added @davidgamez as an FYI.

core/src/main/java/org/mobilitydata/gtfsvalidator/validator/ValidationContext.java
aababilov: Let's fix it while I am here. This comment is pretty confusing.

aababilov:
Thanks a lot for the quick review!

Issue #1292, opened 2022-11-14
Allow the validator to accept a "time of validation" argument


atvaccaro:
### Describe the problem

As part of [the Cal-ITP data pipeline](https://github.com/cal-itp/data-infra/), we collect and validate GTFS and GTFS-RT data over time, which includes executing the Schedule and RT validators against hundreds of feeds daily. Sometimes we need to re-process old data for a variety of purposes such as changing metadata or updating validator versions; usually the best way is to just run a full historical re-processing starting from the raw data. While this is fine for most situations, it does mean that sometimes we are validating data much later than it was originally collected, which affects validation checks that rely on the [validator execution time](https://github.com/MobilityData/gtfs-validator/blob/master/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/FeedExpirationDateValidator.java#L53). (The same situation can occur simply due to retries or latency in our batch processing data pipeline.)

I believe currently there are two affected checks: [feed_expiration_date_7_days](https://github.com/MobilityData/gtfs-validator/blob/1636e44bbc1cefb51fb8cea609638c9a96ed4f3f/RULES.md#feed_expiration_date_7_days) and [feed_expiration_date_30_days](https://github.com/MobilityData/gtfs-validator/blob/1636e44bbc1cefb51fb8cea609638c9a96ed4f3f/RULES.md#feed_expiration_date_30_days).

### Proposed solution

The validator accepts a timestamp argument to use when determining whether a feed has expired, etc. This would allow data pipelines to be more idempotent and less reliant on actual validator execution time.

### Additional context

It looks like we have an [open PR](https://github.com/MobilityData/gtfs-validator/pull/1289) to add additional checks that would be affected by this.

Deferring to @themightychris and @nkkl but I think we (Jarvus) can look at helping with this after the MVP web validator is deployed.

derhuerst:
Would it be possible to just use the input GTFS dataset's time modification time?

wklumpen:
In a similar vein, I'm now getting errors in 4.1.0 that I wasn't in 4.0.0 via `ExpiredCalendarValidator` due to me wanting to validate an expired feed (for research purposes)

Seems like this shouldn't be a system error? Do I need to open a new issue for this?

bradyhunsaker:
It looks like the date is set just once (in a CurrentDateTime object) in ValidationRunner.java. That makes things easier.
This could easily be changed to use an optional date value added to ValidationContext.

[...]

PR #1628, opened 2024-01-02
feat: Add an argument for the cli (--date) that specifies the date to use during verification.

bradyhunsaker:
**Summary:**

Resolves issue #1292.

By specifying a date and time with the --date argument, the validator will use that date instead of the current date. This enables replicating earlier runs and debugging rules that depend on time, like feed_expiration_date_30_days. Only the cli interface has this support added.

The main challenge here was how to handle this gracefully in the output.

A new field "dateForValidation" is added to the JSON report for all runs. I defer to project owners if they prefer a different name; it needs to contrast with the existing field "validatedAt".

The HTML report has a line added only if the validation date is different than the current date.

Some minor refactoring is included to make the new logic simpler, like changing from Date to ZonedDateTime in ValidationRunner.java.

The --date argument requires both a date and a time, because I could not find a way to keep the behavior and code simple when specifying a date only. Adding the minor burden of including the time for the few users of the argument seems better than the workarounds I considered.

A new argument is added to HtmlReportGenerator.generateReport, which is unfortunate given how infrequently this feature is expected to be used. I would have preferred to compare dates inside generateReport, but the run date is passed in a custom format, so it would be awkward to do a comparison inside the function.

**Expected behavior:** 

In all interfaces and all runs, a new JSON field "dateForValidation" is added. In the common default case, this is generally slightly before the date and time in "validatedAt", because it is set earlier in execution.

In the cli, if the --date argument is given a valid date and time, that date and time will be used when checking rules that depend on time, such as checking for expired feeds.

If a date is specified and it is not today, then the HTML report will include a sentence stating the date and time used.

- [X ] Run the unit tests with `gradle test` to make sure you didn't break anything
- [X ] Format the title like "feat: [new feature short description]". Title must follow the Conventional Commit Specification(https://www.conventionalcommits.org/en/v1.0.0/).
- [X ] Linked all relevant issues

### DIFF
--- a/cli/src/main/java/org/mobilitydata/gtfsvalidator/cli/Arguments.java
+++ b/cli/src/main/java/org/mobilitydata/gtfsvalidator/cli/Arguments.java
@@ -21,6 +21,9 @@ import com.google.common.flogger.FluentLogger;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.nio.file.Path;
+import java.time.ZoneId;
+import java.time.ZonedDateTime;
+import java.time.format.DateTimeFormatter;
 import org.mobilitydata.gtfsvalidator.input.CountryCode;
 import org.mobilitydata.gtfsvalidator.runner.ValidationRunnerConfig;
 
@@ -51,6 +54,14 @@ public class Arguments {
           "Country code of the feed, e.g., `nl`. "
               + "It must be a two-letter country code (ISO 3166-1 alpha-2)")
   private String countryCode;
+
+  @Parameter(
+      names = {"--date"},
+      description =
+          "Date to simulate when validating, in ISO_LOCAL_DATE_TIME format like "
+              + "'2001-01-30T12:34'. By default, the current date is used. "
+              + "This option enables debugging rules like feed expiration.")
+  private String dateString;
 
   @Parameter(
       names = {"-u", "--url"},
@@ -110,6 +121,11 @@ public class Arguments {
     }
     if (countryCode != null) {
       builder.setCountryCode(CountryCode.forStringOrUnknown(countryCode));
+    }
+    if (dateString != null) {
+      builder.setZonedDateTime(
+          ZonedDateTime.parse(dateString,
+                              DateTimeFormatter.ISO_LOCAL_DATE_TIME.withZone(ZoneId.systemDefault())));
     }
     if (validationReportName != null) {
       builder.setValidationReportFileName(validationReportName);
--- a/cli/src/main/java/org/mobilitydata/gtfsvalidator/cli/Arguments.java
+++ b/cli/src/main/java/org/mobilitydata/gtfsvalidator/cli/Arguments.java
@@ -21,8 +21,7 @@ import com.google.common.flogger.FluentLogger;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.nio.file.Path;
-import java.time.ZoneId;
-import java.time.ZonedDateTime;
+import java.time.LocalDate;
 import java.time.format.DateTimeFormatter;
 import org.mobilitydata.gtfsvalidator.input.CountryCode;
 import org.mobilitydata.gtfsvalidator.runner.ValidationRunnerConfig;
@@ -58,8 +57,8 @@ public class Arguments {
   @Parameter(
       names = {"-d", "--date"},
       description =
-          "Date to simulate when validating, in ISO_LOCAL_DATE_TIME format like "
-              + "'2001-01-30T12:34'. By default, the current date is used. "
+          "Date to simulate when validating, in ISO_LOCAL_DATE format like "
+              + "'2001-01-30'. By default, the current date is used. "
               + "This option enables debugging rules like feed expiration.")
   private String dateString;
 
@@ -123,9 +122,7 @@ public class Arguments {
       builder.setCountryCode(CountryCode.forStringOrUnknown(countryCode));
     }
     if (dateString != null) {
-      builder.setZonedDateTime(
-          ZonedDateTime.parse(
-              dateString, DateTimeFormatter.ISO_LOCAL_DATE_TIME.withZone(ZoneId.systemDefault())));
+      builder.setDateForValidation(LocalDate.parse(dateString, DateTimeFormatter.ISO_LOCAL_DATE));
     }
     if (validationReportName != null) {
       builder.setValidationReportFileName(validationReportName);
--- a/cli/src/main/java/org/mobilitydata/gtfsvalidator/cli/Arguments.java
+++ b/cli/src/main/java/org/mobilitydata/gtfsvalidator/cli/Arguments.java
@@ -56,7 +56,7 @@ public class Arguments {
   private String countryCode;
 
   @Parameter(
-      names = {"--date"},
+      names = {"-d", "--date"},
       description =
           "Date to simulate when validating, in ISO_LOCAL_DATE_TIME format like "
               + "'2001-01-30T12:34'. By default, the current date is used. "
--- a/cli/src/main/java/org/mobilitydata/gtfsvalidator/cli/Arguments.java
+++ b/cli/src/main/java/org/mobilitydata/gtfsvalidator/cli/Arguments.java
@@ -124,8 +124,8 @@ public class Arguments {
     }
     if (dateString != null) {
       builder.setZonedDateTime(
-          ZonedDateTime.parse(dateString,
-                              DateTimeFormatter.ISO_LOCAL_DATE_TIME.withZone(ZoneId.systemDefault())));
+          ZonedDateTime.parse(
+              dateString, DateTimeFormatter.ISO_LOCAL_DATE_TIME.withZone(ZoneId.systemDefault())));
     }
     if (validationReportName != null) {
       builder.setValidationReportFileName(validationReportName);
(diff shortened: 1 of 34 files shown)


main/src/main/resources/report.html
```diff
@@ -229,6 +229,7 @@ <h1>GTFS Schedule Validation Report</h1>
         <span th:text="${config.gtfsSource}"></span><span
                 th:text="${config.countryCode.isUnknown()} ? '. No country code was provided.' : ', with the country code: ' + ${config.countryCode} + '.'"></span>
         </br>
+        <span th:if="${is_different_date}" th:text="'The date and time used during validation was ' + ${config.zonedDateTime.toLocalDateTime()} + '.'"></span></p>
```
jcpitre, line 232: The format for this date is different from the one when the validator was run, e.g. `2024-01-03 at 17:14:01 EST` vs `2020-01-02T12:34`.
We should consider makeing the format the same.

main/src/main/java/org/mobilitydata/gtfsvalidator/runner/ValidationRunnerConfig.java
```diff
@@ -88,6 +94,8 @@ public abstract static class Builder {
 
     public abstract Builder setCountryCode(CountryCode countryCode);
 
+    public abstract Builder setZonedDateTime(ZonedDateTime zonedDateTime);
```
jcpitre, line 97: The method name is not very instructive. We know the type of what we are setting.
Maybe replace with something like `setDateForValidation`?

main/src/main/java/org/mobilitydata/gtfsvalidator/report/JsonReportSummary.java
```diff
@@ -55,6 +56,7 @@ public JsonReportSummary(
       this.validationReportName = config.validationReportFileName();
       this.htmlReportName = config.htmlReportFileName();
       this.countryCode = config.countryCode().getCountryCode();
+      this.dateForValidation = config.zonedDateTime().toString();
```
jcpitre, line 59: This date and the validatedAt date have different formats. They should be the same.
And this might be outside of the scope of this PR, but since they are destined to be read by a program, they should be easily parsable. ISO 8601 with offset might be more suited for the dates in the json file. e.g. `2024-01-03T17:14:01-05:00`
This could be delegated to another issue.

cli/src/main/java/org/mobilitydata/gtfsvalidator/cli/Arguments.java
```diff
@@ -52,6 +55,14 @@ public class Arguments {
               + "It must be a two-letter country code (ISO 3166-1 alpha-2)")
   private String countryCode;
 
+  @Parameter(
+      names = {"--date"},
```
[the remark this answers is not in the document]
bradyhunsaker: I can't remember why I didn't include it initially. I think I was thinking that this is likely to be a rare argument that doesn't "need" a short form, but I agree that consistency is good. Since there's no other argument starting with d, I'll add "-d".

main/src/main/java/org/mobilitydata/gtfsvalidator/runner/ValidationRunner.java
```diff
@@ -268,9 +266,9 @@ public static void exportReport(
             "Error creating output directory: %s", config.outputDirectory());
       }
     }
-    SimpleDateFormat formatter = new SimpleDateFormat("yyyy-MM-dd 'at' HH:mm:ss z");
-    Date now = new Date(System.currentTimeMillis());
-    String date = formatter.format(now);
+    ZonedDateTime now = ZonedDateTime.now();
```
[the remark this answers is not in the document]
bradyhunsaker: Yes. There have always been two times at play: the time used for validation and the time indicated in the report. In the current code (before this PR), only the report generation time is included in the output.

Note that this part of the code is a refactor. The existing code uses System.currentTimeMillis() while the new code uses ZonedDateTime.now(). In either case, this time included in the report is the time that the report is generated (after validation steps have been done).

The time used for validation is set at the start of processing if the new --date option isn't used. In that common case the two times differ by the processing time, which could be a fraction of a second or could be many minutes for large feeds (maybe even hours? I don't know how long the biggest feeds take).

That was already true and remains true. A difference now is that both times are included in the JSON report, and both are included in the HTML report if the date differs.

I'm open to changing any of that.

main/src/main/resources/report.html
bradyhunsaker: Yes. I admit to noticing this and being a bit lazy. Largely it was because the existing format is a custom format. I like your suggestion later about changing that format, in which case it will be easy for them to use the same format.

main/src/main/java/org/mobilitydata/gtfsvalidator/report/JsonReportSummary.java
bradyhunsaker: I agree that changing the existing date format to a non-custom one would be good. I didn't want to presume to do that, but it sounds like you're comfortable with it. It would make it easy to use matching date formats.

I'll make a separate PR with just that change, with the intent of submitting it before this one. My one concern is whether there is code out there that is parsing this custom date format of validatedAt. Are you concerned about that?

main/src/main/java/org/mobilitydata/gtfsvalidator/runner/ValidationRunner.java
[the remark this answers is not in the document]
bradyhunsaker: I think I misunderstood your original question. Date and time handling is confusing.

The computation of is_different_date converts both ZonedDateTime variables to LocalDate before comparing. It's looking to see if the date (not the time part) is different. In the common case that --date is not specified and the validation run completes on the same calendar date as it starts, is_different_date will be false and the HTML report will not have anything added. As you say, the date+time will always be different, but that isn't what is checked. My goal was to be conservative about changing the human-readable HTML output (although I'm open-minded).

Part of why this is confusing is that I called the argument --date, but you must specify a time as well, so it's really a date+time.

I have a suggestion to reduce the confusion: Refactor the code by replacing "currentDateTime" in ValidationContext with a LocalDate (no time). Currently the type is a ZonedDateTime, which is why I used that type in this PR and require that a time be specified along with a date. But I checked the uses of the ValidationContext, and the rules all convert to LocalDate when using currentDateTime anyway. (example: https://github.com/MobilityData/gtfs-validator/blob/3d5ef6f31839f74fff3338c8a8d9869df88cc7c0/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/DateTripsValidator.java#L76)

I'll create a PR with that refactor so you see what I mean. It won't change the current functionality at all.
After that change (if you agree), then I'll change this PR so that only a date is specified by --date and hopefully it makes everything simpler and clearer. That will also resolve the issue about different timestamp formats; with the change the new one will just be a date. (Though I'll still follow up separately about making the existing one machine readable.)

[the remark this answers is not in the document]
jcpitre: Sorry I missed that the comparison was based only on the date.

main/src/main/java/org/mobilitydata/gtfsvalidator/report/JsonReportSummary.java
bradyhunsaker: No worries. I didn't interpret your comment that way. I like fixing things up when there's agreement to do so.

emmambd:
Thanks for this awesome work @bradyhunsaker! I'm tagging @atvaccaro, @derhuerst, and @wklumpen who expressed interest in this in #1292 in case they have thoughts as well.

main/src/main/java/org/mobilitydata/gtfsvalidator/runner/ValidationRunner.java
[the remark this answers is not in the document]
bradyhunsaker: I just created PR #1636, a refactor to simplify and clarify the existing date handling. I'll wait to see if that is accepted and merged. If so, it will simplify this PR.

bradyhunsaker:
I brought in the changes from PR #1636 and other changes in the master branch. I'll update here when I have the code ready for a re-review.

main/src/main/resources/report.html
bradyhunsaker: Now the date-for-validation is just a date while the run time is a date+time. So there's no mismatch. I would still like to move away from the custom format, but not in this PR since it no longer is an issue.

bradyhunsaker:
OK, ready for review.
I committed changes so that it now specifies just a date (no time) and uses the dateForValidation name.

jcpitre (approved):
This is awesome! Thanks

wklumpen:
Amazing work! Thanks very much for the work. Now I don't have to just ignore the "not within 30 days" warning :)

cli/src/main/java/org/mobilitydata/gtfsvalidator/cli/Arguments.java
```diff
@@ -52,6 +54,14 @@ public class Arguments {
               + "It must be a two-letter country code (ISO 3166-1 alpha-2)")
   private String countryCode;
 
+  @Parameter(
+      names = {"-d", "--date"},
+      description =
+          "Date to simulate when validating, in ISO_LOCAL_DATE format like "
+              + "'2001-01-30'. By default, the current date is used. "
+              + "This option enables debugging rules like feed expiration.")
```
jcpitre, line 62: @bradyhunsaker It's a bit late, but I was wondering about this sentence. I don't see where these debugging rules are enabled.

bradyhunsaker: It's not clear. I meant that the main purpose of this option is debugging. For example, it could be used to debug an error with a rule like feed expiration.

There are not special debugging rules associated with it.
I think the use of the word "enables" was a bad choice. Feel free to remove or edit the description.

PR #1519, opened 2023-06-28
fix: don't throw exception on calendar with no days

Whoops:
**Summary:**

Fix ExpiredCalendarValidator throwing NoSuchElementException when checking a calendar with no days set to 1.

**Expected behavior:** 

Checking a calendar.txt file where a row has no days set to 1 does not crash the validator.

Please make sure these boxes are checked before submitting your pull request - thanks!

- [x] Run the unit tests with `gradle test` to make sure you didn't break anything
- [x] Format the title like "feat: [new feature short description]". Title must follow the Conventional Commit Specification(https://www.conventionalcommits.org/en/v1.0.0/).
- [ ] Linked all relevant issues
- [ ] Include screenshot(s) showing how this pull request works and fixes the issue(s)

PR #1646, opened 2024-01-16
fix: prevent exception raised when serviceId is missing from calendar file

davidgamez:
**Summary:**

This PR fixes the issue raised [here](https://github.com/MobilityData/gtfs-validator/pull/1519#issuecomment-1891809181).

How to reproduce the bug:
- Have a feed with a service ID in the calendar_dates.txt file but missing in calendar.txt(This is a valid use case) and multiple or all calendar expired service IDs. Example: https://gtfs.pro/files/uran/improved-gtfs-dft-gtfs.zip
- Execute the CLI or desktop application
- Behavior: The report is missing all calendar expired notices for the service IDs after the row that contains the missing service ID on the calendar.txt. A _java.util.NoSuchElementException_ exception can be found in the logs.

**Expected behavior:** 
All calendar expired notices are reported.

The root cause of the issue is the improper manipulation of the _Optional.get_ method.

**The fix:**
- Fixed calls to _Optional.get_, ensuring the value is present before calling the method.

**What is new:**
- For feeds that have only calendar_dates.txt files, expired notices will be reported only when all service IDs are expired
- For feeds with missing service IDs on the calendar.txt file, service IDs not in the calendar file will not be reported. 

**Acceptance Tests Results:**
Acceptance tests fail due to the number of expired feeds that have not been reported. All cases are foreign service ID key violations or feeds with only calendar_dates.txt service IDs.

- [x] Add or update any needed [documentation](https://github.com/MobilityData/gtfs-validator/tree/master/docs) to the repo 
- [x] Format the title like "feat: [new feature short description]." Title must follow the Conventional Commit Specification(https://www.conventionalcommits.org/en/v1.0.0/).


emmambd:
Based on internal discussion, next steps are: 

- Update logic so this rule only triggers on feeds that only have calendar_dates.txt when -all- calendars are expired rather than **any**

emmambd:
@davidgamez I'm a bit surprised by the results of the acceptance tests! I'd still expect around 71 feeds to return this warning because their calendar_dates.date fields are all expired. Here's a few examples: 

ca-alberta-lethbridge-transit-gtfs-765: https://storage.googleapis.com/storage/v1/b/mdb-latest/o/ca-alberta-lethbridge-transit-gtfs-765.zip?alt=media

ca-ontario-go-transit-gtfs-727: https://storage.googleapis.com/storage/v1/b/mdb-latest/o/ca-ontario-go-transit-gtfs-727.zip?alt=media

[Full original list of expired feeds is here](https://docs.google.com/spreadsheets/d/1gUPpopy0_kA7NGy345Zjztju9LYAOgiskntxVftR7Zg/edit?usp=sharing). I didn't check them all to see if the producer actually made changes that would've impacted the re-run of the acceptance tests.

emmambd:
@davidgamez I took a quick look at the acceptance test results and this logic looks good to me. 

When we run these analytics before each new release, this can be my list to clean up/check for replacements for feeds in the Mobility Database.

From a QA perspective, so long as we can run https://gtfs.pro/files/uran/improved-gtfs-dft-gtfs.zip successfully, this change can be passed.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/ExpiredCalendarValidator.java
```diff
@@ -54,12 +53,36 @@ public void validate(NoticeContainer noticeContainer) {
...
     for (var serviceId : servicePeriodMap.keySet()) {
       SortedSet<LocalDate> serviceDates = servicePeriodMap.get(serviceId);
-      if (!serviceDates.isEmpty() && serviceDates.last().isBefore(dateForValidation.getDate())) {
-        int csvRowNumber = calendarTable.byServiceId(serviceId).get().csvRowNumber();
-        noticeContainer.addValidationNotice(new ExpiredCalendarNotice(csvRowNumber, serviceId));
+      if (serviceDates.isEmpty()) {
+        continue;
       }
+      if (serviceDates.last().isBefore(dateForValidation.getDate())) {
+        if (calendarTable.byServiceId(serviceId).isPresent()) {
```
jcpitre, line 65: Nitpicking here. In the case you know the calendarTable is empty (isCalendarTableEmpty == true), why bother doing a lookup on it?
```suggestion
        if (!isCalendarTableEmpty && calendarTable.byServiceId(serviceId).isPresent()) {
```

davidgamez: That's true. However, reducing the logic in the if condition helps readability. In either case, empty or not, the _byService(serviceId)_ method will behave the same.

jcpitre (approved):
LGTM!

PR #1653, opened 2024-01-19
fix: Corrected a problem where code was not multithread safe

jcpitre:
**Summary:**

Closes #1647

When running the validator with multiple datasets (like in the web validator) we sometimes have errors in the results because the noticeContainer is shared between threads.

Please make sure these boxes are checked before submitting your pull request - thanks!

- [x] Run the unit tests with `gradle test` to make sure you didn't break anything
- [ ] Add or update any needed [documentation](https://github.com/MobilityData/gtfs-validator/tree/master/docs) to the repo 
- [x] Format the title like "feat: [new feature short description]". Title must follow the Conventional Commit Specification(https://www.conventionalcommits.org/en/v1.0.0/).
- [x] Linked all relevant issues
- [ ] Include screenshot(s) showing how this pull request works and fixes the issue(s)

PR #1657, opened 2024-01-24
Feat: 1639 release process for minor versions

jcpitre:
**Summary:**

Added some documentation about the branching and release strategy.


jcpitre:
@emmambd You can review now.

docs/BRANCHING.md
```diff
@@ -0,0 +1,51 @@
...
+- Generally the master branch is the one where current development is happening.
+- One objective would be to arrive to a point where the master branch can be released very often (every day?).
+  This means the master branch is releasable at all times. I believe this strategy is called **Continuous Delivery**.
+- At this point we don't think it's possible to do that until the testing infrastructure has been improved.
+- Nevertheless, the branching strategy takes this objective into account.
+- To further support this objective, the strategy wil try to minimize the number of long-lasting branches.
+
+## Release numbering
+- example 4.3.0
+- The first number (4 in this example) is incremented for major changes in the software. What major change mean has to be decided.
```
emmambd, line 14: Major = impacts the rules, either errors or warnings.

docs/BRANCHING.md
```diff
@@ -0,0 +1,51 @@
...
+- The second number (3 in this example) is incremented for minor changes in the software.
```
emmambd, line 15: Minor = non-breaking changes for users, and don't impact validation rules.

docs/BRANCHING.md
```diff
@@ -0,0 +1,51 @@
...
+- The third number (0 in this example) is used for maintenance releases.
+  A maintenance is created if some bug has to be corrected before the next minor release.
```
emmambd, line 17: @jcpitre Can we link to an example? Such as this one: https://github.com/MobilityData/gtfs-validator/pull/1653

docs/BRANCHING.md
```diff
@@ -0,0 +1,51 @@
...
+![](BranchingDiagram.png)
+## Feature/fix branches
```
emmambd, line 19: @jcpitre I think I'm fuzzy on when we'd use a feature branch in the validator. Wouldn't we generally rely on release branches?

docs/BRANCHING.md
```diff
@@ -0,0 +1,51 @@
...
+- We will call them feature branches for the rest of this document.
+- They stem from master, are usually short-lived and are merged back to master.
+
+## Maintenance/release branches
+- In general we will try not to create long-lived release branches.
+- A long-lived release branch will be created only if any merge to master that came after the previous release is deemed not suitable to go in the maintenance release.
+- For example the change could be too extensive, or introduce an undue risk in the maintenance release.
+  This is true in particular if the testing infrastructure is not sufficient.
+
+## Hotfixes
```
emmambd, line 31: It might be useful to explicitly state that we should label an issue as a "hotfix" issue, as discussed, so it's clear when a bug warrants this type of branching flow.

docs/BRANCHING.md
```diff
@@ -0,0 +1,51 @@
...
+- Where does the hotfix branch stem from:
+    - Generally from the commit that was tagged for the latest release.
+        - See for example commit A in the diagram.
+    - In keeping with the **Continuous delivery** objective, we will allow the hotfix branch to stem from a commit
+      downstream of the previous release commit if:
+        - we consider that all merges coming from feature branches that happened after the previous release can go in the maintenance release.
+            - For example if commit C from the diagram contains minor changes like typos we could decide that it's OK to include it in the maintenance release.
+            - In that case the hotfix branch can stem from commit C and be merged back into master in commit F.
+            - Commit F is tagged as the maintenance release and the maintenance release is built from there.
+    - If a merge on the master branch cannot be included in the maintenance release (See for example commit I) then we have no choice but to create a maintenance branch
```
emmambd, line 43: @jcpitre So basically, if there's been a commit to master that would need to result in a major release version, then we have to create a maintenance branch?

docs/BRANCHING.md
jcpitre: Feature branches are the way we already do development. Create a short lived branch corresponding to an issue, then merge it back to master. So it's not really related to releases. Should I remove this part?

emmambd: Nope! Just context I didn't have - makes sense now :)

docs/BRANCHING.md
jcpitre: So do you mean adding a rule for example will require a major number change?

emmambd: Correct - any additions or modifications to validation rules.

jcpitre: How often are the rules changed?

emmambd: I would expect at the once every 3 month cadence we talked about (we're talking about modifying equal_shape_dist_traveled_diff_coordinates and trip_distance_exceeds_shape_distance this quarter, for instance, and David's [expired_calendar PR](https://github.com/MobilityData/gtfs-validator/pull/1646) would be a rule modification). We also need to add new fares rules.

docs/BRANCHING.md
jcpitre: We could, except we decided not to do a maintenance release with the PR you mentioned since there was a workaround that did not involve a release. So the example is might not be as convincing as we want.

docs/BRANCHING.md
jcpitre: If there is something we absolutely need to fix in the released software, yes.
Considering that normally you would not want the major commit to make it's way in the maintenance release.
But this can be true for minor commits too. Except it's not as clear cut.

docs/BRANCHING.md
jcpitre: Added something to that effect.

docs/BRANCHING.md
jcpitre: Frankly I can't find any. The best is the one you mentioned. So I can put it in the document and add a comment in #1663 saying that this should lead to a maintenance release but we were lucky to have a workaround.

docs/BRANCHING.md
```diff
@@ -0,0 +1,51 @@
+# Branching Strategy for releases
+
+## General Considerations
+
+- Generally the master branch is the one where current development is happening.
+- One objective would be to arrive to a point where the master branch can be released very often (every day?).
```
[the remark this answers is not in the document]
jcpitre: Removed the section about releasing often or daily

docs/BRANCHING.md
jcpitre: So that means we would never really have minor releases, if there are changes of rules that often. Can we qualify the changes in the rules to make some more important (and impactful) than others?

docs/BRANCHING.md
jcpitre: Added the link

docs/BRANCHING.md
emmambd: @jcpitre I think the reaction would just be to release more :) For example, we could do a minor release for everything we've merged this past month. (Not suggesting it, but going forward after this quarter)

davidgamez (approved):
This LGTM. After this is approved, let's create an issue to align the release pipeline with the branching strategy.

Issue #1611, opened 2023-11-27
Performance issues with trip_distance_exceeds_shape_distance


emmambd:
### Describe the bug

Upon generating analytics for the [4.2 release](https://docs.google.com/spreadsheets/d/1AH4fELPPIB3R3w_TPGUBzSn_T6cVuYpmekW-er1Xogg/edit#gid=0), we found that several datasets failed to run, and we assume that's because of the new `trip_distance_exceeds_shape_distance` error. 

These datasets failed with the message `The configured memory limit was reached`:

1. us-oregon-trimet-portland-streetcar-gtfs-247
2. us-washington-sound-transit-metro-transit-city-of-seattle-king-county-metro-gtfs-267
3. us-illinois-chicago-transit-authority-cta-gtfs-389
4. us-new-jersey-new-jersey-transit-nj-transit-gtfs-508
5. at-wien-wiener-lokalbahnen-wlb-gtfs-648
6. be-vlaams-gewest-de-lijn-gtfs-684
7. ca-alberta-edmonton-transit-system-gtfs-714
8. ca-ontario-toronto-transit-commission-gtfs-732
9. ca-quebec-reseau-de-transport-de-la-capitale-gtfs-757
10. de-berlin-verkehrsverbund-berlin-brandenburg-gtfs-782
11. Ie-dublin-dublin-bus-gtfs-947
12. fr-auvergne-rhone-alpes-cars-region-auvergne-rhone-alpes-transisere-gtfs-985
13. es-madrid-cercanias-madrid-gtfs-993
14. nl-unknown-allgo-keolis-gtfs-1077
15. ee-unknown-abuss-ou-gtfs-1095
16. gb-unknown-transport-for-greater-manchester-arriva-in-the-north-west-gtfs-1103
17. fi-unknown-porvoon-museorautatie-gtfs-1102
18. be-unknown-societe-regionale-wallonne-du-transport-gtfs-1212
19. it-lombardia-agenzia-mobilita-ambiente-territorio-gtfs-1231
20. gr-attiki-athens-urban-transport-organisation-organismos-astikon-sugkoinonion-oasa-gtfs-1228
21. ru-sankt-peterburg-peterburgskii-metropoliten-petersburg-metro-gtfs-1186
22. tw-unknown-taichung-gtfs-1277
23. dk-unknown-rejseplanen-gtfs-1292
24. gb-unknown-chiltern-railways-gtfs-1311
[...]

PR #1676, opened 2024-02-18
feat: reformat trip and shape dist validator

### LINKED ISSUE #1613: Implement threshold for triggering trip_distance_exceeds_shape_distance
### Describe the problem

We've seen `trip_distance_exceeds_shape_distance` trigger as an error from anywhere between 6-7% of feeds in the Mobility Database. Several producers have expressed concern with this new error. 

[Analysis report](https://docs.google.com/spreadsheets/d/1VcKtTgXtEHMQTZIMb43cpCnO38sHXwXI/edit?usp=sharing&ouid=117525155595068812086&rtpof=true&sd=true)

[Python script](https://github.com/MobilityData/mobility-database-catalogs/pull/275)

### Proposed solution

Implement a threshold where this ERROR only generates if the difference between the max trips.txt `shape_dist_traveled` and the max shapes.txt `shape_dist_traveled` is greater than 1.11m, similar to [our plans for equal_shape_dist_traveled_diff_coordinates](https://github.com/MobilityData/gtfs-validator/issues/1638). Rerun the acceptance tests to see how many feeds trigger this rule. 

A warning should be added for when trip_distance_exceeds_shape_distance is less than 1.11m. 

_No response_

### Additional context

_No response_

cka-y:
**Summary:**
Resolves #1613 and #1611 by implementing a distance threshold for the `trip_distance_exceeds_shape_distance` notice.

**Expected Behavior:**
- Introduces a 11.1m threshold for `trip_distance_exceeds_shape_distance`, triggering an `ERROR` for distances $\geq 11.1m$.
- Creates a new notice, `trip_distance_exceeds_shape_distance_below_threshold`, with `WARNING` severity for distances $\lt11.1m$.
- This update streamlines the cross-validation of trips against shapes by adhering to the GTFS specification. This approach has resulted in minimal changes to validation outcomes as evidenced [here](https://github.com/MobilityData/gtfs-validator/pull/1676#issuecomment-1951585993). Any minor discrepancies arise from instances where the feed does not comply with the specification, often leading to `ERROR` level notices such as `decreasing_or_equal_stop_time_distance` and `decreasing_shape_distance`.
  - By capitalizing on the expectation that both stop-time and shape distances should incrementally increase, the validation process is optimized. Instead of evaluating all points, we now only assess the last one, which changes our processing time complexity from linear to constant for these elements.

**Empirical Performance Comparison:**
Considering $n$ as the number of trips, $m$ as the number of stop-times, and $k$ as the number of shapes, the complexity in the worst-case scenario:
- For the `master` branch is $\Omega(k^2nm)$
- For the feature branch (`feat/1613`) is $\Omega(nk)$

**Statistical Performance Comparison:**
The performance improvements are depicted in the graph below. The datasets analyzed are from our catalog, with zipped file sizes of at least 1MB. Sizes have been normalized for a more meaningful comparison of slopes. The performance slope of the feature branch is significantly lower, decreasing from approximately 37 to approximately 25, indicating enhanced efficiency.
time_improvement_comparison


main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TripAndShapeDistanceValidator.java
```diff
@@ -32,55 +35,75 @@
...
-              double maxShapeDist =
+            trip -> {
+              String shapeId = trip.shapeId();
+              // Get distance for trip
+              int nbStopTimes = stopTimeTable.byTripId(trip.tripId()).size();
+              if (nbStopTimes == 0) {
+                return;
+              }
+              GtfsStopTime lastStopTime =
+                  stopTimeTable.byTripId(trip.tripId()).get(nbStopTimes - 1);
```
jcpitre, line 69: There's a bit of a waste of time calling byTripId(trip.tripId()) here and above. Don't know if it's significant.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TripAndShapeDistanceValidator.java
```diff
@@ -32,55 +35,75 @@
...
+              GtfsStop stop = stopTable.byStopId(lastStopTime.stopId()).orElse(null);
+              if (stop == null) {
+                return;
+              }
+              double maxStopTimeDist = lastStopTime.shapeDistTraveled();
+
+              // Get max shape distance for trip
+              GtfsShape maxShape =
```
jcpitre, line 77: I thought shape was also guaranteed to have the greatest distance on the last row. If not there's a error notice for that. Do we need to look for the greatest?

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TripAndShapeDistanceValidator.java
```diff
@@ -32,55 +35,75 @@
...
+                      .orElse(null);
+              if (maxShape == null) {
+                return;
+              }
+
+              double maxShapeDist = maxShape.shapeDistTraveled();
+              double distanceInMeters =
+                  getDistanceMeters(maxShape.shapePtLatLon(), stop.stopLatLon());
+              if (maxStopTimeDist > maxShapeDist) {
+                if (distanceInMeters > DISTANCE_THRESHOLD) {
```
jcpitre, line 89: Hmm, `>` and not `>=` ;)

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TripAndShapeDistanceValidator.java
```diff
@@ -32,55 +35,75 @@
...
+                      .max(Comparator.comparingDouble(GtfsShape::shapeDistTraveled))
+                      .orElse(null);
+              if (maxShape == null) {
+                return;
+              }
+
+              double maxShapeDist = maxShape.shapeDistTraveled();
+              double distanceInMeters =
+                  getDistanceMeters(maxShape.shapePtLatLon(), stop.stopLatLon());
+              if (maxStopTimeDist > maxShapeDist) {
```
jcpitre, line 88: I must admin I am a bit confused. You could have cases where maxStopTimeDist is way bigger than maxShapeDist, but where the coordinates are the same, and this would not generate an error?
@emmambd

emmambd: @jcpitre Discussed in meeting - good point! We talked about the notice defaulting to returning a warning in this case, rather than returning nothing. cc @cka-y for thoughts

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TripAndShapeDistanceValidator.java
```diff
@@ -32,55 +35,75 @@
...
+                if (distanceInMeters > DISTANCE_THRESHOLD) {
+                  noticeContainer.addValidationNotice(
+                      new TripDistanceExceedsShapeDistanceNotice(
+                          trip.tripId(), shapeId, maxStopTimeDist, maxShapeDist, distanceInMeters));
+                } else if (distanceInMeters > 0) {
+                  noticeContainer.addValidationNotice(
+                      new TripDistanceExceedsShapeDistanceBellowThresholdNotice(
```
emmambd, line 95: Bellow should be changed to below.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TripAndShapeDistanceValidator.java
```diff
@@ -98,15 +121,59 @@ static class TripDistanceExceedsShapeDistanceNotice extends ValidationNotice {
...
+      this.maxShapeDistanceTraveled = maxShapeDistanceTraveled;
+      this.maxTripDistanceTraveled = maxTripDistanceTraveled;
+      this.distance = distance;
+    }
+  }
+
+  /**
+   * The distance traveled by a trip should be less or equal to the max length of its shape.
+   *
+   * <p>The distance is less than the 11.1m threshold.
```
emmambd, line 144: Revised to clarify the threshold: `The distance between the last shape point and last stop point is less than the 11.1m threshold.`

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TripAndShapeDistanceValidator.java
```diff
@@ -32,55 +35,75 @@
...
+                }
+              }
             });
   }
 
-  /** The distance traveled by a trip should be less or equal to the max length of its shape. */
+  /**
+   * The distance traveled by a trip should be less or equal to the max length of its shape.
+   *
+   * <p>The distance is greater or equal to the 11.1m threshold.
```
emmambd, line 105: Revised to clarify threshold: `The distance between the last shape point and last stop point is greater than or equal to the 11.1m threshold.`

emmambd (commented):
Notice description + name minor revisions.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TripAndShapeDistanceValidator.java
```diff
@@ -98,15 +121,59 @@ static class TripDistanceExceedsShapeDistanceNotice extends ValidationNotice {
...
+    /** The distance in meters between the shape and the stop. */
+    private final double distance;
+
     TripDistanceExceedsShapeDistanceNotice(
         String tripId,
         String shapeId,
         double maxTripDistanceTraveled,
-        double maxShapeDistanceTraveled) {
+        double maxShapeDistanceTraveled,
+        double distance) {
```
emmambd, line 132: Thoughts on defining this as geoDistanceToShape to match column in https://gtfs-validator.mobilitydata.org/rules.html#stop_too_far_from_shape-rule?

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TripAndShapeDistanceValidator.java
cka-y: I agree it's a bit confusing. We've established a threshold for latitude/longitude coordinates, but our previous analysis shows that it doesn't always logically correspond to the distance traveled. Given we do not infer units, I believe the best we could do, which aligns with our existing default behavior, is to continue issuing a warning.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TripAndShapeDistanceValidator.java
cka-y: done ✅

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TripAndShapeDistanceValidator.java
emmambd: Agreed.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TripAndShapeDistanceValidator.java
cka-y: It is indeed the case. However, based on my analysis, this aspect does not affect performance the same way iterating through `stop_times` does. Therefore, I chose to be a bit more thorough in this step and include this check.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TripAndShapeDistanceValidator.java
cka-y: done ✅

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TripAndShapeDistanceValidator.java
jcpitre: Sorry I think I was not clear enough when I wrote:
> You could have cases where maxStopTimeDist is way bigger than maxShapeDist, but where the coordinates are the same, and this would not generate an error?

It should have ended with: `and this would generate neither error nor warning`

It still seems odd that the situation where maxStopTimeDist > maxShapeDist would not be reported at all.
It is a corner case (coordinates have to be the same but distances not the same) but I think it's also a low hanging fruit.

cka-y: just pushed a fix to address your concerns @jcpitre

jcpitre (approved):
Impressive, again!

PR #1711, opened 2024-03-12
fix: 1707 web validator waits for validation forever

### LINKED ISSUE #1707: Web validator waits for validation forever
### Describe the bug

Trying validation on a uploaded dataset hangs forever with this window:
image

This is intermittent

### Steps/Code to Reproduce

Upload a dataset to the validator and try to validate.
It's intermittent, and does not happen very often.

### Expected Results

A window directing to the validation report

### Actual Results

See above

### Screenshots

image

### Files used

_No response_

### Validator version

4.2.0

### Operating system

irrelevant

### Java version

_No response_

### Additional notes

_No response_

jcpitre:
**Summary:**

Closes #1707 

If there is no timeout on the version endpoint, this PR changes nothing.
Changed the code so it does not generate an exception when the version endpoint is called but times out.
Another problem was that the current version (obtained from the jar files) was discarded if there was a time out calling the version endpoint. 

**Expected behavior:** 

Web validator does not hang forever if there is a timeout while calling the version endpoint.
If the timeout happens, the version is still correct in report.html

** Testing tips **

The timeout is intermittent, and probably only occurs when the validator service is brought up.
To test this, you could add a delay in  [VersionResolver.java](https://github.com/MobilityData/gtfs-validator/blob/dee5802afc2f0e37b0299a5d6dca192c08ee4af8/main/src/main/java/org/mobilitydata/gtfsvalidator/util/VersionResolver.java#L150), e.g.  Thread.sleep(5100) to simulate that the endpoint takes a long time to respond. The delay should be above 5000 ms, since it's the time used for the timeout.

### DIFF
--- a/main/src/main/java/org/mobilitydata/gtfsvalidator/util/VersionResolver.java
+++ b/main/src/main/java/org/mobilitydata/gtfsvalidator/util/VersionResolver.java
@@ -102,6 +102,7 @@ public class VersionResolver {
             logger.atSevere().withCause(ex).log("Error resolving version info");
           }
           VersionInfo info = VersionInfo.create(currentVersion, latestReleaseVersion);
+          Thread.sleep(5100);
           resolvedVersionInfo.set(info);
           return info;
         }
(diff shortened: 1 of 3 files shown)


main/src/main/java/org/mobilitydata/gtfsvalidator/util/VersionResolver.java
```diff
@@ -153,7 +166,7 @@ private Optional<String> resolveLatestReleaseVersion(Optional<String> currentVer
         new URL(
             String.format(
                 "%s?application_type=%s&current_version=%s",
-                LATEST_RELEASE_VERSION_URL, applicationType, currentVersion.orElse("Ï")));
```
jcpitre, line 156: I don't know why this character `"Ï"` was there.

davidgamez (approved):
LGTM!

PR #1803, opened 2024-08-23
feat: 1791 feed infotxt should be added to missing required file when translationstxt exists

### LINKED ISSUE #1791: feed_info.txt should be added to missing_required_file when translations.txt exists
# `Commit Nº 2`
## AUTHOR × DATE
* **isabelle-dr** ( _isabelle-dr_ )
* **July 18, 2024**   @   04:17 AM ETC (08:17 AM UTC)
## MESSAGE
```
Update requirement for feed_info.txt (#460)
* Update requirement for feed_info.txt
* Remove bold from Recommended
```

```
https://github.com/google/transit/commit/cc60dc4275cc44662163efa8cefd3e9e5e504978
```
---

qcdyx:
**Summary:**

Closes #1791 
**Expected behavior:** 
1. when translations.txt exists, if there's no feed_info.txt, create missing required file notice.
image

2. when there's no translation.txt, and no feed_info.txt either, create missing recommended file notice.
image

### DIFF
--- a/core/src/main/java/org/mobilitydata/gtfsvalidator/table/AnyTableLoader.java
+++ b/core/src/main/java/org/mobilitydata/gtfsvalidator/table/AnyTableLoader.java
@@ -32,9 +32,14 @@ public final class AnyTableLoader {
   private static final FluentLogger logger = FluentLogger.forEnclosingClass();
   private static final List<Class<? extends FileValidator>> singleFileValidatorsWithParsingErrors =
       new ArrayList<>();
-
+  private static boolean hasTranslations = false;
   private static final List<Class<? extends SingleEntityValidator>>
       singleEntityValidatorsWithParsingErrors = new ArrayList<>();
+
+  public static void setHasTranslations(boolean translations) {
+    hasTranslations = translations;
+    System.out.println("hasTranslations: " + translations);
+  }
 
   public List<Class<? extends FileValidator>> getValidatorsWithParsingErrors() {
     return Collections.unmodifiableList(singleFileValidatorsWithParsingErrors);
@@ -206,7 +211,11 @@ public final class AnyTableLoader {
         tableDescriptor.createContainerForInvalidStatus(
             GtfsTableContainer.TableStatus.MISSING_FILE);
     if (tableDescriptor.isRecommended()) {
-      noticeContainer.addValidationNotice(new MissingRecommendedFileNotice(gtfsFilename));
+      if (hasTranslations && gtfsFilename.contains("feed_info")) {
+        noticeContainer.addValidationNotice(new MissingRequiredFileNotice(gtfsFilename));
+      } else {
+        noticeContainer.addValidationNotice(new MissingRecommendedFileNotice(gtfsFilename));
+      }
     }
     if (tableDescriptor.isRequired()) {
       noticeContainer.addValidationNotice(new MissingRequiredFileNotice(gtfsFilename));
--- a/core/src/main/java/org/mobilitydata/gtfsvalidator/table/AnyTableLoader.java
+++ b/core/src/main/java/org/mobilitydata/gtfsvalidator/table/AnyTableLoader.java
@@ -32,7 +32,7 @@ public final class AnyTableLoader {
   private static final FluentLogger logger = FluentLogger.forEnclosingClass();
   private static final List<Class<? extends FileValidator>> singleFileValidatorsWithParsingErrors =
       new ArrayList<>();
-  
+
   private static final List<Class<? extends SingleEntityValidator>>
       singleEntityValidatorsWithParsingErrors = new ArrayList<>();
 
--- a/core/src/main/java/org/mobilitydata/gtfsvalidator/table/AnyTableLoader.java
+++ b/core/src/main/java/org/mobilitydata/gtfsvalidator/table/AnyTableLoader.java
@@ -38,7 +38,6 @@ public final class AnyTableLoader {
 
   public static void setHasTranslations(boolean translations) {
     hasTranslations = translations;
-    System.out.println("hasTranslations: " + translations);
   }
 
   public List<Class<? extends FileValidator>> getValidatorsWithParsingErrors() {
--- a/core/src/main/java/org/mobilitydata/gtfsvalidator/table/AnyTableLoader.java
+++ b/core/src/main/java/org/mobilitydata/gtfsvalidator/table/AnyTableLoader.java
@@ -32,13 +32,8 @@ public final class AnyTableLoader {
   private static final FluentLogger logger = FluentLogger.forEnclosingClass();
   private static final List<Class<? extends FileValidator>> singleFileValidatorsWithParsingErrors =
       new ArrayList<>();
-  private static boolean hasTranslations = false;
   private static final List<Class<? extends SingleEntityValidator>>
       singleEntityValidatorsWithParsingErrors = new ArrayList<>();
-
-  public static void setHasTranslations(boolean translations) {
-    hasTranslations = translations;
-  }
 
   public List<Class<? extends FileValidator>> getValidatorsWithParsingErrors() {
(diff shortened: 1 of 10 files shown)


core/src/main/java/org/mobilitydata/gtfsvalidator/table/GtfsFeedLoader.java
```diff
@@ -100,6 +100,9 @@ public GtfsFeedContainer loadAndValidate(
     Map<String, GtfsTableDescriptor<?>> remainingDescriptors =
         (Map<String, GtfsTableDescriptor<?>>) tableDescriptors.clone();
     for (String filename : gtfsInput.getFilenames()) {
+      if (filename.contains("translations")) {
```
jcpitre, line 103: If I am not mistaken, the code depends on the order of the filenames obtained with getFilenames, i.e. it requires that translations is processed before feed_into. I don't think this is guaranteed
It would be better to have one loop that checks for the presence of translations, then the original loop that processes the files.

core/src/main/java/org/mobilitydata/gtfsvalidator/table/GtfsFeedLoader.java
```diff
@@ -100,6 +100,9 @@ public GtfsFeedContainer loadAndValidate(
...
+        AnyTableLoader.setHasTranslations(true);
+      }
       GtfsTableDescriptor<?> tableDescriptor = remainingDescriptors.remove(filename.toLowerCase());
```
jcpitre, line 106: I noticed here that we set the filename to lowercase. Is it because a filename with uppercase letters is allowed?
If it is, I would also make sure we do a lowercase on the filename when checking for the presence of the `translation` string.

jcpitre:
Maybe you could add some tests?

core/src/main/java/org/mobilitydata/gtfsvalidator/table/GtfsTableDescriptor.java
```diff
@@ -12,6 +12,9 @@ public abstract class GtfsTableDescriptor<T extends GtfsEntity> {
   // True if the specified file is required in a feed.
   private boolean required;
 
+  // True if the specified file is conditionally required.
+  private boolean conditionallyRequired;
```
davidgamez, line 16: We already have an annotation to _mark_ the conditional required files, [ConditionalRequired](https://github.com/MobilityData/gtfs-validator/blob/master/model/src/main/java/org/mobilitydata/gtfsvalidator/annotation/ConditionallyRequired.java)

core/src/main/java/org/mobilitydata/gtfsvalidator/table/AnyTableLoader.java
```diff
@@ -32,10 +32,14 @@ public final class AnyTableLoader {
   private static final FluentLogger logger = FluentLogger.forEnclosingClass();
   private static final List<Class<? extends FileValidator>> singleFileValidatorsWithParsingErrors =
       new ArrayList<>();
-
+  private static boolean hasTranslations = false;
```
davidgamez, line 35: This field is not necessary, as the presence of the translation file can be detected by the method [GtfsTableContainer.isMissing](https://github.com/MobilityData/gtfs-validator/blob/b597b7c2521bae18041e2f7c9879304c33c16aa5/core/src/main/java/org/mobilitydata/gtfsvalidator/table/GtfsTableContainer.java#L103)

core/src/main/java/org/mobilitydata/gtfsvalidator/table/AnyTableLoader.java
```diff
@@ -206,7 +210,11 @@ public static GtfsTableContainer loadMissingFile(
         tableDescriptor.createContainerForInvalidStatus(
             GtfsTableContainer.TableStatus.MISSING_FILE);
     if (tableDescriptor.isRecommended()) {
-      noticeContainer.addValidationNotice(new MissingRecommendedFileNotice(gtfsFilename));
+      if (hasTranslations && gtfsFilename.contains("feed_info")) {
+        noticeContainer.addValidationNotice(new MissingRequiredFileNotice(gtfsFilename));
+      } else {
+        noticeContainer.addValidationNotice(new MissingRecommendedFileNotice(gtfsFilename));
+      }
```
davidgamez, line 217: Let's not add specific file logic to this class to keep it clean and generic.

davidgamez:
I think we can follow an already implemented pattern for the conditionally required files. Example: [calendar.txt](https://gtfs.org/schedule/reference/#calendartxt). 
Main steps:
- Remove Recommended annotation from the schema class and add ConditionalRequired
- Create a validator similar to [MissingCalendarAndCalendarDateValidator](https://github.com/MobilityData/gtfs-validator/blob/b597b7c2521bae18041e2f7c9879304c33c16aa5/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/MissingCalendarAndCalendarDateValidator.java#L42)

core/src/main/java/org/mobilitydata/gtfsvalidator/table/GtfsTableDescriptor.java
qcdyx: It is so much better! That was my initial thought, but failed to find the annotation...

core/src/main/java/org/mobilitydata/gtfsvalidator/table/AnyTableLoader.java
```diff
@@ -32,7 +32,6 @@ public final class AnyTableLoader {
   private static final FluentLogger logger = FluentLogger.forEnclosingClass();
   private static final List<Class<? extends FileValidator>> singleFileValidatorsWithParsingErrors =
       new ArrayList<>();
-
```
davidgamez, line 35: [picky]: I would revert this change as it's only a formatting difference.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/MissingFeedInfoValidator.java
```diff
@@ -0,0 +1,36 @@
...
+import org.mobilitydata.gtfsvalidator.annotation.GtfsValidator;
+import org.mobilitydata.gtfsvalidator.notice.MissingRecommendedFileNotice;
+import org.mobilitydata.gtfsvalidator.notice.MissingRequiredFileNotice;
+import org.mobilitydata.gtfsvalidator.notice.NoticeContainer;
+import org.mobilitydata.gtfsvalidator.table.GtfsFeedInfo;
+import org.mobilitydata.gtfsvalidator.table.GtfsFeedInfoTableContainer;
+import org.mobilitydata.gtfsvalidator.table.GtfsTranslationTableContainer;
+
+@GtfsValidator
+public class MissingFeedInfoValidator extends FileValidator {
```
davidgamez, line 13: Please add class comments.

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/MissingFeedInfoValidatorTest.java
```diff
@@ -0,0 +1,43 @@
...
+  @Test
+  public void missingFeedInfoWhenTranslationTableIsPresent() {
+    assertThat(
+            generateNotices(
+                GtfsFeedInfoTableContainer.forStatus(GtfsTableContainer.TableStatus.MISSING_FILE),
+                GtfsTranslationTableContainer.forStatus(
+                    GtfsTableContainer.TableStatus.PARSABLE_HEADERS_AND_ROWS)))
+        .contains(new MissingRequiredFileNotice(GtfsFeedInfo.FILENAME));
+  }
+}
```
jcpitre, line 43: You should consider also testing for false positive, i.e. Nothing missing should give no notice.

qcdyx: great idea!

PR #1798, opened 2024-08-12
fix: Display empty values for feed_start_date and feed_end_date if there's…

qcdyx:
**Summary:**
Closes #1790 
image

**Expected behavior:** 

Display empty values for feed_start_date and feed_end_date if there's no data for these two fields.

Please make sure these boxes are checked before submitting your pull request - thanks!

- [ ] Run the unit tests with `gradle test` to make sure you didn't break anything
- [ ] Add or update any needed [documentation](https://github.com/MobilityData/gtfs-validator/tree/master/docs) to the repo 
- [x] Format the title like "feat: [new feature short description]". Title must follow the Conventional Commit Specification(https://www.conventionalcommits.org/en/v1.0.0/).
- [x] Linked all relevant issues
- [x] Include screenshot(s) showing how this pull request works and fixes the issue(s)

PR #1837, opened 2024-09-10
feat: 1534 service window in summary report

### LINKED ISSUE #1534: feat: Include service window in Summary report
### Describe the problem

Currently, the summary section of the validation report shares the `feed_start_date` and `feed_end_date` when it's provided. However, many feeds don't include `feed_info.txt`, which makes it difficult to know the service window at a glance.

### Proposed solution

- Add full service date range to summary report. [Calculation here](https://github.com/MobilityData/gtfs-validator/blob/master/main/src/main/java/org/mobilitydata/gtfsvalidator/util/CalendarUtil.java)

- Pseudologic for the date range here: 

- When only `calendars.txt` is used: Loop through all services referenced with a trip_id in `trips.txt`. Loop through all those services with a trip_id in`calendars.txt`. check that there have an `start_date` and `end_date` - take the earliest start date and the latest end date
When only `calendar_dates.txt` is used: Loop through all services referenced with a trip_id in `trips.txt`. Loop through all those services in `calendar_dates.txt`. Take the earliest date and the latest date
When both are used: Loop through all services referenced with a trip_id in `trips.txt`. Calculate the earliest start date and the latest end date for each service with a trip_id using `calendars.txt` , and then check to make sure there isn't a service exception in `calendar_dates.txt` on either of those dates

### Design considerations

- The summary report is getting busy [(example here)](https://gtfs-validator-results.mobilitydata.org/50456ba1-92f2-45c5-a6f8-524087834eac/report.html). How do we make sure this info fits and is intuitive?

_No response_

### Additional context

- Include both feed date range and service range
- In the future, calculate based on majority of dates
- Look at other validators' approach to calendar dates, e.g Google Partner Dashboard
_No response_

qcdyx:
**Summary:**
Closes #1534 

**Expected behavior:** 
GTFS Validator report should show Service Date Range under Feed Info section.

1. When only calendar_dates.txt is used 
test dataset: 
[poa_gtfs.zip](https://github.com/user-attachments/files/16949788/poa_gtfs.zip)
) 
image

2.  When only calendar_dates.txt is used 
test dataset: 
[Google_transit.zip](https://github.com/user-attachments/files/16949823/Google_transit.zip)
image

3. When both calendars.txt and calendar_dates.txt are used 
test dataset: 
[Archive.zip](https://github.com/user-attachments/files/16949852/Archive.zip)
image

### DIFF
--- a/main/src/main/java/org/mobilitydata/gtfsvalidator/report/model/FeedMetadata.java
+++ b/main/src/main/java/org/mobilitydata/gtfsvalidator/report/model/FeedMetadata.java
@@ -349,15 +349,15 @@ public class FeedMetadata {
    * @param calendarTable the container for `calendars.txt` data
    * @param calendarDateTable the container for `calendar\_dates.txt` data
    */
-  private void loadServiceDateRange(
-      GtfsTableContainer<GtfsTrip> tripContainer,
-      GtfsTableContainer<GtfsCalendar> calendarTable,
-      GtfsTableContainer<GtfsCalendarDate> calendarDateTable) {
+  public void loadServiceDateRange(
+          GtfsTableContainer<GtfsTrip> tripContainer,
+          GtfsTableContainer<GtfsCalendar> calendarTable,
+          GtfsTableContainer<GtfsCalendarDate> calendarDateTable) {
     List<GtfsTrip> trips = tripContainer.getEntities();
 
     LocalDate earliestStartDate = null;
     LocalDate latestEndDate = null;
-    if (calendarDateTable.isMissingFile() && calendarTable.isParsedSuccessfully()) {
+    if ((calendarDateTable == null) && (calendarTable != null)) {
       // When only calendars.txt is used
       List<GtfsCalendar> calendars = calendarTable.getEntities();
       for (GtfsTrip trip : trips) {
@@ -378,7 +378,7 @@ public class FeedMetadata {
           }
         }
       }
-    } else if (calendarDateTable.isParsedSuccessfully() && calendarTable.isMissingFile()) {
+    } else if ((calendarDateTable != null) && (calendarTable == null)) {
       // When only calendar_dates.txt is used
       List<GtfsCalendarDate> calendarDates = calendarDateTable.getEntities();
       for (GtfsTrip trip : trips) {
@@ -397,40 +397,36 @@ public class FeedMetadata {
           }
         }
       }
-    } else if (calendarTable.isParsedSuccessfully() && calendarDateTable.isParsedSuccessfully()) {
+    } else if ((calendarTable != null) && (calendarDateTable != null)) {
       // When both calendars.txt and calendar_dates.txt are used
-      Preconditions.checkNotNull(calendarTable);
-      Preconditions.checkNotNull(calendarDateTable);
-
       Map<String, ServicePeriod> servicePeriods =
           CalendarUtil.buildServicePeriodMap(
               (GtfsCalendarTableContainer) calendarTable,
               (GtfsCalendarDateTableContainer) calendarDateTable);
-
+      List<LocalDate> removedDates = new ArrayList<>();
       for (GtfsTrip trip : trips) {
         String serviceId = trip.serviceId();
         ServicePeriod servicePeriod = servicePeriods.get(serviceId);
-        if (servicePeriod != null) {
-          LocalDate startDate = servicePeriod.getServiceStart();
-          LocalDate endDate = servicePeriod.getServiceEnd();
+        LocalDate startDate = servicePeriod.getServiceStart();
+        LocalDate endDate = servicePeriod.getServiceEnd();
 
-          if (startDate != null && endDate != null) {
-            if (earliestStartDate == null || startDate.isBefore(earliestStartDate)) {
-              earliestStartDate = startDate;
-            }
-            if (latestEndDate == null || endDate.isAfter(latestEndDate)) {
-              latestEndDate = endDate;
-            }
+        if (startDate != null && endDate != null) {
+          if (earliestStartDate == null || startDate.isBefore(earliestStartDate)) {
+            earliestStartDate = startDate;
           }
-          for (LocalDate date : servicePeriod.getRemovedDays()) {
-            System.out.println("Removed date: " + date);
-            if (date.isEqual(earliestStartDate)) {
-              earliestStartDate = date.plusDays(1);
-            }
-            if (date.isEqual(latestEndDate)) {
-              latestEndDate = date.minusDays(1);
-            }
+          if (latestEndDate == null || endDate.isAfter(latestEndDate)) {
+            latestEndDate = endDate;
           }
+        }
+        removedDates.addAll(servicePeriod.getRemovedDays());
+      }
+
+      for (LocalDate date : removedDates) {
+        if (date.isEqual(earliestStartDate)) {
+          earliestStartDate = date.plusDays(1);
+        }
+        if (date.isEqual(latestEndDate)) {
+          latestEndDate = date.minusDays(1);
         }
       }
     }
--- a/main/src/main/java/org/mobilitydata/gtfsvalidator/report/model/FeedMetadata.java
+++ b/main/src/main/java/org/mobilitydata/gtfsvalidator/report/model/FeedMetadata.java
@@ -383,19 +383,30 @@ public class FeedMetadata {
     feedInfo.put(FEED_INFO_PUBLISHER_URL, info == null ? "N/A" : info.feedPublisherUrl());
     feedInfo.put(FEED_INFO_FEED_CONTACT_EMAIL, info == null ? "N/A" : info.feedContactEmail());
     feedInfo.put(
-        FEED_INFO_FEED_LANGUAGE, info == null ? "N/A" : info.feedLang().getDisplayLanguage());
+            FEED_INFO_FEED_LANGUAGE, info == null ? "N/A" : info.feedLang().getDisplayLanguage());
     if (feedTable.hasColumn(GtfsFeedInfo.FEED_START_DATE_FIELD_NAME)) {
-      LocalDate localDate = info.feedStartDate().getLocalDate();
-      String displayDate =
-          localDate.equals(GtfsFeedInfo.DEFAULT_FEED_START_DATE) ? "N/A" : localDate.toString();
-      feedInfo.put(FEED_INFO_FEED_START_DATE, info == null ? "N/A" : displayDate);
+      if (info != null) {
+        LocalDate localDate = info.feedStartDate().getLocalDate();
+        feedInfo.put(FEED_INFO_FEED_START_DATE, checkLocalDate(localDate));
+      }
     }
     if (feedTable.hasColumn(GtfsFeedInfo.FEED_END_DATE_FIELD_NAME)) {
-      LocalDate localDate = info.feedEndDate().getLocalDate();
-      String displayDate =
-          localDate.equals(GtfsFeedInfo.DEFAULT_FEED_END_DATE) ? "N/A" : localDate.toString();
-      feedInfo.put(FEED_INFO_FEED_END_DATE, info == null ? "N/A" : displayDate);
+      if (info != null) {
+        LocalDate localDate = info.feedEndDate().getLocalDate();
+        feedInfo.put(FEED_INFO_FEED_END_DATE, checkLocalDate(localDate));
+      }
     }
+  }
+
+  private String checkLocalDate(LocalDate localDate) {
+    String displayDate;
+    if (localDate.toString().equals(LocalDate.EPOCH.toString())) {
+      displayDate = "N/A";
+    } else {
+      displayDate = localDate.toString();
+
+    }
+    return displayDate;
   }
 
   /**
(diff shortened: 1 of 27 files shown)


main/src/main/java/org/mobilitydata/gtfsvalidator/report/model/FeedMetadata.java
```diff
@@ -78,10 +82,18 @@ public static FeedMetadata from(GtfsFeedContainer feedContainer, ImmutableSet<St
 
     feedMetadata.setCounts(feedContainer);
 
-    if (feedContainer.getTableForFilename(GtfsFeedInfo.FILENAME).isPresent()) {
```
jcpitre, line 81: This requires that all 4 files be present, but calendar and calendar_date are conditionally required, meaning one of the two could be absent.

main/src/main/java/org/mobilitydata/gtfsvalidator/report/model/FeedMetadata.java
```diff
@@ -377,6 +393,116 @@ private void loadFeedInfo(GtfsTableContainer<GtfsFeedInfo> feedTable) {
...
+   *
+   * @param tripContainer the container for `trips.txt` data
+   * @param calendarTable the container for `calendars.txt` data
+   * @param calendarDateTable the container for `calendar\_dates.txt` data
+   */
+  public void loadServiceDateRange(
+      GtfsTableContainer<GtfsTrip> tripContainer,
+      GtfsTableContainer<GtfsCalendar> calendarTable,
+      GtfsTableContainer<GtfsCalendarDate> calendarDateTable) {
+    List<GtfsTrip> trips = tripContainer.getEntities();
```
jcpitre, line 410: Are you sure tripContainer cannot be null?

main/src/main/java/org/mobilitydata/gtfsvalidator/report/model/FeedMetadata.java
```diff
@@ -377,6 +393,116 @@ private void loadFeedInfo(GtfsTableContainer<GtfsFeedInfo> feedTable) {
...
+      // When only calendars.txt is used
+      List<GtfsCalendar> calendars = calendarTable.getEntities();
+      for (GtfsTrip trip : trips) {
+        String serviceId = trip.serviceId();
+        for (GtfsCalendar calendar : calendars) {
+          if (calendar.serviceId().equals(serviceId)) {
+            LocalDate startDate = calendar.startDate().getLocalDate();
+            LocalDate endDate = calendar.endDate().getLocalDate();
+
+            if (startDate != null || endDate != null) {
```
jcpitre, line 431: Can these be null or will they be some default (e.g. 1970-01-01) if the date is not in the file?

main/src/main/java/org/mobilitydata/gtfsvalidator/report/model/FeedMetadata.java
```diff
@@ -377,6 +393,116 @@ private void loadFeedInfo(GtfsTableContainer<GtfsFeedInfo> feedTable) {
...
+        }
+      }
+    } else if ((calendarDateTable != null) && (calendarTable == null)) {
+      // When only calendar_dates.txt is used
+      List<GtfsCalendarDate> calendarDates = calendarDateTable.getEntities();
+      for (GtfsTrip trip : trips) {
+        String serviceId = trip.serviceId();
+        for (GtfsCalendarDate calendarDate : calendarDates) {
+          if (calendarDate.serviceId().equals(serviceId)) {
+            LocalDate date = calendarDate.date().getLocalDate();
```
jcpitre, line 449: Maybe you should check if the exception_type is addition before using it in the calculation.

main/src/test/java/org/mobilitydata/gtfsvalidator/report/model/FeedMetadataTest.java
```diff
@@ -31,6 +34,14 @@ public class FeedMetadataTest {
   ValidatorLoader validatorLoader;
   File rootDir;
 
+  // Mock data for trips
+  NoticeContainer noticeContainer = new NoticeContainer();
+  ;
```
jcpitre, line 39: typo

main/src/main/java/org/mobilitydata/gtfsvalidator/report/model/FeedMetadata.java
qcdyx: you are tight, they can be unix time 1970-01-01, in that case, we should just display N/A

main/src/main/java/org/mobilitydata/gtfsvalidator/report/model/FeedMetadata.java
qcdyx: When only calendar_dates.txt is used, can we assume that the exception_type is always 1? @emmambd

main/src/main/java/org/mobilitydata/gtfsvalidator/report/model/FeedMetadata.java
jcpitre: Well, if trips.txt is absent it will give a error notice, but do we have a guarantee that the code here will not be reached in that case? If it is, we'll get an NPE which in general is not good.

qcdyx: Ok, it could be empty. I'll add the logic.

main/src/main/java/org/mobilitydata/gtfsvalidator/report/model/FeedMetadata.java
```diff
@@ -379,6 +400,114 @@ private void loadFeedInfo(GtfsTableContainer<GtfsFeedInfo> feedTable) {
...
+        if (date.isEqual(earliestStartDate)) {
+          earliestStartDate = date.plusDays(1);
+        }
+        if (date.isEqual(latestEndDate)) {
+          latestEndDate = date.minusDays(1);
+        }
+      }
+    }
+    DateTimeFormatter formatter = DateTimeFormatter.ofPattern("MMMM d, yyyy");
+    if ((earliestStartDate == null) && (latestEndDate == null)
```
jcpitre, line 495: Won't this generate an NPE if earliestStartDate is null and latestEndDate is not null?

main/src/main/java/org/mobilitydata/gtfsvalidator/report/model/FeedMetadata.java
```diff
@@ -379,6 +400,114 @@ private void loadFeedInfo(GtfsTableContainer<GtfsFeedInfo> feedTable) {
...
+    List<GtfsTrip> trips = tripContainer.getEntities();
+
+    LocalDate earliestStartDate = null;
+    LocalDate latestEndDate = null;
+    if ((calendarDateTable == null) && (calendarTable != null)) {
+      // When only calendars.txt is used
+      List<GtfsCalendar> calendars = calendarTable.getEntities();
+      for (GtfsTrip trip : trips) {
+        String serviceId = trip.serviceId();
+        for (GtfsCalendar calendar : calendars) {
```
jcpitre, line 426: Can't this loop be replaced by something like this: 
GtfsCalendar calendar = ((GtfsCalendarTableContainer) calendarTable).byServiceId(serviceId);

main/src/main/java/org/mobilitydata/gtfsvalidator/report/model/FeedMetadata.java
```diff
@@ -379,6 +400,114 @@ private void loadFeedInfo(GtfsTableContainer<GtfsFeedInfo> feedTable) {
...
+            }
+          }
+        }
+      }
+    } else if ((calendarDateTable != null) && (calendarTable == null)) {
+      // When only calendar_dates.txt is used
+      List<GtfsCalendarDate> calendarDates = calendarDateTable.getEntities();
+      for (GtfsTrip trip : trips) {
+        String serviceId = trip.serviceId();
+        for (GtfsCalendarDate calendarDate : calendarDates) {
```
jcpitre, line 447: Can't this loop be replaced by something like this:
GtfsCalendarDate calendarDate = ((GtfsCalendarDateTableContainer) calendarDateTable).byServiceId(serviceId);

main/src/main/java/org/mobilitydata/gtfsvalidator/report/model/FeedMetadata.java
```diff
@@ -358,7 +375,11 @@ private void loadAgencyData(GtfsTableContainer<GtfsAgency> agencyTable) {
     }
   }
 
-  private void loadFeedInfo(GtfsTableContainer<GtfsFeedInfo> feedTable) {
+  private void loadFeedInfo(
+      GtfsTableContainer<GtfsFeedInfo> feedTable,
+      GtfsTableContainer<GtfsTrip> tripContainer,
```
jcpitre, line 380: With the change done earlier tripContainer, calendarTable and calendarDateTable are not used any more within this method, so they should be removed.

qcdyx: removed!

main/src/main/java/org/mobilitydata/gtfsvalidator/report/model/FeedMetadata.java
qcdyx: Yes, it could, but I prefer to keep the loop as is for the moment. We can create new issue that optimize the code using Optional handling.

main/src/main/java/org/mobilitydata/gtfsvalidator/report/model/FeedMetadata.java
qcdyx: Same here.

emmambd:
Hey @qcdyx - I'm able to do the spot checks now to see if we need service date ranges in cases where feed_info start date and end date apply! 

However, I'm still seeing the bug that was fixed in https://github.com/MobilityData/gtfs-validator/pull/1798. Example report here: 

[report.json](https://github.com/user-attachments/files/17104266/report.json)

Since this was merged to master, why would this be happening?

emmambd (changes requested):
Great work @qcdyx! After doing some spot checks, it makes sense for us to include both the feed start and end date values and the service date range values. 

Minor changes: 

• Change "Service Date Range" to "Service Dates"
• Add tooltip to Service Date Range on the HTML report: 

"The range of service dates covered by the feed, based on trips with an associated service_id in calendar.txt and/or calendar_dates.txt."

emmambd:
@qcdyx Sorry Jingsi, one more minor requirements change and then you can merge 🙈 In thinking about how we'll eventually add majority service window to this report, let's change "service dates" to "service window" so we'll have consistency between these concepts. 

Once changed, all good and ready to merge :)

davidgamez:
@qcdyx  The acceptance tests are reporting a few(20) feeds sources `corrupted`. This means the report was not generated for some reason, and the comparison script cannot compare the previous/latest reports. One of the possible reasons is that the validator throws a critical exception while processing the feed. I tried the `ca-ontario-guelph-transit-gtfs-1300` locally with this branch and indeed raised a NullPointerException in [line](https://github.com/MobilityData/gtfs-validator/blob/9d12dc9f3414c5373e585e164282c4084c91b81b/main/src/main/java/org/mobilitydata/gtfsvalidator/report/model/FeedMetadata.java#L481). In this case the service Id was not found in the `servicePeriods` Map. We can target this specific NullPointerException. 
In addition, I'm suggesting to wrap with a try/catch the full `"discovery"` of the service date between line [429](https://github.com/MobilityData/gtfs-validator/blob/9d12dc9f3414c5373e585e164282c4084c91b81b/main/src/main/java/org/mobilitydata/gtfsvalidator/report/model/FeedMetadata.java#L429) to [507](https://github.com/MobilityData/gtfs-validator/blob/9d12dc9f3414c5373e585e164282c4084c91b81b/main/src/main/java/org/mobilitydata/gtfsvalidator/report/model/FeedMetadata.java#L507), this will make sure that any exception raised due to a null object is properly caught and the report is generated. I am open to suggestions on how to make sure that the report is generated. 

After the changes, let's examine the corrupted files in the acceptance test and ensure that the validator generates the report.

davidgamez:
This looks good for the scope of the issue. I did create a follow-up issue #1848 to address the pending corrupted feeds.

davidgamez (approved):
LGTM

cka-y:
Was the service window added to the json report? It's part of the tasking, but couldn't find the information from running the jar from this branch

qcdyx:
Correct me if I'm wrong but I remembered that we talked about splitting the ticket and focusing on just the service date range on UI when we did the estimate. I recommend merging this PR now, and I’ll create a separate one to add it to the JSON report coz I need to merge master changes into my branch and it takes some time to go thru all the checks. @emmambd @davidgamez

emmambd:
I'm fine with splitting, but the next priority issue will be adding it to the JSON report in that case, since we can't make any changes to the Mobility Database we want without the JSON report functionality. cc @davidgamez for final call

qcdyx:
@emmambd could you approve my PR since merging is blocked due to requested changes.

emmambd:
Did you and @davidgamez discuss? Just waiting for his thoughts on splitting or not before I approve it.

davidgamez:
> Did you and @davidgamez discuss? Just waiting for his thoughts on splitting or not before I approve it.

Yes the JSON changes can be part of a different issue.

emmambd (approved):

PR #1857, opened 2024-09-30
chore: add memory usage records to the JSON report

davidgamez:
**Summary:**

**Expected behavior:** 

### DIFF
--- a/main/src/main/java/org/mobilitydata/gtfsvalidator/runner/ValidationRunner.java
+++ b/main/src/main/java/org/mobilitydata/gtfsvalidator/runner/ValidationRunner.java
@@ -36,6 +36,8 @@ import org.mobilitydata.gtfsvalidator.input.GtfsInput;
 import org.mobilitydata.gtfsvalidator.notice.IOError;
 import org.mobilitydata.gtfsvalidator.notice.NoticeContainer;
 import org.mobilitydata.gtfsvalidator.notice.URISyntaxError;
+import org.mobilitydata.gtfsvalidator.performance.MemoryMonitor;
+import org.mobilitydata.gtfsvalidator.performance.MemoryUsageRegister;
 import org.mobilitydata.gtfsvalidator.report.HtmlReportGenerator;
 import org.mobilitydata.gtfsvalidator.report.JsonReport;
 import org.mobilitydata.gtfsvalidator.report.JsonReportGenerator;
@@ -71,7 +73,11 @@ public class ValidationRunner {
     this.versionResolver = versionResolver;
   }
 
+  @MemoryMonitor
   public Status run(ValidationRunnerConfig config) {
+    MemoryUsageRegister.getInstance().clearRegistry();
+    var memoryBefore =
+        MemoryUsageRegister.getInstance().getMemoryUsageSnapshot("ValidationRunner.run", null);
     VersionInfo versionInfo =
         versionResolver.getVersionInfoWithTimeout(
             Duration.ofSeconds(5), config.skipValidatorUpdate());
@@ -134,7 +140,13 @@ public class ValidationRunner {
     }
     FeedMetadata feedMetadata = FeedMetadata.from(feedContainer, gtfsInput.getFilenames());
     closeGtfsInput(gtfsInput, noticeContainer);
+
+    //    Performance metrics
     feedMetadata.validationTimeSeconds = (System.nanoTime() - startNanos) / 1e9;
+    var after =
+        MemoryUsageRegister.getInstance()
+            .getMemoryUsageSnapshot("ValidationRunner.run", memoryBefore);
+    MemoryUsageRegister.getInstance().registerMemoryUsage(after);
 
     // Output
     exportReport(feedMetadata, noticeContainer, config, versionInfo);
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -18,6 +18,7 @@ plugins {
     id 'java'
     id 'maven-publish'
     id 'signing'
+    id 'io.freefair.aspectj.post-compile-weaving' version '6.4.1' // Add AspectJ plugin
 }
 
 //publishing {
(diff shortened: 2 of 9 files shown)


jcpitre:
Meta comment here: We should consider having a separate report for information related to running the validator, like this and timings. 
While they are related to the _size_ of the datasets, they are not related their _semantics_ or _syntax_, which ultimately is what interests the user. Plus in the long run they will lose their meaning. e.g. ok, it took 10 minutes to run a year ago, but what about today, after a year worth of changes in the code. Same with memory consumption.

docs/ACCEPTANCE_TESTS.md
```diff
@@ -104,6 +104,28 @@ We follow this process:
 
 <img src="/docs/Acceptance-test-process.jpg" width="750">
 
+## Performance metrics within the acceptance reports
+
+There are two man metrics added to the acceptance report comment at the PR level, _Validation Time_ and _Memory Consumption_.
```
jcpitre, line 109:
```suggestion
There are two main metrics added to the acceptance report comment at the PR level, _Validation Time_ and _Memory Consumption_.
```

main/src/main/java/org/mobilitydata/gtfsvalidator/runner/ValidationRunner.java
```diff
@@ -71,7 +73,11 @@ public ValidationRunner(VersionResolver versionResolver) {
     this.versionResolver = versionResolver;
   }
 
+  @MemoryMonitor
   public Status run(ValidationRunnerConfig config) {
+    MemoryUsageRegister.getInstance().clearRegistry();
+    var memoryBefore =
```
jcpitre, line 79: Why do we need to do that here? I thought that because of @MemoryMonitor this would be done automatically in MemoryMonitorAspect.monitorMemoryUsage

davidgamez: This is the case of concurrent threads. If we use the annotation, the AspectJ is called multiple times because multiple threads are created inside the method.

output-comparator/src/main/java/org/mobilitydata/gtfsvalidator/outputcomparator/io/DatasetMemoryUsage.java
```diff
@@ -0,0 +1,72 @@
...
+
+  public DatasetMemoryUsage(
+      String datasetId,
+      List<MemoryUsage> referenceMemoryUsage,
+      List<MemoryUsage> latestMemoryUsage) {
+    this.datasetId = datasetId;
+    this.referenceMemoryUsage = referenceMemoryUsage;
+    this.latestMemoryUsage = latestMemoryUsage;
+    if (referenceMemoryUsage != null) {
+      this.referenceUsedMemoryByKey =
```
jcpitre, line 30: What happens if the list contains the key twice? Can it happen?

jcpitre (approved):

output-comparator/src/main/java/org/mobilitydata/gtfsvalidator/outputcomparator/io/DatasetMemoryUsage.java
davidgamez: Yes, it can happen. When the annotation is used, it's less likely to happen as the key is the complete method signature, including the class name and path. For human-created keys, it's possible. I don't know if adding an exception or interrupting the process is a good idea, as extracting the metrics is a non-essential operation. Logging a warning can be buried within the GitHub action. Do you have any suggestions? _Other option is forcing uniqueness before adding the key to the list, but I landed in the same place, should we interrupt the validation process?_

main/src/main/java/org/mobilitydata/gtfsvalidator/runner/ValidationRunner.java
davidgamez: Added a comment within the file and a section in the acceptance tests documentation.

davidgamez:
> Meta comment here: We should consider having a separate report for information related to running the validator, like this and timings. While they are related to the _size_ of the datasets, they are not related their _semantics_ or _syntax_, which ultimately is what interests the user. Plus in the long run they will lose their meaning. e.g. ok, it took 10 minutes to run a year ago, but what about today, after a year worth of changes in the code. Same with memory consumption.

This is interesting, it might take more work to pass the information across the pipeline when analyzing the execution of the validation process, but it might be cleaner for a "regular" consumer. As an alternative we can create another section(field) in the JSON file that is not the metadata. We keep pouring info into the feed metadata as we go. Let's have more thoughts on it. cc: @emmambd

output-comparator/src/main/java/org/mobilitydata/gtfsvalidator/outputcomparator/io/DatasetMemoryUsage.java
jcpitre: I think it can happen with the comparator when reading a json report file? Please confirm. 
We expect the file to have been written properly by the program, but it's true that a user could modify this file, or there could be a bug when the file is written by the validator. 
I think that just ignoring the second element with the same key with a log could be the way to go. It's a corner case IMO so just a message in the logs should be OK. To me it's more important the the possible exception does not stop the whole process.
I think to do that you'll have to use an explicit loop instead of streams, right?

davidgamez: Yes, this is when the comparator reads the JSON report. Theoretically, the list in the report can have duplicates, so duplicates are still properly defined in a JSON report. In practice, we are currently adding only three keys. I'll add a log. For clarification, there is no exception thrown if duplicates are found in the report, we are just ignoring/overriding the values.

emmambd:
@davidgamez @jcpitre re: your comments, I like how this currently works which is `memoryUsageRecord` as a separate section of the JSON report. I agree if we want to extend these kinds of performance checks, moving them to a separate report might make sense (for example, when we provide more detailed and specific insights into parsing failures). However, I think this is sufficient for now.

output-comparator/src/main/java/org/mobilitydata/gtfsvalidator/outputcomparator/io/DatasetMemoryUsage.java
jcpitre: A bit late, but I was revisiting this and with the way toUnmodifiableMap was used before (toUnmodifiableMap(MemoryUsage::getKey, MemoryUsage::usedMemory)) there was a risk of a IllegalStateException according to [this](https://docs.oracle.com/javase/10/docs/api/java/util/stream/Collectors.html#toUnmodifiableMap(java.util.function.Function,java.util.function.Function))
The 3 parameters function that you used in the end does not throw that exception, so all is well.

PR #1887, opened 2024-10-11
feat: 1883 flex modify stop without stop time to exclude stops associated with a used location group

### LINKED ISSUE #1883: Flex: modify stop_without_stop_time to exclude stops associated with a used location_group
### Describe the problem

Based on the acceptance test changes from https://github.com/MobilityData/gtfs-validator/pull/1834#issuecomment-2346566136, there are changes in fixed stop demand responsive services where a stop_id will not be referenced explicitly in stop_times.txt and therefore would generate the `stop_without_stop_time` warning, even though the stop_id is technically used. 

### Proposed solution

Create an exception for this rule: 

If there's a location_group_id referenced in stop_times.txt, go through the original location_group file and check all the stops associated with that id. If a stop exists in the location_group that is being used in stop_times, don't trigger the warning. 

_No response_

### Additional context

_No response_

qcdyx:
**Summary:**

Closes #1883 

**Expected behavior:** 
exclude stop_without_stop_time notice if stop_id is part of location_group_stops.txt referenced in stop_times.txt.

Before:
image
After:
image


emmambd:
@qcdyx QA is as expected!

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/LocationHasStopTimesValidator.java
```diff
@@ -48,19 +46,41 @@ public class LocationHasStopTimesValidator extends FileValidator {
...
+    Set<String> stopIdsInStopTimesandLocationGroupStops = new HashSet<>();
+    Set<String> locationGroupIdsInStopTimes = new HashSet<>();
+
+    for (GtfsStopTime stopTime : stopTimeTable.getEntities()) {
+      stopIdsInStopTimesandLocationGroupStops.add(stopTime.stopId());
+      locationGroupIdsInStopTimes.add(stopTime.locationGroupId());
+    }
+
+    for (String locationGroupId : locationGroupIdsInStopTimes) {
+      for (GtfsLocationGroupStops locationGroupStop : locationGroupStopTable.getEntities()) {
```
jcpitre, line 72: I think you can use `locationGroupStopTable.byLocationGroupId(locationGroupId)` instead of this second loop.
Something like:
```
      var byLocationGroupId = locationGroupStopTable.byLocationGroupId(locationGroupId);
      for (var locationGroupStop: byLocationGroupId) {
        stopIdsInStopTimesandLocationGroupStops.add(locationGroupStop.stopId());
      }
```
This would insure that each GtfsLocationGroupStops entity is visited only once.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/LocationHasStopTimesValidator.java
```diff
@@ -48,19 +46,41 @@ public class LocationHasStopTimesValidator extends FileValidator {
...
   }
 
   @Override
   public void validate(NoticeContainer noticeContainer) {
+    Set<String> stopIdsInStopTimesandLocationGroupStops = new HashSet<>();
+    Set<String> locationGroupIdsInStopTimes = new HashSet<>();
+
+    for (GtfsStopTime stopTime : stopTimeTable.getEntities()) {
+      stopIdsInStopTimesandLocationGroupStops.add(stopTime.stopId());
+      locationGroupIdsInStopTimes.add(stopTime.locationGroupId());
```
davidgamez, line 68: We should not add null/empty IDs to the sets.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/LocationHasStopTimesValidator.java
qcdyx: More concise and likely to be faster, great find!

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/LocationHasStopTimesValidator.java
qcdyx: added null checks

tzujenchanmbd:
lgtm!

jcpitre (approved):

PR #1892, opened 2024-10-16
feat: 1840 invalid characters

### LINKED ISSUE #1840: Validator Accepts Replacement Character in `stop_name` Field
### Describe the problem

Mobility Data Validator currently accepts the Unicode replacement character (`�`) in the `stop_name` field. This character typically indicates a decoding error and should not be present in valid data. While I've specifically noticed the acceptance of this replacement character, I am unsure if other similar problematic characters are also being accepted. Ideally, the validator should flag or reject `stop_name` fields containing the replacement character and potentially other similar invalid characters, as was done by the older GTFS validator, FeedValidator - https://github.com/google/transitfeed/tree/master, which raised a Unicode error (E33) for invalid values in the stop_name field. Implementing similar checks in future releases would help prevent data issues caused by improperly decoded or malformed input. Could this behavior be reviewed and corrected in future releases please? Thank you for your consideration and for your continued work on this project! 

replacement_char = "\uFFFD"
if replacement_char in stop_name:
	trigger_error("invalid_character_in_stop_name", stop_name)

### Sample GTFS datasets

[auckland_gtfs.zip](https://github.com/user-attachments/files/16991535/auckland_gtfs.zip)

### Severity

_No response_

### Additional context

_No response_

qcdyx:
**Summary:**

Closes #1840 

**Expected behavior:** 
image


core/src/main/java/org/mobilitydata/gtfsvalidator/parsing/RowParser.java
```diff
@@ -137,6 +121,15 @@ public String asString(int columnIndex, GtfsColumnDescriptor columnDescriptor) {
               fileName, getRowNumber(), columnDescriptor.columnName()));
     }
     if (s != null) {
+      // Validate if the string contains invalid characters
+
+      if (containsInvalidCharacters(s)) {
+        noticeContainer.addValidationNotice(
+            new InvalidCharactersNotice(
+                fileName, getRowNumber(), columnDescriptor.columnName(), s));
+        return null;
```
davidgamez, line 130: Let's remove the return statement and continue the validation below. This will allow offending feeds to pop up subsequence notices if they exist.

emmambd: Yes please! Ideally there are no dropped notices because of this new rule.

qcdyx: it's removed.

tzujenchanmbd:
Some examples of correct name in the screenshots:
- Rotterdam, Selma Lagerlöfweg -> Rotterdam, Selma Lagerl�fweg
- Rotterdam, Port Saïdstraat -> Rotterdam, Port Sa�dstraat
- Estación Washington -> Estaci��n Washington

Problem example on maps: https://maps.app.goo.gl/YqnS2Gj9goeWN8GT9

So it seems the issue usually happen on accented characters like "ó", "ö", "ï" in Western European languages.

Perhaps dev team can help confirm, but I guess this is probably because of encoding and decoding mismatch during the data production process. For example, if the text was originally saved using a specific encoding (e.g. ISO-8859-1 or Windows-1252, "legacy" encoding covering characters commonly used in Western European languages, such as accented characters (é, ñ, ü) and special symbols), but is then read using a different encoding (e.g. UTF-8). In this case characters outside the ASCII range (like accented characters) will probably not decode correctly, leading to errors like the replacement character (�).

davidgamez:
> Some examples of correct name in the screenshots:
> 
> * Rotterdam, Selma Lagerlöfweg -> Rotterdam, Selma Lagerl�fweg
> * Rotterdam, Port Saïdstraat -> Rotterdam, Port Sa�dstraat
> * Estación Washington -> Estaci��n Washington
> 
> Problem example on maps: https://maps.app.goo.gl/YqnS2Gj9goeWN8GT9
> 
> So it seems the issue usually happen on accented characters like "ó", "ö", "ï" in Western European languages.
> 
> Perhaps dev team can help confirm, but I guess this is probably because of encoding and decoding mismatch during the data production process. For example, if the text was originally saved using a specific encoding (e.g. ISO-8859-1 or Windows-1252, "legacy" encoding covering characters commonly used in Western European languages, such as accented characters (é, ñ, ü) and special symbols), but is then read using a different encoding (e.g. UTF-8). In this case characters outside the ASCII range (like accented characters) will probably not decode correctly, leading to errors like the replacement character (�).

We assumed that the feeds are in UTF-8, replacement characters and other variations might be due to the fact that is not in proper UTF-8. The _legacy_ Google validator replaced the non-UTF-8 compatible characters with the replacement character. Maybe they have this implemented somewhere in their data pipeline to guarantee that the text is properly rendered in the UI, even with some characters "replaced", [_legacy validator code_](https://github.com/google/transitfeed/blob/104b5a5b339c62a94c1579d7209a41c7c0833e35/transitfeed/loader.py#L259)

emmambd:
Revisions after discussion with @tzujenchanmbd: 

@davidgamez @qcdyx Is it feasible for us to parse the non-UTF-8 characters too? Ideally we could show them to the user in the notice table description, highlighted in bold so they know which characters are causing the problem. 

- Notice name: Invalid character (not plural, to match the style of our other notices) 
- Notice description: Description: This field contains invalid characters, marked in bold. Text must be encoded in [UTF-8](https://en.wikipedia.org/wiki/UTF-8) in order to be valid. When reading text, use the same encoding that was used to save.

davidgamez:
> Revisions after discussion with @tzujenchanmbd:
> 
> @davidgamez @qcdyx Is it feasible for us to parse the non-UTF-8 characters too? Ideally we could show them to the user in the notice table description, highlighted in bold so they know which characters are causing the problem.
> 
> * Notice name: Invalid character (not plural, to match the style of our other notices)
> * Notice description: Description: This field contains invalid characters, marked in bold. Text must be encoded in [UTF-8](https://en.wikipedia.org/wiki/UTF-8) in order to be valid. When reading text, use the same encoding that was used to save.
> 
> I also want to note it looks like feeds with this error will have unparseable files that will mean notices are dropped, from the acceptance tests above.

I suggest creating a different notice for non-UTF-8 text. There are two distinct situations: the first is the presence of a replacement character that is a valid UTF-8 character, and the second is the presence of an invalid UTF-8 character. I suspect that if we have a replacement character, it is due to a tool that transformed the feed and potentially replaced the invalid UTF-8 characters(or any other encoding) to UTF-8 or a different target encoding(violating the spec in this case).

davidgamez:
Regarding the dropped notices, they are expected because of the severity of the notice(Error).

emmambd:
@davidgamez Makes sense. How's this for a revised notice description then, so it's more suggestive and less prescriptive that the issue is non-UTF-8 encoding: 

Notice name: Invalid character (not plural, to match the style of our other notices)
Notice description: Description: This field contains invalid characters, such as the replacement character ("�"). Check that text was properly encoded in UTF-8 as required by GTFS.

davidgamez:
> @davidgamez Makes sense. How's this for a revised notice description then, so it's more suggestive and less prescriptive that the issue is non-UTF-8 encoding:
> 
> Notice name: Invalid character (not plural, to match the style of our other notices) Notice description: Description: This field contains invalid characters, such as the replacement character ("�"). Check that text was properly encoded in UTF-8 as required by GTFS.

The notice name and description make sense to me.

qcdyx:
@davidgamez @emmambd image

emmambd:
LGTM!

davidgamez (approved):
LGTM

PR #1904, opened 2024-10-22
ci: aling matrix dimensions to max concurrent jobs

davidgamez:
**Summary:**

Currently, the number of concurrent jobs is limited by the type of account the organization is holding. As a Team account, we have 60 maximum number of jobs that can concurrently be executed at a time. Having 256 rows in the acceptance tests has two main issues. The first only 60 can be executed concurrently so having more than 60 will make the agents wait and don't add any value as a concurrent job. Secondly, the billable time is computed per second, but the minimum billable time is one minute; all the jobs completed before one minute will be charged one minute, `loosing` few seconds per small job.

**Expected behavior:** 

Acceptance tests passes and the resources are more efficient used.

### DIFF
--- a/scripts/mobility-database-harvester/harvest_latest_versions.py
+++ b/scripts/mobility-database-harvester/harvest_latest_versions.py
@@ -42,7 +42,8 @@ URL_PREFIX = "https://storage.googleapis.com/storage/v1/b/mdb-latest/o/"
 URL_SUFFIX = ".zip?alt=media"
 
 # Github constants
-MAX_JOB_NUMBER = 256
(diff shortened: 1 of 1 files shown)


davidgamez:
The acceptance test was executed two times, and without many stats, it was completed faster than usual.  The time was reduced to ~30 minutes from ~35-45 minutes. Billable time looks down by almost a half.  _This is not a statistical conclusion; we will monitor the performance of the acceptance tests in subsequent PRs._

jcpitre (approved):

Issue #1932, opened 2024-12-06
missing_required_field additions where network.network_id exists


mobilitydataio:
# `Commit Nº 1`
## AUTHOR × DATE
* **Tzu-Jen Chan** ( _tzujenchanmbd_ )
* **December 5, 2024**   @   12:55 PM ETC (05:55 PM UTC)
## MESSAGE
```
[GTFS-Fares v2] Add fare_leg_join_rules.txt (#439)
* Add Effective Fare Leg to term definition
* Add fare_leg_join_rules.txt
* Modify fare_leg_rules for effective fare leg
* Add from/to_stop_id
* Update file desc for from/to_stop_id
* Editorial updates
```
## FILES CHANGED
> ### https://github.com/google/transit/commit/7be58b00bce9955551668d9aa1bf5c8c93e8839a

```
https://github.com/google/transit/commit/7be58b00bce9955551668d9aa1bf5c8c93e8839a
```
---

### Describe the new validation rules

<table>
  <tr>
   <td><strong>Related spec language</strong>
   </td>
   <td><strong>Rationale for rule</strong>
   </td>
   <td><strong>Notice name to update or add</strong>
   </td>
   <td><strong>Pseudologic</strong>
   </td>
   <td><strong>Notice table columns</strong>
   </td>
   <td><strong>Severity</strong>
   </td>
   <td><strong>Notice description</strong>
   </td>
  </tr>
  <tr>
   <td>networks.network_id, fare_leg_join_rules.from_network_id, fare_leg_join_rules.to_network_id
   </td>
   <td>from_network_id and to_network_id are required
   </td>
   <td><a href="https://gtfs-validator.mobilitydata.org/rules.html#missing_required_field-rule">missing_required_field</a>
   </td>
   <td>If the checks for missing_required_column above are complete and do not generate a notice, then:
<p>
If from_network_id is empty and to_network_id is empty:
<p>
<strong>Trigger Notice</strong> and do not check for foreign_key_violation
   </td>
   <td><code>csvRowNumber, </code>
<p>
<code>FieldName, </code>
<p>
<code>Filename</code>
   </td>
   <td>ERROR
   </td>
   <td>A required field is missing.
<p style="text-align: center">
The given field has no value in some input row, even though values are required.
   </td>
  </tr>

  <tr>
   <td>stops.stop_id, fare_leg_join_rules.from_stop_id, fare_leg_join_rules.to_stop_id
   </td>
[...]

PR #1956, opened 2025-01-30
feat: added `fare_leg_join_rules.txt` schema and validation

cka-y:
**Summary:**
- **New Schema:** Introduces the `fare_leg_join_rules.txt` schema.  
- **Foreign Key Validation:** Adds a new notice to verify that every `network_id` found in `fare_leg_join_rules.txt` matches an entry in `networks.txt` or `routes.txt`.  
- **Required Fields:** Adds a notice for missing required fields when `from_stop_id` or `to_stop_id` is present but the other is omitted, ensuring both are supplied consistently.

**Expected behavior:** 
image

### DIFF
--- a/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/FareLegJoinRuleValidator.java
+++ b/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/FareLegJoinRuleValidator.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020 Google LLC, MobilityData IO
+ * Copyright 2025 Google LLC, MobilityData IO
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/FareLegJoinRuleValidator.java
+++ b/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/FareLegJoinRuleValidator.java
@@ -1,0 +1,100 @@
+/*
+ * Copyright 2020 Google LLC, MobilityData IO
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.mobilitydata.gtfsvalidator.validator;
+
+import javax.inject.Inject;
+import org.mobilitydata.gtfsvalidator.annotation.GtfsValidator;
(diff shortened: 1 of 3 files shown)


cka-y:
@emmambd 
The checks that are currently implemented for `to_network_id` and `from_network_id` are 
- both are required
- both represent a FK either from `routes.txt` or `networks.txt`

is there a check that needs to be modified/added to cover #1932?
I followed the descriptions here for the rules: https://gtfs.org/documentation/schedule/reference/#fare_leg_join_rulestxt

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/FareLegJoinRuleValidator.java
```diff
@@ -0,0 +1,100 @@
...
+ *
+ * <p>Generated notices:
+ *
+ * <ul>
+ *   <li>{@link ForeignKeyViolationNotice}
+ *   <li>{@link MissingRequiredFieldNotice}
+ * </ul>
+ */
+@GtfsValidator
+public class FareLegJoinRuleValidator extends FileValidator {
```
davidgamez, line 36: Let's include `shouldCallValidate`(logic based on the missing file method). This does not improve performance, but it's good practice. We can get a log on why some validators don't run without digging too deeply into the code.

skalexch:
@cka-y everything looks good, just one detail I'd like to double check with you and @emmambd. For `from_network_id` and `to_network_id`, we do check for `foreign_key_violation` but we do not check for `missing_required_field`.
There is a code that specifies if the fields are present but that seems to only be one of conditions that check for `foreign_key_violation`.

We might want to show `missing_required_field` before `foreign_key_violation` even when an empty field is probably also a `foreign_key_violation`

cka-y:
@skalexch if i'm understanding properly, you're suggesting to change the display order of the notices so that `missing_required_field` is before `foreign_key_violation`?

skalexch:
@cka-y I'm suggesting the addition of another `missing_required_field` notice that checks that _from_network_id_ and _to_network_id_ are not empty. Then, if they are not empty, we cannot proceed to the `foreign_key_violation` check.

cka-y:
Currently the logic is:
- if `from_network_id` is present -> validate the FK of the id in `from_network_id`
- if `to_network_id` is present -> validate the FK of the id in `to_network_id`
- validate the simultaneous presence of `from_network_id` and `to_network_id` [Here is where the `missing_required_field` notice is generated if one is present and the other is omitted]

I think the current logic does cover your suggestion but I could be mistaking.

skalexch:
@cka-y the `missing_required_field` in this PR validates the simultaneous presence of _from_stop_id_ and _to_stop_id_. The stop ids are conditionally required where if one exists the other has to exist too (which is taken care of in this PR). 

However, _from_network_id_ and _to_network_id_ are required regardless of each other. So even before proceeding to validating the FK, _from_network_id_ and _to_network_id_ need to exist.

Basically, this should trigger a `missing_required_field` notice:

|from_network_id|to_network_id|
|||

cka-y:
@skalexch Yes sorry about that this is actually covered as both are annotated with `@Required` in the schema [here](https://github.com/MobilityData/gtfs-validator/pull/1956/files#diff-49867f2e0f8ba59f18d4ff6e8931b70c00909a26857de10e4025264d298336c4R24-R29). I updated the logic as the check was for `to_stop_id` and `from_stop_id`.

davidgamez (approved):
LGTM

PR #1958, opened 2025-02-03
feat: implemented transfer distance validator

cka-y:
### **Summary**  
This PR introduces an info notice when the transfer distance exceeds **2 km** and a warning when it exceeds **10 km**.

### **Expected Behavior**  
Using [mdb-784](https://mobilitydatabase.org/feeds/mdb-784), we observe the following results:  
image
image

### **Distribution Analysis**  
The graph below shows the distribution of the **max transfer distance** across feeds in the Mobility Database.  
Most transfer distances fall **below the 2 km info threshold**, with only a few exceeding the error threshold of **10 km**.  

transfer_distance_distribution

#### **Top 3 Largest Distances**
1. **[mdb-784](https://mobilitydatabase.org/feeds/mdb-784)** – **5,463 km**  
2. **[mdb-927](https://mobilitydatabase.org/feeds/mdb-927)** – **24 km**  
3. **[mdb-1172](https://mobilitydatabase.org/feeds/mdb-1172)** – **18 km**  

While **mdb-784** contains an extreme outlier, most feeds remain within the defined thresholds (2 km for warnings, 10 km for errors).


emmambd:
@cka-y I know this isn't code review ready yet but I got excited...

1) transfer_distance_too_large should be a warning and transfer_distance_above_threshold should be INFO 
2) Let's change the name of transfer_distance_above_threshold to something easier to understand: transfer_distance_above_2km?

skalexch (commented):
I checked the validator table and the test code. The output looks correct.

emmambd:
@cka-y - Looks great. 

1) Curious why the list of feeds in the acceptance tests for `transfer_distance_too_large` looks smaller than it did in the first tests
2) Can we make an issue to add INFO notices to the acceptance tests in the future? As we add more of the non-spec, threshold-based rules to the validator, being able to view the INFO severity will be important.

cka-y:
@emmambd 
1. In the acceptance tests, I had +17 errors and +4 warnings. The 2KM threshold rule was actually set at the error level, while the 10KM threshold was at the warning level (the PR wasn’t _quite_ ready for review yet which is why i deleted the acceptance test report! 😅). The results remain the same as we lack the info level severities in the acceptance tests.
2. #1962

emmambd:
1) I think I'm still confused why there were 17 `transit_distance_too_large` notices before but now there's only 4.

skalexch (commented):
@cka-y just one question before approving, what happened to mdb-784? It doesn't exist in the +4 new warnings despite the transfer ditsance being 5463 km

cka-y:
@skalexch `mdb-784` (equivalent to `de-unknown-rursee-schifffahrt-kg-gtfs-784`) is actually omitted from the acceptance tests because of its size. [Ref](https://github.com/MobilityData/gtfs-validator/blob/master/scripts/mobility-database-harvester/harvest_latest_versions.py#L34-L38)

jcpitre:
> @cka-y I know this isn't code review ready yet but I got excited...
> 
> 1. transfer_distance_too_large should be a warning and transfer_distance_above_threshold should be INFO
> 2. Let's change the name of transfer_distance_above_threshold to something easier to understand: transfer_distance_above_2km?

If we change the transfer_distance_above_threshold to transfer_distance_above_2km, shouldn't transfer_distance_too_large become transfer_distance_above_10km?

And, to go the other way, is it advisable to have the distance set in the notice name? Are we sure we will never decide that, for example, 1 km is enough of a distance to warrant an info instead of 2km. 
Will we just add a new notice transfer_distance_above_1km in that case, while keeping transfer_distance_above_2km for backwards compatibility?

emmambd:
Hey @jcpitre! Sorry, for the wait, I asked @skalexch to take a look at how this worked in the old deprecated validator first. In the old one, they only have 1 notice and the severity varies (as well as the message which dynamically displays the invalid transfer distance) depending on the result. 

I still think we should keep the behavior as is: `transfer_distance_too_large` for the warning level and `transfer_distance_above_2km` as the info level. If we do change the threshold for the info level, we would deprecate the notice and follow the path defined in https://github.com/MobilityData/gtfs-validator/issues/1964.

jcpitre (approved):
LGTM!

PR #1966, opened 2025-02-04
feat: added `missing_pickup_booking_rule_id` notice

cka-y:
**Summary:**
- Added `missing_pickup_booking_rule_id` WARNING notice when a `stop_times` entity has `pickup_type=2` and missing `pickup_booking_rule_id`

**Expected behavior:** 
image


main/src/main/java/org/mobilitydata/gtfsvalidator/validator/PickupBookingRuleIdValidator.java
```diff
@@ -0,0 +1,54 @@
...
+  @Override
+  public void validate(GtfsStopTime entity, NoticeContainer noticeContainer) {
+    if (entity.hasPickupType()
+        && entity.pickupType() == GtfsPickupDropOff.MUST_PHONE
+        && !entity.hasPickupBookingRuleId()) {
+      noticeContainer.addValidationNotice(
+          new MissingPickupBookingRuleIdNotice(
+              entity.csvRowNumber(),
+              entity.pickupType(),
+              entity.hasDropOffType() ? entity.dropOffType() : null));
```
davidgamez, line 24: I think the logic should align the pickupType with pickup_booking_rule_id and dropOffType with drop_off_booking_rule_id when either of the types is two and its respective id is missing. Unless we have this covered in a different validator.
Screenshot 2025-02-05 at 11 02 14 AM

emmambd: Yes - this should only be affecting feeds with Flex features

cka-y: If I understand properly there's two cases:
- Case 1: `pickup_type = 2` AND `pickup_booking_rule_id` is missing -> Generate warning notice
- Case 2: `drop_off_type = 2` AND `drop_off_booking_rule_id` is missing -> Generate a warning notice
Both cases are independent from one another. 

If that is correct then this is the implemented logic. Not sure why it affects so many feeds.

emmambd: @tzujenchanmbd Mind reviewing this in more depth? I'm wondering if we need to add additional logic like "ONLY in cases where `booking_rules.txt` exists"

tzujenchanmbd: `pickup_type/drop_off_type = 2` have been in GTFS for a long time. Before Flex was officially adopted (i.e., before `booking_rules.tx`t was added to the spec), producers could simply use this field to indicate a service requiring prior booking. On the consumer side, this might have been displayed as a generic "Service requires booking" message without any details on the actual booking rules. This is likely why so many feeds are affected.

Considering that services requiring prior booking should always have associated rules, it seems reasonable for the specification to **recommend** including information about those rules.

However, given the large number of affected feeds and the fact that this warning arises due to a **new feature**, +1 adding a condition to trigger the warning `only if a booking_rules.txt file is present`, to minimize its impact.

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/PickupBookingRuleIdValidator.java
```diff
@@ -0,0 +1,66 @@
...
+          new MissingPickupDropOffBookingRuleIdNotice(
+              entity.csvRowNumber(),
+              entity.hasPickupType() ? entity.pickupType() : null,
+              entity.dropOffType()));
+    }
+  }
+
+  @Override
+  public boolean shouldCallValidate(ColumnInspector header) {
+    return header.hasColumn(GtfsStopTime.PICKUP_TYPE_FIELD_NAME);
```
jcpitre, line 39: Shouldn't you check for dropOffType also here? You should validate if either one of the columns is present. (or to say it differently there's no point in validating if both pickupType and dropOffType columns are absent)

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/PickupBookingRuleIdValidator.java
emmambd: +1 moving forward to the pseudologic change. We can proceed

emmambd:
The acceptance tests are still triggering on a lot of feeds. Maybe the point of criteria should be if any of the Flex features are identified in the feed. Thoughts @skalexch @Sergiodero?

Sergiodero:
I did a quick random check of the feeds (5-6 of them) still triggering this warning and there's two important patters I noticed after looking at the feeds' files: 

- They all had an empty `booking_rules.txt` file
- They were all produced by Trillium

I think the original logic of the rule (trigger if `pickup_type=2` and missing `pickup_booking_rule_id` in `stop_times.txt`) is correct, and if we're trying to be more flexible @tzujenchanmbd's suggestion is a setp in the right direction. 

If we want to be even more flexible we could instead of just checking for `booking_rules.txt` we could look for at least one entry in `booking_rules.txt`. My only concern with this is that the warning will not show up for someone who is not aware that this file exists and could benefit from having it, but I guess that is still true with the latest change to the rule.

Having said that, my guess is that with current state the acceptance test would pass if Trillium would stop adding empty `booking_rules.txt` files, so another option could be to talk to them before or right after releasing the new validation rules. Any thoughts @emmambd @skalexch ?

davidgamez (approved):
LGTM!

emmambd:
@cka-y As discussed in Slack, the logic of this rule is fine as this, without doing an additional check to see if there's a record in `booking_rules.txt`. The rationale behind this is: 

1) For everywhere else in the validator, we automatically trigger a `missing_required_column` notice when a file exists and a required column isn't included. This is good to preserve for consistency. 

2) It's aligned with the updates to a definition of a Field made here: https://github.com/google/transit/pull/414 that providing a header implies the field exists and should be read. 

No further work required on the spec side - this is good to merge :)

PR #1988, opened 2025-02-25
feat: 1976 modifying notice description for invalid currency amount

### LINKED ISSUE #1976: Modifying notice description for invalid_currency_amount
### Describe the problem

`invalid_currency_amount` is a common error on feeds with Fares v2 data. Right now in the validator, it's not obvious that this correlates to `currency_code`, and it also isn't clear that this is often due to formatting issues. 

Image

### Proposed solution

• csvRowNumber should be the second column, not the third
• Replace fieldName with currencyCode 

In description, change to: 

> A currency amount field has a value that does not match the format of its corresponding currency code field.
> 
> Typically, this means the amount did not have the expected number of decimal places. Check the formatting of your `amount` field so it matches the number of decimal places specified by the [`currency_code` value](https://en.wikipedia.org/wiki/ISO_4217#Active_codes). 

_No response_

### Additional context

_No response_

qcdyx:
**Summary:**
Closes #1976 

**Expected behavior:** 
html report
image

json report
image


emmambd:
LGTM from a QA side!

skalexch (commented):
@qcdyx looks good on the spec side too. I'm glad winning day is having a propagating effect!

emmambd:
As @davidgamez flagged in the original issue (#1976), to future-proof this notice we want to revise the columns. This is because multiple amounts could be associated with 1 currency code, e.g if `min_amount` and `max_amount` are adopted in the future. 

Order and changes: 

-fileName
-csvRowNumber
-currencyCode
- Add fieldName, name of the column (`amount`) in this case
- Add replace the `amount` header name with `fieldValue`

qcdyx:
Voila! 
image

davidgamez:
The Test Package Document is failing as expected. This is due to the fact that in this PR we are renaming/update field names. 
Extract from the test documentation:
```
/**
   * Is this test failing? That likely means you've added a new field name to a `Notice` that hasn't
   * been used before. That's not necessarily a bad thing, but we want to enforce some consistency
   * amongst Notice field names because they are communicated and consumed externally.
  ...
```

[More info](https://github.com/MobilityData/gtfs-validator/blob/03cd4dd5ea335cf4d211ba522d39c31b79f22a81/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/NoticeFieldsTest.java#L47)

davidgamez (approved):
LGTM!

PR #1998, opened 2025-03-03
feat: 1980 add fare product with multiple default rider categories

### LINKED ISSUE #1980: Add fare_product_with_multiple_default_rider_categories
### Describe the problem

We adopted rider categories. The PR will be merged soon.


| Related spec language                     | Rationale for rule                                            | Notice name to update or add                       | Pseudologic                                                                                                                                                                                                                               | Notice table columns                                                                                                         | Severity | Notice description                                                                                           |
|-------------------------------------------|----------------------------------------------------------------|---------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------|----------|-------------------------------------------------------------------------------------------------------------|
| rider_categories.rider_category_id, fare_products.rider_category_id, is_default_fare_category        | When multiple rider categories are eligible for a single fare product specified by a `fare_product_id`, there must be exactly one of these eligible rider categories indicated as the default rider category                               | fare_product_with_multiple_default_rider_categories                            | if `rider_categories.txt` does not exist: Do not trigger a notice and ignore subsequent checks <br>For each `fare_product_id`: If multiple `rider_category_id` are associated: <br>If more than one of these rider_category_id has `is_default_fare_category ==1`: Trigger Notice    | FareProductId, CsvRowNumber1,CsvRowNumber2 (for the fare product), RiderCategoryId1,RiderCategoryId2  | ERROR    | A fare product in `fare_products.txt` has more than one eligible default rider category from `rider_categories.txt`.|


### Sample GTFS datasets

[http://developer.trimet.org/schedule/gtfs.zip](https://developer.trimet.org/schedule/gtfs.zip)

### Severity

ERROR

### Additional context

gtfs.org documentation: https://github.com/google/transit/pull/511

qcdyx:
**Summary:**
Closes #1980 

**Expected behavior:** 
image


emmambd:
Looks like the missing_required_column issue is still happening with acceptance tests. cc @davidgamez

skalexch:
@qcdyx I'm not sure about the tests, the test is passing but I feel like it's not validating the proposed behavior.
The fare_product_id should be the same "fare 1" and applied to two default rider_categories "rider 1" and "rider 2"

I also added a comment about the creation of test rider_categories. With that one, the rider_category_id is unique. So one category cannot be default and not default at the same time

emmambd:
@qcdyx I think @skalexch's comment is specifically about your test file, not about the rule logic itself. @skalexch is that an accurate interpretation? Let's move this to Slack if it needs back and forth, but the logic of the rule goes unchanged I believe.

qcdyx:
Thanks @emmambd for clarification. @skalexch I temporarily modified the test dataset to generate a screenshot for Emma to verify the notice name and description. I should've let you know beforehand.

skalexch:
@qcdyx what I meant was the tests are probably not testing what the notice should flag. The notice code looks right (my java isn't great but it seems like you're createing a map of productId -> riderCategoryId if it is default then counting the occurences).
So the test needs to be adjusted to reflect what the notice is supposed to validate.

davidgamez (approved):
LGTM

emmambd:
Waiting on @skalexch for this

skalexch:
@qcdyx thanks for updating. There's one detail left to clear out the false positives.
For the flagged feeds, the same `fare_product_id` was associated with the same `rider_category_id` but had different `fare_media_id` since it's part of the primary key. What the code does is that when it detects that a rider_category_id is default, it adds it to the count of the fare_product_id. So it seems like it's double counting the same ADULT category.
I think a better approach would be having the hashmap value be a set then count the elements in that set. Or just do a check that the category already exists in the value of a hashmap. Then count the elements in the fareProductDefaultCount.get(fareProductId)
Screenshot 2025-03-04 at 4 33 58 PM

davidgamez:
> @qcdyx thanks for updating. There's one detail left to clear out the false positives. For the flagged feeds, the same `fare_product_id` was associated with the same `rider_category_id` but had different `fare_media_id` since it's part of the primary key. What the code does is that when it detects that a rider_category_id is default, it adds it to the count of the fare_product_id. So it seems like it's double counting the same ADULT category. I think a better approach would be having the hashmap value be a set then count the elements in that set. Or just do a check that the category already exists in the value of a hashmap. Then count the elements in the fareProductDefaultCount.get(fareProductId) Screenshot 2025-03-04 at 4 33 58 PM

Hi @skalexch, does this require clarification on the spec?

skalexch:
@davidgamez maybe we can change the description to "When multiple unique rider categories are eligible for a single fare product specified by a fare_product_id, there must be exactly one of these eligible rider categories indicated as the default rider category (is_default_fare_category = 1)."
@tzujenchanmbd will probably have a better phrasing for it. We'll discuss it when he gets back

emmambd:
@skalexch You mean on the spec side? Sounds good. The validator side will be merged this week and we can make changes later post 7.0

skalexch:
@emmambd yes on the spec side. On the validator side once Jingsi resolves the final comment I'll do one final round of QA and I think it will be ready to go.

davidgamez:
> @davidgamez maybe we can change the description to "When multiple unique rider categories are eligible for a single fare product specified by a fare_product_id, there must be exactly one of these eligible rider categories indicated as the default rider category (is_default_fare_category = 1)." @tzujenchanmbd will probably have a better phrasing for it. We'll discuss it when he gets back

My only concern is that is not clear that `fare_media_id` is part of the logic. For me, both interpretations are correct => 
- Can be only one rider category default per fare_product_id and
- Can be only one rider category default per fare_product_id+fare_media_id

Issue #1960, opened 2025-02-04
Acceptance tests summary comment not added to the PR from forks


davidgamez:
### Describe the bug

The acceptance tests summary is not added to the PR when the action is triggered in forks.
The following error is visible in the logs:
```
(node:1) UnhandledPromiseRejectionWarning: HttpError: Resource not accessible by integration
    at /node_modules/@octokit/request/dist-node/index.js:66:23
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
(Use `node --trace-warnings ...` to show where the warning was created)
(node:1) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:1) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```
Example: 
- PR -> https://github.com/MobilityData/gtfs-validator/pull/1959
- GitHub action -> https://github.com/MobilityData/gtfs-validator/actions/runs/13128106708/job/36650937663?pr=1959

### Steps/Code to Reproduce

- Create a fork with a non-contributor user
- Create a PR that triggers the acceptance tests

### Expected Results

The acceptance tests summary comment is added to the PR.

### Actual Results

The acceptance tests summary comment is not added and an error is visible in the logs of the `compare-output` action.

### Validator version

N/A

### Operating system

N/A

mackenziereading19:
I traced the current acceptance workflow. It still runs under pull_request, executes the PR code, and then tries to post the generated summary using GITHUB_TOKEN. For fork PRs that token is read-only, which matches the reported Resource not accessible by integration failure.

I don't think simply granting pull-requests: write fixes fork runs, and moving the acceptance workflow itself to pull_request_target would be unsafe because it executes PR-controlled code.

Would you be comfortable with a two-stage approach: keep the acceptance tests on pull_request, upload the generated summary as an artifact, then use a trusted workflow_run workflow with minimal pull-requests: write permission solely to retrieve that artifact and post it to the originating PR? I can implement that with the privileged workflow avoiding checkout/execution of fork code.

PR #2020, opened 2025-04-18
fix: update TripAndShapeDistanceValidator to ignore shape with no shape_dist_traveled (#2018)

### LINKED ISSUE #2018: trip_distance_exceeds_shape_distance: do not trigger when shapes.shape_dist_traveled column is empty
### Describe the problem

Example feed: https://mobilitydatabase.org/feeds/gtfs/mdb-2458 

Right now, when a trip has a shape_dist_traveled value in stop_times.txt but its associated shape has no shape_dist_traveled value in shapes.txt, the trip_distance_exceeds_shape_distance error is triggered. This should not happen.

### Proposed solution

- Review why this is happening (why do we interpret shapes.shape_dist_traveled value as 0 when the column doesn't exist in shapes.txt?) 
- Modify behavior
- Test with feed to ensure the error no longer triggers
- Additional request: modify notice description to include: 

"A stop time's shape_dist_traveled exceeds the maximum shape_dist_traveled of the shape for the related trip in shapes.txt." 

And references: https://gtfs.org/documentation/schedule/reference/#stop_timestxt

_No response_

### Additional context

_No response_

thelleboid-tsr:
**Summary:**
update TripAndShapeDistanceValidator to ignore shape with no shape_dist_traveled

**Expected behavior:** 

Some GTFS feeds do not provide shape_dist_traveled in shapes.txt In this case, the value is set to 0 in the code, so sorting shapes based on ShapeDistTraveled does not work and usually returns the first shape row.

It leads to an error, the first shape of the shape is usually very far away from the last stop of the trips following the shape.

We can assume the biggest shape_dist_traveled of a given shape should never be 0, otherwise it means the value is not set. In this case, we return early to ensure no validation error will be issued.

Fixes #2018

⚠️ I've not been able to run unit tests on this PR

### DIFF
--- a/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/TripAndShapeDistanceValidatorTest.java
+++ b/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/TripAndShapeDistanceValidatorTest.java
@@ -109,6 +109,24 @@ public class TripAndShapeDistanceValidatorTest {
   }
 
   @Test
+  public void testTripDistanceExceedsShapeDistanceNoShapeDistance() {
+    List<ValidationNotice> notices =
+        generateNotices(
+            createTripTable(2),
+            createStopTimesTable(1, 10.0),
+            createShapeTable(1, 0.0, 10.0),
+            createStopTable(1));
+    boolean found =
+        notices.stream()
+            .anyMatch(
+                notice ->
+                    notice
+                        instanceof
+                        TripAndShapeDistanceValidator.TripDistanceExceedsShapeDistanceNotice);
+    assertThat(found).isFalse();
+  }
+
+  @Test
   public void testTripDistanceExceedsShapeDistanceWarning() {
     List<ValidationNotice> notices =
         generateNotices(
--- a/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TripAndShapeDistanceValidator.java
+++ b/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TripAndShapeDistanceValidator.java
@@ -84,6 +84,11 @@ public class TripAndShapeDistanceValidator extends FileValidator {
               }
 
               double maxShapeDist = maxShape.shapeDistTraveled();
+
+              if (maxShapeDist == 0) {
+                return;
+              }
+
               double distanceInMeters =
                   getDistanceMeters(maxShape.shapePtLatLon(), stop.stopLatLon());
               if (maxStopTimeDist > maxShapeDist) {


main/src/test/java/org/mobilitydata/gtfsvalidator/validator/TripAndShapeDistanceValidatorTest.java
```diff
@@ -40,7 +40,7 @@ private static List<GtfsShape> createShapeTable(
               .setShapePtLat(lonLat)
               .setShapePtLon(lonLat)
               .setShapePtSequence(0)
-              .setShapeDistTraveled(shapeDistTraveled + i)
+              .setShapeDistTraveled(shapeDistTraveled * i)
```
jcpitre, line 43: Thanks for the contribution!
Why this change?
It makes some tests fail:

[TripAndShapeDistanceValidatorTest.testTripDistanceExceedsShapeDistance](https://github.com/MobilityData/gtfs-validator/blob/f1174bd08ed57366ef8a54fe360f89543b5beecd/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/TripAndShapeDistanceValidatorTest.java#L94)
[TripAndShapeDistanceValidatorTest.testTripDistanceExceedsShapeDistanceWarning](https://github.com/MobilityData/gtfs-validator/blob/f1174bd08ed57366ef8a54fe360f89543b5beecd/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/TripAndShapeDistanceValidatorTest.java#L112)

Your test will work if you keep the line as is (i.e +1), since you set `shapeDistTraveled` to 0 when calling `createShapeTable` and only one row is created.

You can run the tests locally with:
`./gradlew test`

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/TripAndShapeDistanceValidatorTest.java
```diff
@@ -108,6 +108,24 @@ public void testTripDistanceExceedsShapeDistance() {
     assertThat(found).isTrue();
   }
 
+  @Test
+  public void testTripDistanceExceedsShapeDistanceNoShapeDistance() {
+    List<ValidationNotice> notices =
+        generateNotices(
+            createTripTable(2),
+            createStopTimesTable(1, 10.0),
```
jcpitre, line 116: I think this will not really test your change. You want to set test data that would normally create a notice but will not because shapeTable.shapeDistTraveled is 0 (please correct me if I am wrong).
But to generate a notice you need the distance to be at least 11.1 meters. And this test sets the distance to 10 meters. So the test as it is now will not generate notices, because the distance falls below the threshold.

I believe that if you set `createStopTimesTable(1, 100.0)` (for example), it will test properly.

skalexch:
@thelleboid-tsr I don't see a problem spec-wise. Once you're able to generate acceptance tests, I can take a look at the data and notices.

emmambd:
Hi @skalexch! @thelleboid-tsr already generated the acceptance tests at https://github.com/MobilityData/gtfs-validator/actions/runs/14531693839 so we can review them there. [There's a separate issue for generating a comment when the acceptance test is created from forked PRs. ](https://github.com/MobilityData/gtfs-validator/issues/1960).

skalexch (commented):
@emmambd thank you. @thelleboid-tsr looking at the datasets, it seems like the error is dropped for both feeds that do not have the column (eg mdb-67) and feeds that have the field but it's empty (eg mdb-1270). All good spec-wise! I'll leave the approval for dev.

jcpitre (approved):
LGTM!
The increased number of corrupted feeds is not related to this PR.

PR #2040, opened 2025-06-21
fix: handle GeoJSON locations when computing distance between stops

### LINKED ISSUE #2031: fast_travel_between_far_stops incorrectly calculates distance when Flex location is between two stops
### Describe the bug

The validator seems to be calculating the distance between two fixed stops incorrectly when a location.geojson record is between those two stop times. The distance is several magnitudes larger than reality (e.g., ~17,000,000 meters vs an expected distance of ~2,000 meters). This is causing false positives for the  `fast_travel_between_far_stops` warning.

### Steps/Code to Reproduce

Run this dataset through the [validator](https://gtfs-validator.mobilitydata.org/):
[greenmtncn-vt-us--flex-v2-TEST.zip](https://github.com/user-attachments/files/20259300/greenmtncn-vt-us--flex-v2-TEST.zip)

### Expected Results

The validation results should calculate the distance between two stops correctly even if there is a Flex zone between those two stops. In this case, it shouldn't throw a warning because the distance is not that far relative to the time.

Notice the distance between the two stops with arrival times of 14:48:00 and 14:56:00, respectively:

Image

### Actual Results

Validation results: https://gtfs-validator-results.mobilitydata.org/98b0cbfc-f424-4c19-97a2-c69f39cfa8a1/report.html

Example:
`stop_times.trip_id=t_5959537_b_83643_tn_1`, `stop_sequence=1` to `stop_sequence=3`

`shape_dist_traveled` in dataset: 2466.4772193312524m

The validation calculates the distance as 17295.69704407396km and throws a warning:

Image

### Screenshots

_No response_

### Files used

_No response_

### Validator version

Production Web UI

### Operating system

MacOS

### Java version

_No response_

### Additional notes

_No response_

sylvansson:
**Summary:**

This PR fixes a bug where we calculate the distance between two stops incorrectly if there's a GeoJSON location between them. We currently fall back to Null Island (0, 0) when we can't get the coordinates of a stop time's stop (in this case, because there's no stop id for a GeoJSON location) which messes up our calculations and can lead to false positives for the fast_travel_between_far_stops notice. See [this comment](https://github.com/MobilityData/gtfs-validator/issues/2031#issuecomment-2902791587).

We also address false negatives for the fast_travel_between_consecutive_stops notice due to only comparing consecutive entries in stop_times.txt.

I'll write some tests once we're happy with the approach.

Closes #2031 

**Expected behavior:** 

- We don't get a fast_travel_between_far_stops notice when two stops are within a reasonable distance but there's a GeoJSON location between them.
- We get a fast_travel_between_consecutive_stops notice when two consecutive stops are too far apart and there's a GeoJSON location between them.

### DIFF
--- a/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/StopTimeTravelSpeedValidatorTest.java
+++ b/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/StopTimeTravelSpeedValidatorTest.java
@@ -17,13 +17,16 @@
 package org.mobilitydata.gtfsvalidator.validator;
 
 import static com.google.common.truth.Truth.assertThat;
+import static org.mobilitydata.gtfsvalidator.validator.StopTimeTravelSpeedValidator.findDistancesKmBetweenStops;
 import static org.mobilitydata.gtfsvalidator.validator.StopTimeTravelSpeedValidator.getSpeedKphBetweenStops;
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.geometry.S2LatLng;
+import com.ibm.icu.impl.Pair;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Optional;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.JUnit4;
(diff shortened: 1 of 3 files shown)


sylvansson:
I see some of the tests are failing. I need to look into that.

main/src/test/java/org/mobilitydata/gtfsvalidator/validator/StopTimeTravelSpeedValidatorTest.java
```diff
@@ -298,6 +301,8 @@ public void farStopsMissingTimeInBetween_yieldsNotice() {
                 GtfsTime.fromString("08:02:30")));
     assertThat(generateNotices(ImmutableList.of(route), ImmutableList.of(trip), stopTimes, stops))
         .containsExactly(
+            createFastTravelBetweenConsecutiveStopsNotice(
```
sylvansson, line 304: Since we now skip stops when we can't compute the distance or speed, we sometimes effectively check non-consecutive stops, which feels slightly misleading. If needed I can adjust the logic to only emit a notice for strictly consecutive entries. 

cc: @davidgamez @emmambd

sylvansson (commented):
I think this is ready for review, sorry for the delay.

jcpitre:
Normally we should see a markdown version of the acceptance report summary inserted in the PR..
Unfortunately the [acceptance_test.yml ](https://github.com/MobilityData/gtfs-validator/blob/master/.github/workflows/acceptance_test.yml)action cannot write to the PR because of a permission problem. 

I copied some of the results here:

<summary><strong>New Warnings</strong> (7 out of 987 datasets, ~1%) ✅</summary>

| Dataset | Notice Code |
|---------|-------------|
| mdb-2135 | fast_travel_between_consecutive_stops |
| mdb-2202 | fast_travel_between_consecutive_stops |
| mdb-2881 | fast_travel_between_consecutive_stops |
| mdb-35 | fast_travel_between_consecutive_stops |
| mdb-68 | fast_travel_between_consecutive_stops |
| mdb-7 | fast_travel_between_consecutive_stops |
| mdb-75 | fast_travel_between_consecutive_stops |

<summary><strong>Dropped Warnings</strong> (4 out of 987 datasets, ~0%) ✅</summary>

| Dataset | Notice Code |
|---------|-------------|
| mdb-2446 | fast_travel_between_far_stops |
| mdb-2447 | fast_travel_between_far_stops |
| mdb-2831 | fast_travel_between_far_stops |
| mdb-2882 | fast_travel_between_far_stops |

Which corresponds to the expected changes.

PR #2094, opened 2026-02-11
feat: service id with no active day validator

cka-y:
**Summary:**
This pull request introduces a new validator to ensure that each service listed in `calendar.txt` has at least one active day of the week. The validator checks for services that are not valid for any day and emits a warning notice if such cases are found. Comprehensive unit tests are also added to verify the validator's behavior.

New validator for service activity:

* Added `ServiceHasNoActiveDayOfTheWeekValidator` class to detect services in `GtfsCalendarTableContainer` that have no active day of the week, emitting a warning notice for each such service.

**Expected behavior:** 
image

### DIFF
--- a/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTimesTripBlockOrderValidator.java
+++ b/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTimesTripBlockOrderValidator.java
@@ -1,0 +1,110 @@
+package org.mobilitydata.gtfsvalidator.validator;
+
+import static org.mobilitydata.gtfsvalidator.notice.SeverityLevel.INFO;
+
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+import javax.inject.Inject;
+import org.mobilitydata.gtfsvalidator.annotation.GtfsValidationNotice;
+import org.mobilitydata.gtfsvalidator.annotation.GtfsValidator;
+import org.mobilitydata.gtfsvalidator.notice.NoticeContainer;
+import org.mobilitydata.gtfsvalidator.notice.ValidationNotice;
+import org.mobilitydata.gtfsvalidator.table.*;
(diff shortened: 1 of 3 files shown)


davidgamez:
> Logic-wise and looking at the tests, everything looks correct. I also checked a couple datasets from the acceptance tests and they check out. One of them even does the bad practice of defining all the `service_ids` with no active days in `calendar.txt` and then specifying the active days fully in `calendar_dates.txt`.
> 
> The acceptance tests are failing but I find it normal that a lot of the datasets are affected by this.

Should we add the calendar_dates.txt to the notice's logic? Is it not a valid use case to have service IDs defined only in calendar_dates.txt in a feed that contains both files?

skalexch:
@davidgamez it is a valid use case to have service IDs defined only in calendar_dates.txt in a feed that contains both files.
I think the scope of this notice is this particular bad practice:
1) Producer defines "service_1" in `calendar.txt` and does not associate any days of week with it.
2) Then, the producer adds the dates for "service_1" in `calendar_dates.txt`.

[The is what the spec actually says.](https://gtfs.org/documentation/schedule/reference/#calendar_datestxt:~:text=Recommended%3A%20Use%20calendar_dates,is%20an%20ID.):
- Both files can be used in conjunction (RECOMMENDED), where `calendar_dates.txt` overrides whatever exists in `calendar.txt`.
- Only `calendar_dates.txt` is used. So the `service_id` in `calendar_dates.txt` does not necessarily have to reference something in `calendar.txt`. Therefore, the practice of having a `service_id` defined with absolutely no days of week in `calendar.txt` just because we want it to be referenced in `calendar_dates.txt` does not add anything.


The case that you mention has more to do with an "unusable_service_id". This is a very rare case where all the dates and added services are negated by `calendar_dates.txt` entries that remove the service. Or even worse cases where all days of week are =0 AND all `calendar_dates.txt` services remove the service as well.

I these cases can be left to an INFO notice for an upcoming release.

PR #2108, opened 2026-02-24
feat: Added MissingRequiredAgencyIdValidator

### LINKED ISSUE #2073: Add notice for agency_id
### Describe the problem

`agency_id` is a conditionally required field, and there is no `missing_agency_id` notice for it. stop_name is also a conditionally required field, and it has its own `missing_stop_name` notice. Wouldn’t it make sense to have a similar notice for `agency_id` as well?

If:
    (number_of_agencies > 1) AND (agency_id = '' OR agency_id is missing):
        trigger a `missing_agency_id` notice

### Sample GTFS datasets

_No response_

### Severity

ERROR

### Additional context

_No response_

jcpitre:
**Summary:**
closes #2073 
added MissingRequiredAgencyIdValidator

### DIFF
--- a/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/FeedValidTodayValidator.java
+++ b/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/FeedValidTodayValidator.java
@@ -1,0 +1,90 @@
+/*
+ * Copyright 2026 MobilityData IO
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.mobilitydata.gtfsvalidator.validator;
+
+import static org.mobilitydata.gtfsvalidator.notice.SeverityLevel.INFO;
+
+import javax.inject.Inject;
+import org.mobilitydata.gtfsvalidator.annotation.GtfsValidationNotice;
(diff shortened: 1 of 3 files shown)


jcpitre:
@skalexch Taking another look at this, it seems that we already have a `missing_required_field` notice triggered for the same conditions as `missing_required_agency_id`
image

In my dataset I have 2 agencies with one having no ID. I get 3 notices of `missing_required_agency_id` AND 3 more of `missing_required_field`

Do we really need both?

Here is the test dataset:
[2073.zip](https://github.com/user-attachments/files/25557371/2073.zip)

skalexch:
@jcpitre I remember having a conversation about this (I think with Emma) but it's been a while. I think the person who raised the issue wanted to highlight that `agency_id` is required when there are multiple agencies, the same way it is highlighted for `missing_stop_name`.
I am for deduplication, only one should be there. The description of the new `missing_required_agency_id` notice is better, so we can remove the original checks for [fare_attributes.txt](https://github.com/MobilityData/gtfs-validator/blob/master/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/FareAttributeAgencyIdValidator.java), [agency.txt](https://github.com/MobilityData/gtfs-validator/blob/1ae8cd7138dee21a971bfa1299fea18ee7ab5f1a/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/AgencyConsistencyValidator.java#L71), and [routes.txt](https://github.com/MobilityData/gtfs-validator/blob/master/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/RouteAgencyIdValidator.java).

One caveat: All the previous checks contain a `missing_recommended_field` notice if the agency count == 1.

Note: I think we should stop at this point for having special case `missing_required_field` notices.

jcpitre:
So essentially the real aim of the issue is replacing `missing_required_field` by `missing_required_agency_id` when it's related to agency id?
Am I right?

jcpitre:
@skalexch Should we also replace the `missing_recommended_field` notice by `missing_recommended_agency_id`

skalexch:
@jcpitre no need to do that for the recommended field, too much specialization. Let's keep it for the required only.

PR #2126, opened 2026-03-05
feat: feed service window validator

cka-y:
**Summary:**

This PR introduces a new validator, `FeedServiceWindowValidator`, which checks the relationship between the feed validity period and service windows in GTFS feeds. 

**New validator for feed and service window alignment:**

* Added `FeedServiceWindowValidator`, which performs two main checks:
  1. Emits a notice if any service window (from `calendar.txt`/`calendar_dates.txt` used by `trips.txt`) extends outside the feed validity period (`feed_info.txt`).
  2. Emits a notice if the feed validity period extends more than 14 days beyond the total service window covered by all services.

**Notice fields:**

* Added new notice classes: `ServiceWindowExtendsPastFeedPeriodNotice` and `FeedValidBeyondTotalServiceWindowNotice`, both at INFO severity, to report the above conditions.

image

image

### DIFF
--- a/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/FeedServiceWindowValidator.java
+++ b/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/FeedServiceWindowValidator.java
@@ -16,7 +16,7 @@
 
 package org.mobilitydata.gtfsvalidator.validator;
 
-import static org.mobilitydata.gtfsvalidator.notice.SeverityLevel.WARNING;
+import static org.mobilitydata.gtfsvalidator.notice.SeverityLevel.INFO;
 
 import java.time.LocalDate;
 import java.time.temporal.ChronoUnit;
@@ -94,7 +94,7 @@ public class FeedServiceWindowValidator extends FileValidator {
     LocalDate totalWindowStart = null;
     LocalDate totalWindowEnd = null;
 
-    // Iterate unique service IDs from trips.txt — one entry per service, no deduplication needed.
+    // Iterate unique service IDs from trips.txt. One entry per service, no deduplication needed.
     for (String serviceId : tripTable.byServiceIdMap().keySet()) {
       ServiceInterval interval =
           serviceIntervalCache.getIntervals(serviceId, calendarTable, calendarDateTable);
@@ -167,7 +167,7 @@ public class FeedServiceWindowValidator extends FileValidator {
 
   /** A service window is not covered by the feed's validity period. */
   @GtfsValidationNotice(
-      severity = WARNING, // TODO tmp should be INFO
+      severity = INFO,
       files =
           @FileRefs({
             GtfsCalendarSchema.class,
@@ -207,7 +207,7 @@ public class FeedServiceWindowValidator extends FileValidator {
 
   /** The feed is valid 14 days beyond its total service window. */
   @GtfsValidationNotice(
-      severity = WARNING, // TODO tmp should be INFO
+      severity = INFO,
       files =
           @FileRefs({
             GtfsCalendarSchema.class,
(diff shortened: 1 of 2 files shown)


main/src/main/java/org/mobilitydata/gtfsvalidator/validator/FeedServiceWindowValidator.java
```diff
@@ -0,0 +1,242 @@
...
+import org.mobilitydata.gtfsvalidator.notice.ValidationNotice;
+import org.mobilitydata.gtfsvalidator.table.*;
+import org.mobilitydata.gtfsvalidator.util.ServiceInterval;
+import org.mobilitydata.gtfsvalidator.util.ServiceIntervalCache;
+
+/**
+ * Validates the relationship between the feed validity period (feed_info.txt) and the service
+ * windows defined in calendar.txt and calendar_dates.txt.
+ *
+ * <p>Two complementary checks are performed in a single pass over all unique service IDs sourced
```
jcpitre, line 38: Cana you explain why you go through trips.txt, whereas [ServiceSpreadValidator](https://github.com/MobilityData/gtfs-validator/blob/master/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/ServiceSpreadValidator.java) doesn't.

cka-y:
# Analysis of affected feeds
## 1. Notice prevalence
image

The graph above shows the notice prevalence across all 1,003 feeds from the acceptance tests. The majority of feeds (788, 78.6%) trigger neither notice, indicating that most agencies correctly align their service windows with their declared feed validity period. Among affected feeds, `service_window_extends_past_feed_period` is the dominant notice appearing in 183 feeds (168 exclusively + 15 co-occurring with `feed_valid_beyond_total_service_window`), compared to only 47 feeds for `feed_valid_beyond_total_service_window` (32 exclusively + 15 co-occurring). Notably, only 15 feeds trigger both notices simultaneously, suggesting the two misalignment patterns are largely independent of each other and stem from different agency-level data practices.

## 2. `service_window_extends_past_feed_period`
image
Analysis of the GTFS feeds from the acceptance tests reveals that service window overflow before the feed start date is significantly more prevalent than overflow past the feed end date (8,577 vs. 2,243 affected services). While median overflow values remain relatively contained (64 days before feed start and 264 days after feed end) the mean for the "before" direction inflates dramatically to 2,503 days due to a distinct cluster of services with implausibly large overflows (~20,000 days), likely attributable to corrupt or default-value dates in a small number of agency datasets. 

## 3. `feed_valid_beyond_total_service_window`
image

The `feed_valid_beyond_total_service_window` notice reveals that the more prevalent mismatch direction is at the feed end i.e. more feeds declare a validity end date that extends beyond their last active service date than feeds whose validity starts before their first active service. While the median trailing gap of 170 days suggests, two feeds (mdb-96 and mdb-525) show extreme outliers of ~2,700 and ~9,700 days respectively, indicating agency-specific anomalies rather than a systemic pattern. The leading gap (feed start before service window) is less common but tends to be larger when it does occur, with a median of 261 days and one notable outlier at ~2,500 days (mdb-1815).

cc: @skalexch

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/FeedServiceWindowValidator.java
cka-y: `trips.txt` has a `byServiceIdMap` which makes it easier to iterate through services. It can be done in other ways for sure.

jcpitre: So any service not referenced by a trip it will not be included. I guess it makes sense.
@skalexch does it? 

It makes me think that maybe ServiceSpreadValidator should be modified since it only uses calendar.txt to get the list of services. Any service not in calendar.txt but present in calendar_dates.txt (it's allowed) will not be used to calculate the spread. 
We should think about harmonizing the way it's done here and in [ServiceSpreadValidator](https://github.com/MobilityData/gtfs-validator/blob/master/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/ServiceSpreadValidator.java).

PR #2139, opened 2026-03-28
feat: warn when trip_headsign matches intermediate stop name

PatrickSteil:
Adds a new validation rule `TripHeadsignValidator` that warns when a trip's `trip_headsign` matches the `stop_name` of an intermediate stop (i.e., any stop that is not the last stop of the trip). This can cause confusion for passengers boarding after that stop, as the headsign suggests the vehicle is heading somewhere it has already passed.
The notice `TripHeadsignMatchesIntermediateStopNotice` is emitted as a `WARNING` and includes:

- the offending `trip_id` and its row number in `trips.txt`
- the `trip_headsign` value
- the `stop_id` and `stop_sequence` of the intermediate stop matching the `trip_headsign`
- the `stop_id` of the actual last stop of the trip

`NoticeFieldsTest` has been updated to register a new notice field name introduced: `tripHeadsign`.

Note that if a trip visits the last stop multiple times, then no warning is shown.

The following screenshot shows a real-world example of the official Swiss GTFS feed, where a trip passed through St. Gallen, and continues to Winterthur. But the `trip_headsign` shows `St. Gallen`.

Screenshot 2026-03-28 at 14 20 28

### DIFF
--- a/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TripHeadsignValidator.java
+++ b/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TripHeadsignValidator.java
@@ -2,10 +2,8 @@ package org.mobilitydata.gtfsvalidator.validator;
 
 import static org.mobilitydata.gtfsvalidator.notice.SeverityLevel.INFO;
 
-import java.util.Comparator;
 import java.util.List;
 import java.util.Optional;
-import java.util.stream.Collectors;
 import javax.inject.Inject;
 import org.mobilitydata.gtfsvalidator.annotation.GtfsValidationNotice;
 import org.mobilitydata.gtfsvalidator.annotation.GtfsValidationNotice.FileRefs;
--- a/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TripHeadsignValidator.java
+++ b/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TripHeadsignValidator.java
@@ -1,0 +1,135 @@
+package org.mobilitydata.gtfsvalidator.validator;
+
+import static org.mobilitydata.gtfsvalidator.notice.SeverityLevel.WARNING;
+
+import java.util.Comparator;
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+import javax.inject.Inject;
+import org.mobilitydata.gtfsvalidator.annotation.GtfsValidationNotice;
+import org.mobilitydata.gtfsvalidator.annotation.GtfsValidationNotice.FileRefs;
+import org.mobilitydata.gtfsvalidator.annotation.GtfsValidator;
+import org.mobilitydata.gtfsvalidator.notice.NoticeContainer;
+import org.mobilitydata.gtfsvalidator.notice.ValidationNotice;
+import org.mobilitydata.gtfsvalidator.table.GtfsStop;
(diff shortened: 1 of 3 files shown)


PatrickSteil:
In this PR https://github.com/patrickbr/gtfstidy/pull/41, i posted the same example but with a screenshot of the trip's route, as well as implementing a fix by overwriting the `stop_headsign`.

emmambd:
Thanks @PatrickSteil for this great rule addition! 

There are some cases where the trip headsign changes mid-trip. In this situation, the [best practices](https://gtfs.org/documentation/schedule/schedule-best-practices/#tripstxt) recommend that feeds set trip.headsign to an intermediate stop and then override it using stop_times.stop_headsign for the later portion.

Screenshot 2026-04-22 at 1 24 31 PM

Due to this case, we'd like to proceed with downgrading this rule from a WARNING to INFO severity, since there are some cases where the trip_headsign matching an intermediate stop name is valid behavior. cc @tzujenchanmbd @skalexch

PatrickSteil:
Thank you for your comment, I changed the WARNING to INFO. I tested i locally on the Swiss Instance, and it is now shown as INFO.
Bildschirmfoto vom 2026-04-24 18-15-23

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TripHeadsignValidator.java
```diff
@@ -0,0 +1,135 @@
...
+      String headsign = trip.tripHeadsign();
+      String tripId = trip.tripId();
+
+      List<GtfsStopTime> stopTimes = stopTimeTable.byTripId(tripId);
+      if (stopTimes.size() < 2) {
+        continue; // Not enough stops to have an intermediate stop
+      }
+
+      // Sort by stop_sequence to find the true last stop
+      List<GtfsStopTime> sorted =
```
davidgamez: The stopTimes list is already sorted by sequence. This is specified in the schema, [link](https://github.com/MobilityData/gtfs-validator/blob/ffff0ec556715d394c35a883c6f609bf28e3b231/main/src/main/java/org/mobilitydata/gtfsvalidator/table/GtfsStopTimeSchema.java#L67). In your local generated code GtfsStopTimeTableContainer, you can find the generated method with the following comment:
```
  /**
   * @return List of org.mobilitydata.gtfsvalidator.table.GtfsStopTime sorted by stop_sequence
   */
```
Even though the list is already sorted, and that helps the sorting algorithm, stop times tend to be a large file; having a stream and sorting will impact performance for larger feeds.

davidgamez (changes requested):
Great contribution! I just added a comment regarding the method's performance easy to fix.
Thanks!

main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TripHeadsignValidator.java
davidgamez: I realize this may not be well-documented for developers, and I apologize for that.

PatrickSteil: Thanks for the feedback! Yes, I was not sure whether i could assume the stop times to be sorted :) I am going to rewrite this

davidgamez (approved):

PR #2165, opened 2026-05-17
notice(non_ascii_or_non_printable_char): use standard fieldName

### LINKED ISSUE #1205: Use standard `fieldName` field for `non_ascii_or_non_printable_char` validation instead of custom `columnName` field
**Describe the bug**
Perhaps this is not a "bug" per se but it is not ideal. The [`non_ascii_or_non_printable_char`](https://github.com/MobilityData/gtfs-validator/blob/master/RULES.md#non_ascii_or_non_printable_char) validation uses a custom `columnName` field to identify the affected column in validation output, whereas it is typical (at least 5+ other validations use this) to use the `fieldName` field.

Confirmed using the [raw RULES.md](https://raw.githubusercontent.com/MobilityData/gtfs-validator/master/RULES.md) file that `non_ascii_or_non_printable_char` is the only validation that populates `columnName`.

This disrupted our pipeline because we had not accounted for this `columnName` field. It is fixable but it seems like it would be preferable to just use a standard field. 

**How we reproduce the bug**
1. Validate the Treasure Island Ferry feed (requires API key): http://api.511.org/transit/datafeeds?api_key={{ API_KEY }}&operator_id=TF
2. See multiple `non_ascii_or_non_printable_char` validations
3. These validations use the `columnName` field rather than the more common `fieldName`

**Expected behaviour**
The information in the `columnName` field for the `non_ascii_or_non_printable_char` validation should instead be provided in the standard `fieldName` field.

**Observed behaviour**
`non_ascii_or_non_printable_char` uses a custom `columnName`.

**Environment versions**
- validator version: 2.0.0
- Java version: N/A
- OS versions: N/A

c-tonneslan:
The serialized output of `NonAsciiOrNonPrintableCharNotice` carries the offending column under the key `columnName`, but every other notice (`InvalidCurrencyNotice`, `InvalidPhoneNumberNotice`, etc.) uses `fieldName`. That forces downstream pipelines to special-case this one rule.

Just renamed the field to match. Callers already pass `cellContext.fieldName()` into the constructor positionally, so no caller signatures change and no behavior changes — only the JSON key the field reflects to.

Closes #1205.

### DIFF
--- a/core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NonAsciiOrNonPrintableCharNotice.java
+++ b/core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NonAsciiOrNonPrintableCharNotice.java
@@ -34,17 +34,17 @@ public class NonAsciiOrNonPrintableCharNotice extends ValidationNotice {
   /** Row number of the faulty record. */
   private final int csvRowNumber;
 
-  /** Name of the column where the error occurred. */
-  private final String columnName;
+  /** Faulty record's field name. */
+  private final String fieldName;
 
   /** Faulty value. */
   private final String fieldValue;
 
   public NonAsciiOrNonPrintableCharNotice(
-      String filename, int csvRowNumber, String columnName, String fieldValue) {
+      String filename, int csvRowNumber, String fieldName, String fieldValue) {
     this.filename = filename;
     this.csvRowNumber = csvRowNumber;
-    this.columnName = columnName;
+    this.fieldName = fieldName;
     this.fieldValue = fieldValue;
   }
 }


skalexch (commented):
Thank you for the change! It makes sense since fieldName is the standard attribute for all other notices using the header name of the column.

@emmambd this requires another review from a dev.

Also, a change needs to be done on the [web page](https://gtfs-validator.mobilitydata.org/rules.html#non_ascii_or_non_printable_char-rule)?

We need to swap columnName for FieldName and change the definition to "The name of the faulty field."

emmambd:
@skalexch The documentation is produced by the notice code, so the notice code is what should include the changed definition. 

FieldName will automatically appear once this PR is merged and released.

davidgamez (approved):
LGTM!

emmambd:
@davidgamez Woohoo! Is this a breaking change since it means downstream users need to remove the special-case for this rule?

core/src/main/java/org/mobilitydata/gtfsvalidator/notice/NonAsciiOrNonPrintableCharNotice.java
```diff
@@ -34,17 +34,17 @@ public class NonAsciiOrNonPrintableCharNotice extends ValidationNotice {
   /** Row number of the faulty record. */
   private final int csvRowNumber;
 
-  /** Name of the column where the error occurred. */
-  private final String columnName;
+  /** Faulty record's field name. */
```
davidgamez, line 37: [nit]: Just to be consistent with other fields' documentation:

```suggestion
  /** The name of the faulty field. */
```

davidgamez:
> @davidgamez Woohoo! Is this a breaking change since it means downstream users need to remove the special-case for this rule?

Yes, this is a breaking change, as any consumer parsing this rule will need to consume the "new" property name.

davidgamez:
@c-tonneslan, please remove `columnName` from the expected field names in the [NoticeFieldsTest test](https://github.com/MobilityData/gtfs-validator/blob/21b27a4ac903bc2bc43d42a89bb3dbb47cc8921d/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/NoticeFieldsTest.java#L69)

davidgamez (changes requested):

PR #2169, opened 2026-06-25
fix: make MemoryUsageRegister thread-safe for concurrent validations

### LINKED ISSUE #2168: MemoryUsageRegister not thread safe
### Describe the bug

The MemoryUsageRegister class uses a singleton instance so that both annotation and manual uses can point to the same instance. This is problematic in concurrent validation scenarios, as the memory report is compromised. Current MobilityData memory reports are not compromised because the API doesn't process calls concurrently.

### Steps/Code to Reproduce

- Start the API server with gtfs-validator-core 8.0.1 and current MemoryUsageRegister singleton implementation.
- Trigger two validation requests at the same time (feed A and feed B), so two ValidationRunner.run(...) executions overlap.
- During execution, both runs call MemoryUsageRegister.getInstance().registerMemoryUsage(...).
- One run reaches clearRegistry() while the other run is still in progress.
- Observe the memory usage snapshots collected for each run.

### Expected Results

MemoryUsageRegister must be thread-safe to allow concurrent validation calls.

### Actual Results

Memory reports are compromised when multiple feeds are validated concurrently.

### Screenshots

_No response_

### Files used

_No response_

### Validator version

8.0.1

### Operating system

any

### Java version

17

### Additional notes

_No response_

themightychris:
## Summary

`MemoryUsageRegister` is a process-global singleton — a single static instance with an unsynchronized `registry` list — and `ValidationRunner.run()` calls `clearRegistry()` at the start of every run. When two validations run concurrently in the same JVM, they append to the same list and one run's `clear()` wipes the other's in-flight snapshots, so `summary.memoryUsageRecords` is corrupted. Validation notices are unaffected, since the `NoticeContainer` is per-run.

This is the bug described in #2168.

## Fix

Thread-confine the register. Both `@MemoryMonitor` interception points — `ValidationRunner.run` and `GtfsFeedLoader.loadAndValidate` — execute on the thread that invokes `run()`, so giving each thread its own register isolates concurrent validations with no locking and no behavior change for the single-threaded CLI/desktop path.

```java
private static final ThreadLocal<MemoryUsageRegister> INSTANCE =
    ThreadLocal.withInitial(MemoryUsageRegister::new);

public static MemoryUsageRegister getInstance() {
  return INSTANCE.get();
}
```

## Test

`MemoryUsageRegisterTest` adds a barrier-driven interleaving that forces the exact corruption (thread B calls `clearRegistry()` after thread A has registered a snapshot). On the old singleton it fails with `expected: A1,A2 but was: B1,A2`; with the fix each thread sees only its own snapshots.

## Validation against a real concurrent service

We exercised this against an HTTP validator service (a Spring wrapper around a shared `ValidationRunner`, as in MobilityData/gtfs-validator-api#1), running 8 concurrent validations of a ~3 MB / ~13k-trip feed:

- **Before:** the 8 runs reported 4 / 13 / 14 / 17 / 24 / 28 / 28 / 31 / 32 `memoryUsageRecords` (the single-run baseline is 4) — cross-contaminated.
- **After:** all 8 runs report exactly 4 records with no cross-run contamination; notices are identical across runs.

Resolves #2168.

### DIFF
--- a/core/src/main/java/org/mobilitydata/gtfsvalidator/performance/MemoryUsageRegister.java
+++ b/core/src/main/java/org/mobilitydata/gtfsvalidator/performance/MemoryUsageRegister.java
@@ -9,7 +9,13 @@ import java.util.List;
 public class MemoryUsageRegister {
   private static final FluentLogger logger = FluentLogger.forEnclosingClass();
 
-  private static MemoryUsageRegister instance = new MemoryUsageRegister();
+  // Thread-confined register: a validation runs entirely on the single thread
+  // that invokes ValidationRunner.run (the @MemoryMonitor-annotated entry points
+  // and the manual snapshots all execute on it), so giving each thread its own
+  // register isolates concurrent validations with no locking and no cross-run
+  // clearRegistry()/append contamination. Fixes #2168.
+  private static final ThreadLocal<MemoryUsageRegister> INSTANCE =
+      ThreadLocal.withInitial(MemoryUsageRegister::new);
   private final Runtime runtime;
   private List<MemoryUsage> registry = new ArrayList<>();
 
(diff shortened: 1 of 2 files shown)


davidgamez (approved):
LGTM!

PR #2181, opened 2026-09-01
feat: report empty required and recommended files

### LINKED ISSUE #1853: Required and recommended notices on files with no rows
# Description
The `missing_required_file` and `missing_recommended_file` notices are triggered when the file is not in the feed. This doesn't cover files with headers and no rows. Having a file in the feed with no data(only headers) defeats the intention of the required and recommended notices, as it doesn't encourage producers to add the data.

# Solution Proposed
Add additional notices `_required_file_empty_` and `_recommended_file_empty_` (or a better name) to express that the files are present but have no rows.

# Alternative Solution
We can also modify the required and recommended notice to catch files with no rows. The downside is that consumers can be confused when providing a file with no rows and getting missing required or recommended file.

Example fee, https://mobilitydatabase.org/feeds/mdb-1147

mackenziereading19:
## Summary

This PR addresses #1853 by distinguishing GTFS files that are present with valid headers but contain no data rows.

The loader already treats a completely empty/headerless CSV as `EmptyFileNotice`. However, a headers-only file is considered parsable and previously produced no notice even when the file was required or recommended.

This change adds:

- `RequiredFileEmptyNotice` (ERROR) for required files with headers but no data rows;
- `RecommendedFileEmptyNotice` (WARNING) for recommended files with headers but no data rows;
- no new notice for optional files with headers but no data rows;
- updated `EmptyFileNotice` documentation so it only describes genuinely headerless files.

The parsed table remains `PARSABLE_HEADERS_AND_ROWS` in all headers-only cases because the file structure and headers are valid; the new notices describe the file-content requirement rather than reclassifying the table as missing or unparsable.

## Tests

Added regression coverage for:

- required headers-only file -> `RequiredFileEmptyNotice`;
- recommended headers-only file -> `RecommendedFileEmptyNotice`;
- optional headers-only file -> no empty-file notice.

Local verification:

- full `:core:test`: pass;
- targeted `NoticeReferenceTest`: pass;
- `:core:spotlessCheck` and `:main:spotlessCheck`: pass;
- `git diff --check`: pass.

Fixes #1853


skalexch (commented):
I tried locally with https://mobilitydatabase.org/feeds/gtfs/mdb-340.

RequiredFileEmptyNotice works well. RecommendedFileEmptyNotice did not work for `shapes.txt` or when I only kept the header for `feed_info.txt`. I am not sure if we tag these files as recommended actually. We tag `feed_info.txt` as Conditionally Required.

By the way, a headerless `shapes.txt` seems to raise MissingRecommendedFile(`shapes.txt`) through the check `shapeTable.getEntities().isEmpty()` ([here](https://github.com/MobilityData/gtfs-validator/blob/21b27a4ac903bc2bc43d42a89bb3dbb47cc8921d/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/MissingShapesFileValidator.java#L38)).

A few suggestions:
- Double check if those Recommended files have a "Recommended" tag (only two files I think: shapes and feed_info. We already have `MissingFeedInfoValidator` which tests for a `shapes.txt` with a header only. We can do the same in 
- If tagging a file as Recommended is not easy, especially for `feed_info.txt`, then remove RecommendedFileEmptyNotice from this PR and add the check in `MissingFeedInfoValidator`.

mackenziereading19:
Thanks for testing this and for the suggestion. I’ve updated the PR accordingly in 8b74e243.

The generic RecommendedFileEmptyNotice has been removed. RequiredFileEmptyNotice remains in CsvFileLoader for genuinely required files.

For feed_info.txt, MissingFeedInfoValidator now treats a headers-only/empty table the same as a missing file, preserving its existing conditional semantics:

no translations.txt → MissingRecommendedFileNotice(feed_info.txt)
translations.txt present → MissingRequiredFileNotice(feed_info.txt)

I left shapes.txt on its existing dedicated path, since MissingShapesFileValidator already treats an empty shapes table as missing and applies the DRT exceptions before emitting MissingRecommendedFileNotice.

Focused verification after the revision:

CsvTableLoaderTest: 8 passed
MissingFeedInfoValidatorTest: 5 passed
MissingShapesFileValidatorTest: 3 passed
NoticeReferenceTest: passed
core/main Spotless checks: passed

skalexch (commented):
Good from spec side. Dev code review is required to approve the PR.

PR #2182, opened 2026-09-01
docs: clarify mixed case notice behavior

### LINKED ISSUE #1799: Adding clarification on how mixed_case_recommended_field works
### Describe the problem

The current notice description of `mixed_case_recommended_field` warning is as follows:

> This field contains customer-facing text and should use Mixed Case (upper and lower case letters) 

However, it does not completely describe [how validator works](https://github.com/MobilityData/gtfs-validator/blob/5a3696c50c08c52f83a201322056d6eb4a415c9b/processor/src/main/java/org/mobilitydata/gtfsvalidator/processor/MixedCaseValidatorGenerator.java) and cause potential confusions.

For example, a stop name `23rd and FM 107` will trigger the warning, since the validator would split text string based on **non-letter characters** into separate text string, but it does contain upper and lower case if looking at it as a whole.

### Suggest a fix or an alternative

Suggestion to fix:
Adding clarification text on notice description.

mackenziereading19:
## Summary

Clarifies the documentation for `mixed_case_recommended_field` so it reflects the validator's current token-based behavior.

The validator splits values on non-letter characters and evaluates the resulting text tokens rather than only checking whether the complete field value contains both upper- and lowercase letters. The updated notice documentation explains that distinction and notes the single-token lowercase case.

No validation logic is changed.

## Tests

- `:core:javadoc`: pass
- targeted `MixedCaseSchemaTest`: pass
- `:core:spotlessCheck`: pass
- `git diff --check`: pass

Fixes #1799

### DIFF
--- a/core/src/main/java/org/mobilitydata/gtfsvalidator/notice/MixedCaseRecommendedFieldNotice.java
+++ b/core/src/main/java/org/mobilitydata/gtfsvalidator/notice/MixedCaseRecommendedFieldNotice.java
@@ -10,10 +10,16 @@ import org.mobilitydata.gtfsvalidator.annotation.GtfsValidationNotice.UrlRef;
  * case letters).
  *
  * <p>This field contains customer-facing text and should use Mixed Case (upper and lower case
- * letters) to ensure good readability when displayed to riders. Avoid the use of abbreviations
- * throughout the feed (e.g. St. for Street) unless a location is called by its abbreviated name
- * (e.g. “JFK Airport”). Abbreviations may be problematic for accessibility by screen reader
- * software and voice user interfaces.
+ * letters) to ensure good readability when displayed to riders. The validator splits the field
+ * value into letter-only tokens at non-letter characters. For values with multiple relevant tokens,
+ * at least one token should contain both upper and lower case letters. A single multi-character
+ * token consisting only of lowercase letters also triggers this notice. As a result, the check
+ * applies to individual text tokens rather than simply checking whether the complete field value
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/mobilitydata/gtfsvalidator/notice/MixedCaseRecommendedFieldNotice.java
```diff
@@ -10,10 +10,16 @@
...
- * letters) to ensure good readability when displayed to riders. Avoid the use of abbreviations
- * throughout the feed (e.g. St. for Street) unless a location is called by its abbreviated name
- * (e.g. “JFK Airport”). Abbreviations may be problematic for accessibility by screen reader
- * software and voice user interfaces.
+ * letters) to ensure good readability when displayed to riders. The validator splits the field
+ * value into letter-only tokens at non-letter characters. For values with multiple relevant tokens,
+ * at least one token should contain both upper and lower case letters. A single multi-character
+ * token consisting only of lowercase letters also triggers this notice. As a result, the check
+ * applies to individual text tokens rather than simply checking whether the complete field value
+ * contains both upper and lower case letters.
```
skalexch: @mackenziereading19 while the description clearly describes what the validator does, it will be complicated to read and understand from the report.

I think a better way to convey it is more like this, feel free to change and improve it.

```suggestion
 * letters) to ensure good readability when displayed to riders. The validator splits the text when it
 * encounters a non-letter character. Therefore, each separate word or token of the text should
 * contain both upper and lower case letters.

